PR #572, opened 2015-01-01
Include rgba function def, and un-ambiguate rgba in use

whitingjp:
Fixed a compile issue on Ubuntu 14.04.  Was complaining that rgba was undefined, and this resolves that.  I am not 100% certain that this is the right color.h containing rgba to include, but unlike including the gfx/color.h one this causes no onwards 'ambiguous definition' issues.

### DIFF
--- a/src/render/median_cut.h
+++ b/src/render/median_cut.h
@@ -8,6 +8,8 @@
 #define RENDER_MEDIAN_CUT_H_INCLUDED
 #pragma once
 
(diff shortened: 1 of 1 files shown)


dacap:
It's the correct color.h :smile: I've move the render code from `doc` to `render` namespace, so the it looks like gcc cannot disambiguate the problem now. Thanks for the contribution @whitingjp!

PR #609, opened 2015-03-15
Small fixes needed to build on Linux

aisamanra:
First of all—wonderful piece of software!

In compiling from source on a Linux machine, there were a few small issues—two files missing `#include<string>` and one that failed to compile because of shadowed template names. All this PR does is fix those.

I've confirmed that this builds on my reasonably up-to-date installs of Fedora, Arch Linux, and Void Linux, and also that the current repo clone did not successfully build on any of them before these changes.

### DIFF
--- a/src/app/pref/option.h
+++ b/src/app/pref/option.h
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "base/signal.h"
+#include <string>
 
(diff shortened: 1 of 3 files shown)


dacap:
Thanks for the contribution!

PR #644, opened 2015-04-29
Build fixes

agateau:
I needed to make those two changes to get Aseprite master to build on my system (Ubuntu 14.10, gcc 4.9.1).

I am not entirely sure why prefixing calls to `notifyActiveDocumentChanged()` with `this->` was necessary, but this is what gcc error suggested, so who am I to argue :)


includeLuka:
Fixed the build problem on my system too.

dacap:
Thanks @agateau !

PR #651, opened 2015-05-03
fix View-Home

blackmiaool:


src/app/ui/main_window.h
```diff
@@ -68,6 +68,7 @@ namespace app {
     void reloadMenus();
     void showNotification(INotificationDelegate* del);
     void showHome();
+    bool isSelectedHome();
```
dacap: Rename this to isHomeSelected()

dacap:
Replace the commit message (you will need to amend the commit and do a push --force). Use some message like "Select Home tab when it's not selected in HomeCommand"

src/app/ui/main_window.h
blackmiaool: I just imitated the "isModifiedTab"...

dacap: Hmm, I think I've to rename that function too. Just as a side note, there are several refactors that I should do but I've never time to do them. And sometimes, there are ways to do the same thing in two different ways (because a refactored-better way was created, but it didn't invalidate the old one). I've to write a document about that.

PR #656, opened 2015-05-04
Fix crash using NewBrush when Home tab is selected

blackmiaool:


src/app/commands/cmd_new_brush.cpp
```diff
@@ -64,6 +64,10 @@ bool NewBrushCommand::onEnabled(Context* context)
 
 void NewBrushCommand::onExecute(Context* context)
 {
+  ASSERT(current_editor);
+  if (!current_editor)
+    return;
+
```
dacap: Is this really needed? The onEnabled() condition should filter this case. (Anyway you may have added this for some special case.)

blackmiaool: 1 open aseprite
2 open a .ase tab
3 go back to home tab
4 click new brash
5 crash

dacap: Nice catch. I think this is a bigger problem. I think that several other commands could be affected by this kind of scenario (Home tab is selected, and we use a command that use the active document view or expected current_editor != nullptr when it isn't).

The real solution is to completely remove the `current_editor` global variable (maybe this is the older global variable in Aseprite, and the more postponed refactor that I've to do).

Anyway, I'll merge this PR, just one thing: Could you change the commit message to something like "Fix crash using NewBrush when Home tab is selected"

dacap: I'll try to find other solution, because a lot of commands can be executed when the Home is selected and a document in other tab.

blackmiaool: Whatever , I fixed the commit message.
_d l m f g9u7672c_kkof

PR #678, opened 2015-05-18
Sort recovered sessions in the recovery view to show the most recent first

agateau:
When I open the recovered session view (which I do a bit too often because of a crash in curl-related crash with longjmp...) I am interested in the latest session, therefore listing the most recent session first instead of having to find it in the list is nicer for me.

I considered sorting the session list in app::crash::DataRecovery, but I felt this was a UI concern so it was more appropriate to do the sort before populating the list. Doing it in DataRecovery would avoid the need for adding the `path()` getter to app::crash::Session though. I am happy to redo it that way if you prefer.

### DIFF
--- a/src/app/crash/session.cpp
+++ b/src/app/crash/session.cpp
@@ -71,6 +71,11 @@ std::string Session::name() const
     return name;
 }
 
+std::string Session::path() const
+{
+  return m_path;
+}
+
 const Session::Backups& Session::backups()
 {
   if (m_backups.empty()) {
--- a/src/app/crash/session.h
+++ b/src/app/crash/session.h
@@ -40,6 +40,7 @@ namespace crash {
     ~Session();
 
     std::string name() const;
+    std::string path() const;
     const Backups& backups();
 
     bool isRunning();
--- a/src/app/ui/data_recovery_view.cpp
+++ b/src/app/ui/data_recovery_view.cpp
@@ -19,6 +19,7 @@
 #include "app/ui/skin/skin_theme.h"
 #include "app/ui/workspace.h"
 #include "base/bind.h"
+#include "base/path.h"
 #include "ui/alert.h"
 #include "ui/button.h"
 #include "ui/entry.h"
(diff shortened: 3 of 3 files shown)


dacap:
Hi @agateau, I'm not quite sure about this descending order of sessions. I was thinking in an ascending order just to know what sprites should be restored (were lost) first. The idea is that a session should be deleted after it's recovered.

dacap:
I think that we should fix that curl-related problem. Maybe a setlongjmp is missing.

agateau:
Oh, I expected recovered sessions would automatically delete themselves when recovered. This is what happens in other creation software, no?

Regarding the curl-related problem: yes we should fix this. I think I caught it in GDB once but I haven't taken the time to investigate more. It's quite random. Have you experienced it as well?

dacap:
> I expected recovered sessions would automatically delete themselves when recovered. This is what happens in other creation software, no?

Yeah. The best way would be something like this: All sprites from a crash sessions are recovered and the session is deleted (automatically). Anyway as the data recovery is in an early development stage, the program could crash when the session is being recovered. (Luckily I've never experienced this, but I think it could happen, e.g. a lot of images to recover, an huge animation with lot of frames, or invalid data, bugs, bad_allocs, etc. several situations that are not fully contemplated.)

Anyway, as you see, is something to improve.

> Regarding the curl-related problem [...] Have you experienced it as well?

No, I haven't. Are you on Linux, right? Are you using a shared libcurl version?  Or just the one included in `third_party` dir? I've to update the source code of the curl library (converting it to a submodule).

agateau:
On Tue, May 19, 2015, at 17:52, David Capello wrote:
> > I expected recovered sessions would automatically delete themselves when recovered. This is what happens in other creation software, no?
> 
> Yeah. The best way would be something like this: All sprites from a crash
> sessions are recovered and the session is deleted (automatically). Anyway
> as the data recovery is in an early development stage, the program could
> crash when the session is being recovered. (Luckily I've never
> experienced this, but I think it could happen, e.g. a lot of images to
> recover, an huge animation with lot of frames, or invalid data, bugs,
> bad_allocs, etc. several situations that are not fully contemplated.)
> 
> Anyway, as you see, is something to improve.

I see. The risks could be mitigated by only removing the session once it
has been fully recovered, or even better, only after it has been fully
recovered and a new session has been saved.

> > Regarding the curl-related problem [...] Have you experienced it as well?
> 
> No, I haven't. Are you on Linux, right? Are you using a shared libcurl
> version?  Or just the one included in `third_party` dir? I've to update
> the source code of the curl library (converting it to a submodule).

Yes, I am on Linux, using the version of libcurl from the `third_party`
dir.

dacap:
I'll take a look to the curl issue (issue #679)

dacap:
Probably I'll modify the way data recovery sessions are being shown. I don't like the current UI so maybe a grid or something similar, I'll see how to solve it (If I made a grid, it's pretty sure it will used for the keyboard shortcuts dialog too.)

It would be nice to have a "Date" column that we can click and re-sort the grid.

PR #699, opened 2015-06-14
Create solarized.gpl

Nsheppard01:
The Solarized Palette


dacap:
Thanks for the PR! :+1:

From: https://github.com/altercation/solarized/tree/master/gimp-palette-solarized

PR #709, opened 2015-06-20
using mouse callback for more fine grained button press detection

farnold:
The problem with the current polling mechanism is that some button presses can be missed if they happen between polling. This is especially noticeable on touchpad devices where some taps aren't recognized. With the intent to solve this problem I used allegro's mouse callback to get a more fine grained feedback of pressed mouse buttons.
Also, I implemented a fallback to the old polling mechanic for devices which only support polling.

I tested on a Retina MacBook Pro running OS X Yosemite.

### DIFF
--- a/src/she/alleg4/mouse_poller.cpp
+++ b/src/she/alleg4/mouse_poller.cpp
@@ -42,6 +42,8 @@ int she_mouse_y;
 int she_mouse_z;
 int she_mouse_b;
 
+bool she_mouse_polling_required;
+
 // Flag to block all the generation of mouse messages from polling.
 bool mouse_left = false;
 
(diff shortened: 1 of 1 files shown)


dacap:
Hi @farnold. Actually [the Windows port does't use the poll mechanism](https://github.com/aseprite/aseprite/blob/f310a51dfb9eaed1835fd4c0d71428d2cc5f589b/src/she/alleg4/she.cpp#L25) for the mouse, so the same behavior needed to be ported to OS X. So thanks for this patch! I'll be testing and merging this PR soon.

Meanwhile, could you sign [our CLA](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla)? Thanks again!

farnold:
You're welcome!
Well then it's for all the other platforms using allegro4. ;)
[CLA is signed](https://github.com/aseprite/opensource/pull/5).

dacap:
> You're welcome!
> Well then it's for all the other platforms using allegro4. ;)

Yeee :smile: 

> CLA is signed

Excellent, it looks fine. I'll need [the email to cla@aseprite.org too](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sending-the-email) (sorry for the inconvenience)

farnold:
Oh, I missed the email part. It's sent now.
Then there was a copy & paste error on my part where only the left mouse button was detected. I edited the commit and fixed it.

PR #795, opened 2015-09-01
Support for the simple lossless and lossy (non animated) WebP Format for #273

sirblackheart:
This pull request contains [simple lossless](https://developers.google.com/speed/webp/docs/riff_container#simple_file_format_lossless) and [simple lossy](https://developers.google.com/speed/webp/docs/riff_container#simple_file_format_lossy) webp file support with the help of libwebp.

For now this means aseprite can open and save static (not animated) lossless and lossy webp files as long as they do not use the [extended format](https://developers.google.com/speed/webp/docs/riff_container#extended_file_format). We still lack support for animation, color profile, EXIF and XMP. But probably only animation is going to be of interest for aseprite for now.

Animation will be implemented in a second step along with support for the  [extended format](https://developers.google.com/speed/webp/docs/riff_container#extended_file_format), exposing the missing features so aseprite could use them if any need arises.

Also a rudimentary [save options dialog](https://github.com/sirblackheart/aseprite/blob/83159aba8a4b984320e4568bc06b5d1e2346e887/data/widgets/webp_options.xml) was implemented so user can save webp in lossless or lossy format. It was made quick and probably still needs some optimization for better UX.

screenshot from 2015-09-01 13-53-55


sirblackheart:
The travis builds failed do to the split_filename_tests bug. Is it possible to redo it without another pull request?

dacap:
Hi @sirblackheart, don't worry about travis. The PR looks good so far. I'll make some comments and test it soon.

data/widgets/webp_options.xml
```diff
@@ -0,0 +1,48 @@
+<!-- ASEPRITE -->
+<!-- Copyright (C) 2014, 2015 by David Capello -->
```
dacap: You can put 2015 and your name here :+1: the same for the webp_format.cpp file.

src/app/file/webp_format.cpp
```diff
@@ -0,0 +1,361 @@
...
+    webp_options.reset(new WebPOptions);
+
+  // Non-interactive mode
+  if (!fop->context || !fop->context->isUIAvailable())
+    return webp_options;
+
+  try {
+
+    // Load the window to ask to the user the JPEG options he wants.
+    UniquePtr<ui::Window> window(app::load_widget<ui::Window>("webp_options.xml", "webp_options"));
```
dacap: You can use the auto-generated widget here. The .h is generated in `your_build_dir/src/app/generated_webp_options.h` and you can use it in this way:
```
#include "generated_webp_options.h"
...
app::gen::WebpOptions window;
...
```
You can find [here](https://github.com/aseprite/aseprite/blob/master/src/app/file/gif_format.cpp#L1247) an example.

src/app/file/webp_format.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+  len = ftell(fp);
+  rewind(fp);
+
+  if (len < 4) {
+    fop_error(fop, "%s is corrupt or not a WebP file\n", fop->filename.c_str());
+    return false;
+  }
+
+  buf = (uint8_t*) malloc(len);
```
dacap: You can use [base_malloc()](https://github.com/aseprite/aseprite/blob/master/src/app/file/jpeg_format.cpp#L153) or a [std::vector](https://github.com/aseprite/aseprite/blob/master/src/app/file/gif_format.cpp#L1108) directly (the last one is preferred, you don't need to worry about base_free()).

src/app/file/webp_format.cpp
```diff
@@ -0,0 +1,390 @@
...
+  return true;
+}
+
+#ifdef ENABLE_SAVE
+struct writerData {
+  FILE* fp;
+  FileOp* fop;
+};
+
+const std::pair<WebPEncodingError, std::string> enc_error_map_data[] = {
```
dacap: I'm not sure about this array of pairs + a map. Is this just to show error messages? Could we have a function with a switch/case to convert `WebPEncodingError` codes to `const char*` strings (not even std::string)?

src/app/file/webp_format.cpp
```diff
@@ -0,0 +1,390 @@
...
+
+static int FileWriter(const uint8_t* data, size_t data_size, const WebPPicture* const pic)
+{
+  return data_size ? (fwrite(data, data_size, 1, ((writerData*)pic->custom_ptr)->fp) == 1) : 1;
+}
+
+
+//Helper functions instead of std::stoi because of missing c++11 on mac os x
+template<typename T>
+static inline std::string ToString(const T& v)
```
dacap: Is this function needed?

src/app/file/webp_format.cpp
```diff
@@ -0,0 +1,390 @@
...
+{
+    std::ostringstream ss;
+    ss << v;
+    return ss.str();
+}
+
+template<typename T>
+static inline T FromString(const std::string& str)
```
dacap: You can use [base::convert_to&lt;int&gt;()](https://github.com/aseprite/aseprite/blob/master/src/base/convert_to.h#L24)

src/app/file/webp_format.cpp
sirblackheart: Both the ProgressReport and FileWriter function are hooked into the libwebp encoder which then calls them while encoding. As soon as bytes to write are ready it directly writes them into the file. It also calls ProgressReport to update on its progress.

dacap: Sorry, it looks like a GitHub UI problem. I clicked the ToString() line. It looks like nobody is using it (anyway you can use `convert_to&lt;std::string&gt;()` if needed)

dacap:
I got the following warnings on VC 2013:
```
webp_format.cpp(128) : warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)
webp_format.cpp(320) : warning C4244: 'argument' : conversion from 'float' to 'int', possible loss of data
webp_format.cpp(320) : warning C4244: '=' : conversion from 'int' to 'float', possible loss of data
webp_format.cpp(328) : warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)
webp_format.cpp(330) : warning C4244: 'argument' : conversion from 'float' to 'int', possible loss of data
webp_format.cpp(338) : warning C4244: '=' : conversion from 'int' to 'float', possible loss of data
webp_format.cpp(344) : warning C4244: 'argument' : conversion from 'float' to 'int', possible loss of data
```

About other issue, I'm not an huge fan of `auto` keyword. I've used it for iterators mainly. Here I think we can use `Image*` and `long` in their respective cases.

dacap:
Looks great :+1: Thanks for your effort Gabriel! I will merge it. Do you think we should close #273?

sirblackheart:
Thanks for having the patience to review my code. I think we should keep #273 open till i get the  extended format implemented.

PR #801, opened 2015-09-05
Fix typo in INSTALL.md

green-big-frog:


dacap:
Thanks @green-big-frog

Issue #806, opened 2015-09-10
Don't convert the layer to Background after Flatten


dacap:
*Layer > Flatten* option should merge all visible layers but nothing else, it shouldn't convert the layer to a Background.

Reported by @enol

PR #812, opened 2015-09-21
Fix for #806; flattening layers won't always convert to background.

Fojar:
Now it will only make the result a background if a visible background was part of the flatten. Otherwise the result will be a regular layer named "Flattened".


dacap:
Looks good @Fojar, thanks!

Issue #838, opened 2015-10-24
Spray tool fails with a spray width of 6 or less


KasumiArai:
1. Select the spray tool
2. Change spray width to 7
3. Spray. It works as intended.
4. Change spray width to 6.
5. Spray. No pixels are changed.
![spraytoolfails](https://cloud.githubusercontent.com/assets/8582123/10710880/ac1de1ba-7a38-11e5-8d11-e6c8c2c0e32b.gif)

Aseprite v1.1.1-beta8 on Windows Vista 32bit

Fojar:
This happens for several combinations of small values of spray width and speed, when the calculation for how many pixels to spray rounds down to zero. @dacap, I have a fix for this that handles the number of pixels to spray as a fraction; can I submit a pull request?

dacap:
Hi @Fojar, a PR would be great as I didn't see this issue yet.

PR #842, opened 2015-10-27
Issue #838

Fojar:
The spray tool will now skip painting sometimes if it needs to, so overall it achieves the intended fractional rate of spraying.

### DIFF
--- a/src/app/tools/point_shapes.h
+++ b/src/app/tools/point_shapes.h
@@ -131,6 +131,7 @@ private:
 
 class SprayPointShape : public PointShape {
   BrushPointShape m_subPointShape;
+  float pixel_remainder = 0;
 
 public:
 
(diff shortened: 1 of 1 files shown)


dacap:
Hi @Fojar, I'll make some comments in the commits. (I haven't tested it yet.)

src/app/tools/point_shapes.h
```diff
@@ -131,6 +131,7 @@ class FloodFillPointShape : public PointShape {
 
 class SprayPointShape : public PointShape {
   BrushPointShape m_subPointShape;
+  float point_remainder = 0;
```
dacap: Rename to "m_pointRemainder" (similar to "m_subPointShape", we use "m_" prefix for member variables)

dacap:
Could you squash both commits (and maybe the extra change) in just one commit?

dacap:
As a side note, after I merge this PR, I'd love to change the random number generator with some of the new standard generators (e.g. Mersenne Twister). But that is something I'll add after this PR.

dacap:
(Other note, don't worry about the Travis failed check, it was my fail with the freetype2 repository. It should be fixed now.)

Fojar:
Sure, I will squash these commits. I'll have to make a new PR for that, correct?

dacap:
>  I'll have to make a new PR for that, correct?

Not really, you can use `git push --force` to overwrite your old commits in your branch.

Fojar:
Ah ok, I wasn't aware of that! Sorry, I already made a new PR, so I'll have to remember that for next time.

dacap:
:+1: :smile:

PR #862, opened 2015-11-15
Fix symmetry mode axis line redraw issue

### LINKED ISSUE #851: The axis line of symmetry: redraw issue
Axis line has a redraw issue in preview window and other duplicated views. 
When we move the axis line,  line still exist on other windows and create multiple copies itself. 
ase_redraw_bug

TheKK:
Hope this is the best way to fix #851.


dacap:
Hi @TheKK, I think the symmetry line shouldn't be visible in the preview. I'm pushing a fix for #851 in a second. Anyway I'll review this patch because it can be useful in other way.

Issue #625, opened 2015-04-11
Undefined reference to symbol 'XGrabPointer' while linking libshe.a


Mailaender:
Another compilation error:

```
/usr/lib64/gcc/x86_64-suse-linux/4.8/../../../../x86_64-suse-linux/bin/ld: ../lib/libshe.a(she.cpp.o): undefined reference to symbol 'XGrabPointer'
/usr/lib64/libX11.so.6: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
```

PR #863, opened 2015-11-15
Fixed undefined reference to symbol 'XGrabPointer' 

Mailaender:
as described in #625. The previous commit linked there made me think this was already fixed in your main tree, but that doesn't seem to be the case. This is essentially the patch from @winterheart but rebased.


dacap:
It should be fixed with #1082

PR #867, opened 2015-11-20
Instructions for updating a copy of the git repo

jpnk:


dacap:
Thanks @jpnk for the contribution!

PR #885, opened 2015-12-03
Show preview correctly when moving multiple cels

### LINKED ISSUE #731: Moving multiple frames: only shows first layer selected until mouse release
When you pick multiple frames and displace them to other position you can only see the first layer selected at the moment. When you release the mouse they suddenly appear in the next position.

---

Cuando desplazas varios frames a la vez a una nueva posición con la herramienta move solo previsualizas el desplazamiento de la primera capa seleccionada. Al soltar el ratón aparecen de repente todas las capas reposicionadas en el nuevo punto. Deberían poder verse todas para controlar mejor la posición.

1pic

2pic

TheKK:
This should fix #731

I've make sure it works when:
0. selecting single cel
1. selecting multiple cels
2. selecting no cel (range is not enabled)
3. selecting single layer/frame
4. selecting multiple layers/frames
5. moving with axis locked
6. the auto-select-layer is on and we select only one cel, then the layer we pointed was selected and moved

The only thing which not very clear to me is the behaviour of item 6. What should moving tool do when auto-select-layer is on and we have multiple cels selected?

### DIFF
--- a/src/app/ui/editor/moving_cel_state.cpp
+++ b/src/app/ui/editor/moving_cel_state.cpp
@@ -36,21 +36,31 @@ using namespace ui;
 MovingCelState::MovingCelState(Editor* editor, MouseMessage* msg)
   : m_canceled(false)
 {
+  ContextWriter writer(UIContext::instance(), 500);
   Document* document = editor->document();
+  DocumentRange range = App::instance()->getMainWindow()->getTimeline()->range();
   LayerImage* layer = static_cast<LayerImage*>(editor->layer());
   ASSERT(layer->isImage());
 
-  m_cel = layer->cel(editor->frame());
-  ASSERT(m_cel); // The cel cannot be null
-  if (m_cel) {
-    m_celStart = m_cel->position();
-  }
-  else {
-    m_celStart = gfx::Point(0, 0);
-  }
-  m_celNew = m_celStart;
+  Cel* currentCel = layer->cel(editor->frame());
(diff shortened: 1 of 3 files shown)


dacap:
> The only thing which not very clear to me is the behaviour of item 6. What should moving tool do when auto-select-layer is on and we have multiple cels selected?

I think that in this case the new layer/cel should be selected, and the previous Timeline selection should be disabled (so we move only one cel). Anyway there are three different cases:

What happen if we have auto-select and:
6.1. we move the current cel which is inside the Timeline selection range
6.2. we move other cel that is inside the Timeline selection range
6.3. we autoselect/move a cel that is outside the Timeline selection range

For 6.3 I think what I said: we should disable the Timeline selection and move only the new auto-selected cel. Anyway 6.1 and 6.2 are trickier.

I'll test this PR and see how it works. We can solve these extra issues later.

TheKK:
Btw, do we suppose to use `app::Transaction.execute(CMD)` rather than `app::DocumentApi`? (as the comment in `src/app/document_api.h` said)
If so, I think I could fix the minor problem in these files as well and prevent using `app::DocumentApi` in the future patch.

dacap:
At the moment we can use both ways. `Transaction::execute` is preferred, but if you find convenient to use some `DocumentApi` member function, you can use it.

In a future the `DocumentApi` should be completely replaced with other `Cmd`s.

Issue #879, opened 2015-11-29
Can 'Jump to Next Frame' looping in tagged frames?


alivegames:
Tag feature is a Lifesaver for me, i use it to manage massive characters. But i have a issue with it.

If i have 'Idle'(Frame 1 to 4) and 'Run'(Frame 5 to 9) two animation, when i finished the last frame of 'Run' (Frame 9) and press the next frame button, editor jump to 'Frame 1'. But I want jump to the first frame of 'Run', it is 'Frame 5' .
![image](https://cloud.githubusercontent.com/assets/6963993/11461877/7016d228-9746-11e5-98f7-2e050788b721.png)

Please tell me if have any way to do this.

And i have another request, when i export file using command, can 'Run' animation first frame start at zero ? Sorry for the trouble
![image](https://cloud.githubusercontent.com/assets/6963993/11461900/c60f3490-9746-11e5-9b38-28ca8e69158a.png)

dacap:
Dup #556

lazcht:
this seems to be bugging the new version again? or is it just me?

(the tag button in the preview box doesn't seem to work as well?)
![2018-12-14_16-35-33](https://user-images.githubusercontent.com/45871611/49995321-5df4e600-ffbe-11e8-976e-e03a378ff33c.gif)

also i'm totally new to github so sorry if i should post a new one instead, let me know. my version is Aseprite v1.2.10-beta4-x64

dacap:
@lazcht sorry but I'm not sure if I understood the GIF 😅 Are you using `,` and `.` keys to move through frames? or the arrow keys? See that there is a difference between arrow keys (that do not loop the tag) and `,` and `.` keys that they do loop inside the tag:

![image](https://user-images.githubusercontent.com/39654/50003147-75fa4480-ff81-11e8-801d-7665acbcc8f2.png)

lazcht:
sorry of the lack of clarity, my problem was the same with the thread starter's. which is, I can't navigate and loop frames inside a tag, instead of looping back on the first frame within tag, it's always back to the first frame. but now i have found my source of problem:

I did change the keyboard shortcut of the next & previous frame with` .` and `,` in `Menus` section, here: 
![image](https://user-images.githubusercontent.com/45871611/50006037-fc913f00-ffde-11e8-9d2c-0339bd099cad.png)
[...]

PR #887, opened 2015-12-05
Add commands to navigate frames with same tag

### LINKED ISSUE #556: Make , and . keys move in the loop section
If a loop is configured, we should be able to use , and . keys to move through the loop (and left/right arrows to move outside the loop)

TheKK:
This should fix #556 (#879)


dacap:
I'm not sure about this solution. Should we add new commands or a user-defined configuration flag to modify `GotoPreviousFrame`/`GotoNextFrame` behavior?

src/app/commands/cmd_goto_frame.cpp
```diff
@@ -88,6 +91,46 @@ class GotoNextFrameCommand : public GotoCommand {
...
+  Command* clone() const override { return new GotoNextFrameWithSameTagCommand(*this); }
+
+protected:
+  frame_t onGetFrame(Editor* editor) override {
+    Sprite* sprite = editor->sprite();
+    frame_t currentFrame = editor->frame();
+    FrameTag* tag = get_animation_tag(sprite, currentFrame);
+    frame_t frameToGo = currentFrame + frame_t(1);
+
+    if (frameToGo > tag->toFrame())
```
dacap: I didn't check this in runtime, but I think `tag` can be nullptr, so this line crashes if the command is used in animations without tags.

TheKK:
I just found that we can pass parameters to commands :P 
Solution which uses user-defined flag should be better than current one, since lots of duplicated code exist now.

dacap:
> I just found that we can pass parameters to commands :P 

:smile: Anyway in this case I'm not talking about command parameters.

I'm talking about a preference option in [pref.xml](https://github.com/aseprite/aseprite/blob/master/data/pref.xml#L68) that you can check using [Preference::instance()](https://github.com/aseprite/aseprite/blob/master/src/app/pref/preferences.h#L45) and that can be changed from [OptionsWindow](https://github.com/aseprite/aseprite/blob/master/src/app/commands/cmd_options.cpp#L39) ([options.xml](https://github.com/aseprite/aseprite/blob/master/data/widgets/options.xml))

The [UndoCommand](https://github.com/aseprite/aseprite/blob/master/src/app/commands/cmd_undo.cpp#L71) is a good example about one command which its behavior changes with a user-defined option.

TheKK:
I'm not a professional digital artist but I think during working I'll need both `goto next frame` and `goto next frame with certain tag` as well and might use them frequently. As my opinion, these two commands should be available at the same time. Defined their behaviour with global preference looks not very reasonable to me.

What do you think?

dacap:
> I think during working I'll need both goto next frame and goto next frame with same tag

You are right. Also it was my original idea in #556 :smile: I'll merge this PR, it looks fine as it is (don't worry about command params).

Thanks for your help! :wink:

TheKK:
Wait! The tag bug is still there!

dacap:
I'll fix it. Don't worry.

PR #890, opened 2015-12-06
Repaint animated sprite only when frame changed

TheKK:
I found CPU bursts crazily while playing animation, and I found this unnecessary repainting.


dacap:
Hi @TheKK, this is a nice catch. Anyway `Widget::invalidate` doesn't repaint the widget, it only indicates that the whole widget must be repainted when the message loop is processed in the future. (Anyway [this line](https://github.com/aseprite/aseprite/blob/master/src/ui/widget.cpp#L1019) could be the more CPU intensive, I'll profile it.)

PR #891, opened 2015-12-08
Now the symmerty handles snap to the edge of editor viewport after zoom in

### LINKED ISSUE #873: Symmetry axis handles disappear as you zoom in
A possible solution is to stick them to viewport borders.

TheKK:
This should fix #873


dacap:
Thanks @TheKK!

TheKK:
Always a pleasure :)

PR #895, opened 2015-12-09
Correct usage of variable arguments

TheKK:
This fix the crash while running aseprite with -v argument. (I only test this on GNU/Linux but I believe this would crash others platforms as well)

We should call va_start() before calling vfprintf() again.


dacap:
Nice fix. I didn't notice it on Windows, it surely depends on the specific C runtime implementation. Thanks again @TheKK!

PR #896, opened 2015-12-09
Ignore cmake build dir

tony:
Prevent build dir from being tracked in git. Common practice / mentioned within INSTALL.md to ``mkdir build`` and ``cmake ..``.


dacap:
I'm used to create the `build` directory outside the source code. But I think this is a good idea for regular users that follow the `INSTALL.md` guide. Thanks again @tony!

PR #900, opened 2015-12-14
Invalidate editor after background setting changed

### LINKED ISSUE #898: Checkered background does not properly redraw when changed
Repro steps:
1. open image with transparent background
2. open settings dialog. Move the dialog so that it only covers part of the transparent area'
3. change the Checkered Background size. Save and close the settings
expected: entirety of checkered background is redrawn immediately
actual: only the area covered by the settings dialog is redrawn

TheKK:
This should fix #898.

Use the same way, `afterChange` signal. as grid settings and onionSkin settings.


dacap:
Thanks for the patch @TheKK!

PR #910, opened 2015-12-20
Add "apply" button to options window

TheKK:
This should be useful for checking changes after some options set, without open options window again and again :)


dacap:
Hi @TheKK, a long time ago I think about this feature and I know that something could not work correctly with this Apply button. Right now I cannot remember what exactly. Anyway I'll be testing this soon so we can merge/close the PR.

PR #919, opened 2015-12-30
Improve build instructions

jpnk:
i changed all the commands from `cmake .. <options>` to `cmake <options> ..` - this is the standard order and on BSD systems the C library requires options to precede other arguments (although i don't know if this affects CMake).

also i was confused by the instructions, i thought i have to use `build/bin/aseprite` and shouldn't run `make install`. maybe it's just me but i still clarified that part.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -55,24 +55,30 @@ the repository clone in a directory called `aseprite-source`):
 
    If you have nmake (MSVC compilers):
 
-        C:\...\aseprite-source\build>cmake .. -G "NMake Makefiles"
+        C:\...\aseprite-source\build>cmake -G "NMake Makefiles" ..
 
    If you have Visual Studio you can generate a solution:
 
-        C:\...\aseprite-source\build>cmake .. -G "Visual Studio 12 2013"
+        C:\...\aseprite-source\build>cmake -G "Visual Studio 12 2013" ..
 
    If you are on Linux:
 
-        /.../aseprite-source/build$ cmake .. -G "Unix Makefiles"
+        /.../aseprite-source/build$ cmake -G "Unix Makefiles" ..
 
    For more information in [CMake wiki](http://www.vtk.org/Wiki/CMake_Generator_Specific_Information).
+   
+   Additionally you can change build settings by passing them on the command line, like so:
 
-3. After you have executed one of the `cmake .. -G <generator>`
+        /.../aseprite-source/build$ cmake -DCMAKE_INSTALL_PREFIX=~/software ..
+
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -55,24 +55,30 @@ the repository clone in a directory called `aseprite-source`):
...
+
+   or later on with a tool like `ccmake`.
+
+3. After you have executed one of the `cmake -G <generator> ..`
    commands, you have to compile the project executing make, nmake,
    opening the solution, etc.
 
 4. When the project is compiled, you can find the executable file
-   inside `build/bin/aseprite.exe`.
+   inside `build/bin/aseprite.exe`. If you invoked `make install` it will be copied to an appropriate location (e.g. `/usr/local/bin/aseprite` on Linux).
```
dacap: I'd prefer this in a *Linux Details* section (like the *Mac OS X details*). Anyway we've to change `build/bin/aseprite.exe` to something like `C:\...\aseprite-source\build\bin\aseprite.exe`

INSTALL.md
```diff
@@ -55,24 +55,30 @@ the repository clone in a directory called `aseprite-source`):
...
+        /.../aseprite-source/build$ cmake -G "Unix Makefiles" ..
 
    For more information in [CMake wiki](http://www.vtk.org/Wiki/CMake_Generator_Specific_Information).
+   
+   Additionally you can change build settings by passing them on the command line, like so:
 
-3. After you have executed one of the `cmake .. -G <generator>`
+        /.../aseprite-source/build$ cmake -DCMAKE_INSTALL_PREFIX=~/software ..
+
+   or later on with a tool like `ccmake`.
```
dacap: I think `cmake-gui` could be named too (the command line is `cmake-gui .`). Also some links to cmake-gui and ccmake docs could be included.

INSTALL.md
jpnk: there's also this
    
    If you are on Linux:
    /.../aseprite-source/build$ cmake -G "Unix Makefiles" ..

and my linux-specific setting changing example in this section. so i'm tempted to just rewrite it. i guess i'll leave a few words about the general process and move the rest into Windows/Linux/Mac OS X details.

Viomi:
Assuming `C:\` is really ostracizing to Linux users. Also I think using `aseprite-source/` instead of just `aseprite/` is rather confusing too, considering that by default `git clone --recursive https://github.com/aseprite/aseprite.git` clones to `aesprite/` by default. (This actually tripped me up into thinking `aesprite-source/` meant `src/`, leading to lots of build errors and confusion.)

PR #924, opened 2016-01-02
Apply brush size when clicking on brush icon

TheKK:
Currently, after we clicking one of the default brush icons (square, circle and line), it won't apply size and angle show on context bar until we change their value manually again.

This PR fix this problem.


dacap:
I'll see if this fixes #978

dacap:
Thanks for this @TheKK, I'm just rewriting the "active tool" logic and I'm just testing the context bar so I've just merged your change.

TheKK:
Always a pleasure! Since I'm free recently, I'll check if there's any thing I could help :)

PR #973, opened 2016-02-21
OS X: fix clang compile error over non-existent simpleini lib

farnold:
I was getting the following build error on OS X 10.11.3:
[ 94%] Linking CXX executable ../bin/aseprite
ld: library not found for -lsimpleini
clang: error: linker command failed with exit code 1 (use -v to see invocation)

Since simpleini appears to only be a header-only library I removed the target_link_libraries command from the CMakeLists.txt. Asesprite now builds fine for me.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,7 +270,6 @@ endif()
 include_directories(${LIBCURL_INCLUDE_DIR})
 
 # simpleini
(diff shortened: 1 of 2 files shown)


dacap:
:+1: I'll take a look to this PR and others in these days.

PR #994, opened 2016-03-02
replace onion settings icon fixes #956

### LINKED ISSUE #956: New onionskin config icon isn't very clear
I'm not sure what others thing, but personally I feel like the old layer window config icon for onionskin was more obvious than the newest, for what the button does. The other icons haven't changed too drastically, but this one has.

i.e. -> This is a fairly standard config icon
image

Whereas this doesn't seem obvious to me at all what it might be for:
image

Also, the old eye icon felt a bit more subtle/smooth in the interface, or less 'in my face' than the newest change.

Any thoughts on these, or on the other icons from anyone else?

sirblackheart:


dacap:
Thanks @sirblackheart! :beers:

PR #995, opened 2016-03-03
Native gtk file dialog

sirblackheart:
This pull request basicly adds the suport for gtk3 dialogs to she
it is disabled by default and can be enabled with
`cmake ... -DWITH_GTK_FILE_DIALOG_SUPPORT=ON ..`
at cmake configuration time and needs the experimental option "Use native file
dialog" in aseprite enabled. It is currently only available on linux and has a external dependencie on
gtkmm >= 3.x

I noticed, that current folder string is empty in https://github.com/aseprite/aseprite/blob/master/src/app/commands/cmd_open_file.cpp#L117
so i can not reproduce the behavior of the Aseprite dialog when opening a file. Aseprites own dialog stays in the same folder where the last file was opened.

And i am sorry for the wrong variables in my last pull request, normally i test my branch in Linux, Windows and OS X, but because of some manual reading of changes i made the mistakes and forgot to test again on Windows. Again i am sorry for the inconvenience.


dacap:
I'll make some comments in the code in a moment. The only thing that will require a lot of testing before I merge this PR are the dependencies (which are really hard to solve/test on `.deb` files for 32-bit, 64-bit, and Steam)

> I noticed, that current folder string is empty in https://github.com/aseprite/aseprite/blob/master/src/app/commands/cmd_open_file.cpp#L117
> so i can not reproduce the behavior of the Aseprite dialog when opening a file. Aseprites own dialog stays in the same folder where the last file was opened.

Maybe a static/global `std::string g_lastUsedDir` might solve this.

> And i am sorry for the wrong variables in my last pull request, normally i test my branch in Linux, Windows and OS X, but because of some manual reading of changes i made the mistakes and forgot to test again on Windows. Again i am sorry for the inconvenience.

No problem :+1:

sirblackheart:
Well i just tried to get this running on Ubuntu 14.04 and it looks rather bad. This will probably be a 16.04 onward feature only.

sirblackheart:
Here some preview of the current state on gnome and ubuntu 15.10. What is probably still disturbing, is that aseprites menu stays open while the dialog is open. And like the last commit mentiones, further testing would have to be done how far the glib support is backward compatible. As a next stepp a preview widget would probably be nice.
file_chooser_gnome
native_ubuntu1510

src/she/CMakeLists.txt
```diff
@@ -97,7 +97,7 @@ if(USE_SKIA_BACKEND)
       find_library(SKIA_OPENGL_LIBRARY opengl32)
       find_library(SKIA_LIBEGL_LIBRARY libEGL.dll PATH "${SKIA_BUILD_OUT_DIR}")
       if(SKIA_LIBEGL_LIBRARY)
-	add_definitions(-DSK_ANGLE=1)
+        add_definitions(-DSK_ANGLE=1)
```
dacap: This line wasn't modified (more than indentation)

src/she/native_dialogs.h
```diff
@@ -25,6 +25,7 @@ namespace she {
     virtual std::string fileName() = 0;
     virtual void setFileName(const std::string& filename) = 0;
     virtual bool show(Display* parent) = 0;
+    static std::string& g_lastUsedDir() { static std::string lastUsedDir{""}; return lastUsedDir; }
```
dacap: Also a std::string doesn't need to be initialized with an empty string (as it's already initialized empty).

src/she/CMakeLists.txt
sirblackheart: like noted in the commit this was a tab removed in a file using spaces

dacap: Oh sorry, I was reviewing the "Files changed" tab (it's like a squashed version of all commits in the PR)

PR #1000, opened 2016-03-04
Experimental native gtk3 file dialog

sirblackheart:
Ok so recap of the situation:
We have now a second cmake flags so this compiles and runs on at least ubuntu 15.10.
At this point i have to ask what the support goal of aseprite will be. As for now ubuntu 15.04
has reached its end of life in February and ubuntu 14.04 will reach its end of life ~~in august~~ on 26. April 2017. But on the other hand we should have ubuntu 16.04 available soon. Which is the next LTS version.
I had the dialog up and running in an alpha version and i did not need the flag for older glib support.

I still have to test this at least on Fedora 22, Debian 7 and Steam OS probably i guess.

On the usability level i already mentioned that the menu stays visible during the period the dialog is open, but that seems to be also the case with the other native dialogs. At least i tested it with the os x one.

Also the positioning is not perfect yet as it just centers on the screen, alternatively it could be positioned near the cursor. Better positioning could be probably achieved by calculations dependent on the absolute position of the aseprite window or even better a proper "parent".

Like i mentioned in the previous pull request it would be nice to have a current dir variable set by the app itself on file opening/saving operations, at least when it makes sense, like when the user opens one of his recent files from the start screen/tab.

The next step here is probably to look into the possibility to write a preview widget. The provided FileChooser Widget used here already has the interfaces to add one.

### DIFF
--- a/src/she/CMakeLists.txt
+++ b/src/she/CMakeLists.txt
@@ -222,6 +222,5 @@ endif()
 
 if(WITH_GTK_FILE_DIALOG_SUPPORT)
   target_link_libraries(she
-    ${GTKMM_LIBRARIES}
-    app-lib)
+    ${GTKMM_LIBRARIES})
 endif()
--- a/src/she/CMakeLists.txt
+++ b/src/she/CMakeLists.txt
@@ -222,5 +222,6 @@ endif()
 
 if(WITH_GTK_FILE_DIALOG_SUPPORT)
   target_link_libraries(she
-    ${GTKMM_LIBRARIES})
+    ${GTKMM_LIBRARIES}
+    app-lib)
 endif()
--- a/src/she/CMakeLists.txt
+++ b/src/she/CMakeLists.txt
@@ -97,7 +97,7 @@ if(USE_SKIA_BACKEND)
       find_library(SKIA_OPENGL_LIBRARY opengl32)
       find_library(SKIA_LIBEGL_LIBRARY libEGL.dll PATH "${SKIA_BUILD_OUT_DIR}")
       if(SKIA_LIBEGL_LIBRARY)
-	add_definitions(-DSK_ANGLE=1)
+        add_definitions(-DSK_ANGLE=1)
       endif()
     endif()
   else()
@@ -193,6 +193,15 @@ if(APPLE)
     osx/native_dialogs.mm)
 endif()
 
+if(WITH_GTK_FILE_DIALOG_SUPPORT AND UNIX AND NOT APPLE AND NOT BEOS)
+  if(WITH_DEPRECATED_GLIB_SUPPORT)
+    add_definitions(-DASEPRITE_DEPRECATED_GLIB_SUPPORT)
+  endif()
+  add_definitions(-DASEPRITE_WITH_GTK_FILE_DIALOG_SUPPORT)
+  list(APPEND SHE_SOURCES
+    gtk/native_dialogs.cpp)
+endif()
+
 add_library(she ${SHE_SOURCES})
 
 target_link_libraries(she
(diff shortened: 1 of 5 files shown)


sirblackheart:
Short update on running this on other distros:

- Running fine:
  - Ubuntu >= 16.04
  - ArchLinux
- Running fine with glib flag:
  - Ubuntu 15.10
  - Fedora 22
- Running with glib flag but into problems:
  - Debian 8.3 (The dialog does not disappear on any action like open or cancel, i had this problem on newer versions when the dialog was not running in the context of a gtk app)
- Not Running:
  - Ubuntu <= 14.04
  - Debian 7

src/she/CMakeLists.txt
```diff
@@ -210,3 +219,9 @@ if(USE_SKIA_BACKEND)
   target_link_libraries(she
     ${SKIA_LIBRARIES})
 endif()
+
+if(WITH_GTK_FILE_DIALOG_SUPPORT)
+  target_link_libraries(she
+    ${GTKMM_LIBRARIES}
+    app-lib)
```
dacap: This is circular dependency: `app-lib` depends on `she`, so `she` cannot depend on `app-lib`. In case we need the thumbnail generator in the `she` side, we should think about a interface that can provide `she` and must be implemented by the `app` side to generate thumbnails.

sirblackheart:
Here some thoughts on the whole thumbnail situation:
There are two possibilities and maybe a combination of both.
Use aseprites own thumbnailer for preview, which is already working with this, lets call it proof of concept. The upside of that is, that we have a better what you see is what you get preview. Also we have support for all formats that aseprite supports. The downside is the relative high loading time when first generating the thumbnail before we can access the cached version. Currently the UI is blocked when loading, but this could be changed. Still aseprites thumbnail loader is slower then just loading a png directly into Gdk::pixbuf.
Which gets us to the alternative approach. Using the built in gtk::image and gdk::pixbuf support we get fast preview and thumbnails for image formats supported by the gdk version the distribution provided.  Also we would get easy animation support for gifs and similar. The downside of this approach is that this excludes exotic formats and aseprites own format. It should possible to write own pixbuf loaders for gdk tough.
So one approach for now could be to use gtks native previews for a fast preview and only rely on aseprites own thumbnailer if the system does not support the file format. And for a later project it could be looked into getting together a "libase" which could be used to put together a gdk pixbuf loader but also for a gnome thumbnailer to get ase thumbnails in nautilus.
What is you opinion on this?

dacap:
I think that we should go for the native thumbnails approach.

There are three major refactors I'd like to do [after v1.1.x series](https://github.com/aseprite/aseprite/milestones/v1.1):

1. Move [code to load/save `.ase` images](https://github.com/aseprite/aseprite/blob/master/src/app/file/ase_format.cpp) into a special library (`libase` or something like that)
1. Move [all code to load/save all supported formats](https://github.com/aseprite/aseprite/tree/master/src/app/file) into [`iff`](https://github.com/aseprite/aseprite/tree/master/src/iff) library (which will use `libase`).
1. Merge `base`, `gfx`, `she`, `ui`, and maybe other modules into a common library to create portable desktop applications (it's probable that `she` will be renamed to `low`, and everything will be inside a new repository/namespace named [`laf`](https://github.com/aseprite/laf))

The last point is a key component. My long term goal with Aseprite is to create a C++ library to develop applications. And inside this library should be a component to create thumbnails in a portable way (Windows Explorer, OS X Finder, and Linux file manager/gdk).

In the best case scenario, all the code inside `app` should be the only code for Aseprite. And it should provide the glue between `libase` and `laf`'s thumbnail generator.

sirblackheart:
Ok i reverted this to basic gtk image opening for now and will open a seperate bug/feature to write down some notes on thumbnail generation on linux.

dacap:
Thanks for your help Gabriel, I'll merge this PR. I didn't test it on Linux yet, but as it's disabled by default, it doesn't look like an issue to already have this merged.

PR #1016, opened 2016-03-14
Added missing includes: stdio.h for rename(2) and limits.h for PATH_MAX

yurivict:
Spotted on FreeBSD 9.2 while compiled with gcc.

### DIFF
--- a/src/base/fs_unix.h
+++ b/src/base/fs_unix.h
@@ -8,6 +8,8 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <stdio.h>
+#include <limits.h>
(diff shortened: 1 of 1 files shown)


src/base/fs_unix.h
```diff
@@ -8,6 +8,8 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <stdio.h>
+#include <limits.h>
```
dacap: Could you please try `<cstdio>` and `<climits>` instead of the C version of these header files (and group them below with `<cstdlib>` and `<ctime>`) to see if they work on FreeBSD?

Issue #901, opened 2015-12-14
Design a more intuitive CLI


dacap:
Right now there are several CLi options that do two different things depending on the context (e.g. `--list-tags` alone will print tags in the stdout, and used with `--data` will list the tags in the JSON file).

It would be a good idea to redesign the CLI with a more "command like" approach:

    aseprite command_name [--modifiers] [files]

dacap:
Anyway remember that #126 will be the preferred way in the future to export sprites. (And the CLI could be something much more simpler like `aseprite --batch mygame.aseproject`)

TheKK:
Do you mean after the project features are completed, we could set preference (e.g. output format and commands to run) for output and use only `aseprite --batch xxx.aseproject` to perform it?

dacap:
With a project, we should be able to see the list of sprites in our project folder(s), right click one file/sprite, add specific export actions for that file (save as/export sprite sheet), and the export should be added to a "list of tasks." Finally, this "list of tasks" could be processed with one "compile/run" button in the UI, or via `--batch projectname`. Note: It should work like `make` comparing modified file dates, and we might have a `--force` option to regenerate everything.

PR #1030, opened 2016-03-23
New cmdl parameters --range and --shrink-to

sirblackheart:
The main reason to add these is to make it possible to use aseprites batch mode for xdg thumbnailing.
As both can probably be useful also in other scenarios i tought it would be nice to add them as command line parameters instead of just exposing a --xdg-thumbnail option.

--frame-range <from:to> lets the user define a range from frame to frame
instead of a frame tag. example --frame-range "0:1" exports 2 frame. frame 0
and frame 1.

--shrink-to <widthxheight> shrinks the sprite into the boundarys of
width and height keeping the original aspect ratio.


dacap:
It looks good to me. Anyway I'll make some tests after merging it.

dacap:
I will change `--shrink-to widthxheight` to `--shrink-to width,height` and the same for `--frame-range from,to` (as `--crop x,y,width,height`)

dacap:
As a side note: I think that, in a near future, we could have a:

    aseprite thumbnail file.ase file.png

command option to avoid loading all frames from `.ase` files to generate a thumbnail. Related to #901

PR #864, opened 2015-11-15
Fixed ftglyph.h not found when using external FreeType2

Mailaender:
For @openSUSE Linux I have to set `-DFREETYPE_INCLUDE_DIR=%{_includedir}/freetype2` so freetype is incorrect.

PR #1035, opened 2016-03-26
Cmake Imporvements

winterheart:
These commits intended to fix some strange compilation error I've met on Linux.

- Using `find_package(FOO REQUIRED)` for system libraries where is possible (`REQUIRED` will fail configuration if library is not found, so `message(FATAL_ERROR ...` is rendurant)
- Replaced `FOO_LIBRARY` to `FOO_LIBRARIES` and `FOO_INCLUDE_DIR` to `FOO_INCLUDE_DIRS` for `find_package` invocation as should be by cmake design (fixes issue #864)
- Added system libcurl linking if used system library.


dacap:
Hi @winterheart, would you like to [sign our CLA](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla)?

winterheart:
Sure, why not. Sent request.

src/app/CMakeLists.txt
```diff
@@ -446,6 +446,7 @@ target_link_libraries(app-lib
   ui-lib
   undo-lib
   ${TINYXML_LIBRARY}
+  ${CURL_LIBRARIES}
```
dacap: This is the only change I think it's not needed (because `net-lib` already depends on `${CURL_LIBRARIES}` in the next commit). I'll merge all the changes and revert this commit to see what happens. Please tell me if it's needed again for some specific scenario.

dacap:
Thanks for your contributions @winterheart!

CMakeLists.txt
```diff
@@ -218,26 +218,26 @@ include_directories(${PIXMAN_INCLUDE_DIR})
...
 endif()
-include_directories(${GIF_INCLUDE_DIR})
+include_directories(${GIF_INCLUDE_DIRS})
 
 if(USE_SHARED_JPEGLIB)
   find_package(JPEG REQUIRED)
 else()
   set(JPEG_LIBRARIES jpeg)
-  set(JPEG_INCLUDE_DIR ${LIBJPEG_DIR})
+  set(JPEG_INCLUDE_DIRS ${LIBJPEG_DIRS})
```
dacap: I just saw that this should be `LIBJPEG_DIR` instead of `LIBJPEG_DIRS`, I'll push a fix.

PR #1042, opened 2016-03-29
Make building tests optional

winterheart:
Added new option ENABLE_TESTS in CMake (ON by default)


cmake/Find-tests.cmake
```diff
@@ -0,0 +1,41 @@
+# - Find tests
+# Find tests and run them
+#
+#  find_test(dir dependencies)
+#  DXGUID_FOUND       - True if dxguid found.
```
dacap: You can remove this line (looks like a copy & paste from `FindDXGuid.cmake`)

winterheart: Done.

dacap:
Thanks @winterheart!

PR #1045, opened 2016-03-31
VGA 13h standard palette

### LINKED ISSUE #1020: Feature request: add the default vga mode 13h palette as a preset
I think that would fit right in. I found a list of hex values here: https://commons.wikimedia.org/wiki/User:Psychonaut/ipalette.sh

winterheart:
Fixes #1020


dacap:
Thanks for this @winterheart

PR #1048, opened 2016-04-04
add linux thumbnailers for Gnome & KDE/Qt

### LINKED ISSUE #1031: Add aseprite mimetype
This would be needed so aseprite can be associated with its own files in the aseprite.desktop file as well as with thumbnail-generators. Not sure if you want to have this in the main repo or rather just added to the distributed package as with the aseprite.desktop file itself like mentioned in #952 .

Attached is an example aseprite.xml (i had to make it a txt so i could drag and drop it) of what this mimetype file could look like. It would have to be installed into "$XDG_DATA_DIRS/mime/packages/aseprite.xml".  $XDG_DATA_DIRS normaly defaults to "/usr/share". As well as a call to update-mime-database usr/share/mime, but that has to be done by the packager.

[aseprite.txt](https://github.com/aseprite/aseprite/files/186937/aseprite.txt)

sirblackheart:
This adds thumbnailer support both for Gnome and KDE5/Qt5. The gnome
thumbnailer is installed by default on linux systems, as it does not add
any additional dependencies at build time and can be used by
GnomeThumbnailFactory. The KDE/Qt thumbnailer need the whole Qt and Kio
stack to build, as well as the distribution dependend variables set.
Because of that it is not build by default. For the thumnailers to work
for all of aseprites supported file formats and additonal mime type file
is installed. Packagers in addition need to call update-mime-database
path-to-mime-database (/usr/share/mime for example) as post install
command.

I created the folder desk as short form for desktop in reference for desktop integration for now, if you prefer misc or install i can rename it.

Also closes #1031


desk/kde/aseprite_thumb_creator.h
```diff
@@ -0,0 +1,16 @@
...
+
+#include <QLoggingCategory>
+Q_DECLARE_LOGGING_CATEGORY(LOG_ASEPRITE_THUMBCREATOR)
+
+#include <kio/thumbcreator.h>
+
+class AsepriteThumbCreator : public ThumbCreator {
+  public:
+    virtual bool create(const QString& path, int width, int height, QImage& img);
+    virtual Flags flags() const;
```
dacap: You can remove `virtual` and add `override` at the end in these two lines.

dacap:
Hi @sirblackheart, I'll review this tomorrow. Would you like to rename the folder to `desktop`?

desktop/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+# Aseprite Desktop Integration Module
+# Copyright (C) 2016  Gabriel Rauter
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 2 as
+# published by the Free Software Foundation.
```
dacap: About this Gabriel, I'll re-license this to MIT in a near future in case this code is moved to laf library: https://github.com/aseprite/laf

sirblackheart:
I hope this solves the licencing situation. I forgot to mention, the kde/qt thumbnailer also needs the cmake [ECM Moduel](http://api.kde.org/ecm/manual/ecm.7.html). This should be the extra-cmake-modules package on ubuntu.

dacap:
You can just use MIT license if you want, no deed for dual license in that case. (Note: I'm not sure if I'll be able to test this until next week, as I'd like to release a new update this week.)

sirblackheart:
I tested the gnome thumbnailer under archlinux and opensuse tumbleweed with both running gnome-shell and the whole gnome stack so far. The kde thumbnailer has been tested on the current beta of kubuntu. At least on kubuntu cmake needs the paramter `-DKDE_INSTALL_USE_QT_SYS_PATHS=on` so the correct `${PLUGIN_INSTALL_DIR} `variable for the kde thumbnailer is set.

dacap:
Thanks @sirblackheart I'll merge this and disable the integration by default, just in case. We can enable it later.

PR #1060, opened 2016-04-13
Clarified instructions for downloading source code

moogDE:
Previous wording did not clearly demonstrate the need to do a recursive clone. It tripped me up. This new wording should prevent others from making the same mistake I did.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -21,12 +21,7 @@ platforms:
 
 # Get the source code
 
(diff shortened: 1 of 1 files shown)


dacap:
Hi @moogDE, thanks for the PR. Anyway I modified that section with a more accurate description (e.g. now we can just download the source code from a zip file). See 392eacdb6ad453eb68ff253a94aa9267b1f1b194

PR #1062, opened 2016-04-13
Fix error when decoding GPL file

RomainGiraud:
Color without name are not added to the palette.

`fstream::good()` method should not be used to test stream, use `fstream::fail()` method instead ([reference](http://www.cplusplus.com/reference/ios/ios/good/)).


dacap:
Hi @RomainGiraud, would you like to [sign our CLA](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla)?

dacap:
Thanks for the contribution @RomainGiraud

PR #1072, opened 2016-04-18
Compilation fixes

winterheart:
- Fixes compilation on Linux when ENABLE_MEMLEAK is ON
- Added freetype to shared libraries for Travis
- CMake optimizations


dacap:
Hi @winterheart, thanks for these changes. They look great. I'll make a comment just in one line...

src/gen/CMakeLists.txt
```diff
@@ -15,4 +15,5 @@ endif()
 
 target_link_libraries(gen
   base-lib
-  ${TINYXML_LIBRARY})
+  ${TINYXML_LIBRARY}
+  ${CMAKE_THREAD_LIBS_INIT})
```
dacap: I think we should add `target_link_libraries(base-lib ${CMAKE_THREAD_LIBS_INIT})` in the `base/CMakeLists.txt` instead of adding it here. (So any application linked with base-lib will be linked with threads libs.)

winterheart: Yes, you're right.

winterheart:
Well, I think, MEMLEAK feature totally broken. I can't build it successfully and pass all tests.
I'll create another PR which contain only CMake refactoring.

PR #1076, opened 2016-04-19
Cmake optimization

winterheart:
* Added freetype to shared libraries for Travis
* CMake optimizations


dacap:
Thanks @winterheart!

PR #1082, opened 2016-04-22
Fixing underlinking for gcc 4.9

winterheart:
Allegro4 backend uses XGrabPointer() from libX11.
Added ${X11_LIBRARIES} for she to resolve underlinking.


dacap:
Thanks @winterheart

Issue #1086, opened 2016-04-23
CMake Big Endian Test fail on Windows


sarcasmloading:
Specs:
Windows 10 Pro 64bit
VS 2015 x64 (installed: common tools for visual c++ 2015 and windows 10 sdk)
CMake 3.5.2
The C/C++ compiler MSVC version is MSVC 19.0.23918.0
Compiling using the Allegro backend, because Skia is hell

I'm trying to compile Aseprite 1.1.4.1 on Windows, and I get this:
`
CMake Error at C:/Program Files (x86)/CMake/share/cmake-3.5/Modules/TestBigEndian.cmake:51 (message):
  no suitable type found
`

I've tried the GUI (admin and no admin), command prompt, VS Dev Command Prompt (vs2015), all to no avail.

I've googled it so many times and all of the answers I've found just say "oh, just empty the build dir". (I've done that every single time)

dacap:
cc @winterheart

winterheart:
Hello, can you please post CMakeError.log and CMakeOutput.log from BUILDDIR/CMakeFiles?

sarcasmloading:
Yep.
CMakeError.log: http://pastebin.com/raw/wRvebWCD
CMakeOutput.log: http://pastebin.com/raw/vu2kLnbw

winterheart:
OK, I think I found what's wrong. Unfortunately, I don't have full Windows build environment, @sarcasmloading, can you check it for me? Try to remove from 
https://github.com/aseprite/aseprite/blob/master/src/CMakeLists.txt#L33 "/SUBSYSTEM:WINDOWS,5.01" substring. It should look like:

```
  if (CMAKE_CL_64)
    # 64 bits
    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /MACHINE:X64")
  else()
    # Add support for Windows XP with 5.01 subsystem
    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /MACHINE:X86")
  endif()
```

@dacap, is Aseprite still supports Windows XP? If yes, then we need move all compilation checks from src/base/CMakeLists.txt to root CMakeLists.txt.

sarcasmloading:
No, that didn't work, sadly. I'm running Windows 10.

winterheart:
Did you cleared cache? Even in Win64 profile still no works?

sarcasmloading:
[...]

PR #1088, opened 2016-04-25
Fix compilation on MSVC for 32bit profile.

winterheart:
Option "/SUBSYSTEM:WINDOWS,5.01" cause failing compilation checkings.
It is near to impossible compile current cmake-project on Windows XP,
and since this statement doing nothing else than CMake can do
themself, let's just delete them.

Tested on Windows 10 / MSVC 2015 on 32 and 64 bit profiles. Fixes
issue #1086.


dacap:
Thanks for this @winterheart

The only thing I'm still thinking about this is why I wasn't able to reproduce this problem. I'll check later in my production VM to see if something else is needed for the WinXP compilation, but I think it should work as it is.

Issue #1089, opened 2016-04-25
Help with compilation on OS X


zwik:
Greetings,

I read the install.md file but I'm getting compilation errors. When on master I get errors when running `cmake -G Ninja ..`

The error I get is:

> CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> Please set them or make sure they are set and tested correctly in the CMake files:
> SKIA_CONFIG_INCLUDE_DIR
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
> SKIA_CORE_INCLUDE_DIR
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
>    used as include directory in directory /Users/zwik/Documents/code/aseprite/src/she
> SKIA_CORE_LIBRARY
>     linked by target "she" in directory /Users/zwik/Documents/code/aseprite/src/she
[...]

Issue #1081, opened 2016-04-22
OS X needs more instructions for building with skia


tony:
Greets,

Following ``INSTALL.md`` @ https://github.com/aseprite/aseprite/blob/8775604e5629bb73d9133e41b46efe3f11d48334/INSTALL.md#mac-os-x-details with the additions. Here is my command:

```
cmake -D "CMAKE_OSX_ARCHITECTURES:STRING=x86_64"  -DUSE_SKIA_ALLEG4=OFF -DUSE_SKIA_BACKEND=ON -DSKIA_DIR=$HOME/study/c++/skia \
-D "CMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.7" \
-D "CMAKE_OSX_SYSROOT:PATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk" \
-D "WITH_HarfBuzz:BOOL=OFF" ..
```

Here is what I get:

```
- Building with PNG
-- Creating directory /Users/me/work/c++/aseprite/build/third_party/freetype2/include/freetype/config
-- Creating file /Users/me/work/c++/aseprite/build/third_party/freetype2/include/freetype/config/ftconfig.h
-- Creating file /Users/me/work/c++/aseprite/build/third_party/freetype2/include/freetype/config/ftoption.h
-- Check if the system is big endian
-- Searching 16 bit integer
-- Using unsigned short
-- Check if the system is big endian - little endian
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
SKIA_CORE_LIBRARY
    linked by target "she" in directory /Users/me/work/c++/aseprite/src/she
    linked by target "she" in directory /Users/me/work/c++/aseprite/src/she
SKIA_EFFECTS_LIBRARY
    linked by target "she" in directory /Users/me/work/c++/aseprite/src/she
SKIA_IMAGES_LIBRARY
    linked by target "she" in directory /Users/me/work/c++/aseprite/src/she
SKIA_OPTS_AVX2_LIBRARY
    linked by target "she" in directory /Users/me/work/c++/aseprite/src/she
SKIA_OPTS_AVX_LIBRARY
    linked by target "she" in directory /Users/me/work/c++/aseprite/src/she
SKIA_OPTS_LIBRARY
    linked by target "she" in directory /Users/me/work/c++/aseprite/src/she
[...]

PR #1097, opened 2016-04-26
Update installation instructions for Skia and OS X

tony:
See also: #1089 #1081

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,6 +9,8 @@
   * [Issues with Retina displays](#issues-with-retina-displays)
 * [Using shared third party libraries](#using-shared-third-party-libraries)
   * [Linux issues](#linux-issues)
+* [Building Skia dependency](#building-skia-dependency)
+  * [Linux, OS X, BSDs](#linux-os-x-bsd)
 
 # Platforms
 
@@ -68,13 +70,15 @@ Aseprite can be compiled with two different back-ends:
 
 ## Linux dependencies
 
-You will need the following dependencies:
+You will need the following dependencies (Ubuntu, Debian):
 
     sudo apt-get update -qq
     sudo apt-get install -y g++ libx11-dev libxcursor-dev cmake ninja-build
 
 The `libxcursor-dev` package is needed to
 [hide the hardware cursor](https://github.com/aseprite/aseprite/issues/913).
+
+Aseprite uses Ninja as the build system on POSIX-like platforms.
 
 # Compiling
 
@@ -95,7 +99,7 @@ the repository clone in a directory called `aseprite`):
 
         C:\aseprite>cd build
 
-   If you have ninja:
+   On Linux / OS X / BSD:
 
         C:\aseprite\build>cmake -G Ninja ..
 
@@ -106,10 +110,6 @@ the repository clone in a directory called `aseprite`):
    If you have Visual Studio you can generate a solution:
 
         C:\aseprite\build>cmake -G "Visual Studio 12 2013" ..
-
-   If you are on Linux:
-
-        ~/aseprite/build$ cmake -G "Unix Makefiles" ..
 
    For more information in [CMake wiki](http://www.vtk.org/Wiki/CMake_Generator_Specific_Information).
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -4,6 +4,7 @@
 * [Get the source code](#get-the-source-code)
 * [Dependencies](#dependencies)
   * [Linux dependencies](#linux-dependencies)
+* [Building Skia dependency](#building-skia-dependency-os-xlinuxbsd)
```
dacap: I'm not planning to add a section to build Skia, sorry

INSTALL.md
```diff
@@ -71,11 +72,49 @@ Aseprite can be compiled with two different back-ends:
...
     sudo apt-get update -qq
-    sudo apt-get install -y g++ libx11-dev libxcursor-dev cmake ninja-build
+    sudo apt-get install -y g++ libx11-dev libxcursor-dev cmake
+    
+Optionally, install ninja for parallelized builds: 
+
+    sudo apt-get install ninja-build
 
 The `libxcursor-dev` package is needed to
 [hide the hardware cursor](https://github.com/aseprite/aseprite/issues/913).
```
dacap: This paragraph should be after the `sudo apt-get install libxcursor-dev`

INSTALL.md
```diff
@@ -71,11 +72,49 @@ Aseprite can be compiled with two different back-ends:
 You will need the following dependencies:
 
     sudo apt-get update -qq
-    sudo apt-get install -y g++ libx11-dev libxcursor-dev cmake ninja-build
+    sudo apt-get install -y g++ libx11-dev libxcursor-dev cmake
+    
+Optionally, install ninja for parallelized builds:
```
dacap: We could put here something like "We use Ninja as the preferred build system, you can install it using:".

INSTALL.md
```diff
@@ -134,14 +173,45 @@ the repository clone in a directory called `aseprite`):
...
+```
+
+run
 
-    -D "CMAKE_OSX_ARCHITECTURES:STRING=x86_64"
-    -D "CMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.7"
-    -D "CMAKE_OSX_SYSROOT:PATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk"
-    -D "WITH_HarfBuzz:BOOL=OFF"
-    -D "SKIA_DIR:PATH=/directory-which-contains-a-compiled-skia"
+`bin/aseprite`
```
dacap: I'm not sure about this section. It adds a lot of noise.

INSTALL.md
tony: Why not? Its a source of confusion when building the project. The library dependency, unlike allegro, isn't available through package managers. It also requires a third party tool (depot tools).

In lieu of that, is linking to the installation instructions on the Skia website OK?

INSTALL.md
tony: I will consolidate this section.

I don't like the noise either - I'd rather keep it brief (because adding this also means there's more to update if a detail changes). I'm trying to balance that and potential ambiguity for newcomers - some of which are not familiar with build systems like CMake.

dacap:
This issue brings me a lot to think about. In one side we could explain all the possibilities (e.g. "you can use make, or ninja, or vs projects, etc.") and all the steps needed to use each alternative. On the other side we could just give a unique list of steps for "non-programmers".

What I think about this? It's quite a tricky situation.

Looking at the amount of recent issues reported to compile Aseprite, it seems that a very low percentage of (C++) programmers are trying to compile Aseprite. We're having a lot of noise with compilations issues.

At the moment, my preference is that INSTALL.md should be focused to programmers.

INSTALL.md
dacap: >  is linking to the installation instructions on the Skia website OK?

Yes, that's is what I think we should do. Linking to the official instructions. (In that way we don't need to keep them in sync)

tony:
@dacap heh, yes. Ideally the issues should be for development related tasks (being a tracker), not Q&A. You wouldn't but a Q&A on bugzilla. 

I've seen Google Groups used as a place to divert discussion for that.

You can use GitHub wiki for community-based documentation. skia keeps per-platform instructions in their repo on their website: https://github.com/google/skia/tree/master/site/user/quick

> Looking at the amount of recent issues reported to compile Aseprite, it seems that a very low percentage of (C++) programmers are trying to compile Aseprite. We're having a lot of noise with compilations issues.

Someone being a C++ programmer, whatever standard that would be to you, wouldn't make everything inherently obvious and unambiguous. I had to go on an expedition to figure out why skia wouldn't build correctly for me, that the CMake can't be used to build without GPU support. So that I had to go grab depot_tools.  You could blame it on me seeing a ``CMakeLists.txt`` and assuming it works - and not reading [cmake/readme.md](https://github.com/google/skia/blob/chrome/m50/cmake/README.md) that says its only experimental.

Incorporate skia adds a big step to building aseprite, probably bigger than configuring and building aseprite itself. I think the increase in the issues is caused by the skia migration - and that docs are a way to decrease the noise.

dacap:
> I had to go on an expedition to figure out why skia wouldn't build correctly for me

Yes, me too, its build system is the most tricky part. Actually the only requirement for Aseprite is that `GYP_DEFINES` must be `skia_gpu` when you compile Skia. Anyway I've compiled Skia so many times that I don't even remember the original instructions or in which state is the official documentation, so my opinions are really biased.

> Incorporate skia adds a big step to building aseprite, probably bigger than configuring and building aseprite itself. I think the increase in the issues is caused by the skia migration - and that docs are a way to decrease the noise.

Yeah. I was having the idea to clean up the whole `INSTALL.md` when I completely remove the Allegro back-end, but seeing the current state, I think you are right, we should add the "how to compile Skia" section in INSTALL (at least to reduce the number of future issue reports).

Could you please move the "skia" section at the end and add a subsection for OSX/Linux,etc? I'll write the Windows instructions then.

dacap:
Also I'm seriously thinking that Ninja should be almost mandatory (if someone wants to use other build system, is because he/she knows about CMake)

tony:
> Also I'm seriously thinking that Ninja should be almost mandatory (if someone wants to use other build system, is because he/she knows about CMake)

I concur, I'll update to reflect that.

tony:
@dacap how about now?

dacap:
It's fine. I'll continue from here and see if I can make this even more simple. Thanks for the help @tony

tony:
yeoopo

dacap:
@tony I've just updated the INSTALL.md instructions with cab10fba1e1ffaed218cfb3856d70b1b5fad5c4d if you can give it a look that would be great, I hope this helps reducing the number of issue reports

PR #1115, opened 2016-05-04
Fixed an unnecessary quotation mark in the OSX building instructions.

dvlpstrcrispin:
In line 168, there was a quotation mark at the end of the line that was causing Bash to expect the user to finish the string, not letting it finish reading the command and compile. It works fine without it, and seems to be a typo.


dacap:
Thanks for this @dvlpstrcrispin

Issue #1134, opened 2016-05-14
Duplicate cel(s) option replaces existing cel


kinow:
Hi, 

If I have three cels, and duplicate the first frame, it will replace the second cel. Instead of replacing the cel, I'd prefer the program to duplicate the first cel and move the other cels to the right.

![duplicate_replacing](https://cloud.githubusercontent.com/assets/304786/15265077/5680b92a-19d0-11e6-8f9d-9d52c2fcee9f.gif)

dacap:
Hi @kinow, actually this is an expected behavior, you can add frames to displace cels. The duplication is done just copying cels at the right side.

PR #1135, opened 2016-05-14
Issue #1134 copy/move existing cels before duplicating

kinow:
Hi, 

Been a loooong time since I did any C or C++ coding, but decided to start learning again. So chose my favorite tools like aseprite, krita and inkscape to start practicing by reading the code.

Yesterday I wanted to duplicate a cel, but it was replacing another existing cel. The issue has been reported in #1134, and this pull request is a first tentative at fixing it :-)

Started learning about aseprite's code base yesterday, so I may have missed something. I'm copying one cel each time, but not sure if that performs well when you have lots of cels. Also still learning about continuous cels.

Thanks for aseprite! And for blogging about C++
Bruno

### DIFF
--- a/src/app/commands/cmd_new_frame.cpp
+++ b/src/app/commands/cmd_new_frame.cpp
@@ -124,6 +124,14 @@ void NewFrameCommand::onExecute(Context* context)
           for (LayerIndex layer = layerBegin; layer <= layerEnd; ++layer) {
             Layer* layerPtr = writer.sprite()->indexToLayer(layer);
             if (layerPtr->isImage()) {
+              LayerImage* layerImagePtr = static_cast<LayerImage*>(layerPtr);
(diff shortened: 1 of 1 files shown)


kinow:
Before:

duplicate_replacing

After (debugging locally in Eclipse + Ubuntu 16.04):

duplicate_moving

dacap:
Hi @kinow thanks for giving a try to this. Anyway I would prefer the current behavior as it is (it might have a configuration setting in the future or more commands to displace cels, for people who want to displace cels instead of replacing.) As I said before, you can add/duplicate frames using `Frame > New Frame` menu (`Alt+N` keyboard shortcut).

kinow:
Not a problem, thanks for looking into this @dacap. Will take a look at the new frame option. Thanks!

PR #1140, opened 2016-05-21
Fixed a serious compiler warning

Mailaender:
Fixes compilation for https://build.opensuse.org/request/show/397276


dacap:
About this one. The idea is that `is_same_pixel` shouldn't be compiled in the generic case (it's specialized for every type `ImageTraits`). What compiler is giving the warning? Does it give a warning for [convert_to](https://github.com/aseprite/aseprite/blob/master/src/base/convert_to.h#L21) too?

Mailaender:
It happens with GCC. Seems to be a simple static analysis. `bool` without a `return` is then "program returns random data" no matter if it actually does it. Changing the function to `void` didn't compile. This was the only severe warning that was triggered. I had to fix it, to continue with the rpm build.

dacap:
OK, I'll merge it, I guess `convert_to` doesn't require the same `return` statement because its return value is a template parameter.

Issue #698, opened 2015-06-10
undefined reference to `app_main(int, char**)' on Fedora 22


martinkg:
compiling the recent git checkout, that fails on Fedora 22 with the following error:

Linking CXX executable ../bin/errno_tests
cd /home/martin/rpmbuild/BUILD/aseprite-f310a51/build/src && /usr/bin/cmake -E cmake_link_script CMakeFiles/errno_tests.dir/link.txt --verbose=1
/usr/bin/c++   -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector-strong --param=ssp-buffer-size=4 -grecord-gcc-switches  -m64 -mtune=generic -pthread -Wall -Wno-switch -std=gnu++11 -O2 -g -DNDEBUG   -Wl,-z,relro  CMakeFiles/errno_tests.dir/app/errno_tests.cpp.o  -o ../bin/errno_tests -rdynamic ../lib/libgtest.so ../lib/libapp-lib.so ../lib/libfixmath-lib.so ../lib/libcfg-lib.so ../lib/libcss-lib.so ../lib/libdoc-lib.so ../lib/librender-lib.so ../lib/libscripting-lib.so ../lib/libundo-lib.so ../lib/libfilters-lib.so ../lib/libui-lib.so ../lib/libshe.so ../lib/libgfx-lib.so ../lib/libbase-lib.so ../lib/libupdater-lib.so ../lib/libsimpleini.so -ljpeg -lz -lpng -lz ../lib/libgiflib.so -ltinyxml -lloadpng -lpixman-1 -lcurl -L/usr/lib64 -lalleg -L/usr/lib64 -lalleg ../lib/libnet-lib.so -lpng -ltinyxml -lloadpng -lpixman-1 -lcurl -Wl,-rpath,/home/martin/rpmbuild/BUILD/aseprite-f310a51/build/lib 
../lib/libshe.so: undefined reference to `app_main(int, char**)'
collect2: error: ld returned 1 exit status
src/CMakeFiles/errno_tests.dir/build.make:119: recipe for target 'bin/errno_tests' failed
make[2]: *** [bin/errno_tests] Error 1

dacap:
Hi @martinkg, I've marked this issue as fixed. But if you find it's not fixed yet, you can reopen it.

martinkg:
have again the problem that this error occurs, on Fedora 22 with recent git.

[...]

PR #1145, opened 2016-05-24
Fix compilation with shared libraries (BUILD_SHARED_LIBS=ON)

winterheart:
There should be app_main() instead main() as defined in libshe.
Seems, this issue somehow related to #698.

(I know this case is uncommon, but enabling KDE/Qt thumbnailer automatically enables BUILD_SHARED_LIBS to ON).

### DIFF
--- a/src/base/CMakeLists.txt
+++ b/src/base/CMakeLists.txt
@@ -66,7 +66,7 @@ if(WIN32)
 endif()
 
 add_library(base-lib ${BASE_SOURCES})
-target_link_libraries(base-lib modp_b64)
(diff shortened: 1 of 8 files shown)


dacap:
Hi @winterheart, those tests shouldn't be compiled with the she library. I think we should fix that problem first.

winterheart:
Sorry for delay.
Maybe better to link components with explicit STATIC directive (i.e. `add_library(foo STATIC ...)`)?

dacap:
Yes, I think that might be a possible solution (to overwrite the given BUILD_SHARED_LIBS=ON), because our libraries must be compiled statically.

PR #1177, opened 2016-07-02
Pixly

zed9h:
First try at implementing the Pixly file format, just loading for now and I tested with only two files, but it could be useful to some people ([for instance this guy](https://trello.com/c/lyMh9uSc/13-import-psd-aseprite-gif-pcx)), so I am pushing right away.

### DIFF
--- a/src/app/file/pixly_format.cpp
+++ b/src/app/file/pixly_format.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2001-2016  David Capello
+// Copyright (C) 2016  Carlo "zED" Caputo
 //
 // This program is free software; you can redistribute it and/or modify
 // it under the terms of the GNU General Public License version 2 as
@@ -17,12 +17,14 @@
 #include "app/ini_file.h"
 #include "app/xml_document.h"
 #include "base/file_handle.h"
+#include "base/convert_to.h"
 #include "base/path.h"
 #include "doc/doc.h"
 
 #include <stdio.h>
 #include <stdlib.h>
 #include <math.h>
+#include <cctype>
 
 #include "png.h"
 
@@ -62,6 +64,36 @@ static void report_png_error(png_structp png_ptr, png_const_charp error)
 {
   ((FileOp*)png_get_error_ptr(png_ptr))->setError("libpng: %s\n", error);
 }
+
+template<typename Any> static Any* check(Any* a, Any* alt = NULL) {
+  if(a == NULL) {
+    if(alt == NULL) {
+      throw Exception("bad structure");
+    } else {
+      return alt;
+    }
+  } else {
+    return a;
+  }
+}
+
+template<typename Number> static Number check_number(const char* c_str) {
+  if(c_str == NULL) {
+    throw Exception("value not found");
+  } else {
+    std::string str = c_str;
+    if(str.empty()) {
+      throw Exception("value empty");
+    }
+    std::string::const_iterator it = str.begin();
+    while (it != str.end() && (std::isdigit(*it) || *it == '.')) ++it;
+    if(it != str.end()) {
+      throw Exception("value not a number");
+    }
+    return base::convert_to<Number>(str);
+  }
+}
+
 
 bool PixlyFormat::onLoad(FileOp* fop)
 {
@@ -158,7 +190,7 @@ bool PixlyFormat::onLoad(FileOp* fop)
       break;
 
     default:
-      fop->setError("Color type not supported\n)");
+      fop->setError("Pixly loader requires a RGBA PNG\n)");
       png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
       return false;
   }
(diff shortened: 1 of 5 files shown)


dacap:
Hi @zed9h, could this use the same png encoder/decoder to avoid duplicating the whole code? I see that it opens an XML file and a png file, so to load the png file you might try to use the png decoder (or calling `load_document()` to load the png file as a temporal `doc::Sprite`/`app::Document`).

Also could you please [sign our CLA](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla)?

zed9h:
@dacap, maybe It would be cleaner if I created a helper to decode/encode the PNG for both file formats, but first I will work on the Pixly format save, which will add more value upfront.

PS: I will read the CLA.

dacap:
I'll take a look at this on next Tuesday. At the moment I recommend you to use `load_document()` if it's possible, so we can avoid modifying the actual png encoder/decoder (the code review from my side would be too much work if you refactor the code).

You can use `Document* tmpDoc = load_document(nullptr, pngfilename)` to "subload" the png file, and `save_document(nullptr, tmpDoc)` to save the png file.

zed9h:
I signed the CLA and finished the Pixly file save; the code ended up much simpler than the original PNG format, because it only handle RGBA, take a look if you think it worth the generalization of the codec. For me it was easier to handle a standard lib code than aseprite internal structure, even when I found the latter pretty clear and organized.

dacap:
@zed9h Remember that there is [an email required](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sending-the-email) in the CLA signing process too.

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,487 @@
...
+  for (y = 0; y < height; y++) {
+    rows_pointer[y] = (png_bytep)png_malloc(png_ptr, png_get_rowbytes(png_ptr, info_ptr));
+    fop->setProgress(0.1 * (double)(y+1) / (double)height);
+  }	
+
+  fprintf(xml_fp, 
+    "<PixlyAnimation version=\"1.5\">\n"
+    "\t<Info "
+    "sheetWidth=\"%d\" sheetHeight=\"%d\" "
+    "totalCollumns=\"%d\" totalRows=\"%d\" "
```
dacap: Is this `totalCollumns` or `totalColumns`?

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,487 @@
...
+    TiXmlElement* xmlFrames = xmlAnim->FirstChild("Frames")->ToElement();
+    int imageCount = strtol(xmlFrames->Attribute("length"), NULL, 10);
+    int frameCount = imageCount / layerCount;
+    sprite->setTotalFrames(frame_t(frameCount));
+
+    for(int i=0; i<layerCount; i++) {
+      sprite->folder()->addLayer(new LayerImage(sprite));
+    }
+
+    int *visible = (int*)malloc(layerCount * sizeof(int));
```
dacap: `visible` can be a `std::vector`, e.g.

    std::vector<int> visible(layerCount, 0);

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,487 @@
+// Aseprite
+// Copyright (C) 2001-2016  David Capello
```
dacap: You can add a new Copyright (C) 2016 [your name] here.

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,487 @@
...
+        break;
+    }
+  }
+
+  {
+    XmlDocumentRef doc = open_xml(fop->filename());
+    TiXmlHandle xml(doc.get());
+    fop->setProgress(0.75);
+
+    TiXmlElement* xmlAnim = xml.FirstChild("PixlyAnimation").ToElement();
```
dacap: What happen if `PixlyAnimation` element does not exist?

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+    TiXmlElement* xmlInfo = xmlAnim->FirstChild("Info")->ToElement();
```
dacap: Same here, what happen if `Info` element does not exist? (Or the attributes in the following lines, `layerCount`, `frameWidth`, `frameHeight`?)

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+    int layerCount  = strtol(xmlInfo->Attribute("layerCount"), NULL, 10);
```
dacap: You should use `base::convert_to<int>(...)` to convert from `std::string`/`const char*` to `int`

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,487 @@
...
+      Layer *layer = sprite->indexToLayer(layer_index);
+
+      const char * duration_str = xmlFrame->Attribute("duration");
+      if(duration_str) {
+        sprite->setFrameDuration(frame, strtol(duration_str, NULL, 10));
+      }
+
+      const char * visible_str = xmlFrame->Attribute("visible");
+      if(visible_str) {
+        visible[(int)layer_index] += std::string(visible_str) == "true";
```
dacap: Add `(`and `)` in the right-side expression:

    visible[(int)layer_index] += (std::string(visible_str) == "true");

dacap:
As a side note, I'll need several `.anim` files to test these routines (original files generated with Pixly), because I don't have an Android device and cannot install Pixly. You can send me some test files at support@aseprite.org (I'll add them into an internal test suite I use for decoders/encoders)

zed9h:
@dacap Great points, I placed many more check()s and improved the error messages. I will send some good files and some bad ones to test the exceptions.

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,535 @@
...
+  png_destroy_read_struct(&png_ptr, &info_ptr, NULL);
+  return success;
+}
+
+#ifdef ENABLE_SAVE
+bool PixlyFormat::onSave(FileOp* fop)
+{
+  const Sprite* sprite = fop->document()->sprite();
+
+  if(sprite->pixelFormat() != IMAGE_RGB || !fop->document()->sprite()->needAlpha()) {
```
dacap: There is no need to check `needAlpha`. Also, you can remove the `pixelFormat()` check if you remove the `FILE_SUPPORT_GRAY `, `FILE_SUPPORT_GRAYA`, and `FILE_SUPPORT_INDEXED` from `onGetFlags()` (and you should remove those constants from `onGetFlags()` as this format doesn't support Grayscale, Indexed)

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,535 @@
...
+      FILE_SUPPORT_LOAD |
+      FILE_SUPPORT_SAVE |
+      FILE_SUPPORT_RGB |
+      FILE_SUPPORT_RGBA |
+      FILE_SUPPORT_GRAY |
+      FILE_SUPPORT_GRAYA |
+      FILE_SUPPORT_INDEXED |
+      FILE_SUPPORT_LAYERS |
+      FILE_SUPPORT_FRAMES |
+      FILE_SUPPORT_PALETTE_WITH_ALPHA;
```
dacap: Remove `FILE_SUPPORT_PALETTE_WITH_ALPHA` too as this format doesn't support indexed images with color palettes (with alpha or without alpha).

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,535 @@
...
+#include "app/ini_file.h"
+#include "app/xml_document.h"
+#include "base/file_handle.h"
+#include "base/convert_to.h"
+#include "base/path.h"
+#include "doc/doc.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
```
dacap: As a minor detail, you can `#include <cstdio>`, `<cstdlib>`, and `<cmath>` here.

zed9h:
If I remove the FILE_SUPPORT_PALETTE_WITH_ALPHA and the auto-generated palette has some alpha on it, there will be a warning every time the file is saved, so I think I will need to keep this flag, even if the palette will not be saved. I also added FILE_SUPPORT_BIG_PALETTES for the same reason. (Not really sure if this is the right way to go.)

I cleaned up some includes that seemed unnecessary.

dacap:
Mhh, I think it's a bug in the code that check conditions in `file.cpp`. I'll be working in a fix for this so `FILE*PALETTE*` flags are checked only for Indexed images. (But I've to think more about those situations.)

The `FILE_SUPPORT_BIG_PALETTES` is in case the file format can save a color palette with more than 256 colors.

You can leave the flags and I'll then modify when the fix is done. Don't worry about that.

src/app/file/pixly_format.cpp
```diff
@@ -0,0 +1,525 @@
...
+    int frameWidth  = check_number<int>(xmlInfo->Attribute("frameWidth"));
+    int frameHeight = check_number<int>(xmlInfo->Attribute("frameHeight"));
+
+    UniquePtr<Sprite> sprite(new Sprite(IMAGE_RGB, frameWidth, frameHeight, 0));
+
+    TiXmlElement* xmlFrames = check(xmlAnim->FirstChild("Frames"))->ToElement();
+    int imageCount = check_number<int>(xmlFrames->Attribute("length"));
+
+    if(layerCount <= 0 || imageCount <= 0) {
+      throw Exception("No cels found");
```
dacap: I see that these `throw` can generate some leaks (mainly the png structures that are not freed). Anyway, don't worry, I'll integrate this with the png encoder/decoder later. I'll merge this code as it is.

dacap: Oh sorry, I've just saw the try{}catch block to avoid leaks.

dacap:
Finally I got a crash [saving this file](https://github.com/aseprite/aseprite/files/346562/crash.ase.zip):

screen shot 2016-07-04 at 12 34 09 pm

Could you please check this issue?

zed9h:
As I suspected when I saw the line of the error, the cel image can be smaller than the sprite size, I will compensate for that.

dacap:
Returning to this one: I'll finally remove `FILE_SUPPORT_BIG_PALETTES` and `FILE_SUPPORT_PALETTE_WITH_ALPHA` flags from `.anim` format as this format doesn't support these features (and the user must be warned that the palette will be lost). I've added the new issue #1182 how to fix the issues commented here.

zed9h:
Of course, if the palette is user created there should be a warning.

@dacap BTW, I didn't explain it properly here, the "collums" (sic) are on the official Pixly format, nothing I could do about it.

dacap:
> BTW, I didn't explain it properly here, the "collums" (sic) are on the official Pixly format, nothing I could do about it

👍   yeah, that's what I thought

PR #1187, opened 2016-07-08
Cel preview

zed9h:
It's not finished yet, I am opening it to start the discussion.

As we talked on #152, 
I placed a button with the eye icon on the right side of the onionskin to toggle the cel preview (it is not being persisted on the DocPref yet), as it covers a lot of important info about the cel I thought that a quick toggle should be pretty important, because both see it and not see it are useful.

I didn't found that easy to make it appear on the preview window, so I did an overlay, I had some problems to invalidate, but it seems alright now.

It seems right to me to cache the surfaces that I am using to plot the thumbnails on the timeline -- even if on my machine its not slow to draw --, but I don't know where exactly to place them, the timeline does not seem to know the lifetime of the document, and even less of the cels, and I am not sure if it's ok to place a surface inside the image object so it can die with it.

What do you think, do I dish this cache? Place it on the image object?


dacap:
Hi @zed9h! Thanks for taking a try to this issue. It will take me some time to review these PRs. Anyway I'll try to make some comments in a quick glance now and then review it in more detail.

About the UI, the popup is pretty close to what I was looking for (before the preview on the same 'Preview' window). I like it so far. On the other side, the preview on the same timeline should be configurable with a separated option or automatically enabled when I add an option in the future to make timeline boxes bigger. I'm not sure about the eye in the same Timeline, I think the thumbnails options might be visible in the timeline configuration popup here:

popup

And we could add a new `Command` to switch the preview status (a command similar to `ShowOnionSkinCommand`).

P.S. I got a crash (an assert failed) when a thumbnail (maybe alpha != 255) was going to be pop-upped (on `drawCelOverlay`):
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff98b82f06 __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff8d04d4ec pthread_kill + 90
2   libsystem_c.dylib             	0x00007fff927556e7 abort + 129
3   aseprite                      	0x0000000104d2a420 base_assert(char const*, char const*, int) + 1040 (debug.cpp:51)
4   aseprite                      	0x0000000104277e4b doc::ImageBits<doc::RgbTraits>::ImageBits(doc::Image*, gfx::RectT<int> const&) + 171 (image_bits.h:37)
5   aseprite                      	0x0000000104277d95 doc::ImageBits<doc::RgbTraits>::ImageBits(doc::Image*, gfx::RectT<int> const&) + 37 (image_bits.h:39)
6   aseprite                      	0x0000000104277cf7 doc::ImageBits<doc::RgbTraits> doc::Image::lockBits<doc::RgbTraits>(doc::Image::LockType, gfx::RectT<int> const&) const + 39 (image.h:60)
7   aseprite                      	0x00000001043af359 doc::LockImageBits<doc::RgbTraits>::LockImageBits(doc::Image const*, gfx::RectT<int> const&) + 41 (image_bits.h:119)
8   aseprite                      	0x00000001043af185 doc::LockImageBits<doc::RgbTraits>::LockImageBits(doc::Image const*, gfx::RectT<int> const&) + 37 (image_bits.h:119)
9   aseprite                      	0x00000001046ac28c void doc::algorithm::image_scale_tpl<doc::RgbTraits, unsigned int (*)(unsigned int, unsigned int)>(doc::Image*, doc::Image const*, int, int, int, int, int, int, int, int, unsigned int (*)(unsigned int, unsigned int)) + 172 (rotate.cpp:49)
10  aseprite                      	0x00000001046abf91 doc::algorithm::scale_image(doc::Image*, doc::Image const*, int, int, int, int, int, int, int, int) + 433 (rotate.cpp:128)
11  aseprite                      	0x00000001045ccd32 app::Timeline::drawCelOverlay(ui::Graphics*, doc::LayerIndex, int) + 1714 (timeline.cpp:1637)
12  aseprite                      	0x00000001045c81a0 app::Timeline::onPaint(ui::PaintEvent&) + 3824 (timeline.cpp:1061)
13  aseprite                      	0x00000001047e9029 ui::Widget::paintEvent(ui::Graphics*) + 569 (widget.cpp:1007)
14  aseprite                      	0x00000001047ea832 ui::Widget::onProcessMessage(ui::Message*) + 978 (widget.cpp:1307)
15  aseprite                      	0x00000001045c20f6 app::Timeline::onProcessMessage(ui::Message*) + 12342 (timeline.cpp:920)
16  aseprite                      	0x00000001047e95aa ui::Widget::sendMessage(ui::Message*) + 106 (widget.cpp:1109)
17  aseprite                      	0x00000001047ba083 ui::Manager::pumpQueue() + 1283 (manager.cpp:1331)
18  aseprite                      	0x00000001047b4e09 ui::Manager::dispatchMessages() + 25 (manager.cpp:584)
19  aseprite                      	0x00000001047ca8a3 ui::MessageLoop::pumpMessages() + 83 (message_loop.cpp:29)
20  aseprite                      	0x00000001047b52ac ui::Manager::run() + 140 (manager.cpp:167)
21  aseprite                      	0x0000000104260aa4 app::App::run() + 404 (app.cpp:695)
22  aseprite                      	0x00000001042572ce app_main(int, char**) + 878 (main.cpp:85)
23  aseprite                      	0x000000010480402d main + 125 (she.cpp:83)
24  aseprite                      	0x0000000104256f54 start + 52
```

src/app/ui/timeline.cpp
```diff
@@ -134,6 +143,9 @@ Timeline::Timeline()
   , m_offset_count(0)
   , m_scroll(false)
   , m_fromTimeline(false)
+  , m_celPreview(false)
+  , m_celPreviewOverlayRect(0,0,0,0)
```
dacap: Minor issue, we use this style: `(0, 0, 0, 0)`

dacap:
About the thumbnail, it shouldn't be inside the `doc::Image`. It's better if it's a completely separated structure/class that only handle thumbnails. It might be separated from the `Timeline` with a nice API to access and update thumbnails. Also this cache object should be able to destroy thumbnails automatically using a document observer.

src/app/ui/timeline.cpp
```diff
@@ -1518,6 +1553,104 @@ void Timeline::drawCel(ui::Graphics* g, LayerIndex layerIndex, frame_t frame, Ce
...
+    ThumbCache::iterator it;
+    if(is_active) {
+      m_celPreviewThumbCache.erase(image_id);
+      it = m_celPreviewThumbCache.end();
+    } else {
+      it = m_celPreviewThumbCache.find(image_id);
+    }
+    if(it != m_celPreviewThumbCache.end()) {
+      thumb_surf = (*it).second.get();
+    } else {
```
dacap: Minor issue, we use this style for else:
```
if (...) {
}
else {
}
```

src/app/ui/timeline.cpp
```diff
@@ -1518,6 +1553,104 @@ void Timeline::drawCel(ui::Graphics* g, LayerIndex layerIndex, frame_t frame, Ce
   // Draw decorators to link the activeCel with its links.
   if (data->activeIt != data->end)
     drawCelLinkDecorators(g, bounds, cel, frame, is_active, is_hover, data);
+
+
+  if(m_celPreview && image) {
+
+    she::Surface *thumb_surf;
+
+#ifdef CEL_THUMB_CACHE
```
dacap: Move all cache handling code into a new class (e.g. inside `ThumbCache`).

src/app/ui/timeline.h
```diff
@@ -291,6 +300,12 @@ namespace app {
 
     AniControls m_aniControls;
 
+    bool m_celPreview;
+    gfx::Rect m_celPreviewOverlayRect;
+    // TODO sortable/cleanable by MRU
+    typedef std::map< doc::ObjectId, base::SharedPtr<she::Surface> > ThumbCache;
```
dacap: `ThumbCache` might be a separate class instead a typedef in other file outside the timeline. There is already a `ThumbnailGenerator` to generate thumbnails for file names, so this `ThumbCache` should be something different.

src/app/ui/timeline.h
```diff
@@ -291,6 +300,12 @@ namespace app {
...
+    ThumbCache m_celPreviewThumbCache;
```
dacap: The cache might be inside the `app::App` object.

src/app/ui/timeline.cpp
```diff
@@ -1518,6 +1553,104 @@ void Timeline::drawCel(ui::Graphics* g, LayerIndex layerIndex, frame_t frame, Ce
...
+
+      thumb_surf = she::instance()->createRgbaSurface(
+        thumb_img->width(),
+        thumb_img->height());
+
+      convert_image_to_surface(thumb_img, m_sprite->palette(m_frame), thumb_surf,
+        0, 0, 0, 0, thumb_img->width(), thumb_img->height());
+
+#ifdef CEL_THUMB_CACHE
+      m_celPreviewThumbCache[image_id] = base::SharedPtr<she::Surface>(thumb_surf);
```
dacap: `m_celPreviewThumbCache` contains `SharedPtr` but they are deleted with `delete`, and a `she::Surface` should be deleted with its `dispose` member.

zed9h:
Fixed the crash, it was due to an aspect ratio issue; I also improved the quality and alignment of the thumbs, but now it really needs that cache and some options to customize speed vs quality.

I will first work on the options, then learn about the document observer. @dacap, you think it should be on App, listen to the modification of Cel images and automatically generate the thumbs, even if they were not seen yet, or the Timeline should request them and App create them on demand?

dacap:
> you think it should be on App, listen to the modification of Cel images and automatically generate the thumbs, even if they were not seen yet, or the Timeline should request the App to create them on demand?

The Timeline should request to the cache that it requires a new thumbnail, the App should know almost nothing about the thumbnail cache (the App controls only the lifetime of the cache), but you can request the cache object from the App instance/singleton. I'll check the latest changes.

src/app/ui/timeline.cpp
```diff
@@ -50,10 +50,18 @@
...
+#include "doc/algorithm/resize_image.h"
+#include "she/surface.h"
+#include "she/system.h"
+#include "doc/conversion_she.h"
 
 #include <cstdio>
 #include <vector>
 
+
+
```
dacap: This is a minor issue, you can remove these lines in case you create other commit for this file.

src/app/ui/timeline.cpp
```diff
@@ -1518,6 +1538,162 @@ void Timeline::drawCel(ui::Graphics* g, LayerIndex layerIndex, frame_t frame, Ce
...
+    double zw = thumb_bounds.w / (double)image->width();
+    double zh = thumb_bounds.h / (double)image->height();
+    double z0 = zw < zh ? zw : zh;
+    double zoom = z0 > 1 ? 1 : z0;
+
+    int w = (int)(image->width()  * zoom);
+    int h = (int)(image->height() * zoom);
+    int x = (int)(thumb_bounds.w * 0.5 - image->width()  * zoom * 0.5);
+    int y = (int)(thumb_bounds.h * 0.5 - image->height() * zoom * 0.5);
```
dacap: Could we replace these `x, y, w, h` with a `gfx::Rect`?

src/app/ui/timeline.cpp
```diff
@@ -1518,6 +1538,162 @@ void Timeline::drawCel(ui::Graphics* g, LayerIndex layerIndex, frame_t frame, Ce
...
+
+    // TODO select alphas and color of background
+    // TODO scale alternatives (bilinear, ...)
+
+    gfx::Rect thumb_bounds = gfx::Rect(bounds).offset(1,1).inflate(-1,-1);
+
+    double zw = thumb_bounds.w / (double)image->width();
+    double zh = thumb_bounds.h / (double)image->height();
+    double z0 = zw < zh ? zw : zh;
+    double zoom = z0 > 1 ? 1 : z0;
```
dacap: I think it would be nice to enclose these `?` expressions with a `(` and `)`. E.g.

    double z0 = (zw < zh ? zw : zh);

src/app/ui/timeline.cpp
```diff
@@ -1518,6 +1538,162 @@ void Timeline::drawCel(ui::Graphics* g, LayerIndex layerIndex, frame_t frame, Ce
...
+    she::Surface *thumb_surf;
+
+    // TODO select alphas and color of background
+    // TODO scale alternatives (bilinear, ...)
+
+    gfx::Rect thumb_bounds = gfx::Rect(bounds).offset(1,1).inflate(-1,-1);
+
+    double zw = thumb_bounds.w / (double)image->width();
+    double zh = thumb_bounds.h / (double)image->height();
+    double z0 = zw < zh ? zw : zh;
```
dacap: This might be replaced with `z0 = MIN(zw, zh);`

src/app/ui/timeline.cpp
```diff
@@ -1518,6 +1538,162 @@ void Timeline::drawCel(ui::Graphics* g, LayerIndex layerIndex, frame_t frame, Ce
...
+    0, 0, 0, 0, overlay_img->width(), overlay_img->height());
+
+  gfx::Color background = color_utils::color_for_ui(docPref().celPreview.background());
+  gfx::Color border = color_utils::blackandwhite_neg(background);
+  g->fillRect(background, bounds);
+  g->drawRgbaSurface(overlay_surf, bounds.x, bounds.y);
+  g->drawRect(border, m_celPreviewOverlayRect);
+
+  overlay_surf->dispose();
+
```
dacap: You can remove these extra blank lines between `{` and `}`.

data/widgets/timeline_conf.xml
```diff
@@ -2,6 +2,27 @@
...
+    <hbox cell_hspan="2">
+      <check id="cel_show_thumb" text="Show thumbnails" cell_hspan="2" />
+      <check id="cel_show_overlay" text="Show on mouse hover" cell_hspan="2" />
+    </hbox>
+
+    <hbox cell_hspan="2">
+      <label text="Thumb quality:" />
+      <radio group="3" text="Nearest" id="cel_quality_nearest" />
+      <radio group="3" text="Bilinear" id="cel_quality_bilinear" />
+      <radio group="3" text="Rotsprite" id="cel_quality_rotsprite" />
```
dacap: I think this "quality" parameter might be displayed as a combo-box as in the resize dialog ([see this](https://github.com/aseprite/aseprite/blob/master/data/widgets/sprite_size.xml#L40) and [this](https://github.com/aseprite/aseprite/blob/master/src/app/commands/cmd_sprite_size.cpp#L171)).

zed9h:
@dacap, I also added the size of the overlay to the prefs, and rearranged it's layout to better use the space.

The preferences and gui systems are pretty nice to work with, but one thing I was not sure were if I had to call `doc()->notifyGeneralUpdate();` on every change callback and also don't know if I had to use `migrate` on the `pref.xml`.

It seems only the cache is missing now.

dacap:
> one thing I was not sure were if I had to call doc()->notifyGeneralUpdate() on every change callback

The general update notification is used to know if the document/sprite/model structure has changed. In this case it shouldn't be necessary to call it because we're changing UI configuration only.

> and also don't know if I had to use migrate on the pref.xml.

The `migrate` field is used to migrate old field names to new ones, so you shouldn't use them for new configuration fields/options.

> PS: I did a ShowCelPreviewThumbCommand and placed on F6 (close to the ShowOnionSkin on F3)

Thinking about this one, why we don't call the whole thing "timeline thumbnails" instead of "cel previews"? In this way the thumbnails could be appear for cels, frames, and layers. So the `"cel_preview"` section in `pref.xml` could be called just `"thumbnails"`.

data/widgets/timeline_conf.xml
```diff
@@ -2,6 +2,23 @@
 <!-- Copyright (C) 2014, 2015 by David Capello -->
 <gui>
 <vbox id="timeline_conf">
+  <separator cell_hspan="2" text="Cel Preview:" left="true" horizontal="true" />
```
dacap: It might be `text="Thumbnails:"`

dacap:
About the `Thumb` and `Zoom` terms in the UI and preference variables (without checking out the code and seeing the UI) I'm not sure what does they mean.

zed9h:
> About the Thumb and Zoom terms in the UI and preference variables (without checking out the code and seeing the UI) I'm not sure what does they mean.

I am not really satisfied with the naming too. Not really sure what to call the `overlay`.

> The general update notification is used to know if the document/sprite/model structure has changed. In this case it shouldn't be necessary to call it because we're changing UI configuration only.

How should I notify the timeline to update the thumbnails?

dacap:
> How should I notify the timeline to update the thumbnails?

It looks like the Timeline should listen to changes in the thumbnails preferences. See that the thumbnails is something in the UI layer, nothing to do with the sprite model ("doc" namespace), e.g. a general update notification would redraw the sprite editor, which isn't required when we switch the thumbnails configuration (only the timeline depends on thumbnails configuration).

[Here](https://github.com/aseprite/aseprite/blob/master/src/app/ui/timeline.cpp#L191) you might start listening changes in `docPref.thumbnails`/`docPref.celPreview` pref section, and [here](https://github.com/aseprite/aseprite/blob/master/src/app/ui/timeline.cpp#L218) you might disconnect the signal slots (use a `base::ScopedConnection` for this).

zed9h:
> Here you might start listening changes in docPref.thumbnails/docPref.celPreview pref section, and here you might disconnect the signal slots (use a base::ScopedConnection for this).

OK, nice.

dacap:
I'll make some general comments now and leave a full review for the next week so you can work on this:

1. The cache class should be in other file (`TimelineThumbnailsCache`, `timeline_thumbnails_cache.cpp/h` or something similar). I've a pending refactor for the Timeline to put the whole implementation in a  `src/app/ui/timeline/`, but that will be for a future version.
1. Try to avoid exposing the cache class functionality/internal implementation to the timeline as much as possible. The best cache API would be something like `cache.getThumbnail(cel)` or `cache.setThumbnail(cel)` (I'm not sure if the Timeline should know more than that about the cache.)
1. You can use the [Object::version()](https://github.com/aseprite/aseprite/blob/master/src/doc/object.h#L26) field to know if the thumbnail should be regenerated.
1. The `DocSurfaceCacheRecentlyUsed` is a `std::multimap`, is this needed? I see the code quickly and the map is not accessed by the key (timestamp). I think you are using this to sort items by timestamp ([but it looks like the order is not guaranteed](http://stackoverflow.com/questions/530462/in-c-does-stdmultiset-keep-a-stable-sorting-order))

zed9h:
(I was writing about this, but you were quicker than me :)

> 1) The cache class should be in other file (TimelineThumbnailsCache, timeline_thumbnails_cache.cpp/h or something similar). I've a pending refactor for the Timeline to put the whole implementation in a src/app/ui/timeline/, but that will be for a future version.

Sure, sure, it's a bit of a boilerplate right now. I was looking for the place to put the new class.
I saw that the documents are managed at `UIContext`, so is it OK to place the cache class as a member of it? Then I can `addObserver` on it to listen to `removeDocument` (to clean the document), as well as `onChange` of the thumbnail prefs, and I was thinking in use the context's `onActiveSiteChange` to (clean the cel being drawn), but it may not be necessary with the `version`.

> 2) Try to avoid exposing the cache class functionality/internal implementation to the timeline as much as possible. The best cache API would be something like cache.getThumbnail(cel) or cache.setThumbnail(cel) (I'm not sure if the Timeline should know more than that about the cache.)

For sure, now that it's working I have a better idea of the interface I will need. The only thing I was not sure were if the cache should generate the surface, or if it should be timeline and just store on the cache. But your suggestion gave me an idea, to index the cache by thumb size as well, so it could be safely used by anyone.

> 3) You can use the Object::version() field to know if the thumbnail should be regenerated.

Great! That was what I was looking for. I was trying to track each individual image with the events from the `DocumentObserver`, but `onRemoveFrame` is called after the cels were deleted, and `onImagePixelsModified` is not called by anyone. The version could be perfect.

> 4) The DocSurfaceCacheRecentlyUsed is a std::multimap, is this needed? I see the code quickly and the map is not accessed by the key (timestamp). I think you are using this to sort items by timestamp (but it looks like the order is not guaranteed)

The order is guaranteed, just not necessarily the stable ordering (items of the same timestamp to be always ordered the same way). Also not sure I required a `multimap`, but the semantics seemed right.

dacap:
> I saw that the documents are managed at `UIContext`, so is it OK to place the cache class as a member of it? Then I can `addObserver` on it to listen to `removeDocument` (to clean the document), as well as `onChange` of the thumbnail prefs, and I was thinking in use the context's `onActiveSiteChange` to (clean the cel being drawn), but it may not be necessary with the `version`.

If the cache is inside the `UIContext` you don't even need to observe the context (the `UIContext` can should be able to update the cache from `UIContext::onRemoveDocument`). Anyway I'm not 100% sure, but you can give a try and put the cache inside `IUContext` instead of `App`.

> The only thing I was not sure were if the cache should generate the surface, or if it should be timeline and just store on the cache.

Same here :smile: I added the `setThumbnail()` just because the Timeline knows the configuration to generate the thumbnail, but the generation might be in the timeline, in the cache, or even other specific function/class to do so.

> But your suggestion gave me an idea, to index the cache by thumb size as well, so it could be safely used by anyone.

OK, in case you create a generic thumbnails cache for `doc::Object`s you can give it a generic name without `Timeline` on it (i.e. just `Thumbnails`). And I can rename the current `ThumbnailGenerator` into something else like `FSThumbnails`.

> > 4) The DocSurfaceCacheRecentlyUsed is a std::multimap, is this needed? I see the code quickly and the map is not accessed by the key (timestamp). I think you are using this to sort items by timestamp (but it looks like the order is not guaranteed)
> 
> The order is guaranteed, just not necessarily the stable ordering (items of the same timestamp to be always ordered the same way). Also not sure I required a `multimap`, but the semantics seemed right.

OK. Anyway if you can avoid using a `multimap` that would be better (as it's not used in other places in the code; you might try a `map`, a `priority_queue`, or just `sort`).

src/app/commands/cmd_toggle_timeline_thumbnails.cpp
```diff
@@ -0,0 +1,50 @@
+// Aseprite
+// Copyright (C) 20016  Carlo "zED" Caputo
```
dacap: You can remove the "zED" part from copyright (I'm not sure about the legal validity about that)

src/app/thumbnails.cpp
```diff
@@ -0,0 +1,315 @@
...
+#include "app/document.h"
+#include "app/color_utils.h"
+#include "render/render.h"
+#include "she/surface.h"
+#include "she/system.h"
+#include "doc/conversion_she.h"
+#include "base/bind.h"
+
+namespace app {
+  namespace thumb {
```
dacap: About this, I really like the idea about the `thumb` namespace :+1:

src/app/thumbnails.h
```diff
@@ -0,0 +1,154 @@
...
+      Sequence timestamp;
+    };
+
+    struct TagRefComp
+    {
+      bool operator()(const Tag* lhs, const Tag* rhs) const;
+    };
+    typedef std::set< Tag*, TagRefComp > RecentlyUsed;
+
+    class Tier {
```
dacap: It would be nice if you can comment in the issue about your design decisions. At a quick glance, I don't understand why there is a full hierarchy of classes here (`Tier`, `Data`, `Directory`, `Surface`, `Image`, `Document`, `Cache`). Also try to use other class names, I know that `thumb` namespace avoid name collisions, but it make harder to read the code. (And example of something we're dealing with: `doc::Document` vs `app::Document`, something I've to refactor because it's a pain to deal with both classes.)

dacap:
One final thing about this (them I'll review this next week): I think we should remove the `quality` parameter, or at least, remove the "rot sprite" possibility to create thumbnails. Mainly because "rot sprite" is really slow and uses a lot of memory and it'll be a real issue for big sprites (try editing 2048x2048 opaque sprites and see the performance of thumbnails when they are generated).

A personal recommendation: You could leave this PR just to add thumbnails into the Timeline, the thumbnails cache can be added them in a future PR. (Also I'd like to work on that merging the current thumbnail generation with a new one for documents/layers/cels, with the possibility to generate thumbnails on a background thread in an async way.)

zed9h:
I was a bit confused with app::Document vs doc::Documento, but once I realized the meaning of the namespaces it were all fine and I ended up following the trend. I will draw a diagram to explain the design decisions and then we discuss naming and this layered structure (btw, I avoided using `Layer` and instead used `Tier` to avoid further confusion).

Yes, `rotsprite` is bringing more problems than solutions, as it's only slightly better than `nearest neighbor` and orders of magnitude slower in this scenery. Thinking more generally about the performance concerns, I was thinking about to place this thumb `Request`s in a queue on separate thread to avoid the problems of slow cache generation, where at first I would return a empty thumb, that later would be filled with the scaled pixels. I would need to notify the widget that requested it somehow, that it's done, possibly by passing a pointer to it on the `Request` and calling `invalidade` when it's request is completely satisfied.

It's pretty easy to make the cache a separate PR, I will just remove all the `Tier`s except the `Data` one. I will do that and we talk about the cache and performance on the second PR.

dacap:
> I will draw a diagram to explain the design decisions and then we discuss naming and this layered structure (btw, I avoided using Layer and instead used Tier to avoid further confusion).

There is no need for a diagram, just a little explanation why the hierarchy and each class is needed is enough. As I see it, it looks like removing the whole hierarchy and virtual methods the code still works (I might be wrong).

zed9h:
I chose to control the lifetime of the cache with the scope of the items on layered maps, and the decision of the classes in between. Each layer with it's own authority, to use `erase` on all thing below it and the destructors would do the job of cleaning up, so:

1. the `Cache` tier is responsible to keep the count of surfaces under a maximum, and also listen for removal of documents with the help of the `UIContext` (inter-document changes);
1. the `Document` tier listen to changes in the document preferences (intra-document changes);
1. the `Image` tier verify if the version and clean all surfaces cached under it ,if it changed;
1. the `Surface` tier is were the data resides, and is responsible to generate, dispose and count it.

The `Request` and `Tag` help out with this passing of parameters between the `Tiers`, the first initializes what would be required and the second helps to categorize the `Data` (for instance, when ir was last used), it's a materialization of the more volatile `Request`.

The interfaces at the beginning (`Tier`, `Directory` and `Data`) were to help thinking about and implementing it consistently, they are not really necessary, could turn into a documentation.

zed9h:
I removed the rotsprite option, removed the abstract classes from thumb namespace and used their names as suffixes of the other ones, I also copied the outline of the design decisions to the issue #152, because the cache code is on #1203 now.

src/app/thumbnails.h
```diff
@@ -0,0 +1,56 @@
...
+  class Image;
+  class Document;
+}
+
+namespace app {
+  class Document;
+
+  namespace thumb {
+
+    typedef std::pair< int, int > Dimension;
```
dacap: Could this `Dimension` type be replaced with a `gfx::Size`? Because it looks like it's acting as a `gfx::Size`

src/app/thumbnails.h
```diff
@@ -0,0 +1,56 @@
...
+    public:
+      Request();
+      Request(const app::Document* doc, const doc::frame_t frm, const doc::Image* img, const Dimension dim);
+      const app::Document* document;
+      const doc::frame_t frame;
+      const doc::Image* image;
+      const Dimension dimension;
+    };
+
+    class SurfaceData {
```
dacap: It looks like this class isn't necessary (there is no state), why don't we convert these `generate`/`fetch` into a couple of (non-member) functions?

src/app/thumbnails.h
```diff
@@ -0,0 +1,56 @@
...
+    public:
+      static she::Surface* generate(Request& req);
```
dacap: It looks like `generate()` is used only inside `thumbnails.cpp`, so there is no real need to put this function in the `.h`. (It should be in an anonymous namespace inside `thumbnails.cpp`.) The same goes for `Request` class as `generate()` is the only function using that class.

src/app/ui/timeline.h
```diff
@@ -291,6 +296,16 @@ namespace app {
 
     AniControls m_aniControls;
 
+    bool m_thumbnailsOverlayVisible;
+    gfx::Rect m_thumbnailsOverlayInner;
+    gfx::Rect m_thumbnailsOverlayOuter;
+    Hit m_thumbnailsOverlayHit;
+    gfx::Point m_thumbnailsOverlayDirection;
+    void updateCelOverlayBounds(const Hit& hit);
```
dacap: We are using an implicit order of members, see that member variables are here, but member functions are before the variables.

src/app/ui_context.cpp
```diff
@@ -213,6 +213,7 @@ void UIContext::onRemoveDocument(doc::Document* doc)
       workspace->removeView(docView);
       delete docView;
     }
+
```
dacap: Remove this change.

src/app/ui_context.h
```diff
@@ -11,6 +11,7 @@
 
 #include "app/context.h"
 #include "doc/documents_observer.h"
+#include "app/thumbnails.h"
```
dacap: It looks like `UIContext` doesn't need `thumbnails.h`

dacap:
Please see that I'm reviewing this PR without seeing the other ones. I prefer to manage this in simple PRs. So we'll first try to clean-up this one, and then advance over the other PRs.

dacap:
I was testing the UI. It looks a little messy. Anyway I think we could merge the PR soon in a separate branch and then we can work more in the UI aspect.

dacap:
About `thumbnails.h`, the content of the file could be something like this:
```
#ifndef APP_THUMBNAILS_H_INCLUDED
#define APP_THUMBNAILS_H_INCLUDED
#pragma once

#include "gfx/rect.h"

namespace doc {
  class Cel;
}

namespace she {
  class Surface;
}

namespace app {
  namespace thumb {

    she::Surface* get_thumbnail(const doc::Cel* cel, const gfx::Rect& bounds);

  } // thumbnails
} // app

#endif
```

You can get the `app::Document` from the `cel` (`cel->sprite()->document()`)

zed9h:
See if it was what you meant; when you merge it, I will merge on the second PR.

src/app/thumbnails.h
```diff
@@ -1,56 +1,23 @@
-// Aseprite
-// Copyright (C) 2016  Carlo Caputo
-//
-// This program is free software; you can redistribute it and/or modify
-// it under the terms of the GNU General Public License version 2 as
-// published by the Free Software Foundation.
-
```
dacap: No need to remove the copyright notice (my comment about the `thumbnails.h` content was just illustrative 😄 )

src/app/thumbnails.h
```diff
@@ -1,56 +1,23 @@
...
-    };
-
-    class SurfaceData {
-    public:
-      static she::Surface* generate(Request& req);
-      static she::Surface* fetch(const app::Document* doc, const doc::Cel* cel, const gfx::Rect& bounds);
-    };
-
-  } // thumbnails
+    she::Surface* get_surface(const doc::Cel* cel, const gfx::Rect& bounds);
```
dacap: Definitely I'd like this function to be called `app::thumb::get_cel_thumbnail` (easier to `grep`). In a future I'll move the file system thumbnails into the `app::thumb` namespace and we can have something like `app::thumb::get_file_thumbnail`.

src/app/ui/timeline.h
```diff
@@ -260,6 +260,10 @@ namespace app {
     DocumentPreferences& docPref() const;
     skin::SkinTheme* skinTheme() const;
 
+    void updateCelOverlayBounds(const Hit& hit);
+    void drawCelOverlay(ui::Graphics* g);
+    void onThumbnailsPrefChange();
+
```
dacap: 👍

zed9h:
Sure :)
I did it in a hurry, didn't even noticed that I removed the copyright.

dacap:
Applying this patch with `git apply` generates some warnings:
```
1187.patch:596: trailing whitespace.
      
1187.patch:842: trailing whitespace.
      thumb_img, image, 
1187.patch:844: trailing whitespace.
      0, 
1187.patch:949: trailing whitespace.
      thumb_img, scale_img, 
1187.patch:957: trailing whitespace.
      x, 
warning: squelched 55 whitespace errors
warning: 60 lines add whitespace errors.
```

dacap:
Mmh, I think those warnings are about intermediate PATCHs, I'll squash everything and see the final result. Don't worry about my previous comment.

dacap:
Merged in [timeline-thumbnails branch](https://github.com/aseprite/aseprite/tree/timeline-thumbnails). Please @zed9h, check the changes in 942300e3502f195195be4185018a6ba2ec29c765 just to see some minor adjustments I made. 

Also there are two big issue to solve: 
1. `Timeline::drawCelOverlay` and `app::thumb::get_cel_thumbnail` contain duplicate code. Why we cannot use `get_cel_thumbnail()` on `drawCelOverlay`?
2. I've found a crash: create a new image with Ctrl+N, close it with Ctrl+W, create a new image with Ctrl+N

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   ???                           	0x00007f9012e10010 0 + 140256768753680
1   aseprite                      	0x000000010825c935 app::Timeline::updateUsingEditor(app::Editor*) + 581 (timeline.cpp:226)
2   aseprite                      	0x0000000108282070 app::UIContext::setActiveView(app::DocumentView*) + 560 (ui_context.cpp:108)
3   aseprite                      	0x00000001081d4f69 app::MainWindow::onActiveViewChange() + 57 (main_window.cpp:260)
4   aseprite                      	0x00000001081d79e5 obs::slot0_mem<void, app::MainWindow>::operator()() + 101 (slot.h:115)
5   aseprite                      	0x0000000107ef32d8 obs::signal0<void>::operator()() + 216 (signal.h:207)
6   aseprite                      	0x0000000108278a08 app::Workspace::setActiveView(app::WorkspaceView*) + 184 (workspace.cpp:98)
7   aseprite                      	0x000000010827875d app::Workspace::addViewToPanel(app::WorkspacePanel*, app::WorkspaceView*, bool, int) + 493 (workspace.cpp:264)
8   aseprite                      	0x000000010827855b app::Workspace::addView(app::WorkspaceView*, int) + 59 (workspace.cpp:66)
9   aseprite                      	0x000000010828291d app::UIContext::onAddDocument(doc::Document*) + 301 (ui_context.cpp:199)
10  aseprite                      	0x000000010836c3b6 void obs::observers<doc::DocumentsObserver>::notify_observers<doc::Document*>(void (doc::DocumentsObserver::*)(doc::Document*), doc::Document*) + 374 (observers.h:37)
11  aseprite                      	0x000000010836c22c void obs::observable<doc::DocumentsObserver>::notify_observers<doc::Document*>(void (doc::DocumentsObserver::*)(doc::Document*), doc::Document*) + 92 (observable.h:34)
12  aseprite                      	0x000000010836a13c void base::Observable<doc::DocumentsObserver>::notifyObservers<doc::Document*>(void (doc::DocumentsObserver::*)(doc::Document*), doc::Document*) + 92 (observable.h:35)
13  aseprite                      	0x0000000108367536 doc::Documents::add(doc::Document*) + 918 (documents.cpp:63)
14  aseprite                      	0x0000000108365746 doc::Document::setContext(doc::Context*) + 102 (document.cpp:40)
15  aseprite                      	0x0000000107f90c68 app::NewFileCommand::onExecute(app::Context*) + 3816 (cmd_new_file.cpp:180)
16  aseprite                      	0x0000000107fe02ff app::Command::execute(app::Context*) + 47 (command.cpp:64)
17  aseprite                      	0x0000000108006c46 app::Context::executeCommand(app::Command*, app::Params const&) + 2022 (context.cpp:77)
18  aseprite                      	0x0000000108079983 app::CustomizedGuiManager::onProcessMessage(ui::Message*) + 4579 (gui.cpp:457)
19  aseprite                      	0x000000010848540a ui::Widget::sendMessage(ui::Message*) + 106 (widget.cpp:1109)
20  aseprite                      	0x000000010845636d ui::Manager::pumpQueue() + 1469 (manager.cpp:1344)
21  aseprite                      	0x0000000108451039 ui::Manager::dispatchMessages() + 25 (manager.cpp:584)
22  aseprite                      	0x0000000108466703 ui::MessageLoop::pumpMessages() + 83 (message_loop.cpp:29)
23  aseprite                      	0x00000001084514dc ui::Manager::run() + 140 (manager.cpp:167)
24  aseprite                      	0x0000000107ef08c4 app::App::run() + 404 (app.cpp:695)
25  aseprite                      	0x0000000107ee710e app_main(int, char**) + 878 (main.cpp:85)
26  aseprite                      	0x000000010849ffbd main + 125 (she.cpp:83)
27  aseprite                      	0x0000000107ee6d94 start + 52
```

zed9h:
I believe both are fixed on my last PR, I will merge it all on a `pseudogames/timeline-thumbnails` branch.

PR #1189, opened 2016-07-08
Pixly2

zed9h:
Now that I am more familiar with the internal structures of the timeline I did some improvements on the Pixly format:

- using the save/load document for the PNG, as you suggested
- trimming the cels and not creating empty ones
- linking neighbor cels with the same image

It really helps managing the timeline of larger files this way.

### DIFF
--- a/src/app/file/pixly_format.cpp
+++ b/src/app/file/pixly_format.cpp
@@ -270,6 +270,11 @@ bool PixlyFormat::onLoad(FileOp* fop)
         throw Exception("looking for cels outside the bounds of the PNG");
       }
 
+      // TODO use Document* tmpDoc = load_document(nullptr, pngfilename) and save_document(nullptr, tmpDoc)
+      // TODO link identical neighbors
+      // TODO if all transparent make empty
+      // TODO trim cel image
+
       base::UniquePtr<Cel> cel;
       ImageRef image(Image::create(pixelFormat, frameWidth, frameHeight));
 
--- a/src/app/file/pixly_format.cpp
+++ b/src/app/file/pixly_format.cpp
@@ -19,6 +19,7 @@
 #include "base/convert_to.h"
 #include "base/path.h"
 #include "doc/doc.h"
+#include "doc/algorithm/shrink_bounds.h"
 
 #include <cmath>
 #include <cctype>
@@ -158,14 +159,10 @@ bool PixlyFormat::onLoad(FileOp* fop)
         throw Exception("looking for cels outside the bounds of the PNG");
       }
 
-      // TODO link identical neighbors
-      // TODO if all transparent make empty
-      // TODO trim cel image
-
+      // read cel images
       base::UniquePtr<Cel> cel;
       ImageRef image(Image::create(IMAGE_RGB, frameWidth, frameHeight));
 
-      // Convert rows_pointer into the doc::Image
       for (int y = 0; y < frameHeight; y++) {
         // RGB_ALPHA
         int y0_down = sheetHeight-1 - y0_up - (frameHeight-1) + y;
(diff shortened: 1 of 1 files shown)


src/app/file/pixly_format.cpp
```diff
@@ -264,37 +153,72 @@ bool PixlyFormat::onLoad(FileOp* fop)
...
+                                  image->maskColor()));
+
+
+        Cel* cel = NULL;
+        if((int)frame > 0) {
+          // link identical neighbors
+          Cel *prev_cel = static_cast<LayerImage*>(layer)->cel(frame-1);
+          if(prev_cel && prev_cel->x() == bounds.x && prev_cel->y() == bounds.y) {
+            Image *prev_image = prev_cel->image();
+            if(prev_image && prev_image->width() == bounds.w && prev_image->height() == bounds.h) {
```
dacap: Replace all this logic to compare two images with `count_diff_between_images` (from `doc/primitives.h`). You can see an example about how to use in the [fli format](https://github.com/aseprite/aseprite/blob/master/src/app/file/fli_format.cpp#L118).

src/app/file/pixly_format.cpp
```diff
@@ -264,37 +153,72 @@ bool PixlyFormat::onLoad(FileOp* fop)
...
+                while(src != src_end) {
+                  if(*(src++) != *(dst++)) {
+                    goto dont_link;
+                  }
+                } // x
+              } // y
+
+              cel = Cel::createLink(prev_cel);
+              cel->setFrame(frame);
+              goto add_cel;
```
dacap: Avoid `goto` with `bool` flags and `break` if it's possible (I'm not sure if it would be necessary after using (`count_diff_between_images`).

src/app/file/pixly_format.cpp
```diff
@@ -244,6 +119,20 @@ bool PixlyFormat::onLoad(FileOp* fop)
...
     }
 
+    // load image sheet
+    Document* sheet_doc = load_document(nullptr, base::replace_extension(fop->filename(),"png").c_str());
+    fop->setProgress(0.5);
+
+    Image* sheet = check(check(check(check(check(sheet_doc)->sprite())->layer(0))->cel(0))->image()); // do I need to check it all?
+
+    if(sheet->pixelFormat() != IMAGE_RGB) {
+      throw("Pixly loader requires a RGBA PNG");
```
dacap: Don't throw a string. It should be `throw Exception(...)`

src/app/file/pixly_format.cpp
```diff
@@ -244,6 +119,20 @@ bool PixlyFormat::onLoad(FileOp* fop)
       sprite->folder()->addLayer(new LayerImage(sprite));
     }
 
+    // load image sheet
+    Document* sheet_doc = load_document(nullptr, base::replace_extension(fop->filename(),"png").c_str());
+    fop->setProgress(0.5);
+
+    Image* sheet = check(check(check(check(check(sheet_doc)->sprite())->layer(0))->cel(0))->image()); // do I need to check it all?
```
dacap: No need to check this after a successful `load_document` with a `png` file (and lets hope it always behaves in the same way 😅 ).

zed9h:
@dacap I switched to debug and noticed an assert failing on the Pixly load, the code is not wrong, but the assert is also reasonable:

```c++
        uint32_t* src_begin = (uint32_t*)sheet->getPixelAddress(x0           , y0_down);
        uint32_t* src_end   = (uint32_t*)sheet->getPixelAddress(x0+frameWidth, y0_down);
        uint32_t* dst_begin = (uint32_t*)image->getPixelAddress(0, y);

        std::copy(src_begin, src_end, dst_begin);
```

```c++
    uint8_t* getPixelAddress(int x, int y) const override {
      ASSERT(x >= 0 && x < width());
      ASSERT(y >= 0 && y < height());
```

As I am only handling RGBA, here is a simple patch to fix it:

```diff
diff --git a/src/app/file/pixly_format.cpp b/src/app/file/pixly_format.cpp
index 75a1e87..a38c03a 100644
--- a/src/app/file/pixly_format.cpp
+++ b/src/app/file/pixly_format.cpp
@@ -173,8 +173,8 @@ bool PixlyFormat::onLoad(FileOp* fop)
       for (int y = 0; y < frameHeight; y++) {
         // RGB_ALPHA
         int y0_down = sheetHeight-1 - y0_up - (frameHeight-1) + y;
-        uint32_t* src_begin = (uint32_t*)sheet->getPixelAddress(x0           , y0_down);
-        uint32_t* src_end   = (uint32_t*)sheet->getPixelAddress(x0+frameWidth, y0_down);
+        uint32_t* src_begin = (uint32_t*)sheet->getPixelAddress(x0, y0_down);
+        uint32_t* src_end   = src_begin + frameWidth;
         uint32_t* dst_begin = (uint32_t*)image->getPixelAddress(0, y);

         std::copy(src_begin, src_end, dst_begin);
@@ -328,7 +328,7 @@ bool PixlyFormat::onSave(FileOp* fop)
             // RGB_ALPHA
             int y0_down = (sheetHeight - 1) - y0 - (frameHeight - 1) + celY + y;
             uint32_t* src_begin = (uint32_t*)image->getPixelAddress(0, y);
-            uint32_t* src_end   = (uint32_t*)image->getPixelAddress(celWidth, y);
+            uint32_t* src_end   = src_begin + celWidth;
             uint32_t* dst_begin = (uint32_t*)sheet->getPixelAddress(x0 + celX, y0_down);

             std::copy(src_begin, src_end, dst_begin);
```

If you prefer a pull request let me know.

PR #1193, opened 2016-07-10
Fixed invalid-desktopfile string list key "Categories" in group "Desktop Entry" does not have a semicolon (';') as trailing character

Mailaender:
Fixes a regression introduced in https://github.com/aseprite/aseprite/commit/a4ae92042d01356abc826b5898e46c8855c4f780.

```
 W: invalid-desktopfile /usr/share/applications/aseprite.desktop value "Graphics;2DGraphics;RasterGraphics" for string list key "Categories" in group "Desktop Entry" does not have a semicolon (';') as trailing character
```


dacap:
Thanks @Mailaender

PR #1200, opened 2016-07-13
OS X is now macOS

JosefKuchar:
I think, it is better for future: https://techcrunch.com/2016/06/13/os-x-is-now-macos-and-gets-support-for-siri-auto-unlock-and-more/


dacap:
Hi @JosefKuchar, actually I prefer to use OS X until macOS is more common. Thanks anyway for the patch!

PR #1208, opened 2016-07-26
Timeline thumbnails

zed9h:
Continuing #152 and addressing the issues at #1187.

If you want me to postpone the PR of the cache let me know.

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -258,11 +258,12 @@
     </section>
     <section id="thumbnails">
       <option id="enabled" type="bool" default="false" />
-      <option id="overlay_enabled" type="bool" default="false" />
-      <option id="overlay_size" type="int" default="5" />
-      <option id="quality" type="doc::algorithm::ResizeMethod" default="doc::algorithm::RESIZE_METHOD_NEAREST_NEIGHBOR" />
       <option id="opacity" type="int" default="255" />
       <option id="background" type="app::Color" default="app::Color::fromRgb(180, 180, 180, 220)" />
+    </section>
+    <section id="overlay">
+      <option id="enabled" type="bool" default="false" />
+      <option id="size" type="int" default="5" />
     </section>
     <section id="onionskin">
       <option id="active" type="bool" default="false" migrate="Onionskin.Enabled" />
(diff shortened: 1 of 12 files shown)


dacap:
I'd prefer to go in little steps. One PR for one specific thing. First, we should fix the issues in #1187 (e.g. you can create one PR with two commits, one to fix the code duplication, other commit to fix the crash). Then there are some things I'd like to discuss about the UI. (Actually there are too much configuration, e.g. background color for overlay, and I think we can simplify it.) After those changes, we can start adding the cache stuff.

As a side note: I squash commits if they don't look "clean", or do "just one thing", or the PR includes several commits adding and removing the same lines. I don't have a specific style/format for commits, but with a short line for the commit it would be enough. If you have to say "this commit does x and y" it means the commit is doing more things that it should (why x and y cannot be separated in one commit for x, and other commit for y?). I generally made that mistake myself, but we should try to avoid adding two or more "logical" changes in the same commit (e.g. "fix issue X, Y, and Z, and refactor some classes" is something really bad because can introduce several bugs at once).

Other minor comment: Do not prefix all commits with the branch name (e.g. "cel-preview"). The branch name already says what it's about, and when we merge the branch, it's will act like a summary of the all the changes related to "cel preview". (Again, these considerations are useful for: 1) it'll be easier for me to review each commit, instead of the all final changes, 2) all commits will be include merging the branch instead of squashing all commits in one big patch).

P.S. Sorry for all this back and forth in the review process, but I should warned you about the complexity of #152 (not one of the best issues for newcomers). Personally I find quite difficult to implement new features because I have to think about several points: 1) avoid introducing new bugs, 2) avoid UI cluttering, 3) refactor code to avoid complexity, 4) all new merges will be a new chunk of code to be maintained (by myself in the general case).

zed9h:
Sorry, didn't explain it properly, those two issues you noticed were the last things I did on `cel-preview3`, and unfortunately they could not be cherry picked. The difficulty I'm in is that your requests are in the reverse order of my implementation, mostly because it was rather exploratory -- for instance, it took some time to realize how the `connections` worked.

The prefix on the commits were because at some point I was working on my master, specially when there were the Pixly ones too, and I was not really sure  how the work flow would go, but as the things got bigger and I started working on branches, I kept prefixing out of habit.

I sure understand your point; such a large, long-standing code must be treated with respect. I think the only clean way to go with this changes is to re-implement it step by step, it would take much longer, but would be better for the process and the final quality.

Good you pointed it all out I will start another PR with only those two commits.

zed9h:
Redirect to #1209

PR #1209, opened 2016-07-26
Timeline thumbnails 1

zed9h:
Continuing #152 and addressing the issues at #1187 and #1208.

For now there are only two commits:

* fixed crash on docPref disconnect at doc remove
* timeline overlay using the thumbs scaler

The next things I could commit:

* connection from the preferences to the conf popup so the cmd keypress would change the checkbox
* improve the quality
   * remove the quality option (hardcode the bilinear scaler)
   * replace the bilinear scaler with a custom box sampling
* provide a better feedback about how the overlay would look with the selected settings
   * separate the preferences of the overlay and the thumbnails
   * connection with the overlay prefs to display it's looks while they the settings are changed 
* thumbnail cache
   * cache structures
   * hooks to cache clear

### DIFF
--- a/src/app/thumbnails.cpp
+++ b/src/app/thumbnails.cpp
@@ -23,7 +23,8 @@
 namespace app {
   namespace thumb {
 
-    she::Surface* get_cel_thumbnail(const doc::Cel* cel, const gfx::Rect& bounds)
+    she::Surface* get_cel_thumbnail(const doc::Cel* cel, gfx::Size thumb_size,
+      gfx::Rect cel_image_on_thumb)
     {
       app::Document* document = static_cast<app::Document*>(cel->sprite()->document());
       doc::frame_t frame = cel->frame();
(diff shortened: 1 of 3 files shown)


dacap:
Thanks @zed9h for this. I've merged the change in `timeline-thumbnails` branch.

dacap:
> * remove the quality option (hardcode the bilinear scaler)

I would prefer Nearest-neighbor in a first version.

> * replace the bilinear scaler with a custom box sampling

Let's leave this for a future. In my mind, I think thumbnails should be generated in three or four possible sizes as fast as possible with a specific algorithm (e.g. 16x16, 32x32, 64x64, 128x128), and when we request a thumbnail, the best-fit should be returned and then resized/scaled when the `she::Surface` is drawn on the display to the specific size we want (e.g. if we want to draw a 40x40, it might be enough to use the 32x32 thumbnail, or 64x64 thumbnail in case it's already generated).

> * provide a better feedback about how the overlay would look with the selected settings

I would like to remove all kind of configuration about the overlay (even background color). As I'd prefer to show the checked background in the overlay instead of a custom/transparent color.

> * separate the preferences of the overlay and the thumbnails

I would prefer to remove as much configuration as possible. And if something must be configured, I think a global configuration in *Edit > Preferences > Timeline* would be a better place.

Issue #974, opened 2016-02-23
Special limitations


dacap:
There should be a way to limit what we can draw in the program (integrated or via plugins/scripts). Some examples:

* [ZX Spectrum attribute clash](https://en.wikipedia.org/wiki/Attribute_clash)
* [Gameboy limitations](http://wayofthepixel.net/index.php?PHPSESSID=ce656168eloguaurhi3tl0e774&topic=5016.msg60635#msg60635)

These restrictions should be checked on real-time.

nodeus:
Yes, this is a very useful feature!

For example, images for the zx spectrum, I draw in Photoshop and check in real time using [img2spec by Jari Komppa](https://github.com/jarikomppa/img2spec)

But photoshop is not something that is designed for pixel art =) And only use Aseprite would be really cool.

If details help, editors for native zx-format under Windows:
— [ZX Paintbrush](http://zx-modules.de/zxpaintbrush/zxpaintbrush.html)
— [ZXGFX](http://vtrdos.ru/pcutilz/ZXGFX07.ZIP)
— [SevenuP](http://metalbrain.speccy.org/down-eng.htm)
— Editor in [RETRO-X](http://members.inode.at/838331/index.html)
— Painter — editor, included in [BMP2SCR](http://members.inode.at/838331/index.html)
— SCREEN$ Paintbox — editor, included in [BASin](https://sites.google.com/site/ulaplus/home/zx-spin-and-basin)

[Some words about zxspectrum gfx making](http://hypr.ru/blog/graphics/320.html) (sorry, only russian)
[A huge collection of zx-graphics.](http://zxart.ee/eng/graphics/)

Ravetracer:
As @nodeus mentioned RETRO-X, it would be very nice and useful to have limitations like the Commodore 64 multicolor mode as Aseprite already has a 2:1 pixel ratio.
So in this mode, we not only have this ratio but also a limitation of 4 different colors in an area of 8x8 pixel (a char) and only one of the 16 colors the C64 provide. 
Editors like "Project One" (http://csdb.dk/release/?id=86776) or Timanthes (http://csdb.dk/release/?id=75871) have these limitations. The only issue with them is, that they're designed only running on Windows systems and this would be a big plus for Aseprite ;).

grauw:
As for the colour channel limitations;

[...]

PR #1210, opened 2016-07-27
Add all 64 colors to the master system palette

pheuberger:
Please see http://www.smspower.org/maxim/HowToProgram/Palette for reference. I also set the `Columns` so it sorts the colors nicely.

### DIFF
--- a/data/palettes/master-system.gpl
+++ b/data/palettes/master-system.gpl
@@ -1,34 +1,67 @@
 GIMP Palette
+Columns: 16
 #
- 85   0  85	Untitled
-170  85 170	Untitled
-255 170 255	Untitled
-  0   0  85	Untitled
- 85  85 170	Untitled
-170 170 255	Untitled
- 85 170 170	Untitled
-  0  85  85	Untitled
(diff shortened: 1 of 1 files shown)


dacap:
Hi @pheuberger, I'll see this PR in these days. About the `Columns` parameter, Aseprite doesn't process it, anyway it's good to have it in the file (maybe in the future we can load the parameter and adjust the palette view).

pheuberger:
Hey @dacap, Ah I didn't know that the `Columns` parameter didn't have an effect. More so the order in which the colors are added and the width of the palette view, right? Right now I'm fairly happy. If I don't reorder the colors I have them actually sorted roughly in the same order I would access them on the Master System.

dacap:
This palette is a [6-bit RGB palette](https://en.wikipedia.org/wiki/List_of_monochrome_and_RGB_palettes#6-bit_RGB). This is related to #974, which should offer in a future a way to automatically handle RGB colors with a limited set of bits (6-bit, 12-bit, etc.). Anyway I'll merge this PR because it matches the master system color possibilities.

PR #1223, opened 2016-08-07
Timeline thumbnails 2: removing preferences

zed9h:
Continuing #152.

I tried to use the `AppRender` to draw it properly, I got entangled on a weird bug that I couldn't pin point yet, I were using a code like this on the thumbnails.cpp:

```cpp
AppRender render;
render.setupBackground(document, image->pixelFormat());
render.setBgCheckedSize(gfx::Size(4,4)); // FIXME hardcoded
render.renderBackground(thumb_img, gfx::Rect(thumb_size), render::Zoom(1, 1));

render.renderImage(
  thumb_img.get(),
  image,
  sprite->palette(frame),
  cel_image_on_thumb.x,
  cel_image_on_thumb.y,
  render::Zoom(cel_image_on_thumb.w, image_size.w),
  opacity, BlendMode::NORMAL);
```

There is at least one design problem using this code, I can't have images that are out of aspect (i.e., force the cel representation on the thumbnail to have at least 1 pixel wide); another problem is that if I use `algorithm::resize_image` I can't use alpha blending; but besides that I am having some other problem that are causing ASSERT failures (on render.cpp: composite_image_scale_down(): `ASSERT(dst_it >= dstBits.begin() && dst_it < dst_end)`), crashes, skews, like the rows of the image are being interpreted with the wrong width:

downscale-skew

I ended up doing did a simple checkered background render, respecting the docPref color.

PR #1224, opened 2016-08-07
Timeline thumbnails 3: zoom

zed9h:
Some improvements to the timeline zoom that maybe useful.

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -257,6 +257,7 @@
       <option id="color2" type="app::Color" default="app::Color::fromRgb(192, 192, 192)" migrate="Option.CheckedBgColor2" />
     </section>
     <section id="thumbnails">
+      <option id="zoom" type="double" default="1" />
       <option id="enabled" type="bool" default="false" />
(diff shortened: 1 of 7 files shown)


zed9h:
I committed some experimental things about the zoom with a layout that allowed to remove the opacity without hiding the cel linking styles:

image

Unfortunately this solution uses more vertical space, but that's how the cel linking looks better. 
I tried with it over the thumbs and it hides too much of them and mixes with the drawings.

Notice that the checkered background is at #1223

dacap:
Good ideas here that I was thinking off: +1 keeping the linked-cels indicators, +1 smaller width on layer eyes/padlock. It was quite hard to make the merge of both PRs+beta 😅  but I think everything is merged correctly (please tell me if you find something wrong).

From now on all future work can be done in beta branch (I'll remove the timeline-thumbnails branch). I'll create issues about this new timeline so we know what can be fixed/improved from here, and you can know if I'm working in something related to this.

dacap:
Merged here 8da12cde1ddee4b7b6d8fc0aa9cde31b9227d89f

Issue #1271, opened 2016-09-26
Timeline thumbnails issues


dacap:
- [ ] overlay image popups too quick/or get in the way sometimes
- [ ] zoom is reset each time after we paint (e.g. on OS X we can change zoom with mouse wheel, after doing that, if we paint, the zoom is reset to the configured zoom level in preferences)
- [ ] vertical scroll jumps in big chunks when we have a big zoom (e.g. Zoom = 10)
- [ ] when we change zoom with mouse wheel we must update the scroll area

dacap:
hi @zed9h, here I've just written down some tasks to improve the thumbnails, if you are going to take one of these items, please make a comment here 👍  (I'll do the same if I start working on one of these items)

zed9h:
Cool, so, you liked this solution? 
It's nice but, I still miss the thumbs on the small timeline. 
Another idea: the overlay could be hidden by default when the timeline is larger.

This week were too busy, I will try to take a look at it in the next couple of weeks.

dacap:
> Cool, so, you liked this solution? 

I liked the solution so far. Other possibility would be to move the thumbnail inside the cels-box, but I'm not 100% sure about this:

![screen shot 2016-10-03 at 11 49 04 am](https://cloud.githubusercontent.com/assets/39654/19041828/0ab1ded4-8960-11e6-9b4a-b7d101710e5e.png)

> It's nice but, I still miss the thumbs on the small timeline. 

Do you mean something like this:
![screen shot 2016-10-03 at 11 54 46 am](https://cloud.githubusercontent.com/assets/39654/19041862/34c26586-8960-11e6-8f40-f6dfb60228ce.png)

We can still have small thumbnails using the Force checkbox in the timeline options.

> Another idea: the overlay could be hidden by default when the timeline is larger.

I'll think more about this, I've to see how it works with different approaches (e.g. different positions for the overlay).

> This week were too busy, I will try to take a look at it in the next couple of weeks.

No problem. Just please make a comment if you start with some of the tasks, I'll do the same.

zed9h:
> Do you mean something like this:
> ![image](https://cloud.githubusercontent.com/assets/7734/19042135/5015e7b2-8961-11e6-9096-8e79f9ddeb5e.png)

Yes, something like this, because the highest usefulness of the thumbnails is for large timelines, where there's also little space to spare.

> Other possibility would be to move the thumbnail inside the cels-box, but I'm not 100% sure about this:
[...]

PR #1288, opened 2016-10-09
Beta

zed9h:
Points from #1271:
- vertical scroll jumps in big chunks when we have a big zoom (e.g. Zoom = 10)
- add zoom with ctrl+mouse wheel (using alt actually, ctrl is used)
- zoom is reset each time after we paint (e.g. on OS X we can change zoom with mouse wheel, after doing that, if we paint, the zoom is reset to the configured zoom level in preferences)
-  when we change zoom with mouse wheel we must update the scroll area

### DIFF
--- a/src/app/ui/timeline.cpp
+++ b/src/app/ui/timeline.cpp
@@ -1063,17 +1063,24 @@ bool Timeline::onProcessMessage(Message* msg)
 
         dx += static_cast<MouseMessage*>(msg)->wheelDelta().x;
 
-        if (msg->ctrlPressed())
-          dx = dz * base_size;
-        else
-          dy = dz * base_size;
-
-        if (msg->shiftPressed()) {
-          dx *= frameBoxWidth() / base_size;
-          dy *= layerBoxHeight() / base_size;
+        if (msg->altPressed()) {
+          double next_zoom = m_zoom + (dz < 0 ? -1 : 1);
+          setZoom(next_zoom);
+          invalidate();
         }
+        else {
+          if (msg->ctrlPressed())
+            dx = dz * base_size;
+          else
+            dy = dz * base_size;
 
-        setViewScroll(viewScroll() + gfx::Point(dx, dy));
+          if (msg->shiftPressed()) {
+            dx *= frameBoxWidth() / base_size;
+            dy *= layerBoxHeight() / base_size;
+          }
+
+          setViewScroll(viewScroll() + gfx::Point(dx, dy));
+        }
       }
       break;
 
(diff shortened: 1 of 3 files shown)


src/app/ui/timeline.cpp
```diff
@@ -88,6 +88,8 @@ enum {
   PART_FRAME_TAG,
 };
 
+static const int base_size = 12;
```
dacap: Here we can add a `// TODO move this to theme dimensions` or include a patch moving this as a Timeline static member (`Timeline::m_baseSize`?) and initialize it with `m_baseSize = skinTheme()->dimensions.timelineBaseSize()` (adding this `timeline_base_size` element in `skin.xml` at the `<dimensions>`)

src/app/ui/timeline.cpp
```diff
@@ -181,13 +180,26 @@ Timeline::~Timeline()
 void Timeline::setZoom(double zoom)
 {
   m_zoom = MID(1.0, zoom, 10.0);
```
dacap: Maybe here we can add a filter

    zoom = MID(1.0, zoom, 10.0);
    if (zoom == m_zoom)
      return;

    m_zoom = zoom;
    ...

But I'm not 100% sure.

src/app/ui/timeline.cpp
```diff
@@ -1061,17 +1073,23 @@ bool Timeline::onProcessMessage(Message* msg)
...
-          dx = dz * frameBoxWidth();
-        else
-          dy = dz * layerBoxHeight();
-
-        if (msg->shiftPressed()) {
-          dx *= 3;
-          dy *= 3;
+        if (msg->altPressed()) {
+          double next_zoom = m_zoom + (dz < 0 ? -1 : 1);
+          setZoomAndUpdate(next_zoom);
```
dacap: Wrap this with

    if (dz) {
      double next_zoom = m_zoom + (dz < 0 ? -1 : 1);
      setZoomAndUpdate(next_zoom);
    }

On OS X `dz` can be 0 (mouse wheel has an acceleration threshold, I think `wheelDelta()` point should be a `gfx::Point<double>` in a near future). Anyway without this `if (dz)` the zoom makes some jumps between -1 and 1.

---

Also I think the `dz` sign is inverted (it should be `m_zoom + (dz < 0 ? 1: -1)`), compare this behavior zooming with the mouse wheel on the sprite editor.

src/app/ui/timeline.cpp
zed9h: If I do this I will have to initialize these values somewhere else:
```cpp
  m_thumbnailsOverlayDirection = gfx::Point(int(frameBoxWidth()*1.0), int(frameBoxWidth()*0.5));
  m_thumbnailsOverlayVisible = false;
```

zed9h:
I think I fixed all the issues.

dacap:
LFTM

I'll add some items to #1271 related to old issues with keyboard+mouse wheel combinations.

Issue #964, opened 2016-02-13
Feature Request: close cell property window with Esc key


OrenjiAkira:
When you double click a cell, the interface opens its cell properties window.

However, it pops inconveniently at times, especially when you're using a tablet while selecting a cell (accidental double clicks are common with tablet users).

Thus, **it would be cool to be able to close it with a handy key, like the Escape key**, instead of having to click it on the tiny "X" close button.

Not sure if there was a similar feature request or not. If there was, I apologize for not looking properly

PR #1302, opened 2016-10-23
Allow Esc key to close Cel and Layer Properties windows

klw0:
Addresses #964, and does the same for the _Layer Properties_ window.

~~This mostly follows the key handling style used in [menu.cpp](https://github.com/aseprite/aseprite/blob/master/src/ui/menu.cpp#L475-L659).~~

### DIFF
--- a/src/app/commands/cmd_cel_properties.cpp
+++ b/src/app/commands/cmd_cel_properties.cpp
@@ -126,7 +126,9 @@ private:
 
       case kKeyDownMessage:
         if (opacity()->hasFocus()) {
-          if (static_cast<KeyMessage*>(msg)->scancode() == kKeyEnter) {
+          KeyScancode scancode = static_cast<KeyMessage*>(msg)->scancode();
+          if (scancode == kKeyEnter ||
+              scancode == kKeyEsc) {
             onCommitChange();
             closeWindow(this);
             return true;
--- a/src/app/commands/cmd_layer_properties.cpp
+++ b/src/app/commands/cmd_layer_properties.cpp
@@ -139,7 +139,9 @@ private:
         if (name()->hasFocus() ||
             opacity()->hasFocus() ||
             mode()->hasFocus()) {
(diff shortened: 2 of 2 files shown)


src/app/commands/cmd_cel_properties.cpp
```diff
@@ -126,9 +126,25 @@ class CelPropertiesWindow : public app::gen::CelProperties
...
+          switch (static_cast<KeyMessage*>(msg)->scancode()) {
+
+            case kKeyEnter:
+              onCommitChange();
+              closeWindow(this);
+              isKeyUsed = true;
+              break;
+
+            case kKeyEsc:
+              closeWindow(this);
```
dacap: I think the `kKeyEsc` should call `onCommitChange()` too, because it acts like a shortcut to close the dialog, but not to cancel the changes in the dialog (e.g. we quickly change the layer name and press `Esc` doesn't mean that we want to discard the layer name change).

I think esc should act exactly like `kKeyEnter`. So this could be just:

    case kKeyEnter:
    case kKeyEsc:
       onCommitChange();
       closeWindow(this)...

src/app/commands/cmd_layer_properties.cpp
```diff
@@ -139,9 +139,25 @@ class LayerPropertiesWindow : public app::gen::LayerProperties
...
+          switch (static_cast<KeyMessage*>(msg)->scancode()) {
+
+            case kKeyEnter:
+              onCommitChange();
+              closeWindow(this);
+              isKeyUsed = true;
+              break;
+
+            case kKeyEsc:
+              closeWindow(this);
```
dacap: Same here.

src/app/commands/cmd_cel_properties.cpp
klw0: That works, I'll update this tonight.

klw0:
@dacap updated.

dacap:
Thanks for your contribution @klw0!

PR #1336, opened 2016-11-30
Added PCC as PCX variation

winterheart:
PCC is old PCX-compatible format, which is still occurs.

See http://www.fileformat.info/format/pcx/egff.htm for reference.


dacap:
Thanks @winterheart!

dacap:
I've just added the extension in the file selector too with 3b7f48224ccb49e918b211ee9beed6dbacedd1ab

PR #1370, opened 2016-12-19
Control for undo tests

winterheart:
Control for building undo test from parent CMakelists.txt


dacap:
Thanks for this change too!

PR #1377, opened 2017-01-02
Add new color wheel for normal maps

robinjam:
This PR adds a new type of colour wheel for creating/editing tangent-space normal maps (similar to the Angle brush in [SpriteIlluminator](https://www.codeandweb.com/spriteilluminator)).

I realise this is a bit of a niche feature, but I figured I'd open this PR anyway in case this is useful to anyone else.

This could be expanded upon by adding a lighting preview to the preview window, which would allow the user to test out their hand-drawn normal map from within Aseprite rather than having to export it first.

## Screenshot

screen shot 2017-01-03 at 12 48 58

### DIFF
--- a/src/app/ui/color_wheel.cpp
+++ b/src/app/ui/color_wheel.cpp
@@ -84,6 +84,22 @@ app::Color ColorWheel::getColorInClientPos(const gfx::Point& pos)
   int v = (pos.y - (m_wheelBounds.y+m_wheelBounds.h/2));
   double d = std::sqrt(u*u + v*v);
 
+  if (m_colorModel == ColorModel::NORMAL_MAP) {
+    float x = float(u) / float(m_wheelBounds.w / 2);
+    float y = -float(v) / float(m_wheelBounds.h / 2);
+    float z = std::sqrt(1 - x*x - y*y);
+    if (z <= 1.f) {
+      return app::Color::fromRgb(
+        int(std::round((x + 1) * 127.5)),
+        int(std::round((y + 1) * 127.5)),
+        int(std::round((z + 1) * 127.5))
+      );
+    }
+    else {
+      return app::Color::fromRgb(128, 128, 255);
+    }
+  }
+
   // Pick from the wheel
   if (d < m_wheelRadius+2*guiscale()) {
     double a = std::atan2(-v, u);
@@ -231,7 +247,7 @@ void ColorWheel::onPaint(ui::PaintEvent& ev)
     }
   }
 
-  if (m_color.getAlpha() > 0) {
+  if (m_color.getAlpha() > 0 && m_colorModel != ColorModel::NORMAL_MAP) {
     int n = getHarmonies();
     int boxsize = MIN(rc.w/10, rc.h/10);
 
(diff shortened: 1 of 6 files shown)


src/app/ui/color_wheel.cpp
```diff
@@ -84,6 +84,22 @@ app::Color ColorWheel::getColorInClientPos(const gfx::Point& pos)
   int v = (pos.y - (m_wheelBounds.y+m_wheelBounds.h/2));
   double d = std::sqrt(u*u + v*v);
 
+  if (m_colorModel == ColorModel::NORMAL_MAP) {
+    float x = float(u) / float(m_wheelBounds.w / 2);
+    float y = -float(v) / float(m_wheelBounds.h / 2);
+    float z = std::sqrt(1 - x*x - y*y);
```
dacap: I think here we should use the distance `d` (distance with the center of the wheel), and the angle `a` (which is already calculated below as `double a = std::atan2(-v, u)`) instead of `x` and `y` (as it's now it looks like a box instead of a wheel)

src/app/ui/color_wheel.cpp
```diff
@@ -84,6 +84,22 @@ app::Color ColorWheel::getColorInClientPos(const gfx::Point& pos)
...
+      );
+    }
+    else {
+      return app::Color::fromRgb(128, 128, 255);
+    }
+  }
+
   // Pick from the wheel
   if (d < m_wheelRadius+2*guiscale()) {
     double a = std::atan2(-v, u);
```
dacap: Inside this condition we should be calculating the color map color depending on `d` and `a`.

src/app/ui/color_wheel.cpp
```diff
@@ -336,16 +352,32 @@ void ColorWheel::onOptions()
...
-  switch (m_harmony) {
-    case Harmony::NONE: none.setSelected(true); break;
-    case Harmony::COMPLEMENTARY: complementary.setSelected(true); break;
-    case Harmony::MONOCHROMATIC: monochromatic.setSelected(true); break;
-    case Harmony::ANALOGOUS: analogous.setSelected(true); break;
-    case Harmony::SPLIT: split.setSelected(true); break;
-    case Harmony::TRIADIC: triadic.setSelected(true); break;
-    case Harmony::TETRADIC: tetradic.setSelected(true); break;
-    case Harmony::SQUARE: square.setSelected(true); break;
+  if (m_colorModel == ColorModel::NORMAL_MAP) {
```
dacap: If the color model is `NORMAL_MAP`, we could just avoid adding `none`, `complementary`, `monochromatic`, etc. menu items inside `menu` variable.

dacap:
Thanks @robinjam, I've just merged this PR. Made some changes in [the formula to calculate the colors](https://github.com/aseprite/aseprite/commit/8f280a50d32c29563dc761f07b670b1097fdb1d8) in the wheel and [added an indicator](https://github.com/aseprite/aseprite/commit/40de823f7cfa988552443d54f4f33451b958393a) to show the picked color in the wheel.

dacap:
Before:

before

[After](https://github.com/aseprite/aseprite/commit/8f280a50d32c29563dc761f07b670b1097fdb1d8):

after

PR #1399, opened 2017-01-28
Shared cmark library

winterheart:
Making optional linking with shared system cmark library with -DUSE_SHARED_CMARK option.


dacap:
Hi @winterheart, this patch isn't passing the Travis build, maybe it needs an `apt-get install cmark` or something like that.

winterheart:
Hi, @dacap, fixed that with second commit.

dacap:
Hi @winterheart, thanks for this. I'll merge it now.

PR #1454, opened 2017-04-14
Fixed crash upon destroying PreviewEditor

mooviies:
On my end there was a crash when creating a new file. I investigated it and found out that when destroying PreviewEditor a call to m_customizationDelegate->dispose(); was making aseprite crash.

This is probably because the virtual table is unloaded before dispose is called.

So I added a setCustomizationDelegate(NULL); directly in a new destructor for PreviewEditor which makes the call to dispose correctly before the destructor of Editor is called.

### DIFF
--- a/src/app/ui/document_view.cpp
+++ b/src/app/ui/document_view.cpp
@@ -162,7 +162,7 @@ public:
 
   ~PreviewEditor()
   {
-	  setCustomizationDelegate(NULL);
+	setCustomizationDelegate(NULL);
   }
 
   // EditorCustomizationDelegate implementation
--- a/src/app/ui/document_view.cpp
+++ b/src/app/ui/document_view.cpp
@@ -160,6 +160,11 @@ public:
     setCustomizationDelegate(this);
   }
 
+  ~PreviewEditor()
+  {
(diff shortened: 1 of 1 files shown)


dacap:
Hi @mooviies, today I'm trying to release a new version so this will wait until next week. Meanwhile could you please [sign the CLA](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla)?

src/app/ui/document_view.cpp
```diff
@@ -160,6 +160,11 @@ class PreviewEditor : public Editor,
     setCustomizationDelegate(this);
   }
 
+  ~PreviewEditor()
+  {
+	setCustomizationDelegate(NULL);
```
dacap: It looks like this line is indented with a tab char, we use 2 spaces for indentation.

mooviies: Alright I'll fix my settings for 2 spaces indentation

dacap:
Thanks @mooviies, I'll check right now because I was able to reproduce the problem on Windows (release mode only).

PR #1455, opened 2017-04-14
Fix 1427

mooviies:
2 things needed to be done to fix that issue about spray paint not doing the exact same thing on each side of a symmetry.

First I changed how rand works in doPointshapePoint. Essentially, a random_seed is created before the stroke loop and srand initialize the seed back to it for each stroke.

Secondly the shape was the same but not symmetric. So I added a horizontal and vertical modifier in Stroke which is 1 by default that is modified when Symmetries generate them to -1 depending on which symmetry is needed. The spray tool then use that to reverse it's spraying direction.

I tried as much as I could to keep the basic logic of the software intact. If there's anything I need to do to keep the conventions let me know! :)

### DIFF
--- a/src/app/crash/backup_observer.cpp
+++ b/src/app/crash/backup_observer.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2001-2016  David Capello
+// Copyright (C) 2001-2017  David Capello
 //
 // This program is distributed under the terms of
 // the End-User License Agreement for Aseprite.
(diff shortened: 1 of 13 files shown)


mooviies:
I needed to add a transformPoint method with 2 more parameters. I would have prefered to send it the Stroke directly but I think you don't want to have PointShape dependent on Stroke. Correct me if I'm wrong.

dacap:
I'll take a look at this issue next week. Anyway there are several problems with symmetry. I've to refactor the code to make symmetry works at the level of pixels instead of mouse points. For example, if we use a brush like this one: 

screen shot 2017-04-14 at 4 23 01 pm

The symmetry doesn't work as expected:

screen shot 2017-04-14 at 4 23 11 pm

So basically the problem must be fixed at the pixel level (I guess) so that should fix the spray tool problem too.

mooviies:
@dacap that make sense. My solution could probably work for that case also, but it implies a case by case solution. So you'd need to take symmetry into consideration for each new tools you add. I'll close the pull request.

PR #1478, opened 2017-04-27
clear_rect.cpp: PVS-Studio: CWE-476 NULL Pointer Dereference.

SvyatoslavRazmyslov:
We have found and fixed a weakness using PVS-Studio tool. PVS-Studio is a static code analyzer for C, C++ and C#: https://www.viva64.com/en/pvs-studio/

We suggests having a look at the emails, sent from @pvs-studio.com.

Analyzer warning: [V595](https://www.viva64.com/en/w/V595/) The 'cel' pointer was utilized before it was verified against nullptr. Check lines: 26, 28. clear_rect.cpp 26

### DIFF
--- a/src/app/cmd/clear_rect.cpp
+++ b/src/app/cmd/clear_rect.cpp
@@ -23,11 +23,14 @@ using namespace doc;
 
 ClearRect::ClearRect(Cel* cel, const gfx::Rect& bounds)
 {
-  app::Document* doc = static_cast<app::Document*>(cel->document());
+  if (!cel)
+    return;
 
-  Image* image = (cel ? cel->image(): NULL);
+  Image* image = cel->image();
   if (!image)
     return;
+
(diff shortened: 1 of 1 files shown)


dacap:
Hi @SvyatoslavRazmyslov, thanks for the PR. I have your emails queued but I didn't find time to answer them yet. Some time ago some people from the CppDepend team contact me with a free key of their product too but I wasn't able to give a try to the product (I don't find time to fix all compiler warnings in the three platforms yet 😞 ).

I'm using macOS most of the time right now but today I'll give a try to PVS-Studio (if it's not today probably it will wait forever, so it must be today). After that I'll write you an email with my impressions.

SvyatoslavRazmyslov:
Hi @dacap,

> I'm using macOS

Since this year, PVS-Studio supports macOS.

Issue #909, opened 2015-12-20
Won't build on OS X


Ivoah:
I tried building aseprite as per INSTALL.md, with installing the old SDK and all, but cmake doesn't exit cleanly
I built it with this command `cmake DCMAKE_OSX_ARCHITECTURES:STRING=i386 -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.4 -DCMAKE_OSX_SYSROOT:STRING=/SDKs/MacOSX10.4u.sdk ..`
and this is the error I got: http://pastebin.com/b6EjgepY

dacap:
Hi @Ivoah, could you try to install `harfbuzz-icu` package?
E.g.

    brew install harfbuzz-icu

Ivoah:
```
Samantha:aseprite Ivo$ brew install harfbuzz-icu
Error: No available formula with the name "harfbuzz-icu" 
==> Searching for similarly named formulae...
Error: No similarly named formulae found.
==> Searching taps...
Error: No formulae found in taps.
Samantha:aseprite Ivo$ 
```

HenrySipp:
I might add that in El Capitan I'm unable to install the 10.4 SDK

Ivoah:
> I might add that in El Capitan I'm unable to install the 10.4 SDK

I installed it just fine on OS X 10.11.2

dacap:
Hi @Ivoah, did you try with `brew install harfbuzz`?
It's strange because I didn't installed it, but maybe I've it installed for some reason.

Ivoah:
I already have harfbuzz installed according to brew

tony:
Tested Jan 2nd, 2016 on El Capitan (10.11):

Download ``xcode314_2809_developerdvd.dmg`` from https://developer.apple.com/downloads/. The title is *Xcode 3.1.4 Developer Tools*.  Then, Ctrl-click ``MacOSX10.4.Universal.pkg``, open, open.

Assuming you have a ``./build.`` directory in the project root, clear out your CMake files and re-run CMake via the command:

```sh
cmake -DCMAKE_OSX_ARCHITECTURES:STRING=i386 \
-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.4 \
-DCMAKE_OSX_SYSROOT:STRING=/SDKs/MacOSX10.4u.sdk ..
```

And ``make`` to build, ``[sudo] make install`` to install the files and resources.

You can also use ninja (``brew install ninja``) with:

[...]

PR #1490, opened 2017-05-26
Update INSTALL.md with how to create application bundle + icon wrapping executable file

haxpor:
Hey,

This changes is for having a convenient way for users that did a manual build of aseprite on macOS system. Instead of finding and clicking on such built executable file after followed build steps, they can setup once and have application bundle + icon similar to how macOS users open other apps.

See image below for example

screen shot 2017-05-27 at 1 13 46 am

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -288,14 +288,15 @@ More information about these steps in the
 
 # Creating an Application Bundle with Icon on macOS
 
-On macOS, after you successfully built aseprite. Now you will have executable aseprite.  
-If you are comfortable with clicking on such executable in which it will open a new terminal, and aseprite application itself, then you're fine, you have to need to do this.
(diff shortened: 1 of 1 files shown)


dacap:
There is a more simple method using the [trial version bundle](https://github.com/aseprite/aseprite/issues/909#issuecomment-168751044). I'll think about this in the near future and see if I merge this PR, or I add another explanation to create the bundle from the trial version, or I just integrate the bundle generation in the repository/compilation process.

haxpor:
Sure, and thanks for reply.
I checked that thread out, and knew what you mean, so it makes sense for users that have trial version or other versions installed already to do that.

dacap:
I'll close this PR and if someone want to contribute with a PR about the method of using the [trial bundle](https://github.com/aseprite/aseprite/pull/1490#issuecomment-304342571) that would be great.

PR #1508, opened 2017-06-06
Adding multiple layers hide show to Layer->Visible (shift-x)

robotron2084:
This adds multiple layer support to Shift-X and Layer->Visible via LayerVisibilityCommand. I did not modify the layer palette to change multiple layers, but I can def add that to this review if you request.

Additionally I forked the repo and did various git incantations then forgot to branch *before* committing my change (i.e. I committed to master then branched). The PR looks OK and hopefully git is forgiving of this.

### DIFF
--- a/src/app/commands/cmd_layer_visibility.cpp
+++ b/src/app/commands/cmd_layer_visibility.cpp
@@ -45,16 +45,31 @@ bool LayerVisibilityCommand::onEnabled(Context* context)
 bool LayerVisibilityCommand::onChecked(Context* context)
 {
   const ContextReader reader(context);
-  const Layer* layer = reader.layer();
-  return (layer && layer->isVisible());
+  const SelectedLayers& selLayers = reader.site()->selectedLayers();
+  bool visible = false;
+  for (auto layer : selLayers)
+  {
+    if(layer && layer->isVisible())
+      visible = true;
+  }
+  return visible;
 }
 
 void LayerVisibilityCommand::onExecute(Context* context)
 {
   ContextWriter writer(context);
-  Layer* layer = writer.layer();
+  const SelectedLayers& selLayers = writer.site()->selectedLayers();
 
-  layer->setVisible(!layer->isVisible());
+  bool anyVisible = false;
+  for (auto layer : selLayers)
+  {
+    if(layer->isVisible())
+      anyVisible = true;
+  }
+  for(auto layer : selLayers)
+  {
+    layer->setVisible(!anyVisible);
+  }
 
   update_screen_for_document(writer.document());
 }
--- a/src/app/commands/cmd_layer_visibility.cpp
+++ b/src/app/commands/cmd_layer_visibility.cpp
@@ -8,6 +8,8 @@
 #include "config.h"
 #endif
 
+#include "app/app.h"
+#include "app/ui/timeline/timeline.h"
 #include "app/commands/command.h"
 #include "app/context_access.h"
 #include "app/modules/gui.h"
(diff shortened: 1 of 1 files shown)


src/app/commands/cmd_layer_visibility.cpp
```diff
@@ -45,16 +45,31 @@ bool LayerVisibilityCommand::onEnabled(Context* context)
 bool LayerVisibilityCommand::onChecked(Context* context)
 {
   const ContextReader reader(context);
-  const Layer* layer = reader.layer();
-  return (layer && layer->isVisible());
+  const SelectedLayers& selLayers = reader.site()->selectedLayers();
+  bool visible = false;
+  for (auto layer : selLayers)
+  {
+    if(layer && layer->isVisible())
```
dacap: Minor style change:
```for (auto layer : selLayers) {
  if (layer && layer->isVisible())
```

src/app/commands/cmd_layer_visibility.cpp
```diff
@@ -45,16 +45,31 @@ bool LayerVisibilityCommand::onEnabled(Context* context)
...
 void LayerVisibilityCommand::onExecute(Context* context)
 {
   ContextWriter writer(context);
-  Layer* layer = writer.layer();
+  const SelectedLayers& selLayers = writer.site()->selectedLayers();
 
-  layer->setVisible(!layer->isVisible());
+  bool anyVisible = false;
+  for (auto layer : selLayers)
+  {
```
dacap: Same style that above.

src/app/commands/cmd_layer_visibility.cpp
```diff
@@ -45,16 +45,31 @@ bool LayerVisibilityCommand::onEnabled(Context* context)
...
+      visible = true;
+  }
+  return visible;
 }
 
 void LayerVisibilityCommand::onExecute(Context* context)
 {
   ContextWriter writer(context);
-  Layer* layer = writer.layer();
+  const SelectedLayers& selLayers = writer.site()->selectedLayers();
```
dacap: Actually now the command doesn't work for the single layer case. This is the case when `selLayers` is empty and `writer.layer()` is the current layer (this happens when we select a layer and then click the sprite editor, the selection outline in the timeline disappears).

You can use something like this here (and in `onChecked`):

    SelectedLayers selLayers;
    auto range = App::instance()->timeline()->range();
    if (range.enabled()) {
      selLayers = range.selectedLayers();
    }
    else {
      selLayers.insert(writer.layer());
    }

robotron2084: oh wow, I can't believe I didn't test that use case. I will code something similar and update the PR.

src/app/commands/cmd_layer_visibility.cpp
robotron2084: Got it! I will do my best to follow the style guide.

src/app/commands/cmd_layer_visibility.cpp
dacap: No problem. I should add a function to get the real set of selected layers and avoid this kind of problems in the future and simplify code (also this little code is duplicated in several places, the `writer.site()` should have a function to solve this for us already).

robotron2084: If that is the better place for this to exist then perhaps I can do that instead. My only thought about that is how best to handle the memory .

dacap: Don't worry, you can now copy the lines, I'll see how to do it right (or if you want, you can then create then another PR for that refactor).

src/app/commands/cmd_layer_visibility.cpp
```diff
@@ -45,10 +47,16 @@ bool LayerVisibilityCommand::onEnabled(Context* context)
...
 {
   const ContextReader reader(context);
-  const SelectedLayers& selLayers = reader.site()->selectedLayers();
+  SelectedLayers selLayers;
+  auto range = App::instance()->timeline()->range();
+  if (range.enabled()) {
+    selLayers = range.selectedLayers();
+  }
+  else {
+    selLayers.insert(const_cast<Layer*>(reader.layer()));
```
robotron2084: I felt a bit dirty const_casting this. I looked for examples of usage with reader.layer() in other commands following the pattern you suggested and didn't find any. I can modify this code to use logic that doesn't cast away the const if you like.

dacap:
This PR was merged with commits squashed (d0c1be871a413f649f11c33549b9984f02fb6ea6). Thanks Chris!

Issue #1453, opened 2017-04-11
Aseprite uses internal headers of Freetype


winterheart:
Since last git additions aseprite uses internal functions of freetype that prevent to use and link external shared library:
```
[ 19%] Building CXX object src/ft/CMakeFiles/ft-lib.dir/stream.cpp.o
/home/winterheart/playground/aseprite/src/ft/stream.cpp:13:40: fatal error: freetype/internal/internal.h: No such file or directory
 #include <freetype/internal/internal.h>
                                        ^
compilation terminated.
```
Can we externalize this?

dacap:
Hi @winterheart, I tried to do that but failed. Anyway I think in a near future I will remove the shared configuration (it makes maintenance unnecessarily more complex) and take a more google chrome-like approach.

skrat:
Is there any solution to this? It fails to build.

dacap:
@skrat you need to[ clone the repository with submodules](https://github.com/aseprite/aseprite/blob/master/INSTALL.md#get-the-source-code)

yurivict:
`USE_SHARED_FREETYPE=ON` is meaningless when internal parts are used. They aren't installed with the shared packages. You should stop using internal parts.

PR #1512, opened 2017-06-07
Fixing Freetype and Harfbuzz system shared libraries support

winterheart:
Fixed linking system Freetype library issue #1453 and added support for system Harfbuzz.


dacap:
Hi @winterheart, I'll take a look at this later. Anyway, the idea behind the `ft::open_stream()` function is that it uses `base::open_file_raw()` so you can load ttf fonts with unicode filenames. From what I see, that functionality would be lost with this patch, is that right?

winterheart:
OK, I got the point. Can we use here FT_New_Memory_Face() instead?

dacap:
I would prefer to use the `FILE` interface because it avoids to load the entire file on memory (at least that is what I think, I'm not 100% sure) and `Arial Unicode.ttf` is 23.3 MB on macOS.

From my point of view this is an issue on the freetype2 library, because their solution to create a customize `FT_Stream` requires their internal header files (and these files shouldn't internal in the first place). Maybe a freetype2 patch?

winterheart:
@dacap hello again.
Can you review this try?

dacap:
Hi @winterheart, sorry for my late response on this one, I'll try to review this after fixing some serious issues on Windows 10 for Surface devices.

PR #1563, opened 2017-08-29
Fix bug when closing Undo History with a keyboard shortcut

### LINKED ISSUE #1423: Undo History disappears when closed using keyboard shortcut
For some reason, every time I close the Undo History window using the keyboard shortcut, I become unable to open it again, while this visual bug persists where it used to show up. Closing it with the X button seems OK though.

screenshot 2017-02-24 18 11 15

* Aseprite version: 1.1.13 Steam version
* System: macOS 10.12.3, MacBook Pro, Early 2015

cebolan:
Closes #1423

Apparently the bug was happening because when the Undo History window was closed using the keyboard shortcut, the message kCloseMessage was never being sent to the window. This is probably not the only cause, but making UndoHistoryCommand close the window instead of just hiding it solves the problem.

I tested it several times, including switching files while the undo history window was open. Everything seems to work as expected.

### DIFF
--- a/src/app/commands/cmd_undo_history.cpp
+++ b/src/app/commands/cmd_undo_history.cpp
@@ -246,7 +246,7 @@ void UndoHistoryCommand::onExecute(Context* ctx)
     g_window = new UndoHistoryWindow(ctx);
 
   if (g_window->isVisible())
-    g_window->setVisible(false);
+    g_window->closeWindow(nullptr);
   else
     g_window->openWindow();
 }


dacap:
Hi @cebolan, thanks for this PR. Would you mind to [sign our CLA](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla)?

cebolan:
I just signed it. (:

dacap:
Thanks @cebolan, I've just merged your patch

Issue #951, opened 2016-02-03
File dialog not listing folders on XFS filesystem


imagn8nmechanic:
In aseprite-1.1-beta1 and higher the file dialog does not list folders on an XFS filesystem. This happens in all instances of the dialog (Open, Save, Save As, Save Copy As, Export, Import).

OS: Linux Mint Debian Edition, 64-bit, Debian version 8.3

aseprite-1.1-beta1: Folders NOT appearing in XFS filesystem
![aseprite v1 1-beta1 no folders in xfs filesystem](https://cloud.githubusercontent.com/assets/10121814/12771840/a7fe038e-c9e9-11e5-83a3-0f12f1211c58.png)

aseprite-1.0.9: Folders appearing in XFS filesystem
![aseprite v1 0 9 - folders appear in xfs filesystem](https://cloud.githubusercontent.com/assets/10121814/12771846/b4591d30-c9e9-11e5-82d4-b5a352e09672.png)

Issue #1530, opened 2017-07-05
aseprite is not able to browse mounted exfat partition


trollworkout:
Aseprite is not able to read read off a exFAT mounted drive.

exfat-utils adds ability to mount exFAT drives (usually USB thumbdrives are exfat)

seems like Aseprite is not able to read exfat or even fat? FS under linux . It thinks every folder is an .ase file and tried to open it instead of opening the folder.

### Aseprite and System version

* Aseprite version: version number, installer/portable/Steam/beta/dev/commit-hash
1.2 beta 11

* System: Windows/macOS/Linux, version, distribution
elementary os loki  Linux 0.4.1 based on ubuntu LTS

![peek 2017-07-05 06-49](https://user-images.githubusercontent.com/6987027/27861289-3bd1efa6-614e-11e7-8d18-ae8a29122f04.gif)

PR #1611, opened 2017-11-09
Avoid dirent.d_type since it is not POSIX compatible.

opeter:
This fixes the empty file dialog bug on XFS/JFS/... from issue #951

### DIFF
--- a/src/app/file_system.cpp
+++ b/src/app/file_system.cpp
@@ -37,6 +37,7 @@
   #define MYPC_CSLID  "::{20D04FE0-3AEA-1069-A2D8-08002B30309D}"
 #else
   #include <dirent.h>
+  #include <sys/stat.h>
 #endif
 
 //////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


dacap:
Hi @opeter, thanks for this PR. Please check our [CLA signing process](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla).

opeter:
Hi @dacap, CLA signing process has been completed and merged a while ago.  (https://github.com/aseprite/opensource/pull/22)

What about the patch?  Seems like it would already fix two issues #1530 and #951.  Is there anything else I need to to from my side?

dacap:
@opeter sorry for the late response in this one, I wasn't able to test the change in Ubuntu but I'll try to do it today.

dacap:
As I cannot test XFS filesystem or exFAT right now, I'll merge the patch, I've tried on Ext4 and it works as usual so basically it can only fix things. Thanks for the contribution @opeter.

dacap:
Cherry picked here 70f0496cd94741f3dc2ad4a962221807edbc89c7

PR #1635, opened 2017-12-16
Update arne32.gpl to v20e5

SagePtr:
Fixed missing black, reordered colors according to http://androidarts.com/palette/16pal_v20-Expanded_v5.gif , few RGB values slightly changed to match v5 palette.


dacap:
Thanks @SagePtr

Issue #1636, opened 2017-12-24
Hex color value corruption when editing number with palette unlocked


jdeeny:
While editing the hex color value, zeros are unexpectedly added to the value, making it very difficult to enter a number.

This only occurs when the palette is unlocked. When the palette is locked, the value can be edited as expected.

![wtf](https://user-images.githubusercontent.com/409859/34324662-03fbfdda-e838-11e7-8b3b-5e170737de3b.gif)

### Aseprite and System version

Two reports of this issue:
User 1:
* Aseprite version: v1.2-dev git
* System: Linux, Fedora 27

User 2:
* Aseprite version: 1.2.6, 1.2.5
* System: Windows 10

oceanhahn:
This issue is related to the bug described in #1632.

jdeeny:
Thanks, I tried searching for an existing issue but missed yours.

Issue #1632, opened 2017-12-12
Bug: Colour picker hex input field autocompletes too early; too often.


oceanhahn:
The text input field in the colour picker is way to hasty in autocompleting hex values, and it has made trying to enter them from keyboard practically impossible. _This problem only seems to occur in indexed colour mode._

As it is, the field only lets me enter three or four of six digits before it autocompletes. When I enter the next character in the hex value, the field clears any excess characters from the _beginning_ of the value (rather than the end), making a mess of the whole thing.

See for yourself:

![hex input](https://user-images.githubusercontent.com/11544843/33908036-1d0de37e-df55-11e7-8bb2-341074c5cb27.gif)

Perhaps the input field could be altered to wait until the input box loses focus before it tries to fill in vacant spaces with zeroes and/or recalculate a max/mix value? Alternatively, clearing excess characters from the _end_ of the hex value instead of the _beginning_ would probably also resolve this issue.

Copy/paste works, of course, and will do in the interim. But direct input from keyboard (rather than clipboard) is nightmarish.

Thanks for considering it! :D

..using ASEprite 1.2.5-x64 on Win10.

dacap:
Also reported on Steam recently: https://steamcommunity.com/app/431730/discussions/0/1639789306557163544/

PR #1637, opened 2017-12-24
Don't send a signal when setting a hex color is typed into the popup

jdeeny:
This fixes the #1636 by preventing the color change signal from being sent while the hex color is typed in.

### DIFF
--- a/src/app/ui/color_popup.cpp
+++ b/src/app/ui/color_popup.cpp
@@ -384,7 +384,7 @@ void ColorPopup::onColorHexEntryChange(const app::Color& color)
   // is writting in the text field.
   m_disableHexUpdate = true;
 
-  setColorWithSignal(color, ChangeType);
+  setColor(color, ChangeType);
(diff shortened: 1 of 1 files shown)


jdeeny:
I closed #1636 because it duplicates #1632, which is also fixed by this PR.

dacap:
Hi @jdeeny, thanks for your contribution, you will need to sign our [CLA to continue](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla). I'll check the fix later and see how it works.

jdeeny:
Thank you @dacap, I didn't realize that was required. I have submitted the PR and sent the email.

dacap:
Hi @jdeeny, this solution has a problem where the palette color is not updated when the hex value is edited (in Palette Edit mode, F4 key). I'll try to fix the original problem generating the signal anyway so the palette can be edited changing the hex value.

PR #1640, opened 2017-12-28
Externalize CMark library

### LINKED ISSUE #1574: Linux build currently installs extra files from vendored libraries
Performing a `ninja` build and install per the docs, I end up with the following extraneous files:

```
/bin/bsdcat
/bin/bsdcpio
/bin/bsdtar
/bin/cmark
/include/cmark.h
/include/cmark_export.h
/include/cmark_version.h
/include/json11.hpp
/lib/cmake/cmark-relwithdebinfo.cmake
/lib/cmake/cmark.cmake
/lib/libcmark.a
/lib/libcmark.so
/lib/libcmark.so.0.27.1
/lib/libjson11.a
/lib/pkgconfig/json11.pc
/lib/pkgconfig/libarchive.pc
/lib/pkgconfig/libcmark.pc
/share/man/man1/bsdcat.1
/share/man/man1/bsdcpio.1
/share/man/man1/bsdtar.1
/share/man/man1/cmark.1
/share/man/man3/cmark.3
```

I ran into similar problems with the (Make-based) [AUR package](https://aur.archlinux.org/packages/aseprite-git/), except that the install _failed_ trying to install `bsdcat` which hadn't been built.  Not sure why the behavior was different.

If the package _had_ built successfully, then Arch still would've refused to install it — I already have `cmark` installed, so Arch would reject the Aseprite package for trying to overwrite another package's file.

I eventually worked around this by deleting relevant chunks of various `CMakeLists.txt` files, but I don't know much about cmake and suspect there might be a better solution.

* Aseprite version: latest master, commit 2e4035312972f2cd48128142ca422df67cb89107
* System: Arch Linux x64, current as of 2017-09-18

winterheart:
Using ExternalProject_Add for third party library instead of using
subdirectory project. Now cmark target runs full cycle of fetching,
building and local installing of library.

Fixes #1574 (partially).

In future we can externalize all libraries by this way.


dacap:
Thanks @winterheart, I didn't know the ExternalProject_Add function. I'll make some comments in the code.

CMakeLists.txt
```diff
@@ -299,7 +299,7 @@ if(USE_SHARED_CMARK)
   find_path(CMARK_INCLUDE_DIRS NAMES cmark.h)
 else()
   add_definitions(-DCMARK_STATIC_DEFINE)
-  set(CMARK_LIBRARIES libcmark_static)
+  set(CMARK_LIBRARIES ${PROJECT_BINARY_DIR}/lib/libcmark.a)
```
dacap: Does it work on Windows? (where the extension is `.lib` instead of `.a`). Also why we use libcmark and not libcmark_static?

third_party/CMakeLists.txt
```diff
@@ -95,22 +96,16 @@ add_subdirectory(fmt)
...
-    list(APPEND copy_cmark_headers ${CMAKE_CURRENT_BINARY_DIR}/cmark/${fn_name})
-  endforeach()
-  add_custom_target(copy_cmark_headers DEPENDS ${copy_cmark_headers})
-  add_dependencies(libcmark_static copy_cmark_headers)
+
+  ExternalProject_Add(
+    cmark_project
+    SOURCE_DIR ${CMARK_DIR}
+    INSTALL_DIR ${CMAKE_BINARY_DIR}
+    CMAKE_ARGS -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR> -DCMARK_TESTS=OFF
```
dacap: Is this `<INSTALL_DIR>` in the parameters right? or in this way we avoid installing cmark files?

third_party/CMakeLists.txt
```diff
@@ -95,22 +96,16 @@ add_subdirectory(fmt)
...
+    BUILD_BYPRODUCTS ${PROJECT_BINARY_DIR}/lib/libcmark.a
```
dacap: Just seeing the `.a` again, is a portable way to do this?

winterheart:
Hi, @dacap 

Yes, indeed *.a is hard-coded. Replaced to cross-platform ${CMAKE_STATIC_LIBRARY_SUFFIX}. 
`libcmark_static` is cmark's project target and with ExternalProject_Add it not available anymore from aseprite's project.
Now cmark's install targets (includes and libraries) will be installed to ${CMAKE_BINARY_DIR} (ie where is build directory). Effectively there no much changes to aseprite build process.

third_party/CMakeLists.txt
winterheart: Yes, this is right, at least as documentation says.

dacap:
It have been a long time from this PR and I don't think I'll be able to review the possibility to make more third party libraries external. So I'll close it by now and in the future we might review the entire handling of third party libraries, but it's not a priority at the moment.

Issue #1660, opened 2018-02-13
Multibyte characters truncated in widget strings


yuxshao:
Essentially [this](https://community.aseprite.org/t/adaptation-to-multibyte-characters/988):
> I’m trying to make Japanese strings file.
> (Make from en.ini ja.ini)
> 
> But multi-byte string wasn’t read multi-byte string correctly.
> 
> [for example]
> (Characters in the ini file -> Characters on the screen)
> “あいうえお” -> “BDFHJ”
> 
> “あ”(code:3042) -> “B”(code:0042)
> “い”(code:3044) -> “D”(code:0044)
> “う”(code:3046) -> “F”(code:0046)
> “え”(code:3048) -> “H”(code:0048)
> “お”(code:304A) -> “J”(code:004A)
> 
> Is it solvable?

In `file.ini` changing `file = &File` to `file = &あいえうお` causes the following output:
![image](https://user-images.githubusercontent.com/9067601/36135606-3e1d01c2-105a-11e8-9e0c-29a7a8c2cee8.png)

### Aseprite and System version

* Aseprite version: 1.3-dev, c3a0c00a623ea748aa091fb4517769a8bdc5deb7
* System: Arch Linux

dacap:
Fixed with #1661

PR #1661, opened 2018-02-13
Stop mnenomic processing from truncating multibyte characters

yuxshao:
Addresses #1660 though I'm not sure if it's the cleanest/most obvious way...

`newText.push_back(*it)` used to push each subsequent symbol casted to a `char` even if it was multibyte. This is avoided by initializing `newText` as a `std::wstring` and converting to `std::string` when setting widget text.

Looks like this on my machine now:
image
Not sure what might be a good way to make a little sharper. Would an extension to pixel fonts for other character sets be a reasonable idea? (or was this behaviour intentional so that widgets only have ASCII text?)

Thanks for the awesome software by the way :)

### DIFF
--- a/src/ui/widget.cpp
+++ b/src/ui/widget.cpp
@@ -1368,7 +1368,7 @@ void Widget::processMnemonicFromText(int escapeChar)
   if (!hasText())
     return;
 
-  std::string newText;
+  std::wstring newText; // for utf8 support during push_back()
   if (!m_text.empty())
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @yuxshao for this fix, I'll give a try and merge the PR soon.

dacap:
@yuxshao I've just merged the PR, thanks for the patch!

Issue #1649, opened 2018-01-14
Broken ellipse sizes


KashouC:
The ellipse tool produces broken circles at very specific sizes. A 118x17 circle is not fully enclosed and missing pixels at the edges.

Aseprite v1.2.6-x64
Windows 7 x64

Chaduke:
Confirmed it occurs at 115 x 18 thru 126 x 18

yuxshao:
@dacap If you haven't already started working on this, I'd be interested in trying to tackle this!

dacap:
@yuxshao I've been working in a new algorithm but never finished it. I'll try your PR in these days, it looks good for me (maybe in a future I'll replace all algorithms, but that's a long term solution).

Issue #868, opened 2015-11-20
Possibility to change angle in rectangles and ellipses


dacap:
So we can create ovals with any angle and rotated rectangles. This might be related to #818 

EDIT: Check #1665, #1649, and [this tweet](https://twitter.com/retronator/status/966619431668502533).

Hammster:
You've already commented on the tweet but i think its not bad for people to have some reference image :) https://twitter.com/retronator/status/966619431668502533

PR #1665, opened 2018-02-15
One possible fix for ellipse gaps

yuxshao:
Addresses #1649 using an implementation of the concepts on [this page](http://members.chello.at/~easyfilter/Bresenham.pdf#%5B%7B%22num%22%3A43%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C102.6%2C755.5%2C0%5D) and [this page](http://members.chello.at/~easyfilter/Bresenham.pdf#%5B%7B%22num%22%3A28%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C109.8%2C724.2%2C0%5D). I think the original also uses this algorithm but traces paths from both the x-pole and y-pole, getting them to meet at the place where the ellipse slopes 45 degrees, but don't quite meet in a few specific cases ([this](https://user-images.githubusercontent.com/9067601/36247378-d2fc072c-1201-11e8-8c07-8199fd304286.png) in particular if it's readable). This one traces a single path from the 'wider' pole down to the 'sharper' pole.

It has a few differences in behaviour from the original - along sharper turns it picks the pixel closer to the ellipse centre a few times, but it doesn't leave any gaps. It also has extra logic to behave the same when `rx` or `ry` `< 2`. I have a test [here](https://github.com/yuxshao/ellipse-test) with sample output showing what I mean.

I'm really sorry if this is throwing away a bunch of really important code! It was just an experiment to see what was going on, and I thought it could lead to a more transparent approach. I don't mean to barge in and make changes to everything (it's just an offering...).

How does it look? Does it compromise the original behaviour/code too much?

### DIFF
--- a/src/doc/algo.cpp
+++ b/src/doc/algo.cpp
@@ -63,19 +63,22 @@ void algo_line(int x1, int y1, int x2, int y2, void* data, AlgoPixel proc)
 /* Additional helper functions for the ellipse-drawing helper function
    below corresponding to routines in Bresenham's algorithm. */
 namespace {
-  int bresenham_ellipse_error(int rx, int ry, int x, int y) {
+  int bresenham_ellipse_error(int rx, int ry, int x, int y)
+  {
     return x*x*ry*ry + y*y*rx*rx - rx*rx*ry*ry;
   }
 
   // Initialize positions x and y for Bresenham's algorithm
-  void bresenham_ellipse_init(int rx, int ry, int *px, int *py) {
+  void bresenham_ellipse_init(int rx, int ry, int *px, int *py)
+  {
     // Start at the fatter pole
     if (rx > ry) { *px = 0; *py = ry; }
     else { *px = rx; *py = 0; }
   }
 
   // Move to next pixel to draw, according to Bresenham's algorithm
-  void bresenham_ellipse_step(int rx, int ry, int *px, int *py) {
+  void bresenham_ellipse_step(int rx, int ry, int *px, int *py)
+  {
     int &x = *px, &y = *py;
     // Move towards the skinnier pole. Having 2 cases isn't needed, but it ensures
     // swapping rx and ry is the same as rotating 90 degrees.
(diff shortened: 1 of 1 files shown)


dacap:
At the moment I'm not sure if merging this PR or using a [new implementation](http://members.chello.at/easyfilter/canvas.html) (in addition to add [rotated ellipses](https://github.com/aseprite/aseprite/issues/868))

src/doc/algo.cpp
```diff
@@ -60,6 +60,45 @@ void algo_line(int x1, int y1, int x2, int y2, void* data, AlgoPixel proc)
...
+/* Additional helper functions for the ellipse-drawing helper function
+   below corresponding to routines in Bresenham's algorithm. */
+namespace {
+  int bresenham_ellipse_error(int rx, int ry, int x, int y)
+  {
+    return x*x*ry*ry + y*y*rx*rx - rx*rx*ry*ry;
+  }
+
+  // Initialize positions x and y for Bresenham's algorithm
+  void bresenham_ellipse_init(int rx, int ry, int *px, int *py)
```
dacap: Instead of using pointers here you can use references `(int rx, int ry, int& x, int& y)`

src/doc/algo.cpp
```diff
@@ -60,6 +60,45 @@ void algo_line(int x1, int y1, int x2, int y2, void* data, AlgoPixel proc)
...
+  {
+    // Start at the fatter pole
+    if (rx > ry) { *px = 0; *py = ry; }
+    else { *px = rx; *py = 0; }
+  }
+
+  // Move to next pixel to draw, according to Bresenham's algorithm
+  void bresenham_ellipse_step(int rx, int ry, int *px, int *py)
```
dacap: Same here: `(int rx, int ry, int& x, int& y)` and no need for the next line `int &x = *px, &y = *py;`

dacap:
I've just merged this, and I'll make the minor fixes in the review. Don't worry and thanks for the PR @yuxshao!

yuxshao:
Hey, thanks so much! I'm still thinking in a C mindset evidently :sweat:
The rotated thing looks so awesome too!

PR #1686, opened 2018-03-20
Initial translation into Russian

mPolr:
Initial translation into Russian language

### DIFF
--- a/data/strings/ru.ini
+++ b/data/strings/ru.ini
@@ -1,0 +1,1219 @@
+# Aseprite
+# Copyright (C) 2016-2018 by David Capello
+
+[advanced_mode]
+title = Предупреждение - Важно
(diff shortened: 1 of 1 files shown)


dacap:
Hi @mPolr, at the moment I'm thinking that handling [translations through extensions](https://github.com/aseprite/aseprite/issues/124#issuecomment-374410225) might be a best option than integrating the translations into Aseprite code. Check these:

* [Japanese](https://twitter.com/clrfnd/status/976021678915399680)
* [Korean](https://twitter.com/etude87/status/976443390928633858)

As I cannot evaluate PRs in other languages correctly, I guess it would work better if the translation is decentralized in a language specific group. (e.g. you could create the Russian translation and distribute it through https://itch.io)

PR #1716, opened 2018-04-30
Update stream.cpp

tcyrus:
Replaced `malloc` + `memset` with `calloc`. This also avoids undefined behavior if `memset` gets called on a `NULL` pointer.


blogcommoda:
normal

dacap:
Thanks @tcyrus, I've fixed this reordering memset after the check just to avoid using calloc (just in case in the future I use config.h and base_malloc).

PR #1721, opened 2018-05-08
Update INSTALL.md

userid0:
fix build skia command on windows

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -275,12 +275,12 @@ fails.)
 
 For VS2015:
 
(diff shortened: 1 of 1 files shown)


dacap:
Hi @userid0, these commands are for the Windows command line (`cmd.exe`). Double-quotes `""` is the escape sequence to specify a single quote `"` inside a string `"..."` in `cmd.exe`. Are you using Bash or something similar? I've to test the commands again on Windows.

dacap:
I'll give a try to this tomorrow. I hope Microsoft didn't change the way `cmd.exe` works.

PR #1800, opened 2018-07-25
Selection of a range in color pallet now scrolls.

### LINKED ISSUE #1796: Color palette does not scroll
The color palette does not scroll while moving the color toward up or down.

aspritebug

* Aseprite version: v1.3-dev
* System: Windows 10

Vcoder89:
Fixes #1796
Simply added an update_scroll call after the mouse dragging is processed.


dacap:
I wasn't able to test this patch, but from a quick peek I can say that we must not change the scroll or invalidate the widget in the same `Timeline::onPaint()` event handler (also, it looks like it might generate an infinite paint/invalidation/validation/paint cycle until we stop dragging the colors).

dacap:
@vcoder89 just in case, @gasparoken already has a fix for this issue and he’ll create a new PR soon

Vcoder89:
That’s great, I’ll find another one to work on..

PR #1816, opened 2018-08-02
Remove input param in Timeline::getDrawable functions

Gasparoken:
Removed unnecesary input argument in functions getDrawableLayers(...) and getDrawableFrames(...) from timeline.cpp and timeline.h


dacap:
👍 looks good

PR #1818, opened 2018-08-03
Fix timeline does not scroll with range dragging (fix #1780)

### LINKED ISSUE #1780: Timeline does not scroll when frames or cels are dragged to its margins.
timeline scroll

The Timeline scrolls during selection for frames, cels, and layers when the cursor is dragged to its outer bounds, but selections cannot be _moved_ this way when they are dragged. This makes it easy to collect many cels, frames, _et cetera_ over large ranges (which is fantastic!), but difficult to _move them long distances_ (not so great).

It would be great if the Timeline could be revised to allow scolling on drag.

Thanks for considering it! :D 
_(Somewhat related to #1344 and possibly related to #1744.)_

..using ASEprite 1.2.8-x64 on Win10.

Gasparoken:


src/app/ui/timeline/timeline.cpp
```diff
@@ -1050,6 +1050,32 @@ bool Timeline::onProcessMessage(Message* msg)
...
+            case STATE_MOVING_RANGE: {
+                frame_t firstDrawableFrame;
+                frame_t lastDrawableFrame;
+                getDrawableFrames(&firstDrawableFrame, &lastDrawableFrame);
+
+                layer_t firstDrawableLayer;
+                layer_t lastDrawableLayer;
+                getDrawableLayers(&firstDrawableLayer, &lastDrawableLayer);
+
+                int newLayer = hit.layer;
```
dacap: This should be `layer_t newLayer = `

src/app/ui/timeline/timeline.cpp
```diff
@@ -1050,6 +1050,32 @@ bool Timeline::onProcessMessage(Message* msg)
...
+                int newFrame = hit.frame;
```
dacap: This should be `frame_t newFrame = `

src/app/ui/timeline/timeline.cpp
```diff
@@ -1050,6 +1050,32 @@ bool Timeline::onProcessMessage(Message* msg)
...
+                    newFrame = firstDrawableFrame - 1;
+                else if (hit.frame > lastDrawableFrame)
+                    newFrame = lastDrawableFrame + 1;
+                if (hit.layer < firstDrawableLayer)
+                    newLayer = firstDrawableLayer - 1;
+                else if (hit.layer > lastDrawableLayer)
+                    newLayer = lastDrawableLayer + 1;
+
+                showCel(newLayer,newFrame);
+
```
dacap: We can remove this blank line

PR #1884, opened 2018-09-13
Enhancement canvas mirror sides (fix #1826)

### LINKED ISSUE #1826: Suggestion: Setting Canvas size, use Shift to mirror sides.
Hey David,

While dragging the edges of the canvas to adjust the canvas size, rather than inputting a value, it would be nice to be able to use something like shift+mouse drag to keep the same distance on both sides.

Cheers!

Gasparoken:
If we hold down shift while drag the edges of the canvas to adjust the canvas size, it will keep the same distance on both sides.


dacap:
👍 cherry picked c05c3dad43e34dda854a39ebfa13d59529767466

PR #1907, opened 2018-10-26
Fix format specifiers

janisozaur:


dacap:
Thanks @janisozaur, there are several places where it happens. I'll close this PR because this is a global change that should be done in all places where a printf-like function is used and a task to be done along with:
1. avoiding all kind of warnings in the compilation with all compilers
2. replacing printf-like functions with `fmt::format()`

Issue #1644, opened 2018-01-01
Bug - CLI - Combination of parameters results in jumbled layer order.


johanwendin:
Hello.

While using the CLI, if you combine the options: 
"--ignore-layer" (to ignore a specific layer, or multiple layers) as well as "--split-layers", the resulting json has the "--list-layers" layer-list order randomized.

Same thing happens if you combine "--split-layers" and specify which layers to include using multiple "--layer"

This also means that it's just not the "layers" meta block that's out-of-order, but the frames list them in the same order as the "layers" meta block, resulting in incorrect sprites if you render them in the order specified within the json.

i.e:

```shell
aseprite -b --split-layers --list-layers filename.ase --format json-array
```

gives: (correct order)
```javascript
"layers": [
 { "name": "_bg", "opacity": 255, "blendMode": "normal" },
 { "name": "legs", "opacity": 255, "blendMode": "normal" },
 { "name": "back_arm", "opacity": 255, "blendMode": "normal" },
 { "name": "body", "opacity": 255, "blendMode": "normal" },
 { "name": "front_arm", "opacity": 255, "blendMode": "normal" },
 { "name": "head", "opacity": 255, "blendMode": "normal" }
]
```

whereas:
```aseprite -b --split-layers --list-layers --ignore-layer _bg filename.ase --format json-array```

gives: (examples, this will vary randomly)

*(first run)*
```javascript
"layers": [
 { "name": "head", "opacity": 255, "blendMode": "normal" },
 { "name": "legs", "opacity": 255, "blendMode": "normal" },
 { "name": "front_arm", "opacity": 255, "blendMode": "normal" },
 { "name": "back_arm", "opacity": 255, "blendMode": "normal" },
 { "name": "body", "opacity": 255, "blendMode": "normal" }
]
```

*(second run)*
```javascript
"layers": [
 { "name": "front_arm", "opacity": 255, "blendMode": "normal" },
 { "name": "legs", "opacity": 255, "blendMode": "normal" },
 { "name": "back_arm", "opacity": 255, "blendMode": "normal" },
 { "name": "head", "opacity": 255, "blendMode": "normal" },
 { "name": "body", "opacity": 255, "blendMode": "normal" }
]
```

*(third run)*
```javascript
"layers": [
 { "name": "back_arm", "opacity": 255, "blendMode": "normal" },
 { "name": "legs", "opacity": 255, "blendMode": "normal" },
[...]

PR #1924, opened 2018-11-10
Fixing bug #1644: Combination of parameters results in jumbled layer …

dncampo:
…order.

This fix iterates over layers, no matter if are visible or not


dacap:
👍 I've [cherry picked](https://github.com/aseprite/aseprite/commit/f6bb446031fc8ef751c1c300b6565639c7995c11) with a modified commit message, but everything looks fine.

PR #1927, opened 2018-11-12
Add default info on status bar (fix #1874)

### LINKED ISSUE #1874: Add default information displayed at the bottom
Okay the title is terrible but I'm not sure what to call it. The context bar at the bottom that shows how large your selection is, cursor coordinates, stuff like that. Right now it defaults to displaying nothing when the mouse is off screen or you have a window open, but I'd like it to always display basic information like image size, and when a selection is up show the selection size as well.

This is something that affects me a lot. Because VERY often I want to name files to include the canvas size, but I always forget to check the size before I open up the save as or export window, and then you can't find it anymore, so you have to close it, memorize the size, then open it up again.

Basically I'd like it to default to the "pen over canvas" context when it doesn't have anything else to show, possibly omitting the pen position.

Gasparoken:


dacap:
This is a recommendation for the future: don't use branch names with `(` or `)`, because they don't work well in the shell (e.g. we need to add quotes to every git command e.g. `git checkout -b "Gasparoken-status-bar-default-text-(fix-#1874)" master`)

dacap:
Cherry picked b3771472c3d1143414a7e9bc294c42e4755e9ad1 with minor changes.

PR #1950, opened 2018-12-13
Add key mapping commands for symmetry

dncampo:
Now it is possible to associate symmetry with a selected shortcut.

### DIFF
--- a/src/app/commands/cmd_symmetry_mode.cpp
+++ b/src/app/commands/cmd_symmetry_mode.cpp
@@ -12,7 +12,12 @@
 #include "app/commands/command.h"
 #include "app/commands/params.h"
 #include "app/context.h"
+#include "app/doc.h"
+#include "app/i18n/strings.h"
 #include "app/pref/preferences.h"
+#include "app/ui_context.h"
+#include "app/ui/context_bar.h"
+#include "app/modules/gui.h"
 
 namespace app {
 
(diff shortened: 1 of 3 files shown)


src/app/commands/cmd_symmetry_mode.cpp
```diff
@@ -22,16 +24,44 @@ class SymmetryModeCommand : public Command {
...
 
 protected:
+  void onLoadParams(const Params& params) override;
   bool onEnabled(Context* context) override;
   bool onChecked(Context* context) override;
   void onExecute(Context* context) override;
+  std::string onGetFriendlyName() const override;
+
+private:
+    int m_mode = int(app::gen::SymmetryMode::NONE);
```
dacap: `m_mode` can be a `app::gen::SymmetryMode`

src/app/commands/cmd_symmetry_mode.cpp
```diff
@@ -22,16 +24,44 @@ class SymmetryModeCommand : public Command {
...
+    default:
+      return Strings::symmetry_toggle();
+  }
+}
+
+void SymmetryModeCommand::onLoadParams(const Params& params)
+{
+  Doc* doc = UIContext::instance()->activeDocument();
+  DocumentPreferences& docPref = Preferences::instance().document(doc);
+  m_mode = int(docPref.symmetry.mode());
```
dacap: The default `m_mode` must be `app::gen::SymmetryMode::NONE`, which actually is (see the `else` case below). So you can remove this line and line 57.

src/app/commands/cmd_symmetry_mode.cpp
```diff
@@ -22,16 +24,44 @@ class SymmetryModeCommand : public Command {
...
+  std::string mode = params.get("orientation");
+  if (mode == "vertical") m_mode ^= int(app::gen::SymmetryMode::VERTICAL);
```
dacap: No need for XOR here (just `m_mode = ...`).

src/app/commands/cmd_symmetry_mode.cpp
```diff
@@ -46,7 +76,14 @@ bool SymmetryModeCommand::onChecked(Context* ctx)
 void SymmetryModeCommand::onExecute(Context* ctx)
 {
   auto& enabled = Preferences::instance().symmetryMode.enabled;
-  enabled(!enabled());
+  DocumentPreferences& docPref = Preferences::instance().document(ctx->activeDocument());
+  docPref.symmetry.mode.setValue(app::gen::SymmetryMode(this->m_mode));
```
dacap: The `docPref.symmetry.mode` must not be modified if `m_mode` is NONE.

You can also use something like `docPref.symmetry.mode(docPref.symmetry.mode() ^ m_mode)` to change the preference value (instead of `setValue`, see that [`Option`](https://github.com/aseprite/aseprite/blob/06352f79e8569aead3320dd277eab4621a6d3929/src/app/pref/option.h#L49)`::operator()` is overloaded [without arguments](https://github.com/aseprite/aseprite/blob/06352f79e8569aead3320dd277eab4621a6d3929/src/app/pref/option.h#L58) to get the value, and [with an argument](https://github.com/aseprite/aseprite/blob/06352f79e8569aead3320dd277eab4621a6d3929/src/app/pref/option.h#L62) to set a value).

src/app/commands/cmd_symmetry_mode.cpp
```diff
@@ -46,7 +76,14 @@ bool SymmetryModeCommand::onChecked(Context* ctx)
...
+  if (this->m_mode == int(app::gen::SymmetryMode::NONE)) {
+    enabled(!enabled());
+  }
+  else {
+    enabled(true);
```
dacap: Inside this `else` we can do the XOR of the `docPref` symmetry mode depending on the `m_mode` value. (because when m_mode is NONE, we don't even need the `docPref` variable).

dacap:
Added in 91c5ed8fd41666f4114830c530ccef721aae2750 with some comments in a2648c26bf802d221d1e7a49dbb63b8fffdd1280 👍

Issue #1501, opened 2017-06-05
Add support to open/save .act files


dacap:
We should be able to load and save .act files.

Quote:
> There is no version number written in the file. The file is 768 or 772 bytes long and contains 256 RGB colors. The first color in the table is index zero. There are three bytes per color in the order red, green, blue. If the file is 772 bytes long there are 4 additional bytes remaining. Two bytes for the number of colors to use. Two bytes for the color index with the transparency color to use. If loaded into the Colors palette, the colors will be installed in the color swatch list as RGB colors.

Source: http://www.adobe.com/devnet-apps/photoshop/fileformatashtml/#50577411_pgfId-1070626

PR #1975, opened 2019-01-08
Add support for ACT palettes (#1501)

SupSuper:
Adds loading/saving Adobe palettes per issue #1501 
ACT palettes only support up to 256 colors so extra colors will be ignored on save.

There was already an [Adobe enum](https://github.com/SupSuper/aseprite/blob/de0f356ed1928d512d55ace55fd95edef7908545/src/dio/file_format.h#L18) in code but it's not used anywhere so might be worth removing to avoid confusion.

### DIFF
--- a/src/app/file/palette_file.cpp
+++ b/src/app/file/palette_file.cpp
@@ -18,6 +18,7 @@
 #include "base/string.h"
 #include "dio/detect_format.h"
 #include "doc/cel.h"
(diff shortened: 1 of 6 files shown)


dacap:
Thanks @SupSuper, I'll take a look to this PR soon.

dacap:
Thanks @SupSuper, I've just merged your PR.

PR #1977, opened 2019-01-09
Add aspect ratio on status bar (fix #1769)

### LINKED ISSUE #1769: Ability to display aspect ratio of current selection/shape tool in the bottom info panel
When making circles and boxes in different sizes it would be handy to be able to tell if they're the same aspect ratio instead of having to count the steps yourself. Not sure if this is as simple to implement as it sounds, since I don't know if aspect ratios might end up really strange and large values, but I don't think it would be too much of an issue.

If there are any good tricks to do this or think about it otherwise I'm open to suggestions.

Gasparoken:


dacap:
Cherry-picked here 3dc76d08c4152249485a3ce84f939fecb04f374a 

I've removed some changes in the `theme.xml` which were unrelated with the feature.

PR #1987, opened 2019-01-21
Add Import Sprite Sheet padding (fix #78)

### LINKED ISSUE #78: Sprite sheet padding on import
_From [Eiye...@gmail.com](https://code.google.com/u/100429032823048932647/) on February 21, 2012 13:48:52_

What do you need to do? I think thzt adding padding to import sprite sheet would be useful. Some tile sheets add spaces between sprite, and thus, importing them with the space is quite annoyant. How would you like to do it? Adding two input boxes to define space width/height would be perfect, don't you agree?

_Original issue: http://code.google.com/p/aseprite/issues/detail?id=78_

Gasparoken:


src/app/commands/cmd_import_sprite_sheet.cpp
```diff
@@ -144,14 +152,36 @@ class ImportSpriteSheetWindow : public app::gen::ImportSpriteSheet
       std::max<int>(1, h));
   }
 
+  gfx::Size getPaddingFromEntries() {
+  int padW = horizontalPadding()->textInt();
```
dacap:
```suggestion
    int padW = horizontalPadding()->textInt();
```
Minor change, the indentation needs a couple of white spaces.

src/app/commands/cmd_import_sprite_sheet.cpp
```diff
@@ -144,14 +152,36 @@ class ImportSpriteSheetWindow : public app::gen::ImportSpriteSheet
...
+  int padH = verticalPadding()->textInt();
```
dacap:
```suggestion
    int padH = verticalPadding()->textInt();
```

src/app/commands/cmd_import_sprite_sheet.cpp
```diff
@@ -144,14 +152,36 @@ class ImportSpriteSheetWindow : public app::gen::ImportSpriteSheet
...
   void onEntriesChange() {
     m_rect = getRectFromEntries();
+    m_padding = getPaddingFromEntries();
 
     // Redraw new rulers position
     if (m_editor) {
       EditorStatePtr state = m_editor->getState();
       if (SelectBoxState* boxState = dynamic_cast<SelectBoxState*>(state.get())) {
         boxState->setBoxBounds(m_rect);
+        if (int(boxState->getFlags()) & int(SelectBoxState::Flags::PaddingRulers)) {
```
dacap: I think this condition is always true, because when we are in the `ImportSpriteSheetWindow`, if `m_editor != nullptr` it means that we have changed the state of `m_editor` to `SelectBoxState` (from `ImportSpriteSheetWindow::captureEditor()`) with this `PaddingRulers` flag enabled. A possible nice change would be to have an assert:
```
ASSERT(int(boxState->getFlags()) & int(SelectBoxState::Flags::PaddingRulers));
```
Anyway as the next line (`boxState->setPaddingBounds()`) already does this, we can just remove this line/if-condition.

Another note: About the code that was already there, I'm not sure why I put a check `if (SelectBoxState* boxState = dynamic_cast<SelectBoxState*>(state.get()))`, it's like the state could be different to `SelectBoxState` in some special cases when `m_editor != nullptr`, but I don't remember why I put that check.

src/app/ui/editor/select_box_state.cpp
```diff
@@ -29,12 +30,16 @@ using namespace ui;
...
   : m_delegate(delegate)
-  , m_rulers(4)
+  , m_rulers((int(flags) & int(Flags::PaddingRulers))? 6: 4)
   , m_rulersDragAlign(0)
   , m_selectingBox(false)
   , m_flags(flags)
 {
   setBoxBounds(rc);
+  if (hasFlag(Flags::PaddingRulers)) {
+    gfx::SizeT<int> padding(0, 0);
```
dacap:
```suggestion
    gfx::Size padding(0, 0);
```
As we talked, we can use `gfx::Size` directly here.

src/app/ui/editor/select_box_state.cpp
```diff
@@ -250,47 +297,238 @@ void SelectBoxState::postRenderDecorator(EditorPostRender* render)
...
+                  if (y + padding.h > sp.y2())
+                    render->fillRect(dark, gfx::Rect( padXTips.back() + padding.w
+                                                    , y
+                                                    , sp.x2() - padXTips.back() - padding.w
+                                                    , sp.y2() - y));
+                  else
+                    render->fillRect(dark, gfx::Rect( padXTips.back() + padding.w
+                                                    , y
+                                                    , sp.x2() - padXTips.back() - padding.w
+                                                    , padding.h));
```
dacap: Minor change: I don't like the style of these lines 😬they could be just:
```
                if (padXTips.back() + padding.w < sp.x2()) {
                  if (y + padding.h > sp.y2())
                    render->fillRect(dark, gfx::Rect(padXTips.back() + padding.w, y,
                                                     sp.x2() - padXTips.back() - padding.w,
                                                     sp.y2() - y));
                  else
                    render->fillRect(dark, gfx::Rect(padXTips.back() + padding.w, y,
                                                     sp.x2() - padXTips.back() - padding.w,
                                                     padding.h));
                }
``` 
Same for other cases below.

src/app/ui/editor/select_box_state.h
```diff
@@ -26,6 +27,7 @@ namespace app {
     // Called each time the selected box is modified (e.g. rulers are
     // moved).
     virtual void onChangeRectangle(const gfx::Rect& rect) { }
+    virtual void onChangePadding(const gfx::SizeT<int>& padding) { }
```
dacap:
```suggestion
    virtual void onChangePadding(const gfx::Size& padding) { }
```
Just `gfx::Size` with a `#include "gfx/fwd.h"` above.

src/app/ui/editor/select_box_state.h
```diff
@@ -59,19 +61,32 @@ namespace app {
...
+    void setFlag(Flags flag);
+    void clearFlag(Flags flag);
 
     // Returns the bounding box arranged by the rulers.
     gfx::Rect getBoxBounds() const;
     void setBoxBounds(const gfx::Rect& rc);
 
+    // Get & Set the size of the padding rulers during Import Sprite Sheet Dialog
+    gfx::SizeT<int> getPaddingBounds() const;
+    void setPaddingBounds(gfx::SizeT<int>& padding);
```
dacap:
```suggestion
    void setPaddingBounds(const gfx::Size& padding);
```

src/app/ui/editor/select_box_state.h
```diff
@@ -59,19 +61,32 @@ namespace app {
...
     void setFlags(Flags flags);
+    void setFlag(Flags flag);
+    void clearFlag(Flags flag);
 
     // Returns the bounding box arranged by the rulers.
     gfx::Rect getBoxBounds() const;
     void setBoxBounds(const gfx::Rect& rc);
 
+    // Get & Set the size of the padding rulers during Import Sprite Sheet Dialog
+    gfx::SizeT<int> getPaddingBounds() const;
```
dacap:
```suggestion
    gfx::Size getPaddingBounds() const;
```

src/app/ui/editor/select_box_state.cpp
```diff
@@ -65,6 +85,30 @@ void SelectBoxState::setBoxBounds(const gfx::Rect& box)
...
+  int w = m_rulers[PV].position() - m_rulers[V2].position();
+  int h = m_rulers[PH].position() - m_rulers[H2].position();
+  if (w < 0)
+    w = 0;
+  if (h < 0)
+    h = 0;
+  return gfx::Size(w, h);
+}
+
+void SelectBoxState::setPaddingBounds(gfx::Size& padding)
```
dacap: I don't like 100% that this function receives the padding as an input/output parameter (because the `padding` argument will be modified). In the best case the `padding` should be `const gfx::Size&` and the `< 0` check should be in `ImportSpriteSheetWindow::getPaddingFromEntries()`.
```suggestion
void SelectBoxState::setPaddingBounds(const gfx::Size& padding)
```

src/app/commands/cmd_import_sprite_sheet.cpp
```diff
@@ -144,16 +152,35 @@ class ImportSpriteSheetWindow : public app::gen::ImportSpriteSheet
       std::max<int>(1, h));
   }
 
+  gfx::Size getPaddingFromEntries() {
+    int padW = horizontalPadding()->textInt();
+    int padH = verticalPadding()->textInt();
+    if (padW < 0) {
+      horizontalPadding()->setText("0");
```
dacap: There is no need to change the text of the `horizontalPadding` field:
```suggestion
```
It's better if we avoid modifying data when a `get` member functions is called.

src/app/commands/cmd_import_sprite_sheet.cpp
```diff
@@ -144,16 +152,35 @@ class ImportSpriteSheetWindow : public app::gen::ImportSpriteSheet
...
+      padW = 0;
+    }
+    if (padH < 0) {
+      verticalPadding()->setText("0");
```
dacap:
```suggestion
```

src/app/ui/editor/select_box_state.cpp
```diff
@@ -250,47 +289,225 @@ void SelectBoxState::postRenderDecorator(EditorPostRender* render)
   const gfx::Color gridColor = skin::SkinTheme::instance()->colors.selectBoxGrid();
   const gfx::Point mainOffset = editor->mainTilePosition();
   gfx::Rect rc = getBoxBounds();
+  gfx::Size padding;
+  if (hasFlag(Flags::PaddingRulers)) {
+    padding = getPaddingBounds();
+    if (padding.w < 0)
```
dacap: I think these `padding.w < 0` checks should be in `getPaddingBounds()`, so you can move 
```
   if (padding.w < 0)
      padding.w = 0;
    if (padding.h < 0)
      padding.h = 0;
```
inside `getPaddingBounds()` before calling `gfx::Size(w, h)` constructor with negative `w` and `h` values (side note: indeed, a `gfx::Size()` should be a `gfx::SizeT<unsigned int>`, to avoid using negative numbers, but that is a different history for other time).

PR #1990, opened 2019-01-25
Fix sheet size extrude export

dncampo:
It fix the problem of exporting a horizontal spriteSheet when it has a lot of frames and using extruding option.
Example:
too-many-frames-for-horizontal-sheet


dacap:
Cherry picked:
* 2e69138890e5a92ba1872decbc2cfe312b196de2 -> dee2dde520d7c710c66089febde5e19dbe42f30a
* eb90a2946e46441f3bf8c0ef6e9d70648f6a62ed -> 6ba71343dc955c5ffed4c6d1d86b42d5253b13b7

dacap:
After this some changes like 86464bc63cb52216fca8677538dcff002523d4ac calculates the texture size correctly when the best fit option is checked, and 67e0da478da465dddaca90cb8e409da33f5ec45c restore the `calculate_sheet_size()` almost to its original form after seeing that the `extrude` parameter is just one extra inner padding option in these algorithms.

PR #1993, opened 2019-01-28
Fix timeline left cel edge bug (fix #1879)

### LINKED ISSUE #1879: Display bug with Linked Cels at left edge of Timeline.
timeline-group-leftedge-display
🔼 <sub>_(All I am doing here is scrolling left and right.)_</sub>

This is a really trivial bug, but when a consecutive block of Linked Cels move off the left edge of the Timeline, the leftmost _visible_ one appears as the _initial_ Linked Cel when it is actually not the first in that block. This only happens when the Timeline is aligned perfectly to the left edge. The bug does not occur at the right edge.

It is a minor issue but maybe it would be an easy fix?
Anyway, thought you ought to know. :y 

..using ASEprite 1.2.9-x86 (installer) on Win10.

Gasparoken:


dacap:
Merged 20aa9992c7192de3423acefe3b2f3f7496cfaa7c

PR #1994, opened 2019-01-28
Fix unnecessary warning saving TGA (fix #1830)

### LINKED ISSUE #1830: Unnecessary warning saving TGA
When saving an RGB Color (not indexed) TGA, there is a warning about not supporting palette with alpha channel. Maybe this isn't needed unless you're saving an indexed image.

Gasparoken:


src/app/file/tga_format.cpp
```diff
@@ -43,7 +43,8 @@ class TgaFormat : public FileFormat {
       FILE_SUPPORT_RGBA |
       FILE_SUPPORT_GRAY |
       FILE_SUPPORT_INDEXED |
-      FILE_SUPPORT_SEQUENCES;
+      FILE_SUPPORT_SEQUENCES |
+      FILE_SUPPORT_PALETTE_WITH_ALPHA;
```
dacap: Check if this new addition doesn't add a new problem saving indexed files that contain palette entries with alpha (because TGA files doesn't support palettes with alpha, a least, that is what I remember). For example, the warning must be shown if the file is indexed and contain entries with alpha.

dacap:
Cherry-picked b3eba1c952b204325399f9660b5ddb2159d2cd0c 👍

PR #1996, opened 2019-01-30
Fix shift after flip throws error  (fix #1873)

### LINKED ISSUE #1873: Edit > Flip then Edit > Shift throws error
## Steps:

 1. New empty file
 2. Select an area (e.g. with `ctrl+A`)
 3. `Edit > Flip Horizontal` (or vertical)
 4. `Edit > Shift > Left`  (or any direction)

## Throws:

> Cannot modify the sprite.
> It is being used by another command.

* Aseprite version: **1.2.9-64 (Steam)**
* System: **win 10**

Gasparoken:


src/app/cmd/shift_masked_cel.cpp
```diff
@@ -40,20 +41,13 @@ void ShiftMaskedCel::onUndo()
...
-  int y = cel->y();
-
-  mask->offsetOrigin(-x, -y);
-  doc::algorithm::shift_image_with_mask(image, mask, dx, dy);
-  mask->offsetOrigin(x, y);
-
-  image->incrementVersion();
+  doc::algorithm::shift_image_with_mask(cel, mask, dx, dy);
+  
+
```
dacap: Remove these two empty lines.

src/app/ui/editor/moving_pixels_state.cpp
```diff
@@ -501,9 +506,25 @@ void MovingPixelsState::onBeforeCommandExecution(CommandExecutionEvent& ev)
...
   // We don't need to drop the pixels if a MoveMaskCommand of Content is executed.
-  if (MoveMaskCommand* moveMaskCmd = dynamic_cast<MoveMaskCommand*>(ev.command())) {
+  if (MoveMaskCommand* moveMaskCmd = dynamic_cast<MoveMaskCommand*>(command)) {
     if (moveMaskCmd->getTarget() == MoveMaskCommand::Content) {
       // Do not drop pixels
+      // Verify Shift condition of the MoveMaskCommand (i.e. wrap = true)
+      if (moveMaskCmd->isWrap()) {
+        int dx = 0;
+        int dy = 0;
+        MoveThing::Direction direction = moveMaskCmd->getDirection();
```
dacap: I think that instead of getting the direction we can the delta directly:
```c++
gfx::Point delta = moveMaskCmd->getDelta(UIContext::instance());
m_pixelsMovement->shift(delta.x, delta.y);
```

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
+    dy = aux;
+  }
+  else if (sin(angle) < -sqrt(2)/2) {
+    double aux;
+    aux = dx;
+    dx = dy;
+    dy = -aux;
+  }
+  Image* imageTmp = Image::create(image->pixelFormat(), image->bounds().w, image->bounds().h);
+  imageTmp->copy(image, gfx::Clip(0, 0, 0, 0, image->bounds().w, image->bounds().h));
```
dacap: Can we just keep the old code to create the image copy?
```c++
// To simplify the algorithm we use a copy of the original image, we
// could avoid this copy swapping rows and columns.
ImageRef crop(crop_image(image, bounds.x, bounds.y, bounds.w, bounds.h,
                         image->maskColor()));	
```
Also there is a memory leak here because the `imageTmp` is a raw pointer (`Image*`, which requires a manual `delete` which is missing). Use a smart pointer in these cases (`ImageRef`) like in the old code (I would use most of the old code as possible).

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
+                         mask->bounds().h);
+
+  // Making one combined image: Image with the Mask Bounds (unfilled spaces were filled with mask color)
+  compImage->copy(cel->image(), gfx::Clip(cel->position().x-compCelBounds.x,
+                                          cel->position().y-compCelBounds.y,
+                                          0, 0,
+                                          cel->bounds().w, cel->bounds().h));
+
+  // Making a copy of only the image which will be shiftted
+  Image* imageToShift = Image::create(compImage->pixelFormat(), maskedBounds.w, maskedBounds.h);
```
dacap: Memory leak. Use `ImageRef imageToShift(...);`

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
-  while (u < 0) u += bounds.w;
-  while (v < 0) v += bounds.h;
-
-  for (int y=0; y<bounds.h; ++y) {
-    for (int x=0; x<bounds.w; ++x) {
-      put_pixel(image,
-                bounds.x + ((u + x) % bounds.w),
-                bounds.y + ((v + y) % bounds.h),
-                get_pixel(crop.get(), x, y));
+  if (cos(angle) < -sqrt(2)/2) {
```
dacap: I'll suppose that this is right 😅

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
+    }
+  }
+
+  // Bounds and Image shrinking (we have to fit compound image (compImage) and bounds (compCelBounds))
+  gfx::Rect newBounds(0, 0, compImage->bounds().w, compImage->bounds().h);
+  if (algorithm::shrink_bounds(compImage, newBounds, compImage->maskColor())) {
+    compCelBounds.offset(newBounds.x, newBounds.y);
+    compCelBounds.setSize(newBounds.size());
+  }
+  Image* finalImage = Image::create(compImage->pixelFormat(), compCelBounds.w, compCelBounds.h);
```
dacap: Use an `ImageRef` preferably so no `(ImageRef)` cast is needed below (which for smart pointers like `ImageRef`, it's better if we use the `ImageRef(...)` syntax instead of the old C-like syntax `(ImageRef)rawPointer`).

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
   }
 }
 
+void shift_image_with_mask(Cel* cel, const Mask* mask, int dx, int dy)
+{
+  ASSERT(!cel->bounds().isEmpty());
+  ASSERT(!mask->bounds().isEmpty());
+
+  // Saving initial cel->position() usefull at the end of this function to adjust the cel bounds in shrink process.
+  gfx::Point initialCelPos(cel->position());
```
dacap: Unused variable, `initialCelPos` can be removed.

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
+
+  // Making a image which bounds are equal to the UNION of cel->bounds() and mask->bounds()
+  int minX = MIN(cel->bounds().x, mask->bounds().x);
+  int minY = MIN(cel->bounds().y, mask->bounds().y);
+  int maxX = MAX(cel->bounds().x+cel->bounds().w, mask->bounds().x+mask->bounds().w);
+  int maxY = MAX(cel->bounds().y+cel->bounds().h, mask->bounds().y+mask->bounds().h);
+  gfx::Rect compCelBounds = gfx::Rect(minX, minY, maxX - minX, maxY - minY);
+  Image* compImage = Image::create(cel->image()->pixelFormat(), compCelBounds.w, compCelBounds.h);
```
dacap: Memory leak. Use `ImageRef compImage(...);`.

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
+
+  // Saving initial cel->position() usefull at the end of this function to adjust the cel bounds in shrink process.
+  gfx::Point initialCelPos(cel->position());
+
+  // Making a image which bounds are equal to the UNION of cel->bounds() and mask->bounds()
+  int minX = MIN(cel->bounds().x, mask->bounds().x);
+  int minY = MIN(cel->bounds().y, mask->bounds().y);
+  int maxX = MAX(cel->bounds().x+cel->bounds().w, mask->bounds().x+mask->bounds().w);
+  int maxY = MAX(cel->bounds().y+cel->bounds().h, mask->bounds().y+mask->bounds().h);
+  gfx::Rect compCelBounds = gfx::Rect(minX, minY, maxX - minX, maxY - minY);
```
dacap: The union of two rectangles can be calculated with the [`gfx::Rect::operator|()`](https://github.com/aseprite/laf/blob/eee4fa27f0f13f7ee03a98283cc41ce1d49c6aa3/gfx/rect.h#L328) (which is a shortcut for [`gfx::Rect::createUnion`](https://github.com/aseprite/laf/blob/eee4fa27f0f13f7ee03a98283cc41ce1d49c6aa3/gfx/rect.h#L247)):
```c++
gfx::Rect compCelBounds = cel->bounds() | mask->bounds();
```
In that way I think you can remove `minX/Y` and `maxX/Y` variables.

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
+    for (int x=initialX; x<finalX; ++x) {
+        put_pixel(compImage,
+                  initialX + (maskedBounds.w + dx + x-initialX) % maskedBounds.w,
+                  initialY + (maskedBounds.h + dy + y-initialY) % maskedBounds.h,
+                  get_pixel(imageToShift, x - initialX, y - initialY));
+    }
+  }
+
+  // Bounds and Image shrinking (we have to fit compound image (compImage) and bounds (compCelBounds))
+  gfx::Rect newBounds(0, 0, compImage->bounds().w, compImage->bounds().h);
```
dacap: Simplify:
```suggestion
  gfx::Rect newBounds = compImage->bounds();
```
See that [`Image::bounds()`](https://github.com/aseprite/aseprite/blob/b8d5e86a8df3744449defb9de00cf3371e7cce22/src/doc/image.h#L52) ([`gfx::ImageSpec::bounds()`](https://github.com/aseprite/aseprite/blob/b8d5e86a8df3744449defb9de00cf3371e7cce22/src/doc/image_spec.h#L41)) already creates a new rectangle with `origin = gfx::Point(0,0)`.

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,111 @@
...
+  if (algorithm::shrink_bounds(compImage, newBounds, compImage->maskColor())) {
+    compCelBounds.offset(newBounds.x, newBounds.y);
+    compCelBounds.setSize(newBounds.size());
+  }
+  Image* finalImage = Image::create(compImage->pixelFormat(), compCelBounds.w, compCelBounds.h);
+  finalImage->copy(compImage, gfx::Clip(0, 0, newBounds.x, newBounds.y,
+                                        compCelBounds.w, compCelBounds.h));
+  
+  // Final cel content assign
+  finalImage->incrementVersion();
```
dacap: Generally we would prefer all `incrementVersion()` calls to be inside `app::cmd` classes, but there are some exceptions. Just add a comment:
```suggestion
  finalImage->incrementVersion(); // TODO this should be in app::cmd module
```

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,101 @@
...
+  gfx::Rect compCelBounds = cel->bounds() | mask->bounds();
+  ImageRef compImage(Image::create(cel->image()->pixelFormat(), compCelBounds.w, compCelBounds.h));
+
+  // Making a Rect which represents the mask bounds of the original MASK relative to
+  // the new COMPOUND IMAGE (compImage)
+  gfx::Point maskCelGap(0, 0);
+  if (cel->bounds().x < mask->bounds().x)
+    maskCelGap.x = mask->bounds().x - cel->bounds().x;
+  else
+    maskCelGap.x = 0;
```
dacap: We can remove the `else` case because `maskCelGap.x` is already initialized `= 0`.

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,101 @@
...
+  if (cel->bounds().y < mask->bounds().y)
+    maskCelGap.y = mask->bounds().y - cel->bounds().y;
+  else
+    maskCelGap.y = 0;
```
dacap: Same here.

src/doc/algorithm/shift_image.cpp
```diff
@@ -20,31 +23,101 @@
...
+  // Making one combined image: Image with the Mask Bounds (unfilled spaces were filled with mask color)
+  compImage->copy(cel->image(), gfx::Clip(cel->position().x-compCelBounds.x,
+                                          cel->position().y-compCelBounds.y,
+                                          0, 0,
+                                          cel->bounds().w, cel->bounds().h));
+
+  // Making a copy of only the image which will be shiftted
+  ImageRef imageToShift(Image::create(compImage->pixelFormat(), maskedBounds.w, maskedBounds.h));
+  imageToShift->copy(compImage.get(), gfx::Clip(0, 0, maskedBounds.x, maskedBounds.y,
+                                                imageToShift->bounds().w, imageToShift->bounds().h));
```
dacap: This can be simplified:
```c++
imageToShift->copy(compImage.get(), gfx::Clip(0, 0, maskedBounds));
```
Because `imageToShift->bounds().w == maskedBounds.w` and `imageToShift->bounds().h == maskedBounds.h`

Issue #1926, opened 2018-11-12
Trim by grid


dacap:
The trim option on Export Sprite Sheet or the CLI could include an alternative behavior to trim by grid tile instead of by pixel (so only if a grid tile is completely empty, we trim it out).

---

*Implementation details:* `DocExporter::captureSamples()` creates the samples for the sprite sheet, if trim option is enabled the `algorithm::shrink_bounds` function is used to trim the bounds of the `sampleRender` image. With this new option we should use an alternative algorithm that shrink by tile grids instead of by pixel (some possible problems: getting the grid settings of the sample document).

dacap:
The only missing part of this issue is exposing the behavior in the CLI with an argument like `--trim-by-grid`

PR #2004, opened 2019-02-05
Trim by grid

dncampo:
Add supports for Trim by grid option when Exporting Sprite Sheet (issue #1926 )


data/pref.xml
```diff
@@ -425,6 +425,7 @@
       <option id="shape_padding" type="int" default="0" />
       <option id="inner_padding" type="int" default="0" />
       <option id="trim" type="bool" default="false" />
+      <option id="grid" type="bool" default="false" />
```
dacap: This option can be called `trim_by_grid` (in the future might be a possible `split_grid` so we have to avoid confusion)
```suggestion
      <option id="trim_by_grid" type="bool" default="false" />
```

data/strings/en.ini
```diff
@@ -521,6 +521,7 @@ border = Border:
 shape = Shape:
 inner = Inner:
 trim = Trim
+grid = by Grid
```
dacap: This can be called:
```suggestion
trim_by_grid = by Grid
```

src/app/commands/cmd_export_sprite_sheet.cpp
```diff
@@ -181,6 +181,7 @@ struct ExportSpriteSheetParams : public NewParams {
   Param<int> shapePadding { this, 0, "shapePadding" };
   Param<int> innerPadding { this, 0, "innerPadding" };
   Param<bool> trim { this, false, "trim" };
+  Param<bool> grid { this, false, "grid" };
```
dacap: This should be called:
```suggestion
  Param<bool> trimByGrid { this, false, "trimByGrid" };
```
Also I'm not sure if the `trim` parameter should be enabled automatically if `trimByGrid` is specified.

data/widgets/export_sprite_sheet.xml
```diff
@@ -17,7 +17,14 @@
...
       <check id="padding_enabled" text="@.padding" />
-      <check id="trim_enabled" text="@.trim" />
+      <check id="trim_enabled" text="@.trim" cell_align="left"/>
+      <grid columns="2" id="trim_container" cell_hspan="1">
+        <hbox>
+          <boxfiller />
+          <boxfiller />
+            <check id="grid_trim_enabled" text="@.grid" cell_align="right"/>
+        </hbox>
+        </grid>
```
dacap: Several changes here are not required (like two consecutive `<boxfiller />`s, or the `cell_align="left"` which does nothing because `cell_*` attributes are useful only when the direct parent XML element is a `<grid>`).

You can replace this with:
```xml
    <vbox>
      <check id="padding_enabled" text="@.padding" />
      <check id="trim_enabled" text="@.trim" />
      <hbox id="trim_container">
        <boxfiller />
        <check id="grid_trim_enabled" text="@.trim_by_grid" />
      </hbox>
      <check id="extrude_enabled" text="@.extrude" />
    </vbox>
```
Removing the `<grid>` (And using a `<hbox>` as the `trim_container`)

src/app/doc_exporter.cpp
```diff
@@ -599,8 +602,17 @@ void DocExporter::captureSamples(Samples& samples)
           sample.setTrimmedBounds(frameBounds = gfx::Rect(0, 0, 0, 0));
         }
 
-        if (m_trimCels)
+        if (m_trimCels) {
+          if (m_trimByGrid) {
+            auto& docPref = Preferences::instance().document(doc);
+            gfx::Point startingPoint(frameBounds.x, frameBounds.y);
+            gfx::Point endingPoint(frameBounds.x + frameBounds.w, frameBounds.y + frameBounds.h);
+            gfx::Point posTopLeft = snap_to_grid(docPref.grid.bounds(), startingPoint, PreferSnapTo::FloorGrid);
```
dacap: Too long lines here, and we can use the `origin()` and `point2()` member functions from `gfx::Rect` as the starting/ending points:
```c++
            gfx::Point posTopLeft =
              snap_to_grid(docPref.grid.bounds(),
                           frameBounds.origin(),
                           PreferSnapTo::FloorGrid);
            gfx::Point posBottomRight =
              snap_to_grid(docPref.grid.bounds(),
                           frameBounds.point2(),
                           PreferSnapTo::CeilGrid);
            frameBounds = gfx::Rect(posTopLeft, posBottomRight);
```

src/app/snap_to_grid.cpp
```diff
@@ -38,12 +38,21 @@ gfx::Point snap_to_grid(const gfx::Rect& grid,
...
     case PreferSnapTo::BoxOrigin:
+    case PreferSnapTo::FloorGrid:
       d = std::div(point.x-dx.rem, grid.w);
       newPoint.x = dx.rem + d.quot*grid.w;
 
       d = std::div(point.y-dy.rem, grid.h);
       newPoint.y = dy.rem + d.quot*grid.h;
       break;
+
+    case PreferSnapTo::CeilGrid:
```
dacap: Not sure if this new `PreferSnapTo::CeilGrid` mode is required or when it's required. Do you have an image/test for this case? Any difference if we use this version in `DocExporter::captureSamples()`?
```c++
            gfx::Point posBottomRight =
              snap_to_grid(docPref.grid.bounds(),
                           frameBounds.point2()
                           + gfx::Point(docPref.grid.bounds().size()),
                           PreferSnapTo::FloorGrid);
```

dncampo: IMHO `PreferSnapTo::CeilGrid` is needed to guarantee the situation of a layer going up to the edge of a grid. For example, look at this sprite of 2 layers:

Layer 1:
1_screenshot from 2019-02-06 00-09-51

Layer 2:
2_screenshot from 2019-02-06 00-10-08

The expected output of "Trim by grid" should be (I think):
3_screenshot from 2019-02-06 00-11-33

However, with the modification you have proposed it produces:
4_screenshot from 2019-02-06 00-10-40

This is because the layer 1 goes up to the edge of a grid, but there is no need of adding an extra one. Adding `+ gfx::Point(docPref.grid.bounds().size()),` in the `snap_to_grid` would add the extra grid gap that it is not needed. This is managed in `PreferSnapTo::CeilGrid:` asking if there is a remainder at the division `std::div(point.x-dx.rem, grid.w);` (it is the same in the y axis).

data/widgets/export_sprite_sheet.xml
dncampo: Done

src/app/commands/cmd_export_sprite_sheet.cpp
dncampo: Done

src/app/doc_exporter.cpp
dncampo: Done

src/app/snap_to_grid.cpp
dacap: Nice example 👍 I'll try to merge the PR today. LFTM

dacap:
Merged squashing commits bdd4202c35a0b96a2942e98d53cf78f237b845d4

Also I've remove the fix word to keep the issue open for the CLI part (maybe a new `--trim-by-grid` parameter).

Issue #1997, opened 2019-02-01
Aseprite v1.2.x


dacap:
**Please make a comment in this issue if you want to start working on one of these tasks.**

This is the set of items that must be seen before releasing the new Aseprite 1.2.x:

- [x] Loading TGA with alpha in palette creates a layer with Background (@Gasparoken)
- [x] Lua: Dialog:file() to add a field to select filenames https://github.com/aseprite/api/issues/7 (@dacap)
- [x] Lua: selection.origin (something to move the selection edges) (@dacap)
- [x] Check that "flash layer" is working correctly (I think this is not working on macOS) (@dacap)
- [x] https://community.aseprite.org/t/errors-with-known-color-profile-srgb/2656 (@dacap)
- [x] Pixels should be blended with normal blend mode when the backdrop is transparent pixel #1096 
   https://steamcommunity.com/app/431730/discussions/2/1739980540113413481/?tscn=1548381128 (@Gasparoken)
- [x] Create a filter command to add an outline to borders (related to #371, #1198) (@dacap)
   - ~~In someway the outline could be calculated with the whole sprite render result (instead of using the active layer only)~~ converted into #2398
- [x] radial gradient (https://community.aseprite.org/t/radial-gradient/2711, https://steamcommunity.com/app/431730/discussions/1/3182216552785250054/ ) (@dacap)
- [x] Edit > Shift feature removing sprite parts permanently (can't undo)
  https://steamcommunity.com/app/431730/discussions/2/1741103267304924778/ (@Gasparoken)
- [x] ~~https://community.aseprite.org/t/cant-browse-folders-when-saving-loading-files/1133~~
- [x] Export name same as file name by default #1964 (@Gasparoken)
- [x] Review the padding checkbox preference for Import Sprite Sheet (@Gasparoken)
- [x] Lua:  API to load/save one frame images only directly (https://github.com/aseprite/api/issues/9) (@dacap)
- [x] Test #1977 manually in release mode for bigger lines (@dacap)
- [x] https://community.aseprite.org/t/aseprite-icon-for-ase-files-is-a-bunny-now/2661 (@dacap)
- [x] Lua: https://community.aseprite.org/t/get-selected-palette-color-s/2512 (@dacap)
- ~~Lua: Image:drawImage() should use NORMAL blend mode (add blend mode)~~ https://github.com/aseprite/api/issues/31
- [x] Sprite sheet JSON exporter outputs full filesystem path in meta.image #1606

[...]

PR #2012, opened 2019-02-12
Fix tga creates Background layer on Load (task #1997)

Gasparoken:


dacap:
From https://en.wikipedia.org/wiki/Truevision_TGA#Header it looks like the TGA header contains some special bits to indicate if the image has or not the alpha channel (and that should determine if the resulting layer should be a background layer or a transparent one):

> Image specification (field 5) has six subfields:
> 
> X-origin (2 bytes): absolute coordinate of lower-left corner for displays where origin is at the lower left
> Y-origin (2 bytes): as for X-origin
> Image width (2 bytes): width in pixels
> Image height (2 bytes): height in pixels
> Pixel depth (1 byte): bits per pixel
> **Image descriptor** (1 byte): **bits 3-0 give the alpha channel depth**, bits 5-4 give direction

This shows us an extra bug where we are [saving TGA files incorrectly](https://github.com/aseprite/aseprite/blob/b3eba1c952b204325399f9660b5ddb2159d2cd0c/src/app/file/tga_format.cpp#L419) for opaque sprites (including the alpha channel when we shouldn't).

Maybe this PR could fix:

1. the `onLoad`: calling `fop->sequenceSetHasAlpha(true);` when the [`descriptor_bits`](https://github.com/aseprite/aseprite/blob/b3eba1c952b204325399f9660b5ddb2159d2cd0c/src/app/file/tga_format.cpp#L223) has bits related to the alpha channel (`(descriptor_bits & 0xf) == 8`).
2. the `onSave`: [save the alpha depth (8)](https://github.com/aseprite/aseprite/blob/b3eba1c952b204325399f9660b5ddb2159d2cd0c/src/app/file/tga_format.cpp#L419) only if the sprite is RGB and is not opaque.

dacap:
It's work great 👍  cherry picked 5cb0adc46a32fe9be269ac470a410ae2a337899e

PR #2023, opened 2019-02-25
Fix pixels normal blended when backdrop is transparent (fix #1096)

### LINKED ISSUE #1096: Pixels should be blended with normal blend mode when the backdrop is transparent pixel
Basically, we should ignore special blending modes when we've to composite pixels over transparent. Instead of applying the layer blending mode to the checked grid (which are transparent pixels for the UI, and do not represent the final result), we should use the normal blending mode when the final pixel/backdrop is transparent. (It is what Photoshop does in this case.)

When we have a layer with non-Normal blending mode (e.g. Overlay blending mode), we've to take special care of pixels that are composite with the transparent background (the gray checked background). For those pixels, the Normal blending mode must be used anyway (because the final render when we export cannot identify that those pixels should be rendered with other blending mode, there is no-blending mode property per pixel in the final render/or file format).

---

Example: Here's the duck is the foreground layer with the `Color` blending mode, and the gradient is at the bottom:

* This is the old blending mode, pixels that are blended over the checked background are blended with the `Color` blending mode anyway (makes sense only for a preview point of view, but not for persistence):
   screen shot 2019-02-26 at 08 14 39

* This is the new blending mode, pixels that are blended over the checked background are blended with the `Normal` blending mode:
   screen shot 2019-02-26 at 08 14 46

* When we export with the old blending mode we get this (pixels are blended with transparent, without check background, so we lost the original pixel information):
   color-old-render

* And when we export with the new blending mode we get this other result (we do not lost the original pixel information):
   color-new-render

In this way the new blending mode is consistent between Visual Preview <-> Export Result on disk.

Gasparoken:


src/app/commands/cmd_change_pixel_format.cpp
```diff
@@ -123,6 +124,7 @@ class ConvertThread : public render::TaskDelegate {
                     m_image->height()));
 
     render::Render render;
+    render.setNewBlend(Preferences::instance().experimental.newBlend());
```
dacap: It would be better if this `newBlend` value is passed as an argument in this `run()` function, because this is executed in a background thread (not the UI main thread), and I'm not sure if the `Preferences` class is thread-safe.

src/app/commands/cmd_change_pixel_format.cpp
```diff
@@ -475,7 +477,8 @@ void ChangePixelFormatCommand::onExecute(Context* context)
           SelectedLayers selLayers;
           for (auto layer : sprite->root()->layers())
             selLayers.insert(layer);
-          job.tx()(new cmd::FlattenLayers(sprite, selLayers));
+          job.tx()(new cmd::FlattenLayers(sprite, selLayers,
+                                          Preferences::instance().experimental.newBlend()));
```
dacap: Same here as in the `cmd_change_pixel_format.cpp`, this lambda is executed in a background thread, I'm not even sure that the code is already thread-safe, but anyway, just in case we can move the `newBlend` outside. You could replace it with something like:
```c++
const bool newBlend = Preferences::instance().experimental.newBlend();
...
job.startJobWithCallback(
  [this, &job, flatten, newBlend] {
    ...
    job.tx()(new cmd::FlattenLayers(sprite, selLayers, newBlend));
...
```

PR #2025, opened 2019-03-08
Fix non-normal blend modes when the backdrop is transparent (fix #1096)

### LINKED ISSUE #1096: Pixels should be blended with normal blend mode when the backdrop is transparent pixel

Gasparoken:


dacap:
I'll merge this commit with some minor changes and then do other commits to fixes tests and other minor things.

dacap:
Merged in f53544842cd2bb7545c00810e427cf88e196bb91 with minor changes. I'll fix the render_tests which is throwing an error.

Issue #1974, opened 2019-01-07
"About Aseprite" menu option disabled on macos in the "Aseprite" menu.


increpare:
Using the system menu (not the one in aseprite), note that "Help - > About" is enabled and works, but the system-native "Aseprite -> About Aseprite" is disabled

<img width="195" alt="bildschirmfoto 2019-01-07 um 15 45 11" src="https://user-images.githubusercontent.com/465632/50774296-5827b300-1293-11e9-83d2-c6bb14350fb3.png">
<img width="385" alt="bildschirmfoto 2019-01-07 um 15 45 08" src="https://user-images.githubusercontent.com/465632/50774298-58c04980-1293-11e9-97e6-14005708929e.png">

(was very slightly confusing when I was looking up the version number)

(It's also not idiomatic that the other options in this menu are also disabled)

###Aseprite and System version

Aseprite version: 1.2.9 (downloaded, IIRC)
System: macOS 10.14.1 (18B75)

dacap:
I'm trying to reproduce this but I'm not able, do you know in what state the program should be to see the disabled system menu?

increpare:
Oh right, it's working for me now as well.  Huh.  Ok, found a repro:

Open a file.

Note that in the MacOS "Aseprite" menu everything is enabled (ok behaviour).

<img width="236" alt="bildschirmfoto 2019-01-07 um 16 41 34" src="https://user-images.githubusercontent.com/465632/50777204-42b68700-129b-11e9-8501-45241282e4fa.png">

go to save as
Open the MacOS "Aseprite" menu, note that the options are disabled (ok behaviour).

<img width="938" alt="bildschirmfoto 2019-01-07 um 16 41 46" src="https://user-images.githubusercontent.com/465632/50777217-48ac6800-129b-11e9-8acc-427c75335627.png">

click the cancel button (or press esc)

Open the MacOS "Aseprite" menu, note that it's still disabled (bug).

(I notice that all dropdown menus are disabled when the "save as" dialog is open, but only the "Aseprite" one remains disabled after this procedure).

dacap:
Oh if it's related to the native file dialog it can have problems (because the feature is still experimental, it's not working correctly on macOS).

tominek:
For anyone else still encountering this issue. Go to Aseprite -> Settings and check option "Show Aseprite menu bar". This will show the in app menu bar with working menus.

PR #2037, opened 2019-03-21
Fix app menu may become disabled on macOS

kawa-yoiko:
Hi! I think here is a valid fix for #1974. Please refer to aseprite/laf#7 — both changes need to be applied for the fix.

The reason for the changes in this repo is that according to the automatic validation policy, modal dialogs somehow disable these menu items, and they need to be re-enabled with a `validate` routine. (Btw, I'd love to hear if there is a better approach ^ ^)

Thanks for the effort and time!

### DIFF
--- a/src/app/app_menus.cpp
+++ b/src/app/app_menus.cpp
@@ -728,6 +728,7 @@ void AppMenus::createNativeMenus()
         UIContext::instance()->executeCommand(cmd);
       }
     };
+    about.validate = [](os::MenuItem* item){ item->setEnabled(true); };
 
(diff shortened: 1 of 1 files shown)


dacap:
Hi @kawa-yoiko! thanks for your patches, I'll take a look to your contribution soon, in the meanwhile, could you please check [the CLA signing process](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla)?

kawa-yoiko:
I made a little tweak so that the behaviour is consistent with both native and LAF dialogs.

Issue #2009, opened 2019-02-10
Add support extra pixel connectivity for Paint Bucket tool


dacap:
We should be able to use [8-connected](https://en.wikipedia.org/wiki/Pixel_connectivity#8-connected) pixels with the paint bucket tool, useful to recoloring outlines.

Requested here: https://steamcommunity.com/app/431730/discussions/1/1776010325128300149/

rscotchmer:
Hi!  I'd like to work on this if that's alright!

dacap:
Hi @rscotchmer, in case you want to make a contribution, please check the [CLA signing process](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla) first.

About this specific issue, my idea is to have a button set (*) with a couple of buttons in this menu:

<img width="270" alt="Screen Shot 2019-03-16 at 11 35 53" src="https://user-images.githubusercontent.com/39654/54476753-bca94600-47df-11e9-84ad-cfb16848f6c1.png">

to switch between both options (4-connected pixels, 8-connected pixels). 

(*) The button set must be something similar to the ones in the brush options menu:

<img width="278" alt="Screen Shot 2019-03-16 at 11 37 38" src="https://user-images.githubusercontent.com/39654/54476771-f2e6c580-47df-11e9-82df-9b17f9ecf6fb.png">

but it can have a couple of icons to show the connection of pixels (+ tooltips).

dacap:
Closed with #2038 (a08cfa9e13f21fb2f9b1b1a20e874c2307168403)

dacap:
![image](https://user-images.githubusercontent.com/39654/56938335-ad543280-6ad7-11e9-8aa0-d086ef6a70e9.png)

PR #2038, opened 2019-03-21
[WIP] Fixes 2009

rscotchmer:
(Will eventually) close issue #2009 by adding the ability to switch between 4-connected and 8-connected pixels when using the paint bucket tool.

I was having trouble finding where to start with adding the option to toggle this on/off in the UI -- if someone could point me in the right direction that'd be much appreciated!

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -46,6 +46,10 @@
       <value id="ACTIVE_LAYER" value="0" />
       <value id="ALL_LAYERS" value="1" />
     </enum>
+    <enum id="PixelConnectivity">
+      <value id="EIGHT_CONNECTED" value="0" />
+      <value id="FOUR_CONNECTED" value="1" />
+    </enum>
     <enum id="EyedropperChannel">
       <value id="COLOR_ALPHA" value="0" />
       <value id="COLOR" value="1" />
@@ -352,6 +356,7 @@
     <section id="floodfill">
       <option id="stop_at_grid" type="StopAtGrid" default="StopAtGrid::NEVER" />
       <option id="refer_to" type="FillReferTo" default="FillReferTo::ACTIVE_LAYER" />
+      <option id="pixel_connectivity" type="PixelConnectivity" default="PixelConnectivity::FOUR_CONNECTED" />
     </section>
   </tool>
 
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -47,8 +47,8 @@
       <value id="ALL_LAYERS" value="1" />
     </enum>
     <enum id="PixelConnectivity">
-      <value id="EIGHT_CONNECTED" value="0" />
-      <value id="FOUR_CONNECTED" value="1" />
+      <value id="FOUR_CONNECTED" value="0" />
+      <value id="EIGHT_CONNECTED" value="1" />
     </enum>
     <enum id="EyedropperChannel">
       <value id="COLOR_ALPHA" value="0" />
--- a/src/app/tools/point_shapes.h
+++ b/src/app/tools/point_shapes.h
@@ -103,6 +103,7 @@ public:
       get_pixel(srcImage, pt.x, pt.y),
       loop->getTolerance(),
       loop->getContiguous(),
+      loop->isPixelConnectivityFourConnected(),
       loop, (AlgoHLine)doInkHline);
   }
 
--- a/src/app/tools/point_shapes.h
+++ b/src/app/tools/point_shapes.h
@@ -103,7 +103,7 @@ public:
       get_pixel(srcImage, pt.x, pt.y),
       loop->getTolerance(),
       loop->getContiguous(),
-      loop->isPixelConnectivityFourConnected(),
+      loop->isPixelConnectivityEightConnected(),
       loop, (AlgoHLine)doInkHline);
   }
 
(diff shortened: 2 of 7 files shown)


dacap:
Hi @rscotchmer, thanks for this PR so far!

About the UI option, you can start seeing something like in `ContextBar::PivotField` class:

Screen Shot 2019-03-21 at 15 15 56

When the `ui::Menu` popup [is displayed]((https://github.com/aseprite/aseprite/blob/3ec3f75d91cdfebd7c99ab0b7e2291d9d681d2be/src/app/ui/context_bar.cpp#L785)), a little `ButtonSet` is created to show those 9x9 options.

data/pref.xml
```diff
@@ -46,6 +46,10 @@
       <value id="ACTIVE_LAYER" value="0" />
       <value id="ALL_LAYERS" value="1" />
     </enum>
+    <enum id="EightConnected">
```
dacap: You could rename this enum to `PixelConnectivity` 👍 
```suggestion
    <enum id="PixelConnectivity">
```
So the pixel connectivity can be 8-connected or 4-connected.

data/pref.xml
```diff
@@ -344,9 +348,11 @@
       <option id="width" type="int" default="16" />
       <option id="speed" type="int" default="32" />
     </section>
+
     <section id="floodfill">
       <option id="stop_at_grid" type="StopAtGrid" default="StopAtGrid::NEVER" />
       <option id="refer_to" type="FillReferTo" default="FillReferTo::ACTIVE_LAYER" />
+      <option id="eight_connected" type="EightConnected" default="EightConnected::FOUR_CONNECTED" />
```
dacap: Same here, you could rename this to "pixel_connectivity":
```suggestion
      <option id="pixel_connectivity" type="PixelConnectivity" default="PixelConnectivity::FOUR_CONNECTED" />
```

rscotchmer:
Hi @dacap !  I'm still working on figuring out why the Travis CI build failed, but in the meantime, is there anything else you want changed?  I made the earlier changes you requested and have since finished adding the option to the UI, and all that.

dacap:
@rscotchmer it looks like some tests are failing but don't worry, it's because I've added some tests in https://github.com/aseprite/tests repository that needs a recent master branch. I'll review your changes just right now.

data/pref.xml
```diff
@@ -46,6 +46,10 @@
       <value id="ACTIVE_LAYER" value="0" />
       <value id="ALL_LAYERS" value="1" />
     </enum>
+    <enum id="PixelConnectivity">
+      <value id="EIGHT_CONNECTED" value="0" />
+      <value id="FOUR_CONNECTED" value="1" />
```
dacap: The only thing I would change, is the order of these constants, I would put `FOUR_CONNECTED` = 0, and `EIGHT_CONNECTED` = 1 (so 0 is the default value). Anyway another change I would make (but I can do it later) is to move this enum to the same place as in `floodfill()` function, so the function can receive the `enum` value instead of a `isEightConnected` boolean value.

src/app/tools/point_shapes.h
```diff
@@ -103,6 +103,7 @@ class FloodFillPointShape : public PointShape {
       get_pixel(srcImage, pt.x, pt.y),
       loop->getTolerance(),
       loop->getContiguous(),
+      loop->isPixelConnectivityFourConnected(),
```
dacap: I'm not sure, but is not `floodfill()` receiving true when it's eight-connected instead of four-connected?

src/app/ui/editor/tool_loop_impl.cpp
```diff
@@ -247,6 +247,16 @@ class ToolLoopBase : public tools::ToolLoop {
     }
     return false;
   }
+
+  bool isPixelConnectivityFourConnected() override {
```
dacap: I think this could be simplified to something like:
```c++
  return m_toolPref.floodfill.pixelConnectivity() == app::gen::PixelConnectivity::FOUR_CONNECTED;
```

src/app/tools/point_shapes.h
rscotchmer: You're absolutely right -- inconsistent naming there on my part.  Changed the function reflect `floodfill()` receiving `true` when it's eight-connected.

data/pref.xml
rscotchmer: I changed the order of the constants as requested, but I couldn't figure out where I needed to move the enum to -- I'd be more than happy to refactor the bool to enum myself if it's easier for you.  I just need a pointer to where to get started :)

dacap:
Thanks @rscotchmer, I'll give a try to these changes next week and merge them. They looks fine :+1:

dacap:
Hi again @rscotchmer, I've merged your changes in a new branch: https://github.com/aseprite/aseprite/tree/8-connected

It's temporal because I have found some issues with the algorithm:

1. In some cases it is a little slow (can take 20 or 30 minutes to fill a region less than 32x32)
2. Sometimes it crashes
3. I think there is an infinite recursion

E.g. Try filling the center of this image:

Sprite-0001

rscotchmer:
Hi @dacap, I've fixed it so it no longer crashes, and can fill large areas without delay.  Let me know if there's anything else that needs to be done!

dacap:
Hi @rscotchmer, I have just merge your change with this commit: a08cfa9e13f21fb2f9b1b1a20e874c2307168403 (is cherry picked with minor indentation changes). Thanks for your contribution!

Issue #2020, opened 2019-02-21
Combining --trim and --sheet-pack causes sprite overlap


DragonDePlatino:
I'm storing all of my sprites in separate .ase files and I'm pulling them onto an atlas whenever I compile. I'm already using `--sheet-pack` but I'd like to save more space with `--trim`. The problem is that when I enable both at once, sprites aren't being packed properly.

When I use just `--sheet-pack`, my atlas looks like this:

![sprite_notrim](https://user-images.githubusercontent.com/9517192/53203380-ab10bc00-35ee-11e9-9f0d-940c4715b45a.png)

This is fine, but there's a lot of wasted space. Especially with large sprites like the flash at the top. If I enable `--trim --sheet-pack`, now my atlas looks like:

![sprite_trim](https://user-images.githubusercontent.com/9517192/53203381-ab10bc00-35ee-11e9-88bf-e1f2eef9b10c.png)

The sprites are being trimmed but they're covering each other up and hanging off the sides of the texture. As an experiment, I tried just `--trim` to see if it worked by itself:

![sprite_nopack](https://user-images.githubusercontent.com/9517192/53203379-ab10bc00-35ee-11e9-811b-558ba7e0ede2.png)

So I can get just `--sheet-pack` or just `--trim` to work, but not both at once. Here is my sprite makefile for reference:

```
ASEPRITE = "C:/Program Files/Aseprite/Aseprite.exe"
ATLAS = sprite.bmp
DATA = $(ATLAS:.bmp=.json)

ASE = $(wildcard *.ase)
SIZE = 512

sprite:$(ATLAS)

$(ATLAS):$(ASE)
	$(ASEPRITE) -b $(ASE) --sheet $@ --trim --sheet-pack --sheet-width=$(SIZE) --sheet-height=$(SIZE) --data $(DATA) --list-tags --list-slices

.PHONY: clean
clean:
	@rm -f $(ATLAS) $(DATA)
```

dacap:
Fixed with 5d5aae73115a36e2eeb0fdf6fb373ce7b61418eb

Issue #1606, opened 2017-11-08
Sprite sheet JSON exporter outputs full filesystem path in meta.image


SaxxonPike:
I propose replacing the absolute filesystem path in `meta.image` with a relative path instead.

When exporting a sprite sheet to JSON, the `meta.image` string contains the full filesystem path of the PNG image. Using absolute paths could reveal sensitive information about the machine on which the export was performed and also cause issues after moving files around if that information is used. Using relative paths instead would avoid this problem and make it easier for content management as well.

The workaround for now is to manually go in and remove this data from the exported JSON. Since making changes in Aseprite and exporting to a new sheet is a task we repeat very frequently, this effort adds up and also lends to making mistakes.

Here's an excerpt from a recently exported file using the Steam release:
```
    "meta": {
        "app": "http://www.aseprite.org/",
        "version": "1.2.4-x64",
        "image": "C:\\Users\\Saxxon\\Documents\\Sprite-0002.png",
        "format": "I8",
```

Here's that same excerpt with the proposed change:
```
    "meta": {
        "app": "http://www.aseprite.org/",
        "version": "1.2.4-x64",
        "image": "Sprite-0002.png",
        "format": "I8",
```

Thanks for your time!

### Aseprite and System version

* Aseprite version: v1.2.4 Steam
* System: Windows 10 x64

dacap:
Taking note about this one. A temporary fix would be to use the [CLI](https://www.aseprite.org/docs/cli/) from the `C:\Users\Saxxon\Documents\` directory in this case (which it looks will generated your desired output):
```
C:\>cd C:\Users\Saxxon\Documents
C:\Users\Saxxon\Documents>"C:\Program Files (x86)\Steam\steamapps\common\Aseprite\aseprite.exe" -b Sprite-0002.ase --data Sprite-0002.json --sheet Sprite-0002.png
```

PR #2039, opened 2019-03-23
Fix spritesheet export path and layout

kawa-yoiko:
The two commits address two separate issues #1606 and #2020, but in order to minimise disturbance I put them together here (^ ^;)

Hope they will help!


dacap:
Thanks @kawa-yoiko! I was never sure about the #1606 fix, I would have preferred the image path relative to the json file, but yesterday I just did that and then I thought that just putting the image filename without any path is the simplest/clean solution.

PR #2044, opened 2019-03-26
Fix export file name updates after SaveAs (fix #1964)

### LINKED ISSUE #1964: Export name same as file name by default
This is a small quality of life issue, but I'd like the ability to make it so that when I go to export a project, it always uses the name of the project by default for the exported file.

Currently, if I make a new project, call it "project1.ase" and then go to export it as png, the default file name will be "project1.png."

If I then save project1.ase as project2.ase, and then go to export project2.ase as a png, the default file name will still be "project1.png." Often I don't notice, and accidentally save over the original file.

I think it'd make more sense that the default file name would be "project2.png" to avoid this issue.

Lots of people have probably gotten used to this quirk, but I use so many different programs that make the default export name the same as the current project name that it's hard to remember to check. I often don't notice the issue until I've overwritten several files, and have to go back and re-export a bunch of sprites. So, if we could have an option to decide which way file names are set when exporting, that would be awesome.

Gasparoken:


src/app/commands/cmd_save_file.cpp
```diff
@@ -179,6 +180,10 @@ void SaveFileBaseCommand::saveDocumentInBackground(
   const std::string& filename,
   const bool markAsSaved)
 {
+  if (Command::id() == CommandId::SaveFileAs() |
```
dacap:
```suggestion
  if (Command::id() == CommandId::SaveFileAs() ||
```

dacap:
It doesn't fix the issue because we are resetting the `Preferences::instance().document(document).saveCopy.filename("")` of the same sprite and not the one of the copy. We have to reset the saveCopy.filename() of the copy (the new created file).

src/app/commands/cmd_save_file.cpp
```diff
@@ -135,6 +136,7 @@ std::string SaveFileBaseCommand::saveAsDialog(
   const std::string& forbiddenFilename)
 {
   Doc* document = context->activeDocument();
+  Preferences::instance().save();
```
dacap: Write a comment about why we have to save the preferences just right now.

src/app/commands/cmd_save_file.cpp
```diff
@@ -168,6 +170,14 @@ std::string SaveFileBaseCommand::saveAsDialog(
     saveDocumentInBackground(
       context, document,
       filename, markAsSaved);
+    Preferences::instance().document(document).saveCopy.filename("");
```
dacap: Here we could use a temporal `auto& docPref = Preferences::instance().document(document)` variable instead of calling `Preferences::instance().document(document)` several times.

Other issue: I guess that we could use something like `docPref.saveCopy.aniDir(docPref.saveCopy.aniDir.defaultValue())`, etc. to reset each value.

dacap:
Cherry-picked 22b58aafaafdfe4ef4e0804c540ce526aa532bc3

PR #2090, opened 2019-06-17
Update INSTALL.md to faster clone, and build instructions skia for python3

HardPlant:
* added --depth 1 args to clone skia dependancy
* added instruction for python3 users

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -281,8 +281,8 @@ several minutes to finish:
 
     mkdir $HOME/deps
     cd $HOME/deps
-    git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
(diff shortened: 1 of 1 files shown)


dacap:
These instructions are now for [skia](https://github.com/aseprite/skia) repo, anyway I would like to keep the instructions as close to the official instructions (from Google) as possible.

dacap:
I saw that the official Skia instructions now use `python2` explicitly, I'll add an issue to add `python2` to that repository, but it's something I've to test yet on all platforms, so not sure if it does work or is really needed (seeing that just `python` does the job in the [workflows](https://github.com/aseprite/skia/blob/aseprite-m81/.github/workflows/build.yml))

PR #2117, opened 2019-07-25
Refactor sprite size command params

Gasparoken:


src/app/commands/cmd_sprite_size.cpp
```diff
@@ -39,6 +41,23 @@ namespace app {
...
+  Param<double> scale { this, 1.0, "scale" };
+  Param<double> scaleX { this, 1.0, "scaleX" };
+  Param<double> scaleY { this, 1.0, "scaleY" };
+  Param<ResizeMethod> method { this, ResizeMethod::RESIZE_METHOD_NEAREST_NEIGHBOR, { "method", "resize-method" } };
+
+public:
+  void setWidth(int w) { width.fromString(base::convert_to<std::string>(w).c_str()); };
+  void setHeight(int h) { height.fromString(base::convert_to<std::string>(h).c_str()); };
+  void setScaleX(double scX) { scaleX.fromString(base::convert_to<std::string>(scX).c_str()); };
+  void setScaleY(double scY) { scaleY.fromString(base::convert_to<std::string>(scY).c_str()); };
```
dacap: These member functions `setWidth`/`setHeight`/etc. are not necessary, you can remove them because there are a way to change `Param` value (using the [`Param::operator()(const T& value)` member function](https://github.com/aseprite/aseprite/blob/4c4c20ace6334685b86d904bfe3fc0b0c50a3db5/src/app/commands/new_params.h#L94)). For example, to change the `width` field you just write:
```c++
int newValue = ...;
params.width(newValue);
```
Example in [CanvasSize](https://github.com/aseprite/aseprite/blob/4c4c20ace6334685b86d904bfe3fc0b0c50a3db5/src/app/commands/cmd_canvas_size.cpp#L342) params.

Basically, a `Param` offers a getter (`int currentValue = params.width()`) and a setter (`params.width(newValue)`).

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -180,17 +199,21 @@ class SpriteSizeJob : public SpriteJob {
 
 };
 
+#ifdef ENABLE_UI
+
 class SpriteSizeWindow : public app::gen::SpriteSize {
 public:
-  SpriteSizeWindow(Context* ctx, int new_width, int new_height) : m_ctx(ctx) {
+  SpriteSizeWindow(Context* ctx, SpriteSizeParams params) : m_ctx(ctx) {
```
dacap:
```suggestion
  SpriteSizeWindow(Context* ctx, const SpriteSizeParams& params) : m_ctx(ctx) {
```
Use `const SpriteSizeParams&` to avoid copying the `SpriteSizeParams` (we can use a reference to the original one from `SpriteSizeCommand`).

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -199,9 +222,23 @@ class SpriteSizeWindow : public app::gen::SpriteSize {
...
+        resize_method = 1;
+      else if (params.method() == doc::algorithm::RESIZE_METHOD_ROTSPRITE)
+        resize_method = 2;
+      else
+        resize_method = 0;
+    }
+    else
+      resize_method = get_config_int("SpriteSize", "Method",
+                                     doc::algorithm::RESIZE_METHOD_NEAREST_NEIGHBOR);
+    set_config_int("SpriteSize", "Method", resize_method);
```
dacap: We shouldn't modifying the configuration here with `set_config_int`, only getting configuration with `get_config_int`.

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -199,9 +222,23 @@ class SpriteSizeWindow : public app::gen::SpriteSize {
     method()->addItem("Nearest-neighbor");
     method()->addItem("Bilinear");
     method()->addItem("RotSprite");
-    method()->setSelectedItemIndex(
-      get_config_int("SpriteSize", "Method",
-                     doc::algorithm::RESIZE_METHOD_NEAREST_NEIGHBOR));
+    int resize_method;
+    if (params.method.isSet()) {
+      if (params.method() == doc::algorithm::RESIZE_METHOD_BILINEAR)
+        resize_method = 1;
```
dacap: There is no need to do this manual conversion for each case between the `doc::algorithm::ResizeMethod` <-> `int`. We can just use:
```c++
resize_method = (int)params.method();
```
because we already were making the assumption that the integer value matches the value of the enum on  `get_config_int()`.

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -199,9 +222,23 @@ class SpriteSizeWindow : public app::gen::SpriteSize {
...
+      else
+        resize_method = 0;
+    }
+    else
+      resize_method = get_config_int("SpriteSize", "Method",
+                                     doc::algorithm::RESIZE_METHOD_NEAREST_NEIGHBOR);
+    set_config_int("SpriteSize", "Method", resize_method);
+    method()->setSelectedItemIndex(resize_method);
+    const bool lock = (params.lockRatio.isSet())? params.lockRatio() : get_config_bool("SpriteSize", "LockRatio", false);
+    set_config_bool("SpriteSize", "LockRatio", lock);
```
dacap: The same here, do not change the configuration when the window is displayed (the configuration is changed only after the user accept the dialog, it's in the `SpriteSizeCommand`). When we construct each window/dialog, we just read/get configuration/preferences.

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -319,15 +328,77 @@ bool SpriteSizeCommand::onEnabled(Context* context)
...
   const ContextReader reader(context);
   const Sprite* sprite(reader.sprite());
-  int new_width = (m_width ? m_width: int(sprite->width()*m_scaleX));
-  int new_height = (m_height ? m_height: int(sprite->height()*m_scaleY));
-  ResizeMethod resize_method = m_resizeMethod;
+  auto& params = this->params();
+
+  double ratio = sprite->width() / double(sprite->height());
+  if (params.scale.isSet()) {
+    params.setWidth(int(sprite->width() * params.scale()));
```
dacap:
```suggestion
    params.width(int(sprite->width() * params.scale()));
```
As we saw above ☝️ we can use the `Param<int>::operator()(const int&)` to change the value of a param (and we can remove the `setWidth()`, etc.)

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -319,15 +328,77 @@ bool SpriteSizeCommand::onEnabled(Context* context)
...
+    else if (params.scaleY.isSet()) {
+      params.setWidth(int(params.scaleY() * sprite->width()));
+      params.setHeight(int(params.scaleY() * sprite->height()));
+      params.setScaleX(params.scaleY());
+    }
+    else {
+      params.setWidth(sprite->width());
+      params.setHeight(sprite->height());
+      params.setScaleX(1.0);
+      params.setScaleY(1.0);
```
dacap: I think these sets for `scaleX` and `scaleY` are not necessary because the default value of `scaleX` and `scaleY` is `1.0` already (the case of `width`/`height` is valid because the default value is `0`, so setting the `sprite->width()/height()` is right 👍).

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -319,15 +328,77 @@ bool SpriteSizeCommand::onEnabled(Context* context)
...
+    }
+  }
+  else {
+    if (params.width.isSet()) {
+      params.setScaleX(params.width() / double(sprite->width()) * 100.0);
```
dacap:
```suggestion
      params.scaleX(params.width() / double(sprite->width()));
```
I think the `100.0` is because we talked about this being a percentage, etc. ☺️  Same in the other cases below 👇

src/app/commands/cmd_sprite_size.cpp
```diff
@@ -319,15 +328,77 @@ bool SpriteSizeCommand::onEnabled(Context* context)
...
+    }
+  }
+  int new_width = params.width();
+  int new_height = params.height();
+  ResizeMethod resize_method = params.method();
 
 #ifdef ENABLE_UI
-  if (m_useUI && context->isUIAvailable()) {
-    SpriteSizeWindow window(context, new_width, new_height);
+  if (ui && context->isUIAvailable()) {
```
dacap:
```suggestion
  if (ui) {
```
No need for `context->isUIAvailable()` because it's already considered when the `ui` variable is defined.

src/app/commands/new_params.cpp
```diff
@@ -55,6 +56,17 @@ void Param<std::string>::fromString(const std::string& value)
   setValue(value);
 }
 
+template<>
+void Param<doc::algorithm::ResizeMethod>::fromString(const std::string& value)
+{
+  if (value == "bilinear" || value == "Bilinear")
```
dacap: Instead of asking for `bilinear` and `Bilinear`, use `base::utf8_icmp(a, b) == 0`, check `Param<doc::ColorMode>::fromString` for an example.

dacap:
As a recommendation @Gasparoken, see the review from the https://github.com/aseprite/aseprite/pull/2117/files tab.

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/cmd_sprite_size.cpp
Gasparoken: Got it!

src/app/commands/new_params.cpp
Gasparoken: Got it! A new function in my mind.

PR #2129, opened 2019-08-01
Prevent CLI render of hidden layers in group when exporting

### LINKED ISSUE #2084: CLI --save-as renders hidden layers when exporting a group
When using the CLI to export a group to a single image file any _hidden_ child layers in that group are also rendered to the resulting image.

For example:

`aseprite -b test_file.aseprite --layer test_group --save-as "result.png"`

(Where 'test_group' is a group, not a layer). The CLI seems to ignore whether or not any of the child layers (or groups) are hidden. The image _does_ render correctly when no specific layer/group is provided through --layer.

* Aseprite version: v1.2.11
* System: MacOS 10.13.6

mrspeaker:
When using the CLI to export a group to a single image file any hidden
child layers in that group are also rendered to the resulting image.

The reason is that `isVisibleHierachy()` is never checked when there
are items in `includeLayers`: as it's assumed the layers are never groups. This
commit checks if the hidden layer was explicitly included (when you want to
render a hidden layer). If not then it is a child of a group, so is not rendered.

I have tested this with many includes and excludes, with groups and
layers and it seems correct: BUT! I do not have a lot of
experience with CPP (I just needed this fix!) so this code might not be canonical.

Resolves #2084

### DIFF
--- a/src/app/cli/cli_processor.cpp
+++ b/src/app/cli/cli_processor.cpp
@@ -102,6 +102,16 @@ void filter_layers(const LayerList& layers,
         (!cof.includeLayers.empty() && !filter_layer(layer, layer_path, cof.includeLayers, true)))
       continue;
 
+    if (!cof.includeLayers.empty() && !layer->isVisibleHierarchy()) {
+      bool isIncluded = false;
+      for (const auto& filter : cof.includeLayers) {
+        if (filter == layer->name()) {
+          isIncluded = true;
+        }
+      }
+      if (!isIncluded) continue;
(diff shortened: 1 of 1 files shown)


dacap:
Hi @mrspeaker, in order to accept the PR, you should [sign our CLA](https://github.com/aseprite/opensource/blob/master/sign-cla.md#sign-the-cla).

mrspeaker:
@dacap - signed!

mrspeaker:
Thanks!

PR #2139, opened 2019-08-26
Google plus removed due to google+ being discontinued

SHADOWELITE7:
Google plus removed due to google+ being discontinued
you can take a look on here because I clicked on the Google+ link on the readme
https://plus.google.com/+AsepriteOrg/posts
image


SHADOWELITE7:
Thanks for seeing this. Yah I do miss google+.

PR #2172, opened 2019-10-11
Fix trim doesn't do anything if there's a pixel in the top left corner (fix #1659)

### LINKED ISSUE #1659: Trim doesn't do anything if there's a pixel in the top left corner
Any canvas size, empty transparent background, place one pixel in the top left corner and trim does nothing, but it works if you place a pixel in any other corner, or if you move the pixel out from the corner it works perfectly.

Aseprite 1.2.6-x64
Windows 7 x64

Gasparoken:
Before this fix, we did use the top-left corner pixel by default as "refColor" in doc::algorithm::shrink_bounds(image, frameBounds, refColor).
Now, we are looking for one color whole border (i.e: solid color border) searching on the sprite square contour.

If the algorithm finds a transparentColor pixel on border frame, automatically "refColor" turns to transparentColor and it goes directly to shrink_bounds function.
If the borders don't contain transparentColor, the algorithm keeps checking:
It search for a solid color border (top border first), then it checks the opposite border (bottom border), then:
1- If the opposite border color is equal to the first border, this color will be the "refColor".
2- If the color of the opposite border is solid, BUT different to the first border we have to trigger a dialog, delegating the choise to the user.
3- If opposite border contains differents colors, we choose the first border color as "refColor".
4- It repeat analisys to sprite left and right columns pixels.
If no border has solid color, trimSprite do nothing.

### DIFF
--- a/src/app/doc_api.cpp
+++ b/src/app/doc_api.cpp
@@ -324,11 +324,59 @@ void DocApi::trimSprite(Sprite* sprite, const bool byGrid)
 
   for (frame_t frame(0); frame<sprite->totalFrames(); ++frame) {
     render.renderSprite(image, sprite, frame);
-
-    // TODO configurable (what color pixel to use as "refpixel",
-    // here we are using the top-left pixel by default)
     gfx::Rect frameBounds;
-    if (doc::algorithm::shrink_bounds(image, frameBounds, get_pixel(image, 0, 0)))
+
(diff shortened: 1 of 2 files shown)


dacap:
This PR was merged with this commit 695d85e56660ece011ef16ee7d024bb8e1b867d0  and followed by 66ccf51eb903e14052373efbbaf87d3e05e6db7b

Issue #1385, opened 2017-01-09
Bug: RGB-to-Indexed colour assignment/approximation problem.


oceanhahn:
There is a problem with the way ASEprite 1.1.13 determines palette indices for very similar values when converting from RGB to indexed colour modes, wherein pixels in an RGB image will be assigned palette ids _similar_ to a colour in the destination palette even if the _exact matching_ colour is listed later in the destination palette.

I discovered this bug when I was copying a flattened section of a larger palettized image into a new document in order to work on it separately. My usual method is to Copy Merged from the source document into a new RGB editor session, quantize it (there are colours from the source not present in the copied pixels and this is a pretty okay way to eliminate them), and then convert the new sprite to Indexed mode. But this time, the conversion mysteriously overwrote a couple of low-contrast colours, even though both were present in the palette. I tried a different way, by copying the old palette to a new Indexed sprite first, _then_ pasting the Copy-Merged pixels in, but to the same effect: the pasted pixels are approximated to a single palette index where their contrast is low.
**Flatting the source image, selecting the new sprite's pixels, and then using New Sprite from Selection avoids this problem,** but it remains a big issue for copy-pasting and mode conversion. That the contrast between clobbered colour values is low also means that the problem could go undetected for a long time if users are not insufferably observant. =_=;

In my case, this happens with these two shades of green, _(12, 31, 32) [#0c1f15]_ and _(8, 25, 16) [#081910]_, in the gif animation below. (Admittedly, even LICEcap has trouble telling them part, but you can see it slightly as the cursor moves over the problem area.)
![quantize problem](https://cloud.githubusercontent.com/assets/11544843/21776617/5c41d640-d669-11e6-849b-83fb851e1559.gif)

You can try it for yourself using this RGB-mode .ase file. Just try converting it to Indexed mode and you should see what I mean. (GitHub doesn't support sharing .ase files, so I'm sending it in this 
[quantize problem sample.zip](https://github.com/aseprite/aseprite/files/694237/quantize.problem.sample.zip)
zip.)

[...]

PR #2231, opened 2019-12-26
Added octree color quantization, an alternative to rgbmap quantization.

Gasparoken:
The main purpose of this addition is to introduce the OctreeMap class like a start point to:
1- Resolve issue #1385 which old rgbmap quantization can't get enought resolution to distinguish color with low contrast.
2- Improve the function create_palette_from_image, to get a better color matching.
3- Improve RGB->INDEXED conversion, to get better color resemblance of the original image.


src/app/file/file.cpp
```diff
@@ -905,7 +905,7 @@ void FileOp::postLoad()
       std::shared_ptr<Palette> palette(
         render::create_palette_from_sprite(
           sprite, frame_t(0), sprite->lastFrame(), true,
-          nullptr, nullptr, m_config.newBlend));
+          nullptr, nullptr, m_config.newBlend, MapAlgorithm::RGBA));
```
dacap: I think we need a `MapAlgorithm::DEFAULT` because here we are using for third time (or maybe more) a default value.

src/app/util/clipboard.cpp
```diff
@@ -385,12 +385,11 @@ void paste(Context* ctx, const bool interactive)
       }
       else {
         RgbMap* dst_rgbmap = dstSpr->rgbMap(site.frame());
-
         src_image.reset(
           render::convert_pixel_format(
             clipboard_image.get(), NULL, dstSpr->pixelFormat(),
             render::Dithering(),
-            dst_rgbmap, clipboard_palette.get(),
+            dst_rgbmap, nullptr, clipboard_palette.get(),
```
dacap: Here we have the  default algorithm too to convert the clipboard image.

src/doc/map_algorithm.h
```diff
@@ -0,0 +1,20 @@
...
+#ifndef DOC_MAP_ALGORITHM_H_INCLUDED
+#define DOC_MAP_ALGORITHM_H_INCLUDED
+#pragma once
+
+namespace doc {
+
+  enum class MapAlgorithm {
+    RGBA,
+    OCTREE
+  };
```
dacap: We can have a `DEFAULT = RGBA` or `DEFAULT = OCTREE` here.

src/doc/map_algorithm.h
```diff
@@ -0,0 +1,20 @@
...
+// Read LICENSE.txt for more information.
+
+#ifndef DOC_MAP_ALGORITHM_H_INCLUDED
+#define DOC_MAP_ALGORITHM_H_INCLUDED
+#pragma once
+
+namespace doc {
+
+  enum class MapAlgorithm {
+    RGBA,
```
dacap: Maybe `RGB5A3`?
```suggestion
    RGB5A3,
```

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+void OctreeNode::fillOrphansNodes(const Palette* palette,
+                                  color_t upstreamBranchColor, int level)
+{
+  for (int i=0; i<8; i++) {
+    if ((*m_children)[i].m_children)
+      (*m_children)[i].fillOrphansNodes(
+        palette,
+        upstreamBranchColor + octetToBranchColor(i, level),
+        level + 1);
```
dacap: Add `{ ... }` to the block inside this `if` statement (generally we would prefer to avoid `{}` only when the block inside the if has one line of code):
```
    if ((*m_children)[i].m_children) { ... }
```

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+    else if (!((*m_children)[i].isLeaf())) {
+      // Here the node IS NOT a Leaf and HAS NOT children
+      // So, we must assign palette index to the current node
```
dacap:
```suggestion
      // So, we must assign a palette index to the current node
```

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+    }
+  }
+}
+
+int OctreeNode::mapColor(int r, int g, int b, int a,
+                         int mask_index, int level)
+{
+  int indexLevel = (  (b >> (7 - level)) & 1) * 4
+                   + ((g >> (7 - level)) & 1) * 2
+                   + ((r >> (7 - level)) & 1);
```
dacap: Same here, `|` instead of `+`.

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+      // to fill the "map holes" (i.e "death tree branches")
+      int currentBranchColorAdd = octetToBranchColor(i, level);
+      int midColorAdd = MID_VALUE_COLOR >> (level + 1);
+      midColorAdd = (midColorAdd << rgba_r_shift)
+                  + (midColorAdd << rgba_g_shift)
+                  + (midColorAdd << rgba_b_shift);
+      color_t branchColorMed = rgba_a_mask
+                             + upstreamBranchColor
+                             + currentBranchColorAdd
+                             + midColorAdd;
```
dacap: I think all these `+` operations should be `|` (in practice the result should be the same, but prefer `|` if you are dealing with bits instead of integer values).

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+  return result - 1;
+}
+
+bool OctreeMap::makePalette(Palette* palette, int colorCount, int levelDeep)
+{
+
+  if (root()->children()) {
+    // We create paletteIndex to get a "global like" variable, in collectLeafNodes
+    // function, the purpose is having a incremental variable in the stack memory
+    // sharend between all recursive calls of collectLeafNodes.
```
dacap:
```suggestion
    // shared between all recursive calls of collectLeafNodes.
```

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+    int paletteIndex = 0;
+    root()->collectLeafNodes(m_leavesVector.get(), paletteIndex);
+  }
+
+  // If we can improve the octree accuracy, makePalette returns false, then
+  // outside from this function we must re-construct the octreeMap all again with
+  // deep level equal to 8.
```
dacap:
```
  // If we can improve the octree accuracy with an extra level, makePalette()
  // returns false, and then the caller must re-call makePalette() with a deep
  // level=8 to re-construct the octreeMap all over again.
```

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+      if (m_leavesVector->size() + auxLeavesVector.size() <= colorCount) {
+        for (int j=0; j < auxLeavesVector.size(); j++)
+          m_leavesVector->push_back(auxLeavesVector[auxLeavesVector.size() - 1 - j]);
+        keepReducingMap = false;
+        break;
+      }
+      else if (m_leavesVector->size() == 0) {
+        // When colorCount is < 8, auxLeavesVector->size() could reach the 8 size,
+        // if this is true and we don't stop the regular removeLeaves algorithm,
+        // the 8 remains colors will collapse in one.
```
dacap:
```suggestion
        // the 8 remaining colors will collapse in one.
```

src/doc/octree_map.cpp
```diff
@@ -0,0 +1,248 @@
...
+        // So, we have to reduce color with other method:
+        // Sort colors by pixelCount (most pixelCount on front of sortedVector),
+        // then:
+        // Blend in pairs from the least pixelCount colors.
```
dacap: Not sure what does `Blend in pairs from the least pixelCount colors` mean here.

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+    {}
+
+    LeafColor(int r, int g, int b, int pixelCount) :
+      m_r((double)r),
+      m_g((double)g),
+      m_b((double)b),
+      m_pixelCount(pixelCount)
+    {}
+
+    ~LeafColor() {}
```
dacap: This destructor is not necessary.

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+      int auxB = (((int)m_b) % m_pixelCount > m_pixelCount / 2)? 1 : 0;
+      return LeafColor(m_r / m_pixelCount + auxR,
+                           m_g / m_pixelCount + auxG,
+                           m_b / m_pixelCount + auxB,
+                           m_pixelCount);
+    }
+
+    color_t LeafColorToColor()
+    {
+      return 0xff000000 + (((int)m_b) << 16) + (((int)m_g) << 8) + (int)m_r;
```
dacap: Use `|` instead of `+`.

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+
+    LeafColor normalizeColor()
+    {
+      int auxR = (((int)m_r) % m_pixelCount > m_pixelCount / 2)? 1 : 0;
+      int auxG = (((int)m_g) % m_pixelCount > m_pixelCount / 2)? 1 : 0;
+      int auxB = (((int)m_b) % m_pixelCount > m_pixelCount / 2)? 1 : 0;
+      return LeafColor(m_r / m_pixelCount + auxR,
+                           m_g / m_pixelCount + auxG,
+                           m_b / m_pixelCount + auxB,
+                           m_pixelCount);
```
dacap: These lines are not aligned with the `(` of the previous line:
```cpp
      return LeafColor(m_r / m_pixelCount + auxR,
                       m_g / m_pixelCount + auxG,
                       m_b / m_pixelCount + auxB,
                       m_pixelCount);
```

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+
+public:
+  OctreeNode()
+  {
+    m_paletteIndex = 0;
+    m_children.reset(nullptr);
+    m_parent = nullptr;
+  };
+
+  ~OctreeNode()
```
dacap: This destructor should not be necessary.

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+    octet += (aux & 2);
+    return octet + ((aux >> 7) & 4);
+  }
+
+  static color_t octetToBranchColor(int octet, int level)
+  {
+    int auxR = (octet & 1) << (7 - level);
+    int auxG = (octet & 2) << (14 - level);
+    int auxB = (octet & 4) << (21 - level);
+    return auxR + auxG + auxB;
```
dacap: Use `|` instead of `+`

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+  {
+  }
+
+  static int getOctet(color_t c, int level)
+  {
+    int aux = c >> (7 - level);
+    int octet = aux & 1;
+    aux = aux >> (7);
+    octet += (aux & 2);
+    return octet + ((aux >> 7) & 4);
```
dacap: It looks like the `octet` is built bit by bit (1, 2, and then 4), so I guess it can be:
```cpp
    int octet = aux & 1;
    aux >>= 7;   // REMOVETHISCOMMENT check that we are using >>= instead of aux = aux >>
    octet |= (aux & 2);
    return octet | ((aux >> 7) & 4);
```
Not sure but I think this code should be equivalent.

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+class OctreeMap {
+
+public:
+  OctreeMap() :
+    m_root(new OctreeNode()),
+    m_leavesVector(new std::vector<OctreeNode*>()),
+    m_modifications(-1)
+  {}
+
+  ~OctreeMap()
```
dacap: Remove this destructor.

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+
+  void addColor(color_t color, int levelDeep = 7)
+  {
+    m_root->addColor(color, 0, m_root.get(), 0, levelDeep);
+  }
+
+  OctreeNode* root() const { return m_root.get(); };
+
+  // makePalette return true if a 7 level octreeDeep is OK, and false if we can add ONE level deep.
+  bool makePalette(Palette* palette, int colorCount, int leveleep = 7);
```
dacap:
```suggestion
  bool makePalette(Palette* palette, int colorCount, int levelDeep = 7);
```

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+  {}
+
+  void addColor(color_t color, int levelDeep = 7)
+  {
+    m_root->addColor(color, 0, m_root.get(), 0, levelDeep);
+  }
+
+  OctreeNode* root() const { return m_root.get(); };
+
+  // makePalette return true if a 7 level octreeDeep is OK, and false if we can add ONE level deep.
```
dacap:
```suggestion
  // Returns true if levelDeep=7 is enough for the given colors,
  // or false if we need an extra level of deep for more precision.
```

src/doc/octree_map.h
```diff
@@ -0,0 +1,182 @@
...
+  void fillOrphansNodes(const Palette* palette, color_t upstreamBranchColor, int level);
+
+  int mapColor(int r, int g, int b, int a,
+               int mask_index, int level);
+
+  void collectLeafNodes(std::vector<OctreeNode*>* leavesVector, int& paletteIndex);
+
+  // removeLeaves(): remove leaves from a common parent
+  // auxParentVector: i/o addreess of an auxiliary parent leaf Vector from outside.
+  // rootLeavesVector: i/o address of the m_root->m_leavesVector
```
dacap:
```cpp
  // Remove leaves from a common parent:
  // - auxParentVector: address of an auxiliary parent leaf Vector from outside.
  // - rootLeavesVector: address of the m_root->m_leavesVector
```

src/doc/sprite.cpp
```diff
@@ -368,6 +370,25 @@ RgbMap* Sprite::rgbMap(frame_t frame, RgbMapFor forLayer) const
   return m_rgbMap;
 }
 
+OctreeMap* Sprite::octreeInit(frame_t frame) const
+{
+  return octreeInit(frame, backgroundLayer() ? RgbMapFor::OpaqueLayer:
+                                           RgbMapFor::TransparentLayer);
```
dacap: This line is not aligned with the previous one:
```cpp
  return octreeInit(frame, backgroundLayer() ? RgbMapFor::OpaqueLayer:
                                               RgbMapFor::TransparentLayer);
```

src/doc/sprite.cpp
```diff
@@ -368,6 +370,25 @@ RgbMap* Sprite::rgbMap(frame_t frame, RgbMapFor forLayer) const
...
+}
+
+OctreeMap* Sprite::octreeInit(frame_t frame, RgbMapFor forLayer) const
+{
+  int maskIndex = (forLayer == RgbMapFor::OpaqueLayer ?
+                   -1: transparentColor());
+
+  if (m_octreeMap == NULL || m_octreeMap->getModifications() != palette(frame)->getModifications())
+    m_octreeMap = new OctreeMap();
```
dacap: As `m_octreeMap` is not an `std::unique_ptr`, there is a memory leak here when `m_octreeMap` is not `nullptr` and the modifications don't match. It needs to do something like:
```cpp
delete m_octreeMap;
m_octreeMap = new OctreeMap;
```
(or we can use a `std::unique_ptr` something that wasn't used for the rgbMap just because I wasn't using unique pointers at the time, I think this code is quite old)

src/render/quantization.cpp
```diff
@@ -44,8 +45,11 @@ Palette* create_palette_from_sprite(
   const bool withAlpha,
   Palette* palette,
   TaskDelegate* delegate,
-  const bool newBlend)
+  const bool newBlend,
+  MapAlgorithm mappingAlgorithm)
 {
+
+  OctreeMap* map = new OctreeMap();
```
dacap: Is this `map` always necessary? e.g. couldn't we create the map only when the algorithm is OCTREE?

There is a memory leak too, two memory leaks to be exact (see below).

src/render/quantization.cpp
```diff
@@ -71,13 +77,32 @@ Palette* create_palette_from_sprite(
...
-  optimizer.calculate(
-    palette,
-    // Transparent color is needed if we have transparent layers
-    (sprite->backgroundLayer() &&
-     sprite->allLayersCount() == 1 ? -1: sprite->transparentColor()));
-
+  if (mappingAlgorithm == MapAlgorithm::OCTREE) {
+    if (!(map->makePalette(palette, palette->size()))) {
+      // We can use an 8bits deep octree map
+      map = new OctreeMap();
```
dacap: Here is the other memory leak, this pointer is never deleted.

src/render/quantization.cpp
```diff
@@ -176,8 +201,11 @@ Image* convert_pixel_format(
 
             if (a == 0)
               *dst_it = new_mask_color;
-            else if (rgbmap)
-              *dst_it = rgbmap->mapColor(r, g, b, a);
+            if (octreeMap)
+                *dst_it = octreeMap->mapColor(r, g, b, a, new_mask_color);
+            else if (rgbmap) {
```
dacap: The added `{ ... }` in this `else if` are not necessary.

src/render/quantization.cpp
```diff
@@ -176,8 +201,11 @@ Image* convert_pixel_format(
 
             if (a == 0)
               *dst_it = new_mask_color;
-            else if (rgbmap)
-              *dst_it = rgbmap->mapColor(r, g, b, a);
+            if (octreeMap)
+                *dst_it = octreeMap->mapColor(r, g, b, a, new_mask_color);
```
dacap: Remove extra indentation:
```suggestion
              *dst_it = octreeMap->mapColor(r, g, b, a, new_mask_color);
```

src/render/quantization.cpp
```diff
@@ -176,8 +201,11 @@ Image* convert_pixel_format(
 
             if (a == 0)
               *dst_it = new_mask_color;
-            else if (rgbmap)
-              *dst_it = rgbmap->mapColor(r, g, b, a);
+            if (octreeMap)
```
dacap: I think it requires an `else` (because the first case `(a == 0)` was already handled:
```suggestion
            else if (octreeMap)
```

src/render/quantization.cpp
```diff
@@ -176,8 +201,11 @@ Image* convert_pixel_format(
...
+                *dst_it = octreeMap->mapColor(r, g, b, a, new_mask_color);
+            else if (rgbmap) {
+                *dst_it = rgbmap->mapColor(r, g, b, a);
```
dacap: Remove extra indentation:
```suggestion
              *dst_it = rgbmap->mapColor(r, g, b, a);
```

src/app/commands/cmd_color_quantization.cpp
```diff
@@ -31,11 +33,70 @@
...
+  MapAlgorithm m_mapAlgorithm;
+};
+#endif
+
 struct ColorQuantizationParams : public NewParams {
   Param<bool> ui { this, true, "ui" };
   Param<bool> withAlpha { this, true, "withAlpha" };
   Param<int> maxColors { this, 256, "maxColors" };
   Param<bool> useRange { this, false, "useRange" };
+  Param<MapAlgorithm> mappingAlgorithm { this, MapAlgorithm::RGBA, "mappingAlgorithm" };
```
dacap: The name of this parameter must match the name of the `ChangePixelFormatCommand` parameter.

PR #2250, opened 2020-01-21
Update Fedora/Linux sections in INSTALL.md

3n-k1:
The Fedora dependencies section was missing several dependencies. In addition, Cmake didn't detect compilers properly, so I added those to the Cmake command.


3n-k1:
After reviewing Aseprite's license, I'd like to request you don't use my PR. Thank you.

PR #2251, opened 2020-01-29
Update Clang version to 8.0.1 in INSTALL.md

Varstahl:
I tried to build Aseprite today on a fresh install, and VS2019+Ninja refused to compile, requiring 8.0.1+. Clang 9 is already available but I took the safe route and tested it with 8.0.1, installation went flawlessly after that.


dacap:
Thanks for this change @Varstahl, anyway there are new instructions to compile Aseprite and Skia. I've updated the LLVM version in the Skia repository:

https://github.com/aseprite/skia#skia-on-windows

But still using the download link to LLVM 9.0.0 from its website (instead of the github page). Not sure why the 9.0.1 is not published yet in their website.

Issue #2233, opened 2019-12-29
Cannot export file with absolute path on Linux


buckle2000:
When exporting, if the path is `/tmp/image.png`, the image will be exported at
 `<current directory>/tmp/image.png`, not at `/tmp/image.png`.

IMO, the software should always use `realpath`.

### Aseprite and System version

Same as in #2228.

What is going on?

PR #2258, opened 2020-02-09
Fix file selector not allowing choosing files by path

### LINKED ISSUE #2228: Can't select file by full path on Linux

* Aseprite version: v1.2.16.3
* System: Linux

When I want to choose a file, if I type in the full path, like `/tmp/test.png`, it will say:

image

This should not happen.

Adrijaned:
Fixes #2228 
Previously was treating path entered into the file selector either as path to a directory, or as a name of the file in current directory. Now allows also for entering path to the file, both relative and absolute.

### DIFF
--- a/src/app/ui/file_selector.cpp
+++ b/src/app/ui/file_selector.cpp
@@ -613,38 +613,33 @@ again:
     }
     // else file-name specified in the entry is really a file to open...
 
-    // check if the filename doesn't contain slashes or other ilegal characters...
-    bool has_invalid_char = (fn.find('/') != std::string::npos);
 #ifdef _WIN32
-    has_invalid_char =
-      has_invalid_char ||
-      (fn.find('\\') != std::string::npos ||
-       fn.find(':') != std::string::npos ||
-       fn.find('*') != std::string::npos ||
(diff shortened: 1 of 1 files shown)


Adrijaned:
I believe this also fixes #2233

dacap:
Hi @Adrijaned, thanks for your contribution, I'll try to take to this issue ASAP in this week.

src/app/ui/file_selector.cpp
```diff
@@ -613,38 +613,33 @@ bool FileSelector::show(
...
-
-      // show the window again
-      setVisible(true);
-      goto again;
+    // check if the filename doesn't contain ilegal characters...
+    // Only if it does not exist
+    // Only on Windows, Linux disallows only `/`, and we can consider that just an embedded path if present
+    if (!enter_folder) {
+      bool has_invalid_char =
+         fn.find(':') != std::string::npos ||
```
dacap: There is a missing `(`, anyway I'll check if I can add support to use full paths on Windows too.
```suggestion
        (fn.find(':') != std::string::npos ||
```

dacap:
Thanks @Adrijaned, we've merged your PR!

PR #2270, opened 2020-02-17
Update Clang link from 8.0.1 to 9.01

Abel-Moremi:
Updated the Clang link as it is out of date.


dacap:
Thanks for this change @Abel-Moremi, anyway there are new instructions to compile Aseprite and Skia. I've updated the LLVM version in the Skia repository:

https://github.com/aseprite/skia#skia-on-windows

But still using the download link to LLVM 9.0.0 from its website (instead of the github page). Not sure why the 9.0.1 is not published yet in their website.

PR #2272, opened 2020-02-17
Fix tile mode glitch with image brushes

Gasparoken:
This fix solves u,v set calculation when 'image brush' are used.

The bug origin is because we considered the set m_u, m_v as constants.

We have several cases when we are working on X axis:
- Scanline entirely contained in a tile, but OUTSIDE of the reference tile.
- Scanline entirely contained in a tile, but INSIDE of the reference tile.
- The left side of a scanline which was sliced (by the tile limit).
- The right side of a scanline which was sliced (by the tile limit).
Each one has its own m_u, so we need to recalculate it before every loop->getInk()->inkHline(...).

We have to take similar considerations when we are working in the Y axis.
We need recalculate m_v according if 'y+scanline.y' is INSIDE or OUTSIDE of the reference tile, each case has its own m_v.

Additionally, we added (x, y) considerations to negative values in app.useTool function.
Before this fix, the Points argument in useTool could result in negative (x, y) values.
Then point_shades processes were incorrect, when tile mode were on.

### DIFF
--- a/src/app/tools/ink.h
+++ b/src/app/tools/ink.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018  Igara Studio S.A.
+// Copyright (C) 2018, 2020  Igara Studio S.A.
 // Copyright (C) 2001-2017  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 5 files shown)


dacap:
Merged in 54bb39a9d39812bf0eaedcb3734fa83dea97e5e7  🎊

PR #2277, opened 2020-02-25
Fix lld-link error when building with clang on Windows

Adriankhl:
Under some settings, `lld-link` is the default linker, but doesn't accept `/LTCG` flag.
Use lld when compiling with clang and replace `/LTCG` by  '-flto=thin' optimization.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -14,7 +14,12 @@ if(MSVC)
   # As Skia is compiled with /GL flag (whole program optimization),
   # the linker prefer a /LTCG parameter to improve link times.
   if(LAF_OS_BACKEND STREQUAL "skia")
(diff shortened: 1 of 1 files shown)


dacap:
Not sure if there is a way to detect the linker (because we could use clang to compile but the original MSVC `link.exe` to link).

Adriankhl:
I was thinking of that as well, but I failed to find any built-in way to detect the linker with cmake. Custom cmake function could work but it maybe overly complicated. 

The problem is that it seems like it is possible to force `clang-cl` to use `lld-link.exe`, but there is no way to force `clang-cl` to use `link.exe`. Therefore, if the default is `lld-link` the build fails.

Maybe a workaround is to add an `USE_LLD` to force this behaviour.

PR #2306, opened 2020-03-24
add css export

zaghaghi:
### Features
* ability to change pixel size
* use CSS3 variables
* generate sample HTML file

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -330,6 +330,12 @@
       <option id="bits_per_pixel" type="int" default="0" />
       <option id="compress" type="bool" default="true" />
     </section>
+    <section id="css">
+      <option id="show_alert" type="bool" default="true" />
+      <option id="pixel_scale" type="int" default="1" />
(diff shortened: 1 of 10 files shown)


dacap:
Thanks @zaghaghi, I'll try to review this right now.

dacap:
I'll merge this PR as it is with some minor changes. It's a good start for a first CSS export, in the future we could improve the output CSS with box-shadow and probably supporting animations with `@keyframes`:

E.g.
* https://medium.com/@araltasher/create-pixel-art-using-css-part-1-ebc9ca2f034
* https://jsfiddle.net/websiter/wnrxmapu/2/ ( source https://stackoverflow.com/questions/50559036/css-box-shadow-animated-pixel-art-flickering )

dacap:
Oh I just saw that unchecking the *CSS3 variables* option this generates the CSS using box-shadow, I think I'll put that as the default option (as the output looks a lot less verbose).

PR #2348, opened 2020-04-22
Fix gif encoder, blended indexed images could be selected to preserve palette orders, but it is not possible.

Gasparoken:


gaganmastana13 (approved):

gaganmastana13 (approved):

gaganmastana13 (approved):

dacap:
I think we can close this PR now because the GIF encoder was modified in the beta repo for v1.3 to be released in the future.

PR #2391, opened 2020-05-21
Improvements to nearest color finding and old dithering

Pigu-A:
I have been noticing that Aseprite produced an image with a weird yellow-ish tone when I convert it into an indexed format. So I decided to give a look into the code too see what went wrong.

I have found the main culprit which is an incorrectly implemented luma calculation in `initBestfit()` in src/doc/palette.cpp that forgets to perform a square root after doing a weighted summing. Making blue barely contribute to the color difference.

However, instead of adding one line to perform a square root, I decided to implement a better color distance metrics instead. Also, I made some fixes to the old dithering method to not generate "un-pattern" when the color is about half-way between two nearest colors.

You can see the difference in the comparison image below. Each column is original, v1.2.18's output and this fix's output in order. All of them is converted to DB32 palette with 4x4 bayer matrix old dithering.

colorcomp


dacap:
Hi @Pigu-A, first please check the steps to [sign the CLA](https://github.com/aseprite/sourcecode/blob/master/sign-cla.md#sign-the-cla).

After that we can review the PR. As a initial comment I can say if you can squash the change in one commit and remove non-changes like [this one](https://github.com/aseprite/aseprite/pull/2391/files#diff-65d7b15c9ad7475c9e300cefe86d15c8L2). Thanks!

Pigu-A:
Ah, I see. I think I can't do it in this account then. I'll close this for now.

PR #2403, opened 2020-05-25
Implement wheel direction customization

### LINKED ISSUE #1801: Add possibility to customize mouse wheel direction
Requested here: https://steamcommunity.com/app/431730/discussions/0/1759104257594233338/

Yanrishatum:
Resolves #1801 
Implements wheel direction customization per-hotkey, completely resolving feature request.

* Adds a checkbox labeled "Reverse wheel" to reverse wheel direction on mouse wheel hotkey (Hidden when editing keyboard hotkeys).
* While editing hotkey: Scrolling wheel vertically will automatically set/reset reverse flag.
* Added `reverseFlag()` to `Accelerator` in order to signal if reverse flag is set.
* `KeyboardShortcuts::getWheelActionFromMouseMessage` now takes pointer to wheel delta argument and if accelerator has reverse flag - negates it.
* Hotkeys with reverse flag set indicate so with ` (Reverse)` label.
* Reverse flag is saved as `reverse` attribute in settings XML.

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -1369,6 +1369,7 @@ alt = Alt
 shift = Shift
 space = Space
 win = Win
+reverse = Reverse Wheel
 assigned_to = Assigned to:
 ok = OK
 cancel = Cancel
--- a/data/widgets/select_accelerator.xml
+++ b/data/widgets/select_accelerator.xml
@@ -16,6 +16,7 @@
           <check text="@.shift" id="shift" />
           <check text="@.space" id="space" />
           <check text="@.win" id="win" />
+          <check text="@.reverse" id="reverse"/>
         </hbox>
 
         <label text="@.assigned_to" />
--- a/src/app/commands/cmd_keyboard_shortcuts.cpp
+++ b/src/app/commands/cmd_keyboard_shortcuts.cpp
@@ -450,9 +450,13 @@ private:
   }
 
   std::string getAccelText(const Accelerator& accel) const {
-    if (m_key && m_key->type() == KeyType::WheelAction &&
-        accel.isEmpty()) {
-      return "(Default Action)";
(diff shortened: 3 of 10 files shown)


dacap:
Hi @Yanrishatum, thanks for your PR. I'll try to review this soon and make comments on each line. I have to say that I don't like the idea to add a new field to the `ui::Accelerator` class related to the mouse, but it might work (I've to check it in more detail). I was thinking more about this mouse wheel reversion like a checkbox to reverse the global behavior (or an option for each mouse wheel action), without associating this configuration with all shortcuts.

In the past I was thinking about integrating Left/Middle/Right/X1/X2 mouse button information into `ui::Message` in some way (or into `app::Key`), but maybe the `ui::Accelerator` is the place to add this. The actual design is a mess, `ui::Accelerator` is in some way "legacy" from old C code, and there are several classes to be integrated in some way:

* `os::Shortcut`
* `ui::Accelerator`
* `app::Key`
* + mouse buttons up/down events & mouse wheel

I wasn't able to dedicate some time to this yet and that is why I'm not sure about touching `ui::Accelerator` yet.

Yanrishatum:
To be frank, I too, not fond of putting that flag into `Accelerator` (at least in a way I did), but I focused the minimal impact of the change, as I'm not very familiar with the code-base. 
Having reverse as a global option is a bad idea. For a reference: Both me and my friend (due to which this PR came to be) don't like that next/prev frame action is counter-intuitive, while I also dislike that you need to scroll down in order to increase brush size, but direction of other actions I'm fine with. If that setting is global - it would not be a solution, and just a switch from one annoyance to another (and rather bigger one, since it just reverses all actions).
And per-action may be another option, but I have no idea how to properly implement it. Well, moving it to `Accelerators` is an option, but that is pretty much same thing, but a bit less customizable.

Issue #1812, opened 2018-07-31
Swap Editor background checkerboard colours as a command.


oceanhahn:
This is kind of a small thing but I would love to have a bindable command that swaps the colours in the Editor workspace checkboard background. This is mainly because the perceptible contrast changes when the colours are swapped, and moreso because there are (many) times when the colour I am working with is very similar to the checkerboard square I am working over top of.

The command would replace having to go to Preferences/Background and changing the colours manually, which is time consuming and includes a lot of somewhat finicky steps, and which must be repeated every time the colours need to be swapped (which can be often).

There are workarounds of course, but it comes up quite frequently and it would be fantastic to be able to do this with a single keystroke instead of a dozen or more mouseclicks.

This command would appear in the View menubar list, and would probably _not_ have a default assignment.

Thanks for considering it! :D 

..using ASEprite v1.2.9 (installer) on Windoze 10.

PR #2408, opened 2020-05-28
Added swap checkerboard background colors command (#1812)

MorganDavid:
Added as a command with no default shortcut and no menu item.
https://github.com/aseprite/aseprite/issues/1812


dacap:
After tha CLA process is finished, about this PR, not sure if the command should be called `SwapCheckerboardColors`, or `SwapCheckedBackgroundColors`, or `SwitchCheckedBackgroundColors` (like `SwitchColors`), or maybe just an argument for `SwitchColors` 🤔

MorganDavid:
Personally, I think swap makes more sense, but to keep it consistent with `SwitchColors` I could change it to `SwitchCheckedBackgroundColors`? I thought about adding params to `SwitchColors` since the behaviour is similar but decided to make a new command because the usage in the application is very different. 

Also, am I doing anything wrong for the Travis CI build to fail?

dacap:
Or don't worry, `SwapCheckerboardColors` it's fine 👍 , we can use it in that way (I might change the `SwitchColors` to `SwapColors` in a future with an alias).

dacap:
About the CI, actually I've fixed it today, and I think this PR was forked from a commit that was already failing, so nothing to worry about.

MorganDavid:
It seems my commits were from the wrong GitHub account, I'll close this and make a new PR.

Issue #2388, opened 2020-05-19
Recent File list sorts items incorrectly at times


KashouC:
I'm not sure what causes this bug, but from time to time the recent files list is not sorted correctly and gets all jumbled. I've managed to recreate the issue in a video here:

https://www.youtube.com/watch?v=hbJQe2eDfkk

Gasparoken:
The 'Recent Files' order depends on the 'SimpleIni.h' code. This order depends on how the 'insert' function manipulates a keyval value. The keyval sorting is according 'char' values, instead of 'integer' values.
'aseprite.ini':
<img width="653" alt="Screen Shot 2020-05-26 at 11 11 59" src="https://user-images.githubusercontent.com/32165475/82920414-9dfb1f00-9f4d-11ea-8398-10df4a3016ed.png">
Keyval pairs during execution:
<img width="669" alt="Screen Shot 2020-05-26 at 11 10 14" src="https://user-images.githubusercontent.com/32165475/82920335-8a4fb880-9f4d-11ea-9258-4825079a2369.png">
@dacap What're you think about how to deal with this change?

dacap:
As we talked the problem is the `std::multimap` implementation and maybe a re-ordering of their nodes (so then the iteration of those nodes gives a different order), and it might depend on the specific `std::multimap` implementation. Possible solutions:
* patch simpleini library to use other kind of structure or iteration to read keys
* use other library to read ini files that keep the order of the keys as they were added (it must be compatible with simpleini, mainly for multilines strings because we use this library to load translations files like `en.ini`)

dacap:
Thinking a little more about this I'm not sure why we depend on the internal order of the Simpleini structures (`std::multimap`) and not on the keys directly 🤔 @Gasparoken check the `enum_config_keys()` from `app/ini_file.cpp` maybe the order of keys returned from that function is the problem.

Gasparoken:
@dacap, I found a good and simple solution in the "laf-base". I forced the string representation of the integer to 4 digit (fill with zeros at the left). I will test it, to see what others things I broke 😅 .

![Screen Shot 2020-05-27 at 15 58 41](https://user-images.githubusercontent.com/32165475/83062442-01ac4780-a035-11ea-84ab-eed5a7469548.png)

dacap:
[...]

PR #2409, opened 2020-05-28
Fix recent files history order on home view when the file count is greater than 10

Gasparoken:
This fix requires a laf update (merge Gasparoken/add-covert branch to include the function `convert_to_4_digits_string` in `convert_to.cpp`), and it can solve issue #2388.


Gasparoken:
It will treated in other way.

PR #2410, opened 2020-05-28
Fix recent files history order on home view when the file count is greater than 10 (issue #2388)

Gasparoken:
This fix is backward compatible (it preserves Recent Files history of the older Aseprite versions).

### DIFF
--- a/src/app/recent_files.cpp
+++ b/src/app/recent_files.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018  Igara Studio S.A.
+// Copyright (C) 2018-2020  Igara Studio S.A.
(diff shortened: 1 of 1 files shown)


dacap:
Hi there @Gasparoken, checking this I'm not sure if this is working, I get this exception when exiting Aseprite:
```
libc++abi.dylib: terminating with uncaught exception of type fmt::v6::format_error: missing '}' in format string
```
I guess the format string given to `fmt::format` is not valid (so not sure if this is even saving the .ini information).

Gasparoken:
It's ready now. Thank you David.

dacap:
Now it works! Even opening the .ini with old versions recent/pinned files/folders still are loaded correctly. Merging it.

PR #2422, opened 2020-06-01
Add crisp shape rendering for svg format

danprince:
The SVG exports that Aseprite produces seem to have visual artifacts between pixels when rendered in current versions of Chrome, Safari, and Firefox.

I noticed that SVGs produced with [Pixels.svg](https://codepen.io/shshaw/full/XbxvNj) don't have this and it seems to be due to the [`shape-rendering: crisp-edges`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/shape-rendering) attribute.

| Current Aseprite Export | With `shape-rendering: crisp-edges` |
| --- | --- |
| image | image |

The CLA seems a bit heavy, so I haven't signed it. It's a small enough change that you can close this PR and have someone else commit it if needs be.

Thanks for the great tool!

### DIFF
--- a/src/app/file/svg_format.cpp
+++ b/src/app/file/svg_format.cpp
@@ -93,7 +93,7 @@ bool SvgFormat::onSave(FileOp* fop)
     fprintf(f, "/>\n");
   };
   fprintf(f, "<?xml version=\"1.0\" encoding=\"UTF-8\" ?>\n");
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @danprince, I'll merge this change as it's.

p.s. In a near future I'll simplify the CLA signing process (just an extra commit in the same repository to accept the licensing terms would be enough).

PR #2426, opened 2020-06-04
Fix locked layer not completely locked (fix #2181)

### LINKED ISSUE #2181: Locked layer not completely locked
If a layer is locked every operation should be disabled.

If I make a selection I can move it around using the arrow keys.
Moving with the mouse is disabled as it should.

* Aseprite version: v1.2.15 compiled
* System: Windows 10 x64

Gasparoken:


src/app/ui/editor/moving_pixels_state.cpp
```diff
@@ -285,6 +285,13 @@ bool MovingPixelsState::onMouseDown(Editor* editor, MouseMessage* msg)
   Decorator* decorator = static_cast<Decorator*>(editor->decorator());
   Doc* document = editor->document();
 
+  // Return if it cannot draw in current layer
+  Layer* layer = editor->getSite().layer();
```
dacap, line 289: Here using `Editor::layer()` will be better (same for other cases):
```suggestion
  Layer* layer = editor->layer();
```
I generally avoid creating the `Site` object (because it contains too much information, range of selected layers/frames from the timeline, etc.).

src/app/ui/editor/moving_pixels_state.cpp
```diff
@@ -532,6 +539,20 @@ void MovingPixelsState::onBeforeCommandExecution(CommandExecutionEvent& ev)
...
+  Layer* layer = m_editor->getSite().layer();
+  if (layer && !layer->isEditableHierarchy() &&
+      (command->id() == CommandId::MoveMask() ||
+      command->id() == CommandId::Flip() ||
+      command->id() == CommandId::Cut() ||
+      command->id() == CommandId::Clear() ||
+      command->id() == CommandId::Rotate())) {
+    StatusBar::instance()->showTip(
+      1000, fmt::format("Layer '{}' is locked", layer->name()));
+    dropPixels();
```
dacap, line 552: I'm thinking if we should `dropPixels()` here or listening for changes in the lock status of the layer i.e. when the lock changes, we drop the pixels, so we are not going to be here in MovingPixelsState with an unlocked layer. Maybe when a "on before command" of `LayerLock` is received, we drop the pixels if the layer is unlocked.

src/app/ui/editor/standby_state.cpp
```diff
@@ -518,10 +531,6 @@ bool StandbyState::onSetCursor(Editor* editor, const gfx::Point& mouseScreenPos)
   if (editor->canDraw()) {
     editor->showBrushPreview(mouseScreenPos);
   }
-  // Forbidden
-  else {
-    editor->showMouseCursor(kForbiddenCursor);
```
dacap, line 523: Are we sure that this will never happen again? Isn't needed for reference layers maybe?

src/app/ui/editor/standby_state.cpp
```diff
@@ -747,6 +756,13 @@ void StandbyState::transformSelection(Editor* editor, MouseMessage* msg, HandleT
     return;
   }
 
+  if (layer && !layer->isEditableHierarchy()) {
```
dacap, line 759: There are too much duplicated code calling `transformSelection()` that already verifies the lock/visible state of the layer, maybe we can move all those verifications inside this function:
```
      if (!layer->isEditableHierarchy()) {
        StatusBar::instance()->showTip(
          1000, fmt::format("Layer '{}' is locked", layer->name()));
        return true;
      }
      else if (!layer->isVisibleHierarchy()) {
        StatusBar::instance()->showTip(
          1000, fmt::format("Layer '{}' is hidden", layer->name()));
        return true;
      }
```

src/app/ui/editor/standby_state.cpp
```diff
@@ -325,6 +325,11 @@ bool StandbyState::onMouseDown(Editor* editor, MouseMessage* msg)
           1000, fmt::format("Layer '{}' is locked", layer->name()));
         return true;
       }
+      else if (!layer->isVisibleHierarchy()) {
+        StatusBar::instance()->showTip(
+          1000, fmt::format("Layer '{}' is hidden", layer->name()));
```
dacap, line 330: I think we can move all these `"Layer '{}' is locked"` and `"Layer '{}' is hidden"` to `data/strings/en.ini` to avoid repeating. They could be in the `[statusbar_tips]` section.

src/app/ui/editor/standby_state.cpp
```diff
@@ -429,6 +434,14 @@ bool StandbyState::onSetCursor(Editor* editor, const gfx::Point& mouseScreenPos)
...
     return true;
 
+  // Forbidden
+  Layer* layer = editor->getSite().layer();
+  if (layer && !layer->isEditableHierarchy()) {
+    editor->showMouseCursor(
+      kCustomCursor, skin::SkinTheme::instance()->cursors.forbidden());
+    return true;
+  }
+
```
dacap, line 444: I'm not sure about this, maybe too restrictive here, because even with the forbidden cursor we can still select pixels.

src/app/util/layer_utils.h
```diff
@@ -14,13 +14,19 @@ namespace doc {
...
   // Calculates a possible candidate to be selected in case that we
   // have a specific "selectedLayer" and are going to delete the given
   // "layerToDelete".
   doc::Layer* candidate_if_layer_is_deleted(
     const doc::Layer* selectedLayer,
     const doc::Layer* layerToDelete);
 
+  // True if the active layer is locked (itself or its hierarchy),
+  // also, it sends a tip to the user 'Layer ... is locked'
+  bool layer_is_locked(Editor* editor);
```
dacap, line 28: I would have used a `Layer* layer` as parameter, but it's fine, we can use the `Editor` (and change it later if it's needed, e.g. if we need to use this function from a command where there is no `Editor`)

src/app/ui/editor/standby_state.cpp
```diff
@@ -299,11 +301,8 @@ bool StandbyState::onMouseDown(Editor* editor, MouseMessage* msg)
         int x, y, opacity;
         Image* image = site.image(&x, &y, &opacity);
         if (layer && image) {
-          if (!layer->isEditableHierarchy()) {
-            StatusBar::instance()->showTip(
-              1000, fmt::format("Layer '{}' is locked", layer->name()));
+          if (layer_is_locked(editor))
```
dacap, line 304: Maybe as `transformSelection()` checks for `layer_is_locked()` there is no need to do this here (same for the other cases where `layer_is_locked` is called just before `transformSelection`)

src/app/ui/editor/moving_pixels_state.cpp
```diff
@@ -285,6 +286,20 @@ bool MovingPixelsState::onMouseDown(Editor* editor, MouseMessage* msg)
...
+                                                      getTransformation(editor));
+  if (editor->isInsideSelection() || handle != NoHandle) {
+    if (layer_is_locked(editor))
+      return true;
+  }
+  else {
+    dropPixels();
+    return StandbyState::onMouseDown(editor, msg);
+  }
+
```
dacap, line 302: What specific case is this filtering? Isn't this related to the condition below? Couldn't we add a `layer_is_locked` check just before `m_pixelsMovement->catchImageAgain(...)` below?  (not sure about this because I don't fully understand why this code is here now)

dacap:
It looks like this failed in the CI: https://travis-ci.org/github/aseprite/aseprite/jobs/696512150#L1769 (when there is no UI, `ENABLE_UI=OFF`), anyway I continue from here and I'll try to fix the CI issue when merging this commit.

dacap:
Cherry picked in 9c37ea41ed6d3fc4b8e5e16067ccd89555886d07 with a couple of changes:

* [Wrapped with `#ifdef/#endif ENABLE_UI`](https://github.com/aseprite/aseprite/commit/9c37ea41ed6d3fc4b8e5e16067ccd89555886d07#diff-218d9d4ae6eb42ee2a2f420436276cc3R49-R53) to avoid using StatusBar when the UI is not available
* Moved the `layer_is_locked()` case when we ask `editor->isInsideSelection()` inside the already existent condition (with this we also make the right-click work even when the layer is locked)

Issue #2440, opened 2020-06-14
Weird issue with self-compiled Skia


enderger:
# Issue: Weird issue with self-compiled Skia
### There is a weird issue involving Skia and compiling Aseprite:
- When I use the `aseprite` Ninja target, everything compiles fine
- When I use the `install` Ninja target, it throws several errors along the lines of:
`FAILED: <name>
<LONG stacktrace>
/usr/bin/ld: /usr/lib64/libfreetype.so.6: error adding symbols: DSO missing from command line
clang-10.0: error: linker command failed with exit code 1 (use -v to see invocation)`

### Aseprite and System version

* Aseprite version: version number, installer/portable/Steam/beta/dev/commit-hash
https://github.com/aseprite/aseprite/commit/34af05411843c96aec4f25b702aa5663cc0f62db

* System: Windows/macOS/Linux, version, distribution
OpenSUSE Tumbleweed, kernel version 5.6.14-1-default

QuinntonCarter:
I'm also running into this issue while trying to compile. I'm not sure how to fix it as I can't find any pointers upon Google searching the error message.

Issue #2387, opened 2020-05-17
Multiple definition errors while linking Aseprite


thechubbypanda:
### Aseprite and System version

* Aseprite version: Release v1.2.18
* System: Linux, Latest, Arch

```
[1/1] Linking CXX executable bin/aseprite
FAILED: bin/aseprite 
[...]

PR #2441, opened 2020-06-14
Add notice to install.md

enderger:
This gives a fix for issues #2440 and #2387 in INSTALL.md on a temporary basis.


enderger:
The CI appears to have failed for some reason, there were only documentation changes in this PR and the failiures were both likely overstressed builds

dacap:
Hi @enderger, I'm not sure about this note because I should be able to reproduce this problem. I'll keep the PR open for some time until I have time to take care of this.

Issue #1699, opened 2018-04-04
Why did you delete USE_SHARED_LIBWEBP?


yurivict:
Shared webp worked very well. Webp can't be just built on FreeBSD, it needs patches. And rebuilding is undesirable, since the package exists. Now I can't update the port because of this.

yurivict:
> If webpdemux and webpmux libraries are distributed with webp, we can add the flag back.

They are distributed with ```webp```:
```
$ pkg info -l webp | grep lib
	/usr/local/lib/libwebp.a
	/usr/local/lib/libwebp.so
	/usr/local/lib/libwebp.so.7
	/usr/local/lib/libwebp.so.7.0.1
	/usr/local/lib/libwebpdecoder.a
	/usr/local/lib/libwebpdecoder.so
	/usr/local/lib/libwebpdecoder.so.3
	/usr/local/lib/libwebpdecoder.so.3.0.1
	/usr/local/lib/libwebpdemux.a
	/usr/local/lib/libwebpdemux.so
	/usr/local/lib/libwebpdemux.so.2
	/usr/local/lib/libwebpdemux.so.2.0.3
	/usr/local/lib/libwebpmux.a
	/usr/local/lib/libwebpmux.so
	/usr/local/lib/libwebpmux.so.3
	/usr/local/lib/libwebpmux.so.3.0.1
```

dacap:
I'll give a try to `find_library` for each individual library.

dacap:
@yurivict I've tested using the shared webp included in Ubuntu 14.04 but [it failed](https://travis-ci.org/aseprite/aseprite/builds/368198450). You can give a try to the [shared-webp](https://github.com/aseprite/aseprite/tree/shared-webp) branch.

winterheart:
aseprite requires >=webp-0.5.2, but Ubuntu 14.04 has only 0.4.4. Appropriate version can be found in Ubuntu 17.10.

yurivict:
Please consider bringing back ```USE_SHARED_WEBP```.
I patched cmake files to use the external ```webp```, and it builds with it fine.

cata0309:
libwebpmux cannot be found on Solus Linux but it is installed on the system both regular and dev versions of package(with headers and libs). What to do ?

Xtendera:
Based on the mentioned issue, it seems like this issue can be closed for the moment. @dacap

Issue #1681, opened 2018-03-17
Please add USE_SHARED_LIBARCHIVE, USE_SHARED_FMT, USE_SHARED_STRINGENCODERS, USE_SHARED_MUJS


yurivict:
These packages exist as standalone packages and can be used as shared.

dacap:
Sorry but the idea of the official distribution is to have almost no dependencies so Aseprite can be used in most distributions.

dacap:
Oh sorry again 😄, I guess you are talking about adding these options in the CMakeLists.txt!

yurivict:
Yes, to CMakeLists.txt -)

merwok:
Would this be a simple change to do?  I don’t know cmake, but I could see how USE_SHARED_LIBPNG is used and adapt it for libarchive at least, unless there are more places to change that can't be found by searching for USE_SHARED_

yurivict:
I am very reluctant to update a FreeBSD package because excessive bundling causes a lot of extra-work.

dacap:
At the moment there are no plans to include these flags, patches will be accepted but those patches will need to keep things working on all platforms (Windows, macOS, Linux).

PR #2523, opened 2020-09-26
Shared libs

Drauthius:
Make it possible to compile and link against the system's libarchive and webp.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ option(USE_SHARED_CURL    "Use your installed copy of curl" off)
 option(USE_SHARED_GIFLIB  "Use your installed copy of giflib" off)
 option(USE_SHARED_JPEGLIB "Use your installed copy of jpeglib" off)
 option(USE_SHARED_ZLIB    "Use your installed copy of zlib" off)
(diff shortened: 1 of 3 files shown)


ryandesign:
This would fix #1699.

Your fix looks similar to a previous attempt to fix this, from https://github.com/orivej/aseprite/commit/6e3f74107c40029584c524d7458f5c5674c933c3 and https://github.com/orivej/aseprite/commit/ea87e65b357ad0bd65467af5529183b5a48a8c17, except that your fix is missing some changes that were part of https://github.com/orivej/aseprite/commit/6e3f74107c40029584c524d7458f5c5674c933c3 and I don't know if they were important. The previous fix worked for me on macOS; I have not tried this new fix.

ryandesign:
This would be a step toward fixing #1681.

Drauthius:
My changes have worked on my Gentoo Linux system for quite some time now. I don't see the reason for the extra lines in orivej@6e3f741, unless webp can be split up into different include directories or libraries on some systems. Either way, I feel like since these settings are opt-in, they shouldn't break any current builds, and that if they aren't comprehensive enough, future pull requests could expand upon them to work with more systems.

PR #2535, opened 2020-10-08
Make LibArchive as shared library dependency

winterheart:
Add -DUSE_SHARED_LIBARCHIVE option to enable shared library usage.


ryandesign:
A similar fix was already submitted in #2523.

ryandesign:
This would be a step toward fixing #1681.

dacap:
Hi there! thanks @winterheart and @ryandesign, I'll review this soon in my Linux VM, I think I prefer this PR instead of #2523 because it adds only support for libarchive (and the webp is more complex, because there are incompatibilities between versions, what library to link, etc.).

dacap:
This PR was closed automatically because we deleted the `master` branch (now know as `main` branch).

Anyway I'm not sure if we should keep adding more USE_SHARED flags or remove them all: #2843

From my point of view is that more options bring more problems.

ISSOtm:
This allows working around an issue I have building Aseprite: I require `-DCMAKE_BUILD_TYPE=None` (and since Aseprite defaults to `RelWithDebInfo` if it is unspecified, I must specify it explicitly); however, `libarchive` rejects that option (it only defaults to `Debug` if the cache entry is unset... but Aseprite's own `CMakeLists.txt` prevents that.)

This relates to the points I made in https://github.com/aseprite/aseprite/issues/2843#issuecomment-1003416723.

PR #2594, opened 2020-12-20
Update thumbnail_handler.cpp

Aayush-Bachwana-25:
Indented the previous code.


dacap:
Hi @Aayush-Bachwana-25, we are not currently accepting PRs only to re-indent the code (even more when the style is different from the current style we are using). Thanks anyway!

Issue #2584, opened 2020-12-04
Pen Pressure + Symmetry bug


alexpandanet:
Some pixels will either be removed or added when using vertical symmetry and pen pressure in tandem. In particular, pen pressure affecting pen size.
![image](https://user-images.githubusercontent.com/75462439/101117510-58767900-35a4-11eb-8282-9a3270822cf2.png)

Aseprite v1.2.25-x64

Windows 10 Enterprise 10.0.18363

dacap:
Maybe related to #1427

dacap:
This should be fixed in v1.2.26 https://community.aseprite.org/t/aseprite-v1-2-26/8720

PR #2599, opened 2020-12-29
Fix pen pressure size + symmetry

Gasparoken:
This fix improves the symmetry drawing when symmetry and the pen pressure size are on.

### DIFF
--- a/src/app/tools/symmetries.cpp
+++ b/src/app/tools/symmetries.cpp
@@ -38,8 +38,13 @@ void HorizontalSymmetry::generateStrokes(const Stroke& mainStroke, Strokes& stro
   strokes.push_back(mainStroke);
 
   Stroke stroke2;
+  const bool isDynamic = loop->getDynamics().isDynamic();
   for (const auto& pt : mainStroke) {
     Stroke::Pt pt2 = pt;
+    if (isDynamic) {
+      brushSize = pt2.size;
+      brushCenter = (brushSize - brushSize % 2) / 2;
+    }
     pt2.x = m_x - ((pt.x-brushCenter) - m_x + 1) - (brushSize - brushCenter - 1);
     stroke2.addPoint(pt2);
   }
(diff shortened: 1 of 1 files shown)


src/app/tools/intertwine.cpp
```diff
@@ -69,7 +69,8 @@ void Intertwine::doPointshapeStrokePt(const Stroke::Pt& pt, ToolLoop* loop)
     main_stroke.addPoint(pt);
 
     Strokes strokes;
-    symmetry->generateStrokes(main_stroke, strokes, loop);
+    DynamicsOptions dynamics = loop->getDynamics();
+    symmetry->generateStrokes(main_stroke, strokes, loop, &dynamics);
```
dacap, line 73: As the `generateStrokes()` already receives the `loop`, we can get the dynamics inside `generateStrokes()` instead of adding a new parameter.

src/app/tools/symmetries.cpp
```diff
@@ -40,14 +41,19 @@ void HorizontalSymmetry::generateStrokes(const Stroke& mainStroke, Strokes& stro
   Stroke stroke2;
   for (const auto& pt : mainStroke) {
```
dacap, line 42:
```suggestion
  const DynamicsOptions dynamics = loop->getDynamics();
  for (const auto& pt : mainStroke) {
```

src/app/tools/symmetries.cpp
```diff
@@ -40,14 +41,19 @@ void HorizontalSymmetry::generateStrokes(const Stroke& mainStroke, Strokes& stro
...
     Stroke::Pt pt2 = pt;
+    if (dynamics->isDynamic()) {
+      brushSize = pt2.size;
+      brushCenter = (brushSize - brushSize % 2) / 2;
```
dacap, line 46: It looks like the brush center (brush->center()) is lost here, but I might be wrong. Maybe some tests could be created about this in https://github.com/aseprite/tests/blob/c4252301bf0272433d9c22d8922b536e963f602c/scripts/tools.lua#L391

dacap (changes requested):

src/app/tools/symmetries.cpp
Gasparoken: I ran the tests and it was OK. To test a pen with dynamic size, we need first an emulation of pressure in app.useTool. I think, may be in a future.

src/app/tools/symmetries.cpp
```diff
@@ -40,6 +40,10 @@ void HorizontalSymmetry::generateStrokes(const Stroke& mainStroke, Strokes& stro
   Stroke stroke2;
   for (const auto& pt : mainStroke) {
     Stroke::Pt pt2 = pt;
+    if (loop->getDynamics().isDynamic()) {
```
dacap, line 43: Please cache this boolean value to avoid calling getDynamics() on each for() step.
```
bool isDynamic = loop->getDynamics().isDynamic();
for (...
```

dacap:
First commit to fix https://github.com/aseprite/aseprite/issues/2584 (it works for symmetrical brushes, although the square brush is not working 🤔).

Issue #2179, opened 2019-10-15
[Request] Save Index palettes


leopardheart982:
A simple way to save and retrieve an indexed color palette from one sprite so that you can later use the same palette for another would be really convenient. Say you make a sprite that uses #FFF200, #9bfd71, #c8c600, and #6ccf44 for its colors. You index these colors, and under Color Mode > More Options there would be the option to save the indexed colors with a + button. You can give the indexed color palatte a name like "Yellow Flower", then next time you make a sprite, you go into Color Mode > More Options again and choose your pre-saved color palatte, and the index with all its colors are loaded.

KashouC:
You can actually do this already. You can then load it as a file, or you can add it to the palette folder and it will show up in the list of palette presets.

![image](https://user-images.githubusercontent.com/17066488/66845538-76cc8000-ef70-11e9-9080-7f819089a239.png)

dacap:
The "Load Palette"/"Save Palette" can be used, also a script can be used. Anyway other users have asked about the possibility to have a set of palettes inside the same sprite (and being able to switch between them).

PR #2601, opened 2020-12-30
Add palette option to save as preset

mezz:
This works the same as saving a palette normally, but opens the file dialogue to the user's palette preset directory.
This streamlines the workflow for artists who want to create and iterate on multiple palettes, and demystifies some of the behavior around creating custom palette presets.

A new menu option appears:
Screen Shot 2020-12-29 at 8 06 44 PM

The dialogue opens to the folder that stores palette presets:
Screen Shot 2020-12-29 at 8 07 22 PM

The new preset shows in the quick picker:
Screen Shot 2020-12-29 at 8 07 57 PM

I have signed the CLA here: https://github.com/aseprite/sourcecode/pull/58

### DIFF
--- a/src/app/ui/palette_popup.cpp
+++ b/src/app/ui/palette_popup.cpp
@@ -42,6 +42,7 @@ PalettePopup::PalettePopup()
   addChild(m_popup);
 
   m_paletteListBox.DoubleClickItem.connect(base::Bind<void>(&PalettePopup::onLoadPal, this));
+  m_paletteListBox.Change.connect(base::Bind<void>(&PalettePopup::onSearchChange, this));
   m_popup->search()->Change.connect(base::Bind<void>(&PalettePopup::onSearchChange, this));
   m_popup->loadPal()->Click.connect(base::Bind<void>(&PalettePopup::onLoadPal, this));
   m_popup->openFolder()->Click.connect(base::Bind<void>(&PalettePopup::onOpenFolder, this));
--- a/data/gui.xml
+++ b/data/gui.xml
@@ -1078,19 +1078,18 @@
         <param name="type" value="normal-map-wheel" />
       </item>
       <separator />
+      <item command="ColorQuantization" text="@.create_palette_from_current_sprite" group="palette_generation" />
       <item command="LoadPalette" text="@.load_palette" />
+      <item command="LoadPalette" text="@.load_default_palette">
+        <param name="preset" value="default" />
+      </item>
       <item command="SavePalette" text="@.save_palette" />
       <item command="SavePalette" text="@.save_palette_as_preset">
         <param name="save_as_preset" value="true" />
       </item>
-      <item command="LoadPalette" text="@.load_default_palette">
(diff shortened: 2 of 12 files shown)


shixianli158:
reload it

dacap:
Hi @mezz, thanks for this PR, I'll take a look and probably merge it as it is (at least the first commit, I'm not still sure about the 2nd one that modifies the signal from Extensions). Also I thinking about a change in the menu to something like:

Screen Shot 2021-01-04 at 13 45 00

But still I have to think more about it.
Probably this is a fix for #2179

mezz:
Thanks! I'm glad to follow up and fix this up, so just let me know what you think is good and I can try taking it in that direction. 
I'm new to the code base and wasn't sure which approach to take to making the preset palette scroll box reload, so I did what I think is simplest to get it working. I'm interested to know if there's a cleaner way to do it.

dacap:
Probably we could [add a new signal in App class](https://github.com/aseprite/aseprite/blob/a7419c78131a3e6309923a6de6aae0a83c05e4b8/src/app/app.h#L123) like:

    obs::signal<void()> PalettePresetsChange;

And [connect to that signal](https://github.com/aseprite/aseprite/blob/a7419c78131a3e6309923a6de6aae0a83c05e4b8/src/app/ui/palettes_listbox.cpp#L124) too in `PalettesListBox`:

```c++
  m_extPaletteChanges =
    App::instance()->extensions().PalettesChange.connect(
      [this]() { reload(); });
  m_extPresetsChanges =
    App::instance()->PalettePresetsChange.connect(
      [this]() { reload(); });
```

mezz:
Thanks for the review. I have pushed the changes, so it's using the new signal and the menu is better organized:
Screen Shot 2021-01-04 at 9 50 29 AM

src/app/ui/palette_popup.cpp
```diff
@@ -42,6 +42,7 @@ PalettePopup::PalettePopup()
   addChild(m_popup);
 
   m_paletteListBox.DoubleClickItem.connect(base::Bind<void>(&PalettePopup::onLoadPal, this));
+  m_paletteListBox.Change.connect(base::Bind<void>(&PalettePopup::onSearchChange, this));
```
dacap, line 45: Actually reviewing this line I checked that `ui::ListBox::Change` signal is generated when the selection is changed (on `ui::ListBox::onChange`), so this connection is a little confusing. I'll merge this PR as it is but adding a new signal when the resources are completely loaded so we can call `onSearchChange()` indirectly (instead of using the `ListBox::Change` signal).

dacap (approved):

dacap:
Thanks for your PR @mezz, I'll merge it in a couple of minutes.

dacap:
Merged https://github.com/aseprite/aseprite/commit/d4072fac9468cf3b76bd123c012d42ae478c0e1a

mezz:
Thanks!

PR #2614, opened 2021-01-10
Fix bug when after pasting a color to the ColorBar Palette the ColorS…

Zeralith:
…elector is not updated properly

### DIFF
--- a/src/app/ui/color_bar.cpp
+++ b/src/app/ui/color_bar.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2020  Igara Studio S.A.
+// Copyright (C) 2018-2021  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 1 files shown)


dacap:
Hi @Zeralith, I cannot reproduce this bug. Is there a list of steps to reproduce the bug when the patch is not applied?

In case this patch is needed, I think something like this on `ColorBar::onPaletteViewPasteColors` would be preferred:

```diff
diff --git a/src/app/ui/color_bar.cpp b/src/app/ui/color_bar.cpp
index 649d5bf9d..960e98783 100644
--- a/src/app/ui/color_bar.cpp
+++ b/src/app/ui/color_bar.cpp
@@ -702,6 +702,7 @@ void ColorBar::onPaletteViewPasteColors(
   }
 
   setPalette(&newPalette, "Paste Colors");
+  invalidate();
 }
 
 app::Color ColorBar::onPaletteViewGetForegroundIndex()
```

Zeralith:
Hi @dacap, running `invalidate()` is just not enough here. I tried this before. For some odd reason, running `setBounds(bounds())` is the only way I could seem to solve this.

Running `setBounds(bounds())` after the `setPalette(&newPalette, "Paste Colors");` line on `ColorBar::onPaletteViewPasteColors()` works as well, as you described.

To explain the bug, I will make a video and send it to you. It's a very nuanced bug to the untrained eye 😝

Stand-by.

Zeralith:
@dacap, here you go! This video shows the bug & fix in action:

https://youtu.be/2pUV8VJSk-s

Zeralith:
Hey @dacap, did you need me to do anything else with this PR, or were you going to edit it?

Did my video make sense describing the bug and the fix?

Thank you.

dacap:
Hi @Zeralith, sorry I missed your PR for the release. I'll give a try in these days (and maybe find an alternative fix). My main issue with this patch is detecting the "Paste Colors" as:
```c++
  if (actionText == "Paste Colors") {
```

PR #2618, opened 2021-01-15
Fix deleting last nested group causes crash

Gasparoken:


dacap:
Cherry picked 3035a5713d99121a2eb096032341b8177edf6387

PR #2631, opened 2021-01-31
ARM build fixes

Zaryob:
Related with issue [#2630](https://github.com/aseprite/aseprite/issues/2630)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -200,6 +200,19 @@ Run `cmake` with the following parameters and then `ninja`:
 In this case, `$HOME/deps/skia` is the directory where Skia was
 compiled or uncompressed.
 
+---
+**NOTE**
+
+For armv7l and older arm version's building, add this cmake parameters:
+```
+-DCMAKE_CXX_FLAGS:STRING="-march=armv7-a -mfloat-abi=hard -mfpu=neon -mtune=cortex-a15 -Wno-psabi"  \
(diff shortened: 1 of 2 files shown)


fabinshen (approved):
The change seems good and straightforward.

SerenaLexis:
Please unsubscribe me ! I’m getting hundreds of emails that mean nothing to
me

Zaryob (commented):
reviewed

dacap:
Hi @Zaryob, from the patch it looks like we have to patch or update libcurl, but I saw that it doesn't include the curlbuild.h file anymore in its repository:
https://github.com/curl/curl/commit/73a2fcea0b4adea6ba342cd7ed1149782c214ae3

I think that we should convert the curl code into a submodule and use the latest version from it (and then patch libcurl if it's needed).

Zaryob:
> Hi @Zaryob, from the patch it looks like we have to patch or update libcurl, but I saw that it doesn't include the curlbuild.h file anymore in its repository:
> [curl/curl@73a2fce](https://github.com/curl/curl/commit/73a2fcea0b4adea6ba342cd7ed1149782c214ae3)
> 
> I think that we should convert the curl code into a submodule and use the latest version from it (and then patch libcurl if it's needed).

Got it is a good new. This type of arm fix is needed for either raspi or Apple m1 macs
Good works.

PR #2646, opened 2021-02-25
Fix crash using Sprite:newCel() (fix #2635)

### LINKED ISSUE #2635: Crash using Sprite:newCel()
Running this script crashes the program:
```lua
a = Sprite(256, 256)
b = Sprite(256, 256)
-- Uncomment this to avoid the crash
--a:newFrame(1)
a:newCel(a.layers[1], 2, b.cels[1].image)
```

Gasparoken:


dacap:
Hi 👋  This is really strange, the `m_renderEngine` shouldn't be nullptr at that point (the m_renderEngine should be != nullptr if a Editor is alive), do you know when does this happen? at the exit of the program? after calling `Editor::destroyEditorSharedInternals()`?

PR #2669, opened 2021-03-17
Fix undo exception while moving an image

Gasparoken:
Original post: https://community.aseprite.org/t/pretty-nasty-bug-specific/8290

### DIFF
--- a/src/app/ui/editor/moving_cel_state.cpp
+++ b/src/app/ui/editor/moving_cel_state.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2020  Igara Studio S.A.
+// Copyright (C) 2020-2021  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -13,6 +13,7 @@
 
 #include "app/app.h"
 #include "app/cmd/set_cel_bounds.h"
+#include "app/commands/command.h"
 #include "app/context_access.h"
 #include "app/doc_api.h"
 #include "app/doc_range.h"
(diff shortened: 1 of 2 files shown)


dacap:
Hi @Gasparoken, you can intercept commands using `Context::BeforeCommandExecution`. Please check how `MovingPixelsState::onBeforeCommandExecution()` is doing it (you need a signal connection).

That will be better than adding the new cancel() method call and the global call from the CustomizedGui.

dacap:
It looks better now! 😄  The only issue I've found is that while moving the cel, if we press the Escape key an assert fails (in release mode the program crashes with a "segmentation fault").

src/app/ui/editor/moving_cel_state.cpp
```diff
@@ -326,4 +313,54 @@ gfx::Point MovingCelState::intCelOffset() const
...
+        modified = true;
+      }
+    }
+  }
+  return modified;
+}
+
+void MovingCelState::onBeforeCommandExecution(CommandExecutionEvent& ev)
+{
+  if (ev.command()->id() == CommandId::Undo()) {
```
dacap, line 344: What happen if we use Redo? or other commands?

src/app/ui/editor/moving_cel_state.cpp
```diff
@@ -326,4 +313,54 @@ gfx::Point MovingCelState::intCelOffset() const
...
+  }
+  else if (ev.command()->id() == CommandId::Cancel()) {
+    restoreCelStartPosition();
+    gfx::Point point(0, 0);
+    MouseMessage mouseMsg(MessageType::kMouseUpMessage,
+                          PointerType::Mouse,
+                          MouseButton::kButtonLeft,
+                          KeyModifiers::kKeyNoneModifier,
+                          point);
+    m_editor->getState()->onMouseUp(m_editor, &mouseMsg);
```
dacap, line 363: Why do we need to call onMouseUp() of the current state (which is this MovingCelState)? instead of executing the same code that onMouseUp() does? (which looks pretty similar to the undo case)

I'm not sure what was causing the crash with the previous commit.

dacap:
Closing this PR because it was merged in 42201a47832854446f8192a0c4196bf020869922

PR #2700, opened 2021-04-16
Fix empty language update

### LINKED ISSUE #2557: Language option will be reset when other option has changed
Steps to reproduce:
1. Open preferences and change a language option to anything but English then press `OK` button. Now you can see the UI in your specified language.
2. Open preferences again and check `Experimental > Use native file dialog` then press `OK` button.
3. Open preferences again and uncheck `Experimental > Use native file dialog` then press `OK` button.

Now you could see the UI in English after step 3, but sometimes it's happened after step 2.
I didn't try to check on all settings, but I'm sure there are more settings on the list.
Aseprite v1.2.25 on Win x64

Camouflager:
Fix #2557

### DIFF
--- a/src/app/commands/cmd_options.cpp
+++ b/src/app/commands/cmd_options.cpp
@@ -541,9 +541,11 @@ public:
       sendMessage(msg);
     }
 
-    // Update language
-    Strings::instance()->setCurrentLanguage(
-      language()->getItemText(language()->getSelectedItemIndex()));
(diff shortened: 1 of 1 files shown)


Camouflager:
The issue was when preference window is opened it will only initialize the sections that are active. So the `general` section might or might not be initialized at all when user clicks `ok` or `apply`. And the current language setting is always set to respect values in the combobox when the config is saved, which might be null.

PR #2710, opened 2021-04-25
Fix external folder manipulation (fix #232)

### LINKED ISSUE #232: unable to see new folders created while the file browser is open
_From [st...@sprixelsoft.com](https://code.google.com/u/105605591483612568850/) on April 30, 2013 18:34:58_

What steps will reproduce the problem?

1. open an aseprite file
2. select "file->save as"
3. in windows, create a new folder in your directory for a new animation
4. go back to aseprite and attempt to save in this folder

What is the expected output? What do you see instead?
I would expect the file browser to update to recognize any new folders that have been created

What version of the product are you using? On what operating system?
0.9.5 on windows 7 x64 

_Original issue: http://code.google.com/p/aseprite/issues/detail?id=232_

Enfyve:
Updates folder upon navigation (up, back, forward, in).

Also introduces a new button to refresh the current folder view when pressed or when hotkey pressed (F5 or Ctrl+R)

Scenarios tested:
- navigate in
- navigate up
- navigate forward
- navigate backward
- click refresh button
- press F5
- press Ctrl+R
- change file type
- change view


dacap:
Hi @Enfyve, from what I can see from the PR it will invalidate cached items in all actions, does removing [this condition](https://github.com/aseprite/aseprite/blob/141bc434bf0fa55ff031d7edfc4e9ec352a2b13a/src/app/file_system.cpp#L426) achieve the same effect?

Enfyve:
@dacap removing the condition alone does not, however further adjusting the condition does achieve the same effect in most scenarios except when adjusting views

``` diff
+++  if (isFolder() ||
---  if (isFolder() &&
      // if the children list is empty, or the file-system version
      // change (it's like to say: the current m_children list
      // is outdated)...
+++      !m_children.empty()) {
---      (m_children.empty() ||
---       current_file_system_version > m_version)) {
```

dacap:
Finally merged with several changes/fixes:

* https://github.com/aseprite/aseprite/commit/f41c6bc1c51b64f406836834c1ade77fcac82b1f
* https://github.com/aseprite/aseprite/commit/debd22014c44d94e4eab8d8a4ae1a8af090c0121
* https://github.com/aseprite/aseprite/commit/db84c02e942a58e1bd92c131a7d86e7921047356
* https://github.com/aseprite/aseprite/commit/9248f534470044d5a73eb8ffdd966120bbdaf006
* https://github.com/aseprite/aseprite/commit/c7827410bd1b109f12180564a6aa19c270e27b2d

Mainly the list of files is not refreshed each time we change between folders (so the cache is not invalidated). Anyway a pending change to finally fix #232 would be something like #3275

dacap:
And finally I almost forgot, thanks @Enfyve for this patch and the patience.

PR #2730, opened 2021-05-23
Add Layer Name length field to spec.

brickviking:
Adding extra field for length of layer name which is in the file format, but not in the document that describes the format.


dacap:
Hi @brickviking, actually the field is there because the `Layer name` is a `STRING`, and a `STRING` is defined as:

* `STRING`:
  - `WORD`: string length (number of bytes)
  - `BYTE[length]`: characters (in UTF-8)
  The `'\0'` character is not included.

So adding this new `WORD` will create two string length fields (which is not the case). Thanks anyway.

brickviking:
Ah, right. That'll teach me for failing to read the doc correctly. Thanks for looking at it anyway.

PR #2763, opened 2021-06-11
Ellipses for 16x16 were better with the previous algorithm

Gasparoken:
Also, 'ellipse_algo ()' function supports using additional straight pixels to allow the algorithm to be able to draw rounded squares


dacap:
Cherry picked 1eace2489125933f1975def6d5f3eea3344c4dc3 👍

PR #2782, opened 2021-06-21
Fix potential memory leak when exception thrown

endjkv:
Fix potential memory leak when `prefWidget->setPref(pref)` / `Tx tx(writer.context(), operationName, ModifyDocument)` throws an exception. I use `unique_ptr` to avoid memory leak on `tx` when `(*tx)(cmd)` throws an exception.

I'm not sure whether the default constructor of `Tx` will throw exceptions or not. If it will, there are some more code to be modified. (In file src/app/script/sprint_class.cpp)

### DIFF
--- a/src/app/ui/color_bar.cpp
+++ b/src/app/ui/color_bar.cpp
@@ -75,6 +75,7 @@
 
 #include <cstring>
 #include <limits>
+#include <memory>
 
 namespace app {
 
@@ -1182,9 +1183,16 @@ void ColorBar::updateCurrentSpritePalette(const char* operationName)
           cmd->execute(UIContext::instance());
         }
         else {
-          Tx tx(writer.context(), operationName, ModifyDocument);
-          tx(cmd);
-          tx.commit();
+          std::unique_ptr<Tx> tx;
+          try {
+            tx = std::unique_ptr<Tx>(new Tx(writer.context(), operationName, ModifyDocument));
+          }
+          catch (...) {
+            delete cmd;
+            throw;
+          }
+          (*tx)(cmd);
+          tx->commit();
         }
       }
     }
--- a/src/app/widget_loader.cpp
+++ b/src/app/widget_loader.cpp
@@ -203,7 +203,13 @@ Widget* WidgetLoader::convertXmlElementToWidget(const TiXmlElement* elem, Widget
         // Automatic bind <check> widget with bool preference option
         if (pref) {
           auto prefWidget = new BoolPrefWidget<CheckBox>("");
-          prefWidget->setPref(pref);
+          try {
+            prefWidget->setPref(pref);
+          }
+          catch (...) {
+            delete prefWidget;
+            throw;
+          }
           widget = prefWidget;
         }
         else {


src/app/ui/color_bar.cpp
```diff
@@ -1182,9 +1183,16 @@ void ColorBar::updateCurrentSpritePalette(const char* operationName)
           cmd->execute(UIContext::instance());
         }
         else {
-          Tx tx(writer.context(), operationName, ModifyDocument);
-          tx(cmd);
-          tx.commit();
+          std::unique_ptr<Tx> tx;
+          try {
+            tx = std::unique_ptr<Tx>(new Tx(writer.context(), operationName, ModifyDocument));
```
dacap, line 1188: There is no need to create a pointer for Tx (it can be just a `Tx` as it was, we prefer value types instead of pointers).

src/app/ui/color_bar.cpp
```diff
@@ -1182,9 +1183,16 @@ void ColorBar::updateCurrentSpritePalette(const char* operationName)
...
+          }
+          catch (...) {
+            delete cmd;
```
dacap, line 1191: It's a known issue that in some places there are some leaks on some `Cmd`s before adding then to the `Tx`. We think that we need a more general solution to this problem (e.g. creating a custom `CmdPtr` which could be an alias/similar to `std::unique_ptr`).

src/app/widget_loader.cpp
```diff
@@ -203,7 +203,13 @@ Widget* WidgetLoader::convertXmlElementToWidget(const TiXmlElement* elem, Widget
         // Automatic bind <check> widget with bool preference option
         if (pref) {
           auto prefWidget = new BoolPrefWidget<CheckBox>("");
-          prefWidget->setPref(pref);
+          try {
+            prefWidget->setPref(pref);
```
dacap, line 207: In this case the `setPref` can throw an exception but we would prefer to handle these cases with smart pointers/unique_ptr instead of using throw/catch.

dacap (changes requested):
Hi @endjkv, thanks for these patches. I've made some comments. Remember to take a look to the [CLA sign process too](https://github.com/aseprite/sourcecode/blob/main/sign-cla.md#sign-the-cla).

src/app/ui/color_bar.cpp
endjkv: Thanks for the reply!

But if we just use a `Tx`, as the below code:
```c++
try {
  Tx tx(writer.context(), operationName, ModifyDocument);
  tx(cmd);
  tx.commit();
} catch (...) {
  delete cmd;
  throw;
}
```
There might be a double free on `cmd` if `tx(cmd)` throws an exception. (`Transaction::execute` will free `cmd`)

Using a pointer is a bit weird but I'm not sure if there are any other ways to do this.

src/app/widget_loader.cpp
endjkv: I agree it's better to use smart pointers to avoid these kinds of memory leaks. This PR tries to fix the memory leak in a relatively simple way (remain the same semantics). It would be better If those raw pointers are to be replaced with smart pointers in future.

src/app/ui/color_bar.cpp
endjkv: Oh, I'm sorry. I'll have a look at the CLA.

dacap: The best option is to avoid try/catch at all using a std::unique_ptr. I've pushed a new commit with a fix for this situation: https://github.com/aseprite/aseprite/commit/89f8e6f275869d5c5f7a27b1f5d8676526bd2a3e

src/app/widget_loader.cpp
dacap: I've pushed a fix with a smart pointer for this case: https://github.com/aseprite/aseprite/commit/6157b004e40aa0fbe954b5b6c1539c3b0be37435

endjkv:
Good.

Besides, please have a look at `Sprite_newGroup`, `Sprite_newLayer` and `Sprite_newTag`. (in src/app/script/sprint_class.cpp)

I'm not sure if the default constructor would throw an exception, just to let you know.

dacap:
Each `new` that is not contained in in smart pointer is a potential memory leak (because it can throw std::bad_alloc). Aseprite requires a general refactor to fix all these leaks, but the possibility to occur of these issues vs the amount of work for the refactor (+possible new bugs because the refactor) vs the benefit of these fixes is something we always take care. So fixing this kind of memory leaks is a really low priority. Anyway thanks for your comments.

PR #2816, opened 2021-07-06
Fix eyedropper tool picks a pixel of mask color, the FG color changes to the entry 0 (fix #2813)

### LINKED ISSUE #2813: When eyedropper tool picks a pixel of mask color, the FG color changes to the entry 0
Even if the mask color is present on the palette, when the eyedropper picks a pixel of mask color it results in the selection of palette color entry 0.

* Aseprite version: 1.3beta4
* System: macOS Catalina 10.15.3
maskColorAndEyedropper

Gasparoken:
Even if the mask color is present on the palette.

### DIFF
--- a/src/app/ui/palette_view.cpp
+++ b/src/app/ui/palette_view.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2020  Igara Studio S.A.
+// Copyright (C) 2018-2021  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -972,7 +972,8 @@ int PaletteView::findExactIndex(const app::Color& color) const
   switch (color.getType()) {
 
     case Color::MaskType:
-      return (current_editor ? current_editor->sprite()->transparentColor(): -1);
+      return (currentPalette()->findMaskColor() >= 0) ?
+        currentPalette()->findMaskColor() : -1;
 
     case Color::RgbType:
     case Color::HsvType:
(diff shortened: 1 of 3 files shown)


dacap:
It works great in RGB, but I'm still not sure what it should do in Indexed. This is the regression I was talking the other day in Indexed mode:

Screen Shot 2021-07-12 at 10 41 25

Selecting "Mask" color will not select the transparent color in the color bar (but a alpha=0 color in the palette). This can be seen as a correct result, but before this the idea is that the "Mask" color = transparent sprite color (in indexed mode).

src/app/ui/palette_view.cpp
```diff
@@ -972,7 +972,8 @@ int PaletteView::findExactIndex(const app::Color& color) const
   switch (color.getType()) {
 
     case Color::MaskType:
-      return (current_editor ? current_editor->sprite()->transparentColor(): -1);
+      return (currentPalette()->findMaskColor() >= 0) ?
+        currentPalette()->findMaskColor() : -1;
```
dacap, line 976: Minor change to avoid double findMaskColor (anyway given the regression I'm not 100% sure if this should be the final solution):
```suggestion
    case Color::MaskType: {
      int i = currentPalette()->findMaskColor();
      return (i >= 0 ? i: -1);
    }
```

Gasparoken: Yeah, I. think RGB and Grayscale should apply this. I'll fix soon.

Gasparoken:
I'll close this PR, and create a new one (updated)

PR #2840, opened 2021-07-16
Fix crash when pressing ESC while moving a selection (fix #2829) 

### LINKED ISSUE #2829: Crash when pressing ESC while moving a selection
Reported:
>I was using the marquee tool, with an existing selection in the rectangular mode. I was attempting to click the top right corner of the selection while holding shift to resize the selection. However I miss-clicked and moved the whole selection instead. I then pressed ESC to cancel this movement (still with LMB and shift held) at which point the crash occurred.

It seems that just pressing the ESC key while moving a selection reproduces this crash. I reproduce it on v1.3-beta5.

Crash report:
* https://igarastudio.zendesk.com/agent/tickets/2476

martincapello:


dacap:
@martincapello could you please make this PR from the `main` branch, not from the `master` branch?

martincapello:
I thought I did it...will double check and fix it. Thanks.

martincapello:
You are right, I used the master branch. Is that branch still relevant? Should we delete it? It is a bit misleading to have master and main I think...I wil fix this PR anyways. Thank you!

dacap:
> Is that branch still relevant? Should we delete it?

The best case would be to finally delete it, but that would break some links that were made in forums/blog posts/etc. that are pointing to the master branch files. From time to time I try to keep it updated but I think the best would be to keep it outdated in some previous point of history.

martincapello:
Found this https://github.blog/changelog/2020-07-17-links-to-deleted-branches-now-redirect-to-the-default-branch/
It seems we could delete the master branch...not sure if we really want to try it out 😬

dacap:
👍 thanks @martincapello, I've just merged it.

About deleting the `master` branch, I'll delete it now and then see if something goes wrong, but if they auto-redirect the links that might work.

PR #2847, opened 2021-07-20
Add the ability to set the continuity when creating a new layer

sumibi-yakitori:
With this change, if all existing layers are in continuous mode, when you create a new layer, it will set that layer to continuous mode.

I use a lot of layers and frames, so this feature is necessary for me.

### DIFF
--- a/src/app/commands/cmd_new_layer.cpp
+++ b/src/app/commands/cmd_new_layer.cpp
@@ -221,11 +221,11 @@ void NewLayerCommand::onExecute(Context* context)
       fmt::format(Strings::commands_NewLayer(), layerPrefix()));
     DocApi api = document->getApi(tx);
     bool afterBackground = false;
-
 #ifdef ENABLE_UI
     // Since the timeline status changes when a layer is added, get the necessary value here.
     bool allLayersContinuous = App::instance()->timeline()->allLayersContinuous();
 #endif
+
     switch (m_type) {
       case Type::Layer:
         layer = api.newLayer(parent, name);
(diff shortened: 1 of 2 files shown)


dacap:
Hi @sumibi-yakitori, sorry for the late response but I don't know if this would be a proper solution. My fear is that people will start experiencing a different behavior for New Layer (e.g. with one layer, e.g. the Background, that layer already dictates how the new layers will be created). Not sure if there might be a better solution (e.g. a script to create a "New Continuous Layer")

sumibi-yakitori:
@dacap 
Okay, I understand that you are busy.
I will try to create such a script for now.
(Also, even if this PR has a chance of being adopted, it may have needed to have the ability to switch from the Aseprite settings to do so.)

> (e.g. with one layer, e.g. the Background, that layer already dictates how the new layers will be created)

But what does this mean?
I would think that under Aseprite's current behavior, no matter which continuity mode the Background layer is in, the layer created afterwards will always be a non-continuous layer.

dacap:
> But what does this mean?

With this PR, if I create a new sprite, then set the Background as continuous (because I want a continuous background), then creating new layers will create them as continuous too (but probably the user doesn't expect this, as foreground layers generally are not be continuous, so the user should start investigating about how to unlink cels).

I think about this because probably a lot of users that click the "continuous layer" icon may not know what it does. Sometimes in the past I took decisions just to avoid confusing/unexpected behavior for regular/new users, e.g.:
* Use RGBA by default in File > New (instead of Indexed, as Indexed mode is confusing for most new users)
* Snap to Grid is activated by accident and a lot of users don't know how to disable, we've included a visible popup to explicitly disable it

sumibi-yakitori:
Ahhhh, I see. I understand very well. Thank you for the explanation 😄 

It certainly seems to me that defaulting to features for power users, which is what I would prefer, would be confusing to beginners.

Issue #1093, opened 2016-04-26
Saving animation takes too long time


dacap:
> Way to reproduce:
1) Open Aseprite
2) Create new animation 2048x1536 Indexed with Transparent color
3) Add 200 frames using Alt+N keyboard shortcut
4) Save animation to file
5) It takes too long for Aseprite to save 200 identical empty frames, showing dialog "Saving to file" with "Working" and progress bar.

> Expected: empty frames should take no time while saving (or loading), any amount of identical frames with same CEL should take nearly the same small amount of time while saving (or loading).

Reported here: http://steamcommunity.com/app/431730/discussions/2/357284131799724546/

dacap:
The performance to save files on Windows was improved in #2863 just avoiding `fseek()` on ase_file_write_start_chunk(). If we completely remove the `fseek()` usage in the future it might improve even more (e.g. for small chunks we could prepare the chunk on memory and write the whole chunk with its header in one step).

PR #2863, opened 2021-08-02
Replace fseek with putl/w to reduce io flushes

Amarcolina:
ase_file_write_start_chunk needs to skip forward the size of the chunk header, as these values will be written in later.  Using fseek was causing performance issues on my Windows machine due to causing an io flush on every chunk, for projects with many (thousands) of chunks.  Replacing with the equivalent put commands in ase_file_write_close_chunk results in significant speedup for certain test projects.

I have uploaded my test project below, which emphasizes the problem.  It is a project with many cells, as the issue seems directly corelated with chunk count, as each cell is written as a single chunk no matter how small they may be.  The project has 100 layers, and 100 frames.  The file is ~350 Kb on my machine due to the small canvas size (16x16).  But despite this it takes ~20 seconds to save!  With the fix here the file saves in just moments (less than half a second).

possibly related to #1093 

[SavingStressTest.zip](https://github.com/aseprite/aseprite/files/6914332/SavingStressTest.zip)

### DIFF
--- a/src/app/file/ase_format.cpp
+++ b/src/app/file/ase_format.cpp
@@ -526,7 +526,8 @@ static void ase_file_write_start_chunk(FILE* f, dio::AsepriteFrameHeader* frame_
   chunk->type = type;
   chunk->start = ftell(f);
 
-  fseek(f, chunk->start+6, SEEK_SET);
+  fputl(0, f);
+  fputw(0, f);
 }
 
 static void ase_file_write_close_chunk(FILE* f, dio::AsepriteChunk* chunk)


dacap:
Thanks @Amarcolina, I'll give a try to this today on Windows and macOS and merge your PR if everything goes fine. But it LGTM so far.

dacap:
Thanks for this patch @Amarcolina, I've tested it and the performance improves a lot on Windows. I'll leave #1093 open so we might replace the `fseek()` usage completely in a future (e.g. to write chunks in memory and write the file chunk by chunk).

PR #2874, opened 2021-08-07
Fix translucent backgrounds in color selection

ruerob:
This fix adds the background settings of the document to the color palette, the color selection tools, and the color buttons. 
I've found a to-do comment in the source code that suggested this change.

I hope this is helpful and would be happy to apply more changes in the future
ruerob

### DIFF
--- a/src/app/modules/gfx.cpp
+++ b/src/app/modules/gfx.cpp
@@ -39,10 +39,6 @@ using namespace gfx;
 
 namespace {
 
-// TODO hard-coded values, use pref.xml values
-gfx::Color gridColor1 = gfx::rgba(128, 128, 128);
-gfx::Color gridColor2 = gfx::rgba(192, 192, 192);
-
 void draw_checked_grid(ui::Graphics* g,
                        const gfx::Rect& rc,
                        const gfx::Size& tile,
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @ruerob I'll start reviewing this change, I'm not sure if `current_editor != nullptr` in all cases, so I'll check that and try to merge this commit (probably with a little extra patch to take care of that situation).

dacap:
I've checked and there are some issues with the ColorSelector which uses [the draw_alpha_slider() from a background thread](https://github.com/aseprite/aseprite/blob/4c6835c4392be0e898ae5d7041964e628564074d/src/app/ui/color_selector.cpp#L465) and that could generate some crashes (e.g. accessing preferences from a non-UI thread).

I've fixed that temporarily with this patch:  https://github.com/aseprite/aseprite/commit/013792a215c1d0740f57af3248fc831b4caf8f22

But this is the result:

Screen Shot 2021-09-01 at 12 08 01

We can fix this later. I've cherry-picked this PR here: https://github.com/aseprite/aseprite/commit/ac31ee35c0bdb5a71dc50d82605400adb1da649f
Thanks @ruerob!

PR #2876, opened 2021-08-08
Fix color switching on empty palette space click (fix #2776)

### LINKED ISSUE #2776: Clicks on empty space of palette colors/tiles scope changes the foreground color to the most bottom-right color/tile.
When we change the color palette scope to the tiles palette scope (and vice versa) by clicking on the empty space of the palette scope, the selected color/tile is changed to the most bottom-right color/tile, instead of holding the last foreground selected color / tile chosen before the scope change.
Same behavior in the same palette scope: the moost bottom-right color is selected when clicking in the empty space of the color palette scope.
Suggestion: clicks on empty space of palette colors/tiles do not change the foreground color.

* Aseprite version: 1.3 beta4
* System: MacOS Catalina 10.15.3

ruerob:
Fix for issue https://github.com/aseprite/aseprite/issues/2776.

The old behavior was, that if a user clicked on empty space in the palette view, the last palette entry was selected. And used as foreground or background color, depending on the clicked mouse button.

Now the click on empty space gets the currently selected color deselected and the foreground and background color stay the same.

### DIFF
--- a/src/app/ui/palette_view.cpp
+++ b/src/app/ui/palette_view.cpp
@@ -333,6 +333,11 @@ bool PaletteView::onProcessMessage(Message* msg)
           break;
       }
 
+      if(m_hot.color >= currentPalette()->size()){
+        deselect();
+        break;
+      }
+
       captureMouse();
 
       // Continue...


dacap:
I've just merged this change but [moving it inside the specific case when we click a color](https://github.com/aseprite/aseprite/commit/b5414c5d952f152cdf336f048d88469a91b75d57), because in other case resizing the palette (`Hit::RESIZE_HANDLE`) was broken.

dacap:
And thanks for the patch and patience @ruerob!

ruerob:
I'm happy I could help.

Issue #2869, opened 2021-08-05
Make thicker the cursor and selection marching ants when [screen scale = 100%] and  [UI scale >= 200%]


Gasparoken:
When the editor has a 100% screen scale and 200% UI scale:
The cursor is too thin on the screen (mine is 13") and hard to find.
The selection marching ants, could perhaps be thicker to improve visibility.

Possible solution: double size cursor and selection border.

Aseprite version: 1.3-beta5
This issue started with #2862

PR #2879, opened 2021-08-09
Fix cursor and selection size on 100% screen scale (fix 2869)

ruerob:
Fix suggestion for issue #2869.

I don't know if it is okay if everything is just scaled on the smallest screen scale, but I'm really satisfied with the sizes at 200%.

I changed:

- The size of the pen cursor by setting the scale to 2.
- The size of the selection cursor by adding scaling to the pixel putting. 
- The size of the selection by drawing the same selection one pixel right and down again.

I hope this is helpful and wish you a nice day.
ruerob

### DIFF
--- a/src/app/ui/editor/brush_preview.cpp
+++ b/src/app/ui/editor/brush_preview.cpp
@@ -519,7 +519,7 @@ void BrushPreview::createCrosshairCursor(ui::Graphics* g,
   }
 
   os::Window* window = m_editor->manager()->display();
-  const int scale = window->scale();
+  const int scale = (window->scale()<2)?2:window->scale();
   os::CursorRef cursor = nullptr;
 
   // Invalidate the entire cache if the scale has changed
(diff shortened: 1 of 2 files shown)


dacap:
Hi @ruerob, thanks for your contribution, I think you could use the `guiscale()` to duplicate (or triplicate, etc.) the thick of each edge.

For `Editor::drawMask()` I guess `os::Paint::strokeWidth()` could be set.

ruerob:
Hi @dacap, thanks for the reply!

I tried `os::Paint::strokeWidth()` but that messed up the marching ants.

I'll try `guiscale()` and see how it looks.

ruerob:
With strokeWidth(2) the lines are two pixels thick but look like this:

image

On the upper edge, the upper pixels won't march at all, and the lower pixels are moving like expected. That doesn't look pretty in my eyes.

ruerob:
Thank you @dacap!

Now I've added `ui::guiscale()` for scaling the select cursor.

dacap:
Thanks @ruerob to dedicate some time to this issue, I'll review this today and see if merge part of it. The issue with the marching ants pattern looks that could be solved making the pattern bigger (by guiscale()), but that requires modifying the laf library too. We can leave that part for the future.

dacap:
After reviewing this I think the resulting cursor (mainly for the selection tool) is not a desirable result:

Screenshot from 2022-09-28 17-45-22

Also I'm not sure about changing the marching ant effect/path stroke width.

Screenshot from 2022-09-28 17-48-16

I'll left this PR to be reviewed in the future but we don't have a resolution for this yet.

dacap:
I have rebased the changes and an extra #include to make it work https://github.com/dacap/aseprite/tree/ruerob-fix-2869

dacap:
Thinking about this option, as cursors are created [using "get pixel" from the screen](https://github.com/aseprite/aseprite/blob/4981f05f0badff06d43cc47d6cd63861225f77db/src/app/ui/editor/brush_preview.cpp#L532), and then doing a negative using black or white, we cannot make the current cursor bigger (as this would require [a bigger cache of cursors](https://github.com/aseprite/aseprite/blob/4981f05f0badff06d43cc47d6cd63861225f77db/src/app/ui/editor/brush_preview.cpp#L61)), we should use another kind of cursor in these cases.

dacap:
Closing this as we need a different approach.

PR #2902, opened 2021-08-19
Added Duration to filename formatter

jestarray:
https://www.aseprite.org/docs/cli/#filename-format

Added support for templating the `{duration}` of a frame in the filename formatter


dacap:
Hi @jestarray, looking at the diffs I saw that something seems wrong [in the filename_formatter.h](https://github.com/aseprite/aseprite/pull/2902/files#diff-efbf71a9c61d3fdd3799ec8d89fabb9bbcb4f0ca13502b06ccfcdb4f47c2c260): it looks like the file content was duplicated (instead of just adding the new duration() field.

jestarray:
Oh god my sleep.... I think I did  this to 2 of the files and thought I fixed both of them when I only fixed 1. Okay! it should be good now, sorry.

dacap:
Thanks @jestarray!, squash merged here: https://github.com/aseprite/aseprite/commit/59e01493cb6118852b0684fd532f58fa79dd6987

PR #2909, opened 2021-08-24
Fix crash after saving a GIF file from an indexed image which transparent index is gone

Gasparoken:
Before this fix, it was possible to lose the transparent index at Indexed Color Mode.
It was reproducible doing:
- New file, Indexed mode, Black background.
- Erase all the palette, add 40 colors slots (these all will be black)
- Press Remap button
- Go to Sprite > Properties...  set transparent color to 11.
- Erase color entries 2 to 20
- Press Remap button

You'll see the transparent index is gone (transparent index = -1)
To reproduce the crash, additional steps are needed:
- Right click to layer and select 'Layer from Background'
- Make a color RGBA(0, 0, 0, 0) and add to the palette
- Paint on canvas with this color (it'll act like a eraser)
- File > Save As... save in GIF format.
- Press OK

Crash!

This crash was reported on ticket 2620 and 2621.
Aseprite version 1.2.28

### DIFF
--- a/src/app/cmd/set_palette.cpp
+++ b/src/app/cmd/set_palette.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2020  Igara Studio S.A.
+// Copyright (C) 2019-2021  Igara Studio S.A.
 // Copyright (C) 2001-2015  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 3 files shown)


dacap:
Hi @Gasparoken! I'll merge this PR as it is (I wasn't able to fully test it), anyway there are a couple of things to refactor for the future:

1. Now `cmd::SetPalette` contains a similar logic than `cmd::SetTransparentColor` embedded
2. The m_oldTransparentIndex and m_newTransparentIndex are not used if the color mode isn't indexed (or if the transparent color is not changed)

It looks like it would be better to add an extra `cmd::SetTransparentColor` when needed but maybe there are too many places to check.

dacap:
Cherry-picked https://github.com/aseprite/aseprite/commit/dd2d226264b0960145a7e04a97a4aa3669330ad3

PR #2921, opened 2021-08-28
fixed restoring grid preference on undo/redo (fix #2872)

### LINKED ISSUE #2872: Undoing grid preferences confuses the 'Preferences' window
Undoing (Ctrl+Z) Grid's width / height / offset restores the *displayed* grid, but not the values in Preferences.

Steps to reproduce:
1. Set Grid to 2×2
2. Set Grid to 4×3
3. Press Ctrl+Z
4. Open Preferences
5. See that it still says “4×3”

iamOgunyinka:
Problem: the grid settings always save the newly set bounds of a sprite/document in the document's preference but does not do restore the old/new bounds back to the document's preference when we undo or redo changes.

This solution: In the Undo and Redo methods of the SetGridBounds, we can save the old/new bounds into the document's preference.

### DIFF
--- a/src/app/cmd/set_grid_bounds.cpp
+++ b/src/app/cmd/set_grid_bounds.cpp
@@ -13,6 +13,7 @@
 #include "app/doc.h"
 #include "app/doc_event.h"
 #include "app/doc_observer.h"
+#include "app/pref/preferences.h"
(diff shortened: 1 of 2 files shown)


iamOgunyinka:
@dacap got it.

Issue #2901, opened 2021-08-19
Grays are Treated as Reds in Hue Gradient and Sort By Hue


behreajj:
Hello,

The version and platform I'm using doesn't pertain here, as this is a problem general to a lot of software.

Probably as an artifact of treating zero as a substitute for invalid, not a number, infinity or whatever result you want to assign to n / 0, gray colors are treated as reds in hue operations. There are certain cases where this is probably intended, such as HSV+ and HSL+ hue adjustments. There are other cases, as in the color picker, where this is avoided - when a color's saturation is dragged to zero, its hue is preserved for a time.

In hue gradients it leads to results like this
![redGrayCyanGradient](https://user-images.githubusercontent.com/12038348/130108250-4a299abc-6f1c-4cfe-9059-702ac7de3974.png)

Green and cyan are most adversely impacted, as they are furthest away from red on the color wheel (depending on which color wheel you use, the complement of red varies).

You could argue that this is conventional: it is how many other software packages handle hue gradients. I tested GIMP, Krita, Blender. I think it would make more sense to default to an sRGB gradient in cases where either the origin or the destination is desaturated. Think for example of a case where a gradient for a metallic surface is being created.

Sorting a gradient like this by hue

![beforeSort](https://user-images.githubusercontent.com/12038348/130108826-c4780d6e-a871-4eb0-be49-c2e25c7376b7.png)

gives this result

![afterSort](https://user-images.githubusercontent.com/12038348/130108850-73a81a43-8b8b-4f86-8e49-ee17ce69cd62.png)

Ideally, the grays would be clumped together before any of the other hues.

PR #2923, opened 2021-08-28
Color Sort For Hue, Alpha

behreajj:
Removes switch case with fall through. Removes nested switch cases. Color sorting method uses fallbacks for HSV channels when primaries are invalid or equivalent.

Partial fix for #2901.

### DIFF
--- a/src/doc/sort_palette.cpp
+++ b/src/doc/sort_palette.cpp
@@ -21,123 +21,176 @@
 
 namespace doc {
 
-using namespace gfx;
+  using namespace gfx;
 
-namespace {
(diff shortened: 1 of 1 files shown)


dacap:
Hi @behreajj, I've just cherry-picked your changes in d14cf4038da99ea6e12a953b3d82ea3fea986911 with some minor changes:

* Use `== 0` instead of `< 1`
* And removed the formatting changes (I'm not sure if some changes were made with clang-format, but the clang-format is not yet ready to use and we prefer to create specific patches to migrate everything to clang-format in a near future)

Issue #2924, opened 2021-08-30
Tilemap layers don't save blend mode and opacity in .aseprite files


dacap:
This bug is on the `beta` branch. If we change the blend mode or opacity of a tilemap layer, it's not saved correctly in the `.aseprite` file.

Reported here: https://steamcommunity.com/app/431730/discussions/2/3040480634941318651/

dacap:
This is already fixed with https://github.com/aseprite/aseprite/commit/d502876cd3cb6e1e580e0eb54115294161b8c0d8

PR #2927, opened 2021-08-31
Fixed saving of blend mode and opacity in aseprite files

iamOgunyinka:
Fixed bug referenced in #2924 when blend mode and opacity are incorrectly saved because their default values are changed.

### DIFF
--- a/src/app/file/ase_format.cpp
+++ b/src/app/file/ase_format.cpp
@@ -691,8 +691,8 @@ static void ase_file_write_layer_chunk(FILE* f, dio::AsepriteFrameHeader* frame_
   // Default width & height, and blend mode
   fputw(0, f);
   fputw(0, f);
-  fputw(layer->isImage() ? (int)static_cast<const LayerImage*>(layer)->blendMode(): 0, f);
-  fputc(layer->isImage() ? (int)static_cast<const LayerImage*>(layer)->opacity(): 0, f);
+  fputw(layer->isImage() ? (int)dynamic_cast<const LayerImage*>(layer)->blendMode(): 0, f);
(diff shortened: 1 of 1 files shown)


dacap:
Hi @iamOgunyinka! I've just tested it and it looks like the problem is still present. E.g. I tried to log this:

```c++

  TRACEARGS("--DEBUG--",
            "\n  const_cast static_cast:",
            (int)static_cast<LayerImage*>(const_cast<Layer*>(layer))->blendMode(),
            (int)static_cast<LayerImage*>(const_cast<Layer*>(layer))->opacity(),
            "\n  const_cast dynamic_cast:",
            (int)dynamic_cast<LayerImage*>(const_cast<Layer*>(layer))->blendMode(),
            (int)dynamic_cast<LayerImage*>(const_cast<Layer*>(layer))->opacity(),
            "\n  static_cast:",
            (int)static_cast<const LayerImage*>(layer)->blendMode(),
            (int)static_cast<const LayerImage*>(layer)->opacity(),
            "\n  dynamic_cast:",
            (int)dynamic_cast<const LayerImage*>(layer)->blendMode(),
            (int)dynamic_cast<const LayerImage*>(layer)->opacity());
```

and the output for a layer with opacity=85 and blendMode=2 was:

```c++
--DEBUG-- 
  const_cast static_cast: 2 85 
  const_cast dynamic_cast: 2 85 
  static_cast: 2 85 
  dynamic_cast: 2 85
```

So it looks like this doesn't make a difference, and the problem might be in the loading/decoder process (`src/dio/aseprite_decoder.cpp`).

iamOgunyinka:
@dacap that's strange. The result of

iamOgunyinka:
@dacap you're right. I'm looking at the decoder right now.

PR #2929, opened 2021-08-31
Fixed saving of blend mode and opacity in aseprite files

iamOgunyinka:
Fixed bug referenced in #2924 when blend mode and opacity are incorrectly saved because their default values are changed.


dacap:
Thanks @iamOgunyinka! I've just cherry-picked your patch here: d502876cd3cb6e1e580e0eb54115294161b8c0d8

Issue #2928, opened 2021-08-31
Color index of individual color bigger than palette size


Gasparoken:
How to reproduce this bug:
- New file (color mode: Indexed, transparent background)
- Modify the color which index is 10, turn to RGBA(0, 0, 0, 0)
- Paint all the canvas with Paint Bucket Tool with color index 10 (you will see nothing, because index 10 is transparent)
- Resize the palette to 4 colors (dragging the end of palette in the color bar)
- Press the Remap button.
Finally, all the canvas is painted with color index 10 (transparent), but the maximum palette index is 3 (0, 1, 2, 3, four colors)
Then if you try to export this sprite to PNG, libpng error appears  and the png isn't saved.
And if you try to export this sprite to GIF, Aseprite will crash.

Issue originally reported on ticket 2650

Aseprite version: 1.2.29
OS: macOS 10.15.3 Catalina

iamOgunyinka:
Sounds interesting, want me to take it on, @Gasparoken? it should be fun

dacap:
Related to #2842 and #2912

dacap:
I'm not 100% but probably part of this issue was fixed with https://github.com/aseprite/aseprite/commit/dd2d226264b0960145a7e04a97a4aa3669330ad3 too.

Issue #2912, opened 2021-08-24
Flatten Visible layers uses indices outside the palette range as opaque colors instead of transparent color


dacap:
If we have a layer with an index outside the palette range, the editor shows that pixels as a transparent entry, but after Flatten Visible, the pixels will overwrite the pixels below it (just like any other opaque pixel, but the result will be like a transparent pixel). Example file: 
[index0123.aseprite.zip](https://github.com/aseprite/aseprite/files/7042379/index0123.aseprite.zip)

Before flatten:

<img width="534" alt="Screen Shot 2021-08-24 at 19 12 27" src="https://user-images.githubusercontent.com/39654/130697018-e593d5c5-6eac-4820-aa6b-8ee4793f51b9.png">

After flatten:

<img width="529" alt="Screen Shot 2021-08-24 at 19 12 30" src="https://user-images.githubusercontent.com/39654/130697024-5aeb940a-52f2-4b3c-983f-e466b4a2b56c.png">

dacap:
Related to #2842

iamOgunyinka:
@dacap with the attached file, I can see what you're talking about but w/o the attached file, how do I reproduce this?

dacap:
@iamOgunyinka a similar file can be reproduced with these steps (it's a little complex to replicate but can be done):

1. Create an indexed image
2. draw something
3. create a new layer
4. draw something with a color with high index (e.g. if the palette contains 32 colors, you can draw something with index 31)
5. then resize the palette to something smaller (e.g. 16 colors):

    <img width="388" alt="Screen Shot 2021-08-30 at 09 41 58" src="https://user-images.githubusercontent.com/39654/131340640-1f3c0419-39f7-4041-958f-c4df86f8dd24.png">

Issue #2842, opened 2021-07-17
libpng: Wrong palette index exceeding num_palette


metnias:
I was getting this same issue in v1.3 beta so I reverted back to v1.2
but as some of the features are backported to v1.2 now, this critical bug followed
and now I cannot export my work anymore.

As I draw, I change the palette with a multiple layered project,
and use the transparent colour of the current palette to erase which reliably causes the issue.
Exporting to PNG does not work at all,
and exporting to other formats also breaks the image cause some of the layers now are no longer transparent.

Now I have to screenshot the program and trim the image :(

### Aseprite and System version

* Aseprite version: 1.2.28-x64, Steam
* System: Windows 10

dacap:
Hi @metnias, could you please share with us privately a .aseprite file that you use to export to .png to support@aseprite.org? So we can check this issue you are talking about (we didn't receive other bugs report about this).

metnias:
Just sent the aseprite files that are causing issues to support@aseprite.org.
Thanks for the quick response!

dacap:
@metnias I'm seeing your files and this was not caused because we back-ported some features from v1.3 to v1.2. Aseprite v1.2.27 already showed this message so I'm a little confused. What is the Aseprite version that you was previously using? Maybe you have been using an even older Aseprite version? (v1.2.9 or v1.1.13? which are still available on Steam)

We'll check if we can fix some of these use cases for the following versions. It looks like the Flatten command and Export should replace any index outside the palette range with the transparent index (or interpret it as transparent). The only problem could appear if the transparent index is really visible in the final render and the index is outside the palette range (probably in that only case we should expand the palette with an extra index to represent the transparent color).

> For developer reference, the ticket is: https://igarastudio.zendesk.com/agent/tickets/2497

metnias:
For `029 reachout.aseprite`, that's an older project but also having a presumably similar issue so I wanted to include it in this report. For reference, it's made on June 12th, but I was testing v1.3 beta from time to time before so I don't remember which version I used for this project. I'm sorry for the confusion.

But `serverlogo 2107.aseprite` is created earlier today with the latest stable version on Steam (1.2.28-x64) and broken even more, which is why I've decided to report this.

I'd just assumed this is because of backporting because I hadn't seen `libpng: Wrong palette index exceeding num_palette` in earlier version v1.2 but only with v1.3beta, in my personal experiences.

dacap:
[...]

PR #2934, opened 2021-09-02
Fixed issue regarding a pixel having bigger value than the available palette

iamOgunyinka:
This fixes the bug reported in #2928, #2912 and #2842.

The bug occurs when a pixel in an index image is bigger than the index of the available palettes. It has caused crashes and unsuccessful export in some file formats.


dacap:
Thanks @iamOgunyinka! Just merged

dacap:
Forgot to mention, I've added a couple of tests for this here: https://github.com/aseprite/tests/commit/0ac10ee18d295d144480286426140339604519d2

PR #2941, opened 2021-09-06
XML parser and CloseEvent additions 

martincapello:
Add support to the <image> tag and the "visible" attribute.
Add support to cancelable CloseEvent.


dacap:
Merged with minor fix 7d248c33d7374d679f87be36cb69ae2b182303b9

The base class is always initialized first, i.e. the code of the base class ctor will be called before the initialization of member variables.

Issue #2877, opened 2021-08-09
Performance issue/lag drawing in "Auto" mode on a tilemap layer


dacap:
We don't know the exact problem but we have a file to reproduce this issue in the ticket (the file was sent privately via support email address, please keep it private).

Ticket: https://igarastudio.zendesk.com/agent/tickets/2539

dacap:
Fixed with ed8e2c8392b01fbe844bdab11a912a6f298c2d1d

PR #2945, opened 2021-09-08
Fixed performance issue/lag drawing in 'Auto' mode on a tilemap layer

iamOgunyinka:
The problem occurs when one or more tile has an index outside of the available tile set. This PR fixes this problem when drawing in auto mode and the crash that happens when a tile is moved and a `remap` command is issued.


dacap:
Thanks @iamOgunyinka! I've just merged the commit in https://github.com/aseprite/aseprite/commit/ed8e2c8392b01fbe844bdab11a912a6f298c2d1d with some minor changes (mainly adding a reference to #2877 in the commit). P.S: There are some typos in the commit description, that's my fault 🖐️ 😅

PR #2947, opened 2021-09-10
Fix disappearing tilemap layers on resizing canvas (fix #2942)

### LINKED ISSUE #2942: [bug] Canvas Size change makes tilemap layers disappear
I just wanted to crop my image created in Aseprite 1.3 but when I did it, I realized tilemap layers are no longer rendered.

👇  Here on a screenshot below you can see how I am preparing for a canvas size change. Numbers visible on the screenshot are chosen in a way to have all layers included, so you can see how only tilemap ones are not rendered.

before canvas size change

👇  And here is a screenshot with a canvas size change result, no tilemaps visible.

after canvas size change

ℹ️  Let me know if you need a source file from me.

* Aseprite version: `v1.3-beta6`, from Steam
* System: macOS

iamOgunyinka:
Details of bug can be seen in #2942 -- when a change to the canvas size makes the tilemap layers disappear completely.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -56,8 +56,13 @@ jobs:
         if [[ "${{ runner.os }}" == "Linux" ]] ; then
           export XVFB=xvfb-run
         fi
+        if [[ "${{ github.base_ref }}" == "beta" ]] || [[ "${{ github.ref }}" == "beta" ]] ; then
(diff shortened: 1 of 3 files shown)


dacap:
Adding some missing scenarios so far in the issue for future reference:

canvasresize-bug

PR #2960, opened 2021-09-22
Fix Alt+mouse wheel to scroll through tiles when we are in tiles mode (fix #2956)

### LINKED ISSUE #2956: Fix Alt+mouse wheel to scroll through tiles when we are in tiles mode
Reported here: https://community.aseprite.org/t/alt-scroll-doest-work-in-the-tile-palette/10970/1

Gasparoken:
Before this fix, Alt+mouse wheel to scroll through tiles didn't work as it did in the color palette.
Also, fixed tile deselection when user removes the focus of the tileset (for example, doing click in the editor).

### DIFF
--- a/src/app/ui/color_bar.cpp
+++ b/src/app/ui/color_bar.cpp
@@ -409,6 +409,22 @@ void ColorBar::setBgColor(const app::Color& color)
     onColorButtonChange(color);
 }
 
+void ColorBar::setFgTile(doc::tile_t tile)
+{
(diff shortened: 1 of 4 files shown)


dacap:
This was already merged here https://github.com/aseprite/aseprite/commit/4e12f4dcc237aac79145502305f2e47b28c33f0a

nkoder:
does it applies to `[` and `]` shortcuts too?

PR #2962, opened 2021-09-22
Fix incorrect cropping rectangle calculation in canvas size command in lua script

Gasparoken:
Prior to this fix, CanvasSize command via lua script results in bad canvas and image clipping.


Gasparoken:
This fix is ​​required to successfully run any app.command.CanvasSize {limits = Rectangle (..., ..., ..., ...)}

dacap:
Nice catch @Gasparoken! I've just cherry-picked this change in `main` branch (83c18399a4c7f3c3628930b2cb4fc02f5e6e7669) and merge `main` to `beta` (so it's fixed in both branches now).

PR #2964, opened 2021-09-23
Fixed disappearing tilemap layers on resizing canvas(fix #2942)

### LINKED ISSUE #2942: [bug] Canvas Size change makes tilemap layers disappear

iamOgunyinka:
Details of bug can be seen in #2942 -- when a change to the canvas size makes the tilemap layers disappear completely.
Referenced also in a [closed PR](https://github.com/aseprite/aseprite/pull/2947).

### DIFF
--- a/src/app/doc_api.cpp
+++ b/src/app/doc_api.cpp
@@ -261,17 +261,36 @@ bool DocApi::cropCel(LayerImage* layer,
       newCelBounds.offset(-bounds.origin());
 
       gfx::Point paintPos(newCelBounds.x - newCelPos.x,
-                          newCelBounds.y - newCelPos.y);
+        newCelBounds.y - newCelPos.y);
 
+      const color_t bg = image->pixelFormat() == IMAGE_TILEMAP ?
+                           notile :
+                           m_document->bgColor(layer);
       newCelPos = newCelBounds.origin();
 
-      // Crop the image
+      doc::Grid grid;
+      if (layer->isTilemap()) {
+        const Tileset* tileset = static_cast<LayerTilemap*>(layer)->tileset();
+        grid = tileset->grid();
+
+        const gfx::Rect  roiRect(bounds.origin() - cel->position(),
+                                 newCelBounds.size());
+        const gfx::Size newImageSize = grid.canvasToTile(roiRect).size();
+        newCelBounds.setSize(newImageSize);
+
+        const gfx::Point prevPaintPos = paintPos;
+        paintPos = grid.canvasToTile(paintPos);
+
+        const gfx::Point celPaintPixelPos = grid.tileToCanvas(paintPos);
+        newCelPos += (celPaintPixelPos - prevPaintPos);
+      }
+
+      // crop the image
       ImageRef newImage(
         crop_image(image,
                    paintPos.x, paintPos.y,
                    newCelBounds.w, newCelBounds.h,
-                   image->pixelFormat() == IMAGE_TILEMAP ?
-                     notile : m_document->bgColor(layer)));
+                   bg));
 
       // Try to shrink the image ignoring transparent borders
       gfx::Rect frameBounds;
@@ -284,10 +303,11 @@ bool DocApi::cropCel(LayerImage* layer,
             crop_image(newImage.get(),
                        frameBounds.x, frameBounds.y,
                        frameBounds.w, frameBounds.h,
-                       image->pixelFormat() == IMAGE_TILEMAP ?
-                         notile : m_document->bgColor(layer)));
-
-          newCelPos += frameBounds.origin();
+                       bg));
+          if (layer->isTilemap())
+            newCelPos += grid.tileToCanvas(frameBounds.origin());
+          else
+            newCelPos += frameBounds.origin();
         }
       }
       else {
@@ -295,7 +315,7 @@ bool DocApi::cropCel(LayerImage* layer,
         return false;
       }
 
-      // If it's the same iamge, we can re-use the cel image and just
+      // If it's the same image, we can re-use the cel image and just
       // move the cel position.
       if (!is_same_image(cel->image(), newImage.get())) {
         replaceImage(cel->sprite(),


src/app/doc_api.cpp
```diff
@@ -261,17 +261,46 @@ bool DocApi::cropCel(LayerImage* layer,
...
+        else
+          newCelPos.x += celPosDifference.x;
+
+        if (imageSizeInPixel.h >= newCelBounds.h)
+          newCelPos.y -= celPosDifference.y;
+        else
+          newCelPos.y += celPosDifference.y;
+
+        paintPos = newPaintPos;
+        newCelBounds.setSize(imageTileSize);
```
Gasparoken, line 295: This can be simplified. Position arithmetics and some CanvasToTiles conversion is enough.
Take care of this variables:
1. `paintPos` will be the start coordinate in tile units refered to the tilemap image (for example: paintPos = (0, 0) is the most left-top tile of the cel image). It's used in `crop_image` function.
2. `newCelBounds` is used to define the crop size (in tile units) of `crop_image` function.
3. `newCelPos` it'll the "final" cel position in pixels, but be careful,  in the following `shrink`, `newCelPos` may need some adjust.

src/app/doc_api.cpp
```diff
@@ -261,17 +261,46 @@ bool DocApi::cropCel(LayerImage* layer,
...
+      const color_t bg = image->pixelFormat() == IMAGE_TILEMAP ?
+                           notile :
+                           m_document->bgColor(layer);
       newCelPos = newCelBounds.origin();
 
-      // Crop the image
+      doc::Grid grid;
+      if (layer->isTilemap()) {
+        const Tileset* tileset = static_cast<LayerTilemap*>(layer)->tileset();
+        grid = tileset->grid();
```
Gasparoken, line 274: This `grid` variable has `m_origin`=(0, 0). Then `grid.canvasToTile( some canvas bound )` will result in bounds (in tile units) referred to a grid  aligned with the canvas. So, `grid.canvasToTile( some canvas bound )` in the case which the tilemap has an offset of some pixels (>0 and less than one tilesize) will result in wrong tile bounds.

src/app/doc_api.cpp
```diff
@@ -284,18 +313,19 @@ bool DocApi::cropCel(LayerImage* layer,
             crop_image(newImage.get(),
                        frameBounds.x, frameBounds.y,
                        frameBounds.w, frameBounds.h,
-                       image->pixelFormat() == IMAGE_TILEMAP ?
-                         notile : m_document->bgColor(layer)));
-
-          newCelPos += frameBounds.origin();
+                       bg));
+          if (layer->isTilemap())
+            newCelPos += grid.tileToCanvas(frameBounds.origin());
```
Gasparoken, line 318: This is the final adjust. My recommendation is to debug this when cases where no shrink is needed are resolved.

src/app/doc_api.cpp
iamOgunyinka: Thank you for the review. After taking another look at the code, I realized the value of the below arithmetic would always be equal to `paintPos` and thus makes `roiOringin` unnecessary. The only single occurrence of it can be substituted with `paintPos`
```
const gfx::Point roiOrigin = bounds.origin() - cel->position();
const gfx::Rect  roiRect(roiOrigin, bounds.size());
```
to give this.
```
const gfx::Rect  roiRect(paintPos, bounds.size());
```
However, I think the other calculations are necessary so I'll walk you through my thought process. Perhaps in trying to explain it, you and/or I can come up with a better alternative to make it more elegant than it is now.

`imageTileSize` and `newPaintPos`, as you mentioned, are quite straightforward -- they're the conversion of the image/coordinate from pixel to tile unit and they're used as-is on lines 294 and 295 respectively. We, however, need to get the position of the `cel` right, otherwise the positioning of the image view would be incorrect. 

The result of using the point designated by `newPaintPos` on an image gives the entire part of the tile at that index. However, the bounds origin specified may have touched only a small part of the tile at that starting image index (or call it coordinate/point). So we needed a way to get how far out or in, from that tile we need to "go". Thus, we convert the `newPaintPos` back to pixel and find the difference between the "original" position specified in the bounds and the calculated position returned in `newPaintPos`. This is what I was trying to accomplish here.
```
celPaintPixelPos = grid.tileToCanvas(newPaintPos); // <-- the origin of the tile (in pixel) starting that image
celPosDifference = paintPos - celPaintPixelPos; // <- the difference between the start and the paintPos
```

Then, we adjusted the `celPosition`'s `x` and `y` coordinates based on the newly created image size.

src/app/doc_api.cpp
iamOgunyinka: Adjusting the `grid`'s origin here would affect the calculation of the `newPaintPos`. We only need the origin, one time, to get the right size of the new image. All the other calls to grid need the absolute positioning to work.

src/app/doc_api.cpp
iamOgunyinka: When there's no shrink, this check `if (frameBounds != newImage->bounds())` would fail and the second cropping would not happen.

src/app/doc_api.cpp
Gasparoken: I see this:

src/app/doc_api.cpp
```diff
@@ -261,17 +261,46 @@ bool DocApi::cropCel(LayerImage* layer,
...
+
+        paintPos = newPaintPos;
+        newCelBounds.setSize(imageTileSize);
+      }
+
+      // crop the image
       ImageRef newImage(
         crop_image(image,
                    paintPos.x, paintPos.y,
                    newCelBounds.w, newCelBounds.h,
```
Gasparoken, line 302: Remember: to crop correctly a tilemap image with `crop_image` function, `paintPos.x`, `paintPos.y`, `newCelBounds.w`, `newCelBounds.h` these should be in tile units.

(x, y) `paintPos`  is the top-left corner tile coordinate from the `image` where the crop begins.

(w, h) `newCelBounds` is the size (in tile units) of the part of the `image` to being cropped.

Gasparoken: When you do this:
```
const Tileset* tileset = static_cast<LayerTilemap*>(layer)->tileset();
grid = tileset->grid();
```
`grid.origin()` will be (0, 0) because the `static_cast<LayerTilemap*>(layer)->tile set()` doesn't return grid origin information, that's why we need to adjust the grid.origin. If you didn't adjust the `grid.origin()` the `canvasToTile` function will return incorrect coordinates.
Look at the following example:  correct `canvasToTile` return values (2x2 pixel individual tile size):
canvasToTile

src/app/doc_api.cpp
dacap: I didn't read all these comments, but I wanted to say that it's important to set a correct grid origin (generally the tilemap cel position, as @Gasparoken said) so `grid.canvasToTile()` and `grid.tileToCanvas()` functions work as expected (in other case converting between canvas <-> grid/tiles coordinates will not work). All the code in Aseprite is programmed in such way, and that's the expected behavior.

For example, `grid.tileToCanvas(Point(0, 0))` should return the origin position of the grid (the tilemap cel position) instead of `Point(0, 0)`. And `grid.canvasToTiles(cel->position())` should return `Point(0, 0)`.

> Adjusting the grid's origin here would affect the calculation of the newPaintPos.

From the code `paintPos` is a delta in pixels (`newCelBounds.origin() - newCelPos`) and you are converting that delta to tiles (`newPaintPos = grid.canvasToTile(paintPos)`). I can see why in this case the grid should have origin=(0,0). My personal recommendation would be that you should try to avoid converting deltas in pixels directly with the grid (it will not work). Instead you should convert canvas positions (in pixels) to tiles, and then calculating the delta of tiles (maybe something like `grid.canvasToTile(newCelBounds.origin()) - grid.canvasToTile(newCelPos)`, but I'm not sure).

PR #2980, opened 2021-10-03
Interprocessing via WebSocket

lampysprites:
Suggestion of adding a websocket client available to the scripting API, and a few other things that make it useful for live syncing with other applications.

## Motivation

When using aseprite with some other applications, needing to save/export a file and reload it feels rather limiting. It’d be great to share data in real time automatically, in both directions. The patch is based on my in-house solution for updating textures in Blender. I also noticed similar desires in [LDTk](https://github.com/deepnight/ldtk) (watching for .ase file change) and [Sprytile](https://github.com/Sprytile/Sprytile) (reloading periodically). There’s also some possibilities beyond data syncing. While websocket isn’t the most elegant way of interprocessing, it’s cross-platform, well supported, and easy to understand.

## Changes in the Lua API
 - `WebSocket` type that implements a client (using https://github.com/machinezone/IXWebSocket)
 - `onchange` callback to Sprite and Site, added in order to react to the changes
 - `Image:bytes` property for getting and setting image data as a byte string

All changes [doc.md](https://github.com/aseprite/aseprite/files/7274198/apichange.md)

Demo: [external preview app](https://github.com/lampysprites/aseprite-interprocessing-demo)

## Notes and further discussion

- CMake: Websocket dependency and functionality can be disabled with `-DENABLE_WEBSOCKETS=OFF` . To avoid a warning, replace `third_party/IXWebSocket/CMakeLists.txt` (solution offered by dacap) - [diff.txt](https://github.com/aseprite/aseprite/files/7274210/diff.txt)

- Perhaps networking should prompt for a permission the same way as file ops do now.

- It’s a bit dangerous that a forgotten connection will be only closed during garbage collection. So, for example, assigning it to a global variable will keep it, and callback will be called after the script finishes. It also might be done on purpose, so I guess closing becomes a developer’s responsibility.

- Sprite’s onChange lifespan is kinda messy since internally it’s one per document, but is cleaned during GC. I haven’t realized the contradiction until writing the docs. Thinking of removing the gc at all, or just checking that it’s not some other callback being removed. Would love to hear some suggestions.

- I don’t think there’s a way to check if it’s safe to modify an image asynchronously, but in my previous blender-specific patch it was a rare event. At times it crashed, but usually only an  “another operation is happening” message showed.

- Currently, only client implementation is exposed to the API. IXWebSocket already implements the server, so adding it is not that hard, but feels a bit excessive? Needs more opinions.

-------

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -66,3 +66,6 @@
 [submodule "src/tga"]
 	path = src/tga
 	url = https://github.com/aseprite/tga.git
+[submodule "third_party/IXWebSocket"]
+	path = third_party/IXWebSocket
+	url = https://github.com/machinezone/IXWebSocket
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ option(ENABLE_MEMLEAK      "Enable memory-leaks detector (only for developers)" 
 option(ENABLE_NEWS         "Enable the news in Home tab" on)
 option(ENABLE_UPDATER      "Enable automatic check for updates" on)
 option(ENABLE_SCRIPTING    "Compile with scripting support" on)
+option(ENABLE_WEBSOCKET    "Compile with websocket support" on)
 option(ENABLE_TESTS        "Compile unit tests" off)
 option(ENABLE_BENCHMARKS   "Compile benchmarks" off)
 option(ENABLE_TRIAL_MODE   "Compile the trial version" off)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -74,6 +74,11 @@ if(ENABLE_SCRIPTING)
   add_definitions(-DENABLE_SCRIPTING)
 endif()
 
+if(ENABLE_WEBSOCKET)
+  # Needed for "app" and "main"
+  add_definitions(-DENABLE_WEBSOCKET)
+endif()
+
 ######################################################################
 # Aseprite Libraries (in preferred order to be built)
 
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -673,7 +673,7 @@ if(REQUIRE_CURL)
 endif()
 
 if(ENABLE_SCRIPTING)
-  target_link_libraries(app-lib lua lauxlib lualib)
+  target_link_libraries(app-lib lua lauxlib lualib ixwebsocket)
 endif()
 
 if(ENABLE_UPDATER)
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -147,6 +147,10 @@ if(ENABLE_SCRIPTING)
       commands/cmd_open_script_folder.cpp
       ui/devconsole_view.cpp)
   endif()
+  if(ENABLE_WEBSOCKET)
+    set(scripting_files_ws
+      script/websocket_class.cpp)
+  endif()
   set(scripting_files
     commands/cmd_run_script.cpp
     script/app_command_object.cpp
@@ -189,8 +193,8 @@ if(ENABLE_SCRIPTING)
     script/tool_class.cpp
     script/values.cpp
     script/version_class.cpp
-    script/websocket_class.cpp
     shell.cpp
+    ${scripting_files_ws}
     ${scripting_files_ui})
 endif()
 
@@ -674,7 +678,11 @@ if(REQUIRE_CURL)
 endif()
 
 if(ENABLE_SCRIPTING)
-  target_link_libraries(app-lib lua lauxlib lualib ixwebsocket)
+  target_link_libraries(app-lib lua lauxlib lualib)
+
+  if(ENABLE_WEBSOCKET)
+    target_link_libraries(app-lib ixwebsocket)
+  endif()
 endif()
 
 if(ENABLE_UPDATER)
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -189,6 +189,7 @@ if(ENABLE_SCRIPTING)
     script/tool_class.cpp
     script/values.cpp
     script/version_class.cpp
+    script/websocket_class.cpp
     shell.cpp
     ${scripting_files_ui})
 endif()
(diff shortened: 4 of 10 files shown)


Eternal-plasma:
Very Cool Idea!

dacap:
@lampysprites thanks a lot for this implementation, I'll try to review it and test it ASAP (maybe today and tomorrow). I remember your Aseprite branch so I'm pretty sure this implementation took you a lot of time.

lampysprites:
Glad to hear that. Surprizingly, without needing to do the actual data syncing it became a relatively tiny amount of work.  

Please do it at your pace, I can do it later if the review takes a while.

Sorry about spamming the commits, there was a careless mistake

dacap:
I've just pushed these changes into https://github.com/aseprite/aseprite/tree/interprocessing with a couple of fixes, I'll check if the CI works and then review the changes and patch what is necessary. Personally I don't like the `Site.onChange` because the `Site` object acts like a struct (to save the current site) but it doesn't have behavior. I'll see if I can move this to `app.events.onsitechange` or something similar (still thinking about the API).

dacap:
I've been thinking about the Lua API, and I'll make some changes in https://github.com/aseprite/aseprite/tree/interprocessing

I'll see if this is possible:

* `Sprite.events:on(eventName, function)`
* `Sprite.events:off(function)` and `Sprite.events:off(eventName, function)`
* `app.events:on(eventName, function)`

After this I'll send you a PR to https://github.com/lampysprites/aseprite-interprocessing-demo to use the new API. But the general idea would be something like:

```lua
spr = Sprite(32, 32)
spr.events:on('change', onSpriteChange)
app.events:on('sitechange', onSiteChange)
...
spr.events:off(onSpriteChange)
app.events:off(onSiteChange)
```

dacap:
@lampysprites I've just merged your PR with some extra changes. Thanks a lot for sharing this feature and your help implementing this. :rocket:

dacap:
I've updated the documentation too:

* https://github.com/aseprite/api/blob/next-version/api/websocket.md#websocket
* https://github.com/aseprite/api/blob/next-version/api/events.md#events

Thanks for this too :heart:

lampysprites:
Thank you very much for your work!!

I've still got a few small changes, although not entirely sure that's the way the things should be solved:


- Full trust permission doesn’t stop the alert from appearing again because `FileAccessMode::Full` doesn’t have the bit for `FileAccessMode:OpenSocket`. Changing `security.cpp:202` to something like `if (dlg.full()->isSelected() && resourceType != ResourceType::WebSocket)` will save a websocket permission without granting file access and networking both at once.


- While writing scripts for this api, the exponential backoff got quite annoying. When the server stops for 0.1 * 2^n seconds, the next attempt takes another 0.1 * 2^n seconds. Since this solution is mostly meant for app communication, perhaps it’s okay to limit it like
```cpp
    // websocket_class.cpp:40
    auto ws = new ix::WebSocket();
    ws->setMaxWaitBetweenReconnectionRetries(5000); // milliseconds
```


- I was trying to avoid exposing the registry - a wrong index to unsubscribe may free a ref that is stored elsewhere, and looks like it could be allocated again later for a different function? Lua doc is ambiguous on it but the luaL_ref/unfer keep a “freelist” of recycled references. In 85f3164 I tried to place the ids inside an userdata to avoid creation/modification, but it seemed very excessive. Instead, how about checking that it’s an existing listener is being unsubscribed? Then it only breaks if the listeners are the same kind:
```cpp
    //events_class:293
    if (evs->hasListener(callbackRef)) {
      evs->remove(callbackRef);
      luaL_unref(L, LUA_REGISTRYINDEX, callbackRef);
    }
```

- Something I couldn’t figure out - aseprite stops updating when the mouse is out of the window, and it delays the socket callback until it's back. It’s still usable like this, but if there’s a cheap way to kick the ui loop, I hope you can add it. Happens on windows build, in case it matters.

lampysprites:
> I don't what could be best here. Maybe we can expose those properties ...

That's why I was not sure. It would make sense to wait for several complaints about sockets not connecting before doing anything I guess :) 

> Oh yes, this can happen when there is nothing in the UI that needs to be updated with a timer ...

That was the case, here's the patch: [timer.diff.txt](https://github.com/aseprite/aseprite/files/7312432/timer.diff.txt)

No idea what interval should be though, but it has to fire to update the ui.

snippets to test with the demo
```
// main.cpp, inside the loop

if (tigrKeyDown(screen, TK_SPACE)) {
  for (auto ws : serv.getClients()) {
    ws->sendText("A");
  }
}


-- client.lua, message handler

if t == WebSocketMessageType.Text then
    app.alert(message)
```

API-Beast:
Two things I noticed while using this:

1) Why isn't the API selection exposed to Lua? It's just a custom string that's send to the server so it knows how to deal with the client if there is different API versions available. It's not a big deal, but it seems weird to leave out this feature.
2) The URLs in the examples should probably use the WebSocket (ws://) protocol instead of the HTTP (http://) protocol. As I understand it that websocket even allows the second is a compatibility feature to deal with old HTTP servers like Apache or NGinx and will have some overhead. (The overhead being an additional HTTP handshake and subsequent negotiation to see if the HTTP connection can be "upgraded" to the WebSocket protocol.)

dacap:
Hi @API-Beast,

> 1. Why isn't the subprotocol selection exposed to Lua? It's just a custom string that's send to the server when connecting so it knows how to deal with the client if there is different subprotocols available. It's not a big deal since an empty string is also a valid subprotocol and it's up to the server how to handle this, but it seems weird to leave out this feature.

I'm not sure, but I think we could add it in the [WebSocket() constructor](https://github.com/aseprite/api/blob/next-version/api/websocket.md#websocket-1) or adding a member function to add subprotocols (just like [the IXWebSocket API](https://github.com/aseprite/IXWebSocket/blob/6d8495bd73e7390443ec61b973d85eb5847e44e0/ixwebsocket/IXWebSocket.h#L61)). It looks more like an array of strings, but we could just split a CSV string.

> 2. The URLs in the examples should probably use the WebSocket (ws://) instead of the HTTP (http://). As I understand it that websocket even allows the second is a compatibility feature to deal with old HTTP servers like Apache or NGinx and will have some overhead. (The overhead being an additional HTTP handshake and negotiation to see if the HTTP connection can be upgraded to the WebSocket protocol.)

I'll give a try to some examples, but it would be better if the examples use ws:// directly :+1:

PR #2989, opened 2021-10-06
Replace the curl directory by a submodule

martincapello:
Replaces the current curl version in the third_party/curl directory for a submodule in the same path with a newer version of curl (7.79.1).


martincapello:
Oh yeah, I forgot about that! Will update it

PR #3000, opened 2021-10-13
Add initial support for loading .psd photoshop files in Aseprite

iamOgunyinka:
These series of commit attempts loading some Photoshop files directly in Aseprite. The initial support should be able to open RGB, Indexed and Grayscale images, with 8 or 16 bits per channel.

### DIFF
--- a/src/app/file/psd_format.cpp
+++ b/src/app/file/psd_format.cpp
@@ -100,8 +100,8 @@ FileFormat* CreatePsdFormat()
 class PsdDecoderDelegate : public psd::DecoderDelegate {
 public:
   PsdDecoderDelegate()
-    : m_currentLayer(nullptr)
-    , m_currentImage(nullptr)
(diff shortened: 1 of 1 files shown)


dacap:
Hi @iamOgunyinka, send a new commit with the `src/psd` updated. Thanks.

dacap:
Merged in https://github.com/aseprite/aseprite/commit/026eeb96b9e4cfd276facdb00235a98d53b59641 with some minor changes.

Some future work required:

1. Merge `onImageScanline()` on one member function
2. Check if both `hasTransparency()` member functions can be merged in function in some one way
3. Use `ImageIterators` on `onImageScanline()` and move all the needed code from `PsdDecoderDelegate::putPixel()` to `onImageScanline()`, it'll be faster if we process scanline by scanline (using iterators/pointers) instead of pixel by pixel (using getPixel/putPixel).

PR #3009, opened 2021-10-18
Interprocessing fixes

lampysprites:
Hello again!

Out of three commits here, first two implement the solutions for connection backoff and update timer [from the initial IPC PR](https://github.com/aseprite/aseprite/pull/2980#issuecomment-938670554) (thanks!)

The last commit avoids a freeze that happened randomly (and regularly) when the tab, or the app is closed. As far as I can tell, it is a deadlock in `ClosedDocs::addClosedDoc`'s unique lock and `SpriteEvents::~SpriteEvents` call to `doc()->remove_observer`. Also noticed that for it to happen, the doc has to be released from ClosedDocs, doesnt matter if it's timed or immediately, otherwise the destructor is not called. Honestly I don't see what's wrong there.

To reproduce:
1. disable Settings>Files>Keep Closed Sprite In Memory (or set to 10s, harder to catch but same thing happens)
2. open a file and run the following script, it creates spriteEvents 
``` lua
app.activeSprite.events:off(app.activeSprite.events:on("change", function() end))`
```
3. Then close the file. Repeat open/script/close several times depending on today's luck.
4. If the day is exceptionally lucky, runthe script several times and open multiple sprites.
5. Not unsubscribing events makes it less likely (?!)

Regarding the solution, I only removed the destructor involved. It seems that observer and event classes are destroyed together, and the global pointer clears in onDestroy, so no subscription remains anyway? Still I have a suspicion that the destructor was there for a reason.

With these changes, I don't run into any critical issues while doing half-duplex communication.

(For a while, the networking thread and this issue were the main suspect, perhaps it'll strike later:
https://github.com/machinezone/IXWebSocket/issues/316)

-------------------

### DIFF
--- a/src/app/script/websocket_class.cpp
+++ b/src/app/script/websocket_class.cpp
@@ -30,8 +30,8 @@ namespace {
 // Additional "enum" value to make message callback simpler
 #define MESSAGE_TYPE_BINARY ((int)ix::WebSocketMessageType::Fragment + 10)
 
-static ui::Timer* timer;
-static std::set<ix::WebSocket *> connections;
+static std::unique_ptr<ui::Timer> g_timer;
+static std::set<ix::WebSocket *> g_connections;
 
 int WebSocket_new(lua_State* L)
 {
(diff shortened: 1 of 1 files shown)


dacap:
I'm going to investigate, I was able to reproduce the deadlock with the steps you gave. I think it's related to the `onDestroy` listener of SpriteEvents, which can be called from the ClosedDocs background thread. It's quite complex but we'll see if I can find a fix.

dacap:
I think the deadlock issue should be fixed now: https://github.com/aseprite/aseprite/commit/4dcdcfe387d613400758f2ba7eedcef923a09631

PR #3013, opened 2021-10-19
Add a check to allow exit command execution only when the main window is the currently running

martincapello:


dacap:
Cherry-picked https://github.com/aseprite/aseprite/commit/c5fccb1499bdda8dc7db569a942e7cde12d5b0e7

I was able to reproduce the bug and check that this PR fixed it using the Frame Properties dialog as the foreground window and then pressing the main window close button.

PR #3048, opened 2021-11-11
[lua] Add "filenamechange" event to Sprite.events

lampysprites:
Can be convenient for watching app state. I think it could possibly just emit "change" and expect the scripter to put an if/else.

One other use I can think of is keeping some custom data in an external file, if text .data is not enough.

`Document::setFilename` currently fires the event without checking if the string actually changed, and I think it's ok.

Doc can be :

* `'filenamechange'`: When `Sprite.filename` is modified by user saving the file or a script. It happens even if the name string remains the same.

-------------------

### DIFF
--- a/src/app/script/events_class.cpp
+++ b/src/app/script/events_class.cpp
@@ -192,7 +192,7 @@ class SpriteEvents : public Events
                    , public DocUndoObserver
                    , public DocObserver {
 public:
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @lampysprites 🙏  Just merged and updated the [API docs for the next version](https://github.com/aseprite/api/blob/next-version/Changes.md#v1231).

PR #3049, opened 2021-11-11
Fix confusion bug when layer name is an integer(fix #3045)

### LINKED ISSUE #3045: when changing the layer name to a number, aseprite API gets confused
# Bug:
when running the following simple lua code:
```lua
for i, layer in ipairs(app.activeSprite.layers) do
    print(i..") "..layer.name)
end
```
which lists all layers and their index, Aseprite gets confused if you rename one of the layers to a single digit, and the enumeration doesn't work properly anymore:

https://user-images.githubusercontent.com/4385118/140664667-8fb017fb-312e-458f-a92f-a3c6539d1926.mp4

I have also experienced that sometimes layers don't get enumerated at all (can't reproduce).

### Steps to reproduce:
1. Create a new aseprite file.
2. Create several new layers.
3. Run the lua command
    -> everything should be fine
4. rename "Layer 3" to "1" and "Layer 4" to "2"
5. Run the lua command 
    -> "Layer 1" will be displayed as "1" and "Layer 2" as "2" (same name as the actual "1" and "2").
6. Continue adding layers
    -> added layers are displayed out of order

## Update:
1. If you rename a layer to a number which is higher than the layer count it will add an "artificial" layer which doesn't actually exist (7 layers on the left, 8 layers enumerated):

image

2. Or just assign it a random name (8 present 3 times in the enumeration):

image

### System
Aseprite 1.3 beta-7 on windows 10 installed from Steam

iamOgunyinka:
This PR fixes a bug in the scripting engine where having an integer as a layer name confuses the `Layers_index`, so we add a check that is a lot stricter and avoids conversion from integer to strings.

### DIFF
--- a/src/app/script/layers_class.cpp
+++ b/src/app/script/layers_class.cpp
@@ -61,7 +61,7 @@ int Layers_index(lua_State* L)
   auto obj = get_obj<LayersObj>(L, 1);
 
   // Index by layer name
-  if (lua_isstring(L, 2)) {
+  if (lua_type(L, 2) == LUA_TSTRING) {
     if (const char* name = lua_tostring(L, 2)) {
       for (ObjectId layerId : obj->layers) {
         Layer* layer = doc::get<Layer>(layerId);


dacap:
Thanks @iamOgunyinka, I wasn't able to test this yet, anyway a couple of things:

1. Could you please use the following style `... (fix #3045)` to end the commit summary?
2. It would be nice if we can add a new test in https://github.com/aseprite/tests/blob/main/scripts/layers.lua to check the bug and the fix

iamOgunyinka:
I'll add tests in the morning.

dacap:
Thanks @iamOgunyinka for this fix! I've just cherry-picked in `main` branch (as this bug is in `main` branch): https://github.com/aseprite/aseprite/commit/12e61d33be4b6a660f62b30d493fd124614e5732

PR #3056, opened 2021-11-16
Add support for sprite slicing

iamOgunyinka:
This PR adds support for slicing. Due to the nature of the preference, the default slice color, if any, defaults to blue.

### DIFF
--- a/src/app/file/psd_format.cpp
+++ b/src/app/file/psd_format.cpp
@@ -11,6 +11,7 @@
 #include "doc/image_impl.h"
 #include "doc/layer.h"
 #include "doc/palette.h"
+#include "doc/slice.h"
 #include "doc/sprite.h"
 #include "psd/psd.h"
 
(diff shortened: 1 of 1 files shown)


src/app/file/psd_format.cpp
```diff
@@ -364,16 +367,31 @@ class PsdDecoderDelegate : public psd::DecoderDelegate {
       const int timeMs = frameInfo.duration * 10;
       m_sprite->setFrameDuration(frame_t(i), timeMs);
     }
-    // at this point, we're supposed to be able to set the activeFrame
+
+    auto& slices = m_sprite->slices();
+    for (const auto& slice : m_slices.slices) {
```
dacap, line 372: Couldn't we do this just in `onSlicesData()` without storing the slices in `m_slices`?

dacap:
I've just tested and it looks like the cel is not correctly aligned, and "disabled" slices shouldn't be imported:

Screen Shot 2021-12-01 at 09 57 53

Result:

Screen Shot 2021-12-01 at 09 57 41

Expected:

Screen Shot 2021-12-01 at 09 59 33

iamOgunyinka:
@dacap thanks for your review, I'll take another look at this.

dacap:
Attaching the slices.psd file inside a zip: [slices.psd.zip](https://github.com/aseprite/aseprite/files/7641554/slices.psd.zip)

PR #3059, opened 2021-11-17
Save and retain color profile when saving palettes (fix #2893)

### LINKED ISSUE #2893: Saved Palette Doesn't Retain Color Profile
Hi,

I have an issue to report.

When palettes are saved to file formats that are capable of holding color profiles, such as .aseprite and .png, the exported palette files does not retain the color profile of the original sprite. I expect that it would. Otherwise the colors in the palette file will not match those of the original sprite. To reproduce:

- Create a new sprite
- Select a palette.
- Prepare to observe potential change in palette appearance.
- Assign or convert the sprite to a non-standard color profile contained in an .icc file.
- Observe change in palette appearance, if any.
- Click on the hamburger menu in the palette panel.
- Save Palette.
- Select a file format that can retain a color profile (for example, .aseprite or .png).
- Open the newly created palette file.
- Compare the palette visually with the palette in the original sprite.

I'm not sure what the expectation would be for formats which don't retain a color profile, such as .pal and .gpl. Likely candidates would be convert or assign to None or to SRGB. If there is no conensus on the right option, there could be a warning about the loss of information or a choice presented to the user.

(A similar warning may be helpful for loss of the alpha channel in .pal and .gpl formats, though the latter is complicated by Aseprite's extension for the format. See https://github.com/aseprite/aseprite/issues/2878.)

Thank you for considering this.
Best,
Jeremy

iamOgunyinka:
This PR allows retaining color profile (in file format that is capable of holding it), when saving palettes from palette panel.

### DIFF
--- a/src/app/file/palette_file.cpp
+++ b/src/app/file/palette_file.cpp
@@ -118,7 +118,8 @@ Palette* load_palette(const char* filename,
   return pal;
 }
 
-bool save_palette(const char* filename, const Palette* pal, int columns)
+bool save_palette(const char* filename, const Palette* pal, int columns,
+                  const gfx::ColorSpaceRef& cs)
 {
   dio::FileFormat dioFormat = dio::detect_format_by_file_extension(filename);
   bool success = false;
@@ -154,11 +155,12 @@ bool save_palette(const char* filename, const Palette* pal, int columns)
       int h = (pal->size() / w) + (pal->size() % w > 0 ? 1: 0);
 
       Context tmpContext;
+      gfx::ColorSpaceRef colorSpace = (cs ? cs: gfx::ColorSpace::MakeNone());
       Doc* doc = tmpContext.documents().add(
         new Doc(Sprite::MakeStdSprite(
                   ImageSpec((pal->size() <= 256 ? doc::ColorMode::INDEXED:
                                                   doc::ColorMode::RGB),
-                            w, h), pal->size())));
+                            w, h, 0, colorSpace), pal->size())));
 
       Sprite* sprite = doc->sprite();
       doc->sprite()->setPalette(pal, false);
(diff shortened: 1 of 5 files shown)


src/app/script/palette_class.cpp
```diff
@@ -243,7 +243,7 @@ int Palette_saveAs(lua_State* L)
     if (!ask_access(L, absFn.c_str(), FileAccessMode::Write, ResourceType::File))
       return luaL_error(L, "script doesn't have access to write file %s",
                         absFn.c_str());
-    save_palette(absFn.c_str(), pal, pal->size());
+    save_palette(absFn.c_str(), pal, pal->size(), nullptr);
```
dacap, line 246: If the palette is associated to a sprite (`PaletteObj::spriteId`) we can get the color space from the sprite and use it here.

dacap (changes requested):
Hi @iamOgunyinka, the SavePaletteCommand looks good to me. I'm adding just a minor comment in the scripting side to modify the patch.

As a side note, please add the ` (fix ...)` to the commit summary itself: `Save and retain color profile when saving palettes (fix ...)`. And add a space between the parenthesis `... palettes(fix ...` -> `... palettes (fix ...`

src/app/file/palette_file.cpp
```diff
@@ -154,11 +155,12 @@ bool save_palette(const char* filename, const Palette* pal, int columns)
       int h = (pal->size() / w) + (pal->size() % w > 0 ? 1: 0);
 
       Context tmpContext;
+      gfx::ColorSpaceRef colorSpace = cs ? cs: gfx::ColorSpace::MakeNone();
```
dacap, line 158:
```suggestion
      gfx::ColorSpaceRef colorSpace = (cs ? cs: gfx::ColorSpace::MakeNone());
```

dacap:
I'll try to cherry-pick this patch for `main` branch (as this is an issue of the stable version).

dacap:
Thanks Joshua, I've just cherry-picked this into `main` branch in https://github.com/aseprite/aseprite/commit/3c7b34438985d8ce3527c9b9d2087a6336614abe

I've changed the summary as I just saw that it doesn't fully fix #2893, there is still a missing part to load the color profile when the palette is loaded: https://github.com/aseprite/aseprite/issues/2893#issuecomment-971468769

PR #3060, opened 2021-11-18
Fix tilemap layers empty after converting to a color profile (fix #3052)

### LINKED ISSUE #3052: TileMap Layers Empty After Converting To A Color Profile
Hello,

I found a bug again. This is in version 1.3-beta7x64 on Windows 10. Converting to a color profile causes a tile map layer to become transparent (or empty). To reproduce, 

1. Select any layer with visual content.
2. Convert the layer to a tile map layer.
3. Go to `Sprite > Properties`.
4. Select a color profile other than the sprite's current profile.
5. Press the `Convert` button.
6. Switch to eye drop tool, sample layer; should match the `Empty` tile.

The visual result may be similar to the one in the issue https://github.com/aseprite/aseprite/issues/2739 (the most closely related issue I noticed), though the cause may be different.

So far as I tested, assigning a color profile does not cause this issue. However, the colors in the tile set do not alter in appearance, as the palette color bar and canvas do. While an `.icc` file is not needed to test the conversion problem, using an `.icc` that differs from `sRGB` can make the problem with assignment easier to see.

The issue is compounded if settings in `Edit > Preferences > Color` use conversion, or are set to `Ask` and the user chooses a conversion. A sprite with a tile map layer can be saved with the `.aseprite` extension, then opened with an empty tile map layer.

Thanks,
Jeremy

Gasparoken:
Before this fix, tilemap layers disappeared after converting the Sprite's Color Profile in the Sprite Property dialog.

### DIFF
--- a/src/app/cmd/convert_color_profile.cpp
+++ b/src/app/cmd/convert_color_profile.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2020  Igara Studio S.A.
+// Copyright (C) 2018-2021  Igara Studio S.A.
 //
 // This program is distributed under the terms of
 // the End-User License Agreement for Aseprite.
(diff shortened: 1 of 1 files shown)


dacap:
Cherry-picked b5a787a20fe33fcaf995056028dcd61486856f3b

PR #3065, opened 2021-11-24
Fix bounds getting reset when the dialog is closed (fix #3018)

### LINKED ISSUE #3018: Dialog.bounds gets reset if the dialog is closed (cannot pass it to a second dialog)
A common practice when writing the UI for a script with multiple tabs is to pass the bounds of the first dialog (dlgA) to a new dialog (dlgB) in order to move it to the dlgA's position. This does not work as expected in 1.3-beta6. The (x, y) for **dlgB**.bounds is reset to (0, 0), moving the second tab to the top left corner of the app's window. 

To reproduce, use the [Wizard Example script ](https://raw.githubusercontent.com/aseprite/Aseprite-Script-Examples/master/Wizard%20Example.lua) from the sample scripts. It occurs with any script that uses the same approach. 

**(edit) IMPORTANT:** This only seems to occur with the "UI with multiple windows" option enabled (in `Edit > Preferences > Experimental > UI with multiple Windows`)

The workaround seems to be to cache the previous dialog's bounds and pass this copy to the new dialog. The second time the dialogs are loaded, their bounds will retain their new position, but it's not the one received from the first one.

Please see the attached gifs.

* Aseprite version: 1.3-beta6 dev
* System: Windows 10 64bit English

### Attachments
No script, just try it out with `Wizard Example.lua`

#### 1.3-beta 6
wizard dialog 1 3-beta6

#### 1.2.29
wizard dialog 1 2 29

iamOgunyinka:
This PR attempts to fix the bug reported extensively [here](https://github.com/aseprite/aseprite/issues/3018). Comments about the problem and why this fix works is detailed in the [comment](https://github.com/aseprite/aseprite/issues/3018#issuecomment-976390803) as well.

### DIFF
--- a/src/ui/window.cpp
+++ b/src/ui/window.cpp
@@ -411,8 +411,20 @@ void Window::closeWindow(Widget* closer)
     return;
 
   m_closer = closer;
-  if (m_ownDisplay)
+  if (m_ownDisplay) {
     m_lastFrame = m_display->nativeWindow()->frame();
+    if (!m_lastFrame.isEmpty() && parent()) {
+      const auto parentDisplay = this->parent()->display();
+      const gfx::Point parentOrigin =
+        parentDisplay->nativeWindow()->contentRect().origin();
+      const gfx::Point origin =
+        this->display()->nativeWindow()->contentRect().origin();
+      const int scale = parentDisplay->scale();
+      const gfx::Rect newBounds((origin -parentOrigin) / scale,
+        m_lastFrame.size());
+      setBounds(newBounds);
+    }
+  }
 
   manager()->_closeWindow(this, true);
 


LorenzoQuiroz (approved):
LGTM; small fix that would be appreciated by many users

dacap:
Thanks @iamOgunyinka, I've just cherry-picked this patch in https://github.com/aseprite/aseprite/commit/f9d41665f0b17e61778b77c59e85922f70a53ff7

Just moved the fix to `ui::Manager::_closeWindow()` because I notice a parallel with `_openWindow()`:
* _openWindow sets the bounds origin to 0,0 of the ui::Window when the os::Window is created
* So _closeWindow should restore the ui::Window bounds origin when the os::Window is destroyed

PR #3066, opened 2021-11-24
Fix Shift+N behavior(fix #2957)

### LINKED ISSUE #2957: Fix Shift+N behavior in v1.3-beta
Reported here: https://community.aseprite.org/t/add-new-layer-broken/10760

<kbd>Shift+N</kbd> should work exactly as in Aseprite v1.2.28 (`main` branch).

iamOgunyinka:
Fix the bug in the timeline when focus is lost in the timeline and a new layer is added. Instead of stacking the new layer on top of the currently selected layer, it gets added to the top of the entire layer stack.

### DIFF
--- a/src/app/ui/timeline/timeline.cpp
+++ b/src/app/ui/timeline/timeline.cpp
@@ -1782,8 +1782,8 @@ paintNoDoc:;
 
 void Timeline::onBeforeCommandExecution(CommandExecutionEvent& ev)
 {
-  m_savedCounter = (m_document ? *m_document->undoHistory()->savedCounter():
-                                 std::numeric_limits<int>::min());
+  m_savedVersion = (m_document ? m_document->sprite()->version():
+                                 std::numeric_limits<doc::ObjectVersion>::max());
 }
 
 void Timeline::onAfterCommandExecution(CommandExecutionEvent& ev)
@@ -1792,8 +1792,8 @@ void Timeline::onAfterCommandExecution(CommandExecutionEvent& ev)
     return;
 
   // TODO improve this: no need to regenerate everything after each command
-  const int currentCounter = *m_document->undoHistory()->savedCounter();
-  if (m_savedCounter != currentCounter) {
+  const doc::ObjectVersion currentVersion = m_document->sprite()->version();
+  if (m_savedVersion != currentVersion) {
     regenerateRows();
     showCurrentCel();
     invalidate();
--- a/src/app/ui/timeline/timeline.h
+++ b/src/app/ui/timeline/timeline.h
@@ -21,6 +21,7 @@
 #include "base/debug.h"
 #include "doc/frame.h"
 #include "doc/layer.h"
+#include "doc/object_version.h"
 #include "doc/selected_frames.h"
 #include "doc/selected_layers.h"
 #include "doc/sprite.h"
@@ -396,7 +397,7 @@ namespace app {
     // Value of DocUndo::savedCounter() before executing a
     // command. Used to compare after executing a command to avoid
     // regenerating all rows if it's not necessary.
-    int m_savedCounter;
+    doc::ObjectVersion m_savedVersion;
 
     // Data used to display each row in the timeline
     std::vector<Row> m_rows;


dacap:
Hi @iamOgunyinka, maybe we can change `m_savedCounter` to a `doc::ObjectVersion m_savedVersion;` defined in `doc/object_version.h` to avoid those casts.

iamOgunyinka:
@dacap please ignore this PR for now, I'll fix in the morning.

iamOgunyinka:
@dacap I had to use the `std::numeric_limits<T>::max()` because I tested and saw that the `::min()` returns 0 for the unsigned type and we may need a non-zero digit to designate the _start_ of the timeline.

dacap:
Cherry-picked in https://github.com/aseprite/aseprite/commit/ff851157ccaadd66ef648c0f69bcddb7d2f62d02 with some modifications (updated comment about `m_savedVersion` and have used 0 for the saved version when there is no document, I think it will not be a problem, because when we switch documents other mechanism updates the timeline rows).

iamOgunyinka:
> Cherry-picked in [ff85115](https://github.com/aseprite/aseprite/commit/ff851157ccaadd66ef648c0f69bcddb7d2f62d02) with some modifications (updated comment about `m_savedVersion` and have used 0 for the saved version when there is no document, I think it will not be a problem, because when we switch documents other mechanism updates the timeline rows).

Using `0` introduced a new bug, David. To reproduce, open Aseprite newly and open an existing document. The timeline looks like the image below:
image

Perhaps we can revert to using the `::max()` or any other non-zero digit.

dacap:
I wasn't able to reproduce it 🤔  I'll give a try to this tomorrow on Windows.

dacap:
@Gasparoken could you please check if you can reproduce [this bug](https://github.com/aseprite/aseprite/pull/3066#issuecomment-986621317)?

Gasparoken:
Hi @iamOgunyinka.

> Using 0 introduced a new bug, David. To reproduce, open Aseprite newly and open an existing document. The timeline looks like the image below:

I cannot reproduce the new bug. Can you currently reproduce it?

iamOgunyinka:
How to reproduce:

1. Create a new sprite, 32x32, Indexed, Transparent.
2. The default layer is an image layer, `Layer 1`, so draw anything on it and then hide it.
3. Add a new tilemap layer, draw anything on it too.
4. Add another layer image on the tilemap layer and draw on it, then hide this layer (i.e. `Layer 2`)

The structure of the layers is:
`Layer 2` -> not visible
Tilemap Layer -> visible
`Layer 1` -> not visible.

5. Select the second layer, the tilemap, and then save the file.

Then close Aseprite and re-open it. On fresh re-opening, select this file from the list of recent files and open.

Gasparoken:
Thanks @iamOgunyinka , I was able to reproduce it. I'm finding out why this is possible (I'm testing Timeline::onPaint() , nothing is drawn in this case, because the clipping regions are empty).

Gasparoken:
The bad visualisation in a fresh Aseprite launch is due to a bad initial vertical scroll position calculation on the time line.
The default `m_bounds` of a new fresh TimeLine instance is `gfx::Rect(0, 0, 0, 0)`.
Then in `void Timeline::updateScrollBars()` `m_viewportArea` will be calculated taking the default `m_bounds` as base.
`void Timeline::showCel(layer_t layer, frame_t frame)` will set `m_hbar.pos` and `m_vbar.pos` according a `celBounds` calculation.
The following line is correct only if the top-upper layer was the active layer while saving the file:
https://github.com/aseprite/aseprite/blob/ed595eb6b1fa0967732de7c9f35a658df569b3e6/src/app/ui/timeline/timeline.cpp#L3699

I think your initial solution about this line is correct:
Instead this
https://github.com/aseprite/aseprite/blob/ed595eb6b1fa0967732de7c9f35a658df569b3e6/src/app/ui/timeline/timeline.cpp#L1784
Use something like this when `m_document` is NULL:

`m_savedVersion = (m_document ? m_document->sprite()->version(): std::numeric_limits<int>::max());`

dacap:
Thanks for the explanation and the steps to reproduce the bug. I was able to reproduce it here too.

I'm particularly interested in this bug because I know that I saw this paint-error in the timeline using the official version several times in the past (before this patch, and in some other kind of scenario that I don't know how to reproduce). This case is not reproducible in the official version because we're already [doing this trick](https://github.com/aseprite/aseprite/blob/621bc209ac5cfde7e543ffed483b06e53ef9594b/src/app/ui/timeline/timeline.cpp#L1786), but I'm interested in checking the inner issue that is causing this problem (which might fix the other unknown issue that leads to the same paint-error situation).

I'll try to dedicate some extra time tomorrow to find the source of this bug before we put the same workaround, e.g. with the workaround the bug is fixed "by chance", but there are some other things that don't make sense (as the official version already is doing) like regenerating the timeline rows two consecutive times when switching between documents (one in updateUsingEditor() and other in onAfterCommandExecution()).

Issue #1366, opened 2016-12-14
gui.xml was not found


StigOlavsen:
After building from the repo, rev 4fada3a, on OSX 10.11.5, everything works fine when running aseprite from the build directory.

However after doing ninja install, and trying to run aseprite from the installed location in /usr/local/bin/aseprite, it gives the error gui.xml was not found.

The data folder was installed to /usr/local/share/aseprite/data, but aseprite seems unable to locate it?

dacap:
Hi @h0rdak, actually I'm not sure how the "ninja install" command might work as CMakeLists.txt files are not prepared to do so (I never use "ninja install", the executable is on the `build/bin` directory).

h0rdak:
I've attached the build manifest, showing what gets installed where. 

[install_manifest.txt](https://github.com/aseprite/aseprite/files/653332/install_manifest.txt)

As I said, it works fine if I run it from the build directory, but that's an inconvenient location to execute from every time. 

I was able to make an app bundle instead, so I regard the problem as solved on my part, but I would expect this to be an issue on linux as well, as it would install in a similar manner.

kasbert:
This helped for me:

`diff --git a/src/app/resource_finder.cpp b/src/app/resource_finder.cpp
index 4ba8cf567..10ae69f80 100644
--- a/src/app/resource_finder.cpp
+++ b/src/app/resource_finder.cpp
@@ -100,6 +100,9 @@ void ResourceFinder::includeDataDir(const char* filename)
   sprintf(buf, "../Resources/data/%s", filename);
   includeBinDir(buf);  // $BINDIR/../Resources/data/filename (inside a bundle)
 
+  // $BINDIR/../share/aseprite/data/filename (installed in /usr/ or /usr/local/)
+  sprintf(buf, "../share/aseprite/data/%s", filename);
+  includeBinDir(buf);
 #else
 
   // $HOME/.config/aseprite/filename
`

PR #3070, opened 2021-11-26
Add Dockerfile to make compilation easier for end-users.

bjorm:

Motivation:
I found it quite inconvenient to piece together all the invidual steps to get aseprite (and skia, I had linking problems with the pre-built binaries) compiled on linux. With this Dockerfile linux users can just build a container image and in the end copy the aseprite binary from a container without to much hassle and without messing up their systems. 

I am not too familiar with this ecosystem, otherwise I could have built it in such a way that the binary would be written to a local folder which could be mounted during the image build step. That way there would be no need to copy the file from a container.

Also I am not sure about whether the resulting binary could be run on any x86_64 linux?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,48 @@
+FROM fedora:35 as base
+
+ENV CC=/usr/bin/clang
+ENV CXX=/usr/bin/clang++
+ENV PATH="/tmp/depot_tools:${PATH}"
+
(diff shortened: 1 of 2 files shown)


bjorm:
This needs to be considered https://github.com/aseprite/aseprite/issues/1366

dacap:
Hi @bjorm, thanks for your PR, but at the moment there are no plans to include (and keep updated) a Dockerfile in the official repository.

PR #3071, opened 2021-11-26
Fix panning canvas out of view clips in Auto Guides (fix #2994)

### LINKED ISSUE #2994: Panning Canvas Out of View Clips Auto Guides
In the image below the line extending from the left edge of the layer edges (or cel image) should continue to the left edge of the sprite canvas with a label that says 48px.

diagnosticInfoCutoff

I'm using Aseprite v1.3-beta6-x64 on Windows 10.

To reproduce:
1. Move cursor to outside the layer edge (or cel image in scripting API terms).
2. Hold down `Ctrl` to display layer edges and auto guides.
3. Pan the view until the sprite canvas is partially out of view.
4. Pan the view so all of the sprite canvas is in view.

Thanks! Best,
Jeremy

iamOgunyinka:
This PR is meant to solve the issue detailed [here](https://github.com/aseprite/aseprite/issues/2994). I'll add further technical comment about this in few hours.

### DIFF
--- a/src/app/ui/editor/editor.cpp
+++ b/src/app/ui/editor/editor.cpp
@@ -923,10 +923,9 @@ void Editor::drawSpriteUnclippedRect(ui::Graphics* g, const gfx::Rect& _rc)
 
   // Draw active layer/cel edges
   if ((m_docPref.show.layerEdges() || m_showAutoCelGuides) &&
-      // Show layer edges only on "standby" like states where brush
-      // preview is shown (e.g. with this we avoid to showing the
-      // edges in states like DrawingState, etc.).
-      m_state->requireBrushPreview()) {
+      // Show layer edges and possibly cel guides only on states that
+      // allows it (e.g scrolling state)
+      m_state->allowLayerEdges()) {
     Cel* cel = (m_layer ? m_layer->cel(m_frame): nullptr);
     if (cel) {
       gfx::Color color = color_utils::color_for_ui(Preferences::instance().guides.layerEdgesColor());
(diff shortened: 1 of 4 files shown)


Gasparoken:
I review it and looks ok for me.

dacap:
I wasn't able to test this patch yet, but looking at it, it doesn't feel like `ScrollingState` requires the the brush preview, it might fix the original issue but just "by chance" (or maybe we need an extra concept for editor states to allow this fix). Some explanation about this fix would be great.

iamOgunyinka:
@dacap perhaps there's a need for a different kind of concept but here's what brought up this patch:

in the editor, we only draw a cel guide or layer edges when specified (in the document preference or the editor's `m_showAutoCelGuide`) **and** when the current state [requires a brush preview](https://github.com/aseprite/aseprite/blob/ff851157ccaadd66ef648c0f69bcddb7d2f62d02/src/app/ui/editor/editor.cpp#L925). However, the brush preview defaults to false in the scrolling state (rightly so?). So every time there's an update on the screen/editor via the scrolling state, if the drawn guide pans out of view, they're never redrawn when they're brought back into focus. Thus, the reason this bug comes up.

dacap:
Thanks for the explanation @iamOgunyinka, as you have noticed, it looks like I've already made the mistake in the past mixing the `requireBrushPreview()` with this "display edges/auto-guides" concept (there, were you've referenced and [here too](https://github.com/aseprite/aseprite/blob/e7575f237384b8afb2ed89e7dc2e7d9c45d9c05a/src/app/ui/editor/editor.cpp#L2823)).

Probably the only correct place where this member function is used is [here](https://github.com/aseprite/aseprite/blob/e7575f237384b8afb2ed89e7dc2e7d9c45d9c05a/src/app/ui/editor/brush_preview.cpp#L223) in `BrushPreview`, and we could add a new `EditorState::allowLayerEdges()` function similar to `requireBrushPreview()`.

dacap:
Cherry-picked in 31bf651d8b8284d84fed14ad9b4e1eeaa6df40eb

PR #3077, opened 2021-11-29
Fix allowing alpha channel if layer is converted from indexed (fix #3073)

### LINKED ISSUE #3073: Background Layer in RGB Mode Allows Alpha If Converted from Indexed
Hi,

This is a bug report. I'm using Aseprite v1.3beta7-x64 on Windows 10.

afterRgbConversion

To reproduce:
1. Create a new sprite that is set to indexed color mode.
2. Ensure that the palette contains at least one element with alpha channel less than 255.
3. Apply translucent or transparent pixels to the layer cel image with a brush, fill, etc.
4. Right click on the first and only layer in the timeline.
5. Select `Convert To ... Background` in the context menu.
6. Observe that the layer is underlined in the timeline, and that its name may be changed to "Background."
7. Go to `Sprite > Color Mode > RGB Color` to convert to RGB.

I observed this in version 1.2.30 as well. The user interface for converting a layer to background in steps 4 to 5 was slightly different: right click `Background from Layer`.

Thanks,
Jeremy

iamOgunyinka:
This PR attempts to fix #3073, a situation where alpha channel is still present in a background layer converted from indexed to RGB.

### DIFF
--- a/src/render/quantization.cpp
+++ b/src/render/quantization.cpp
@@ -329,8 +329,13 @@ Image* convert_pixel_format(
 
             if (!is_background && c == image->maskColor())
               *dst_it = rgba(0, 0, 0, 0);
-            else
-              *dst_it = palette->getEntry(c);
+            else {
+              const uint32_t p = palette->getEntry(c);
+              if (is_background)
+                *dst_it = rgba(rgba_getr(p), rgba_getg(p), rgba_getb(p), 255);
+              else
+                *dst_it = p;
+            }
           }
           ASSERT(dst_it == dst_end);
           break;


dacap:
> @dacap what do you think of this fix? Is it good enough or do you prefer a blend `bg_color <---> pixel color`?

Hi Joshua, what is the bg_color? Do you mean the [background color](https://www.aseprite.org/docs/color-bar/#background-color) in the color bar? If that is the case, no, I think the given solution is correct.

It would be nice to have a automated test in the `tests` repository to check this fix.

iamOgunyinka:
>  Do you mean the background color in the color bar? If that is the case, no.

I had a different understanding of this when I made [this comment](https://github.com/aseprite/aseprite/pull/3077#issuecomment-981784703) and a lot of changes would have to be made to implement what I had in mind. Please, ignore the second part now.

Side note: I'll make a PR to test this change, too.

dacap:
Hi @iamOgunyinka, was you able to create a tests for this PR?

iamOgunyinka:
@dacap yes, I opened a [PR in tests](https://github.com/aseprite/tests/pull/5).

dacap:
Thanks for the tests @iamOgunyinka and this patch, I've just [cherry-picked it](https://github.com/aseprite/aseprite/commit/f5678055914c984bf9d04464604edfb9ebcccbab) into `main` branch and modified the summary a little (the commit was pointing to issue 3078, when it was 3073)

PR #2727, opened 2021-05-17
Fixed issue #2030 (close all command doesn't show work about to close)

MorganDavid:
Changes active view before the dialog box appears for closing said view. #2030

PR #3089, opened 2021-12-06
Fix lost focus on closing tab (fix #3080)

### LINKED ISSUE #3080: Saving Sprite Close Prompt, Tab Focus
When you close Aseprite and have multiple documents open.
If they are not saved, Aseprite will prompt you with
`[Warning] Saving changes to the sprite "SpriteName" before quitting?`

Problem is, it doesn't change the tab focus to the document it's asking to close.
So you can see and make sure that you want to close the document.

* Aseprite version: 1.3-beta7, Steam/beta
* System: Windows 10 x64

iamOgunyinka:
When you close Aseprite and have multiple documents open. The expected behavior is to switch to each documents' tabs so user could see the "content" and be able to determine if that doc should be saved. This PR makes sure a tab is switched to before the closing of that tab.

### DIFF
--- a/src/app/commands/cmd_close_file.cpp
+++ b/src/app/commands/cmd_close_file.cpp
@@ -73,6 +73,7 @@ protected:
     }
 
     for (auto docView : docViews) {
+      workspace->setActiveView(docView);
       if (!workspace->closeView(docView, m_quitting))
         break;
     }


dacap:
I'll review this, but taking care of a [previous PR](https://github.com/aseprite/aseprite/pull/2727) too and compare both behaviors. More info [here](https://github.com/aseprite/aseprite/issues/3080#issuecomment-987059197).

dacap:
I've chosen the fix in #2727 with an extra patch (a2328a37935b261a3a78b0886e1f1d4c401fce34) because it works in both cases: Closing all files + closing just the non-active file (pressing the cross button of a tab). But both solutions were possible.

PR #3090, opened 2021-12-06
Fix bug when drawing on images with no parent cel (fix #3069)

### LINKED ISSUE #3069: [scripting] app.range.editableImages nil Cel error on drawImage
When using drawImage on an Image object from app.range.editableImages or app.range.images the following error occurs:
> Using a nil 'Cel' object

Demo script (run with an image open):
```
-- works
-- local images = {app.activeCel.image}

-- error: "Using a nil 'Cel' object" on drawImage
local images = app.range.editableImages
-- local images = app.range.images

for _,image in ipairs(images) do
    assert(image ~= nil)
    local work = Image(image)
    image:drawImage(work)
end
print("DONE!")
```

Code works fine when given app.activeCel.image instead.

* Aseprite version: v1.3-beta7-x64, Steam
* System: Windows 10 Pro, Version 10.0.19043 Build 19043

iamOgunyinka:
The check on some images that are standalone and are thus unattached to cels (e.g. they could be cloned from cel images) usually generate errors but this isn't the desired behavior. This PR _relaxes_ this restriction.

Technical detail: calling `ImageObj::cel` performs a value/type check on the bject the user is operating on and thus throw an exception in cases where a `Cel` object is expected. In these two cases though, we do not need to throw an exception since having no `Cel` is an expected behavior.

### DIFF
--- a/src/app/script/image_class.cpp
+++ b/src/app/script/image_class.cpp
@@ -231,7 +231,7 @@ int Image_drawImage(lua_State* L)
 
   // If the destination image is not related to a sprite, we just draw
   // the source image without undo information.
-  if (obj->cel(L) == nullptr) {
+  if (doc::get<doc::Cel>(obj->celId) == nullptr) {
     doc::copy_image(dst, src, pos.x, pos.y);
   }
   else {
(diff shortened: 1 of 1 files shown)


dacap:
Hi @iamOgunyinka, it looks like the issue is bigger than it looks. There are two problems:

1. `Images_index` is pushing raw `doc::Image` objects into the Lua stack instead of `ImageObj` (so this generates crashes when we try to use the given image)
2. Each pushed image (`ImageObj` now) must be pushed with the related `Cel::id` (instead of `Image::id`) from the current `app.range`

I've just pushed a fix for #3069: e7575f237384b8afb2ed89e7dc2e7d9c45d9c05a

PR #3096, opened 2021-12-14
Fix colorbar palette indicator when view is changed (fix #3082)

### LINKED ISSUE #3082: Colour Picker Indicators Vanish When Switching From Tileset-Active Document Workspaces to Non-Tileset Workspaces.
When the Tile Picker is active (document has tiles; tile-placement mode is active; the Tile Picker has focus versus the Colour Picker), the Tile Picker indicators (the small triangles in the top-left/bottom-right corners of Colour Picker and Tile Picker swatches) do not reappear in the Colour Picker when tabbing to a document Workspace *without* tiles. The indicators do not reappear when new Colours are selected by any means (Eyedropper, direct selection from the Colour Picker, hotkey, mousewheel).

The problem seems to be that the ASEprite UI has the 'Picker indicators "locked" to Tile Map/Placement mode, and when the software switches to a Workspace with no tiles, the indicators fail to return to the Colour Picker and vanish with the Tile Picker when it is hidden. All drawing tools continue to work normally, but there is no indication in the Colour Picker as to which Foreground and Background colours are currently active.

Switching to any Workspace with a Tilemap-enabled document causes the Tile Picker to reappear as normal, complete with indicators. Selecting a colour from the Colour Picker (which returns focus to that selector), then tabbing back to the document Workspace without tiles solves the problem (the indicators are once again visible in the Colour Picker).

This recording may be clearer. 🔽 
tile picker colour picker focus bug

..using ASEprite v1.3-b6 on Win10.

iamOgunyinka:
This PR fixes the vanishing palette indicator when views/tabs gets switched from documents in tilemap mode to documents that may or may not have tilemaps. Details of and comments on this bug that gave birth to this PR can be found [here](https://github.com/aseprite/aseprite/issues/3082).

### DIFF
--- a/src/app/ui/color_bar.cpp
+++ b/src/app/ui/color_bar.cpp
@@ -971,8 +971,13 @@ void ColorBar::onRemapTilesButtonClick()
 
 bool ColorBar::onIsPaletteViewActive(PaletteView* paletteView) const
 {
+  // As the m_tilemapMode is kept to restore it if we go back to a tilemap layer,
+  // there is a possibility where we are in a regular layer and the tilemap
+  // mode is a tile. So we need an extra check to see if the tilemap is editable.
+  // A palette view is active iff the tile is not editable.
   return
     (paletteView == &m_paletteView && m_tilemapMode == TilemapMode::Pixels) ||
+    (paletteView == &m_paletteView && m_tilemapMode == TilemapMode::Tiles && !canEditTiles()) ||
     (paletteView == &m_tilesView && m_tilemapMode == TilemapMode::Tiles);
 }
 


iamOgunyinka:
@dacap the concept behind this PR is that when we switch from a tile mode to a pixel mode, the global color bar stays in `TilemapMode::Tile` mode but the palette view is correctly switched. This check's logic is as thus: if
- we are in a view whose palette is pixel (`paletteView == &m_paletteView`) but
- the tilemap mode of the view is not pixel but tiles (`m_tilemapMode == TilemapMode::Tiles`) ... and
- we cannot edit this tile `!canEditTiles()`

then, we must be in a pixel mode. The pixel indicator in the color bar's palette is drawn *iff* the [palette view is active](https://github.com/aseprite/aseprite/blob/ff851157ccaadd66ef648c0f69bcddb7d2f62d02/src/app/ui/palette_view.cpp#L845).

dacap:
@iamOgunyinka [I know](https://github.com/aseprite/aseprite/issues/3082#issuecomment-993030046), I was telling that some information could be included as a comment in the code itself:
```c++
// As the m_tilemapMode is kept to restore it if we go back to a tilemap layer,
// there could be the possibility where we are in a regular layer and the tilemap
// mode.... etc.
```

dacap:
Cherry-picked with minor changes: https://github.com/aseprite/aseprite/commit/539e115b7dbdd080e65b33aa099572bc4c14d4c6

PR #3102, opened 2021-12-17
Fix spray painting behavior when mask color equals color (fix #3063)

### LINKED ISSUE #3063: Index mode spray does not handle masked color as expected
In indexed mode, when drawing with the masked color with the spray tool, on a non background layer, in "simple ink" mode, if the masked color does not have alpha 0, it either does not work or it attempts to draw another color.
See demonstration:

aseprite

Spray does draw a color (the one closest to the masked color) with right click if you have mask on right and random color on left, but not if you do that the other way around.

This behaviour is different from the pencil tool, the pencil tool behaves as I expect. Therefor the spray tool behaviour is unexpected.

* Aseprite version: v1.3-beta6-x64
* System: Windows 10

iamOgunyinka:
In indexed mode, when spraying (using spray tool, simple ink) a mask color on the canvas, the color is not drawn. This is not the expected behavior and this PR fixes that. In simple paint ink mode, when the trace policy "overlaps", we could use the faster `CopyInkProcessing`. How to reproduce this bug and comments about this is detailed [here](https://github.com/aseprite/aseprite/issues/3063).

Addendum(2021-12-20): I am in the process of writing a Lua script to test this PR.
Addendum(2021-12-21): Done writing the test script (ensuring the correctness of the inks). @dacap reviews needed.

### DIFF
--- a/src/app/tools/inks.h
+++ b/src/app/tools/inks.h
@@ -133,7 +133,9 @@ public:
               //
               // TODO this trace policy is configured in
               //      ToolLoopBase() ctor, is there a better place?
-              loop->getTracePolicy() != TracePolicy::Overlap) {
+              (loop->getTracePolicy() != TracePolicy::Overlap ||
+               (loop->getTracePolicy() == TracePolicy::Overlap &&
+                m_type == Simple))) {
             color_t color = loop->getPrimaryColor();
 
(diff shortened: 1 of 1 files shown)


dacap:
Closing this PR as #3063 was finally fixed with #3521

PR #3129, opened 2022-01-14
Fix crash resulting from deleting topmost layer via script (fix #3087)

### LINKED ISSUE #3087: Deleting a Layer via Script Allows User to Delete All Layers in Sprite
Hello,

This is a bug report. I was able to reproduce this in both v1.2.30 and v1.3beta-7 on Windows 10. A group layer cannot be deleted manually if as a result the sprite would contain no layers. However, with the line of code `app.activeSprite:deleteLayer(app.activeLayer)`, 

deleteAllLayers

the result is a sprite with no layers.

If the same script is run on a layer that is not a group layer, v1.2.30 crashes immediately. v1.3beta-7 shows the same as pictured above. v1.3beta-7 will crash if there is an attempt to add a new layer with `Shift+N`.

Thanks for looking into this,
Jeremy

iamOgunyinka:
Deleting the only and topmost layer in a sprite should not be possible -- this is the behavior when we use Aseprite UI but this can be done via scripting. This PR fixes this bug and the desired behavior is restored.

### DIFF
--- a/src/app/script/sprite_class.cpp
+++ b/src/app/script/sprite_class.cpp
@@ -375,6 +375,10 @@ int Sprite_deleteLayer(lua_State* L)
     }
   }
   if (layer) {
+    const auto& rootLayers = sprite->root()->layers();
+    if (rootLayers.size() == 1 && rootLayers[0] == layer)
+      return luaL_error(L, "You cannot delete all layers");
+
     Tx tx;
     tx(new cmd::RemoveLayer(layer));
     tx.commit();


Gasparoken:
@dacap, I looked at it and it looks fine to me. I tested and works OK.

dacap:
Sorry that I didn't review this yet, but I was thinking about the solution. It might be risky to change this behavior in the API at this moment (but not too risky). These are my thoughts and possible solutions:

1. There might be a user script that temporarily delete all layers and then add layers from scratch (merging this change will break such script). This is not a big deal, we could just break it and move forward.
2. We could create another "possible solution" that after a transaction/or the script execution, if the sprite doesn't contain layers, we add a dummy layer just to avoid crashes. I don't like this solution too much as in the middle of the script we could have a "broken" sprite without layers and some commands could crash the program anyway expecting a non-empty layers list.
3. Add support for sprites without any layer to the whole program (this requires an exhaustive check of all commands).

The last solution is the one I like the most (it would make the program a little more stable in extreme cases, e.g. what happen if we load a hand-made .aseprite file without layers). This mean that a sprite could be completely empty (without layers), and the program shouldn't crash. The internal logic of several widgets (mainly the Timeline) is not prepared for this case but it would be nice to support this case. (Even more, there are plans to add a button to add new layers in the Timeline it self, after the last layer, so with a sprite without layers we could see a button to add a new layer there.)

Still not sure if it worth the effort.

Gasparoken:
Thank you David!
I think 3 would be great, but imo it doesn't worth the effort. Perhaps in the future with other priorities.
The 2 sounds like a patch for me (adding this singularity for this specific case). This is valid, but I don't like it too.
I think 1 is the best cost/benefit ratio.

dacap:
I'll close this PR and take the issue from here. Thanks

Gasparoken:
Just one last comment. During my review, I noticed that it doesn't check if the deleted layer belongs to the associated sprite.

dacap:
This function is full of bugs 😅 

Would you like to take that issue @Gasparoken about deleting a layer that doesn't belong to the sprite? If `layer->sprite() != sprite` return a `luaL_error`, etc.

About this PR, I think I'll merge this patch next month as it is and then review the behavior for a near future. For example, at the UI level (and scripting level), just removing the last existing layer could delete it (without error) and we could insert a new empty layer. So deleting the last layer would generate a new one but completely empty (without cels, etc.). At the moment this is not possible in the current Timeline and it looks like could be a nice addition to just delete everything (e.g. selecting all layers and delete them).

Gasparoken:
> Would you like to take that issue @Gasparoken about deleting a layer that doesn't belong to the sprite? If `layer->sprite() != sprite` return a `luaL_error`, etc.

Yeah, @dacap ! No problem.

PR #3141, opened 2022-01-20
Fix bug that creates unlimited slice keys from existing slices (fix #3122)

### LINKED ISSUE #3122: Creating Slice On Later Frame Then Saving File Leads to "Unsupported chunk" warning.
Hello,

This is a bug report.
I encountered the bug when using Aseprite v1.3-beta7-x64 on Windows 10.

Here are the steps I took. I included some observations; many may not be relevant.
1. Open a new sprite with a transparent background, RGB color mode.
2. Add new frames via the '+' button in the bottom right corner.
3. Observe that, given a discontinuous layer, the intersection of the layer and frames creates new cels. Turning on layer edges indicate transparent images the size of the sprite.
4. Switch to the slice tool.
5. On the final frame, add a slice.
6. Switch to previous frames; observe that the slice bounds persists across earlier frames.
7. Try to change the bounds with the slice tool. This did not work for me.
8. Save the sprite with the `.aseprite` extension.
9. Close and reopen Aseprite.
10. Open the file.
11. Observe error message, "Warning: Unsupported chunk type 0 (skipping)," written twice in the console.

errorMessage

There are a few issues that I found after searching which are potentially related. I posted this issue because I didn't see any with steps for reproduction related to saving and reopening a sprite, though there was mention of corrupted files. https://github.com/aseprite/aseprite/issues/2285 , https://github.com/aseprite/aseprite/issues/2284 .

Thanks for looking into this.
Best,
Jeremy

iamOgunyinka:
Paragraph one is about the PR and paragraph 2 is how the PR relates to the [bug](https://github.com/aseprite/aseprite/issues/3122).

When you create a slice in a frame and edit it (by resizing or change its position) in another frame, the expected behavior is to _duplicate that slice_ in the current frame and allow the editing. Technically, it's simply [adding one more](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/doc/keyframes.h#L120-L126) `SliceKey` to the `Keyframe` of the slice. IIRC, the total number of keys in a single slice should be in the range `0 >= SliceKeys < TotalFramesInSprite`. However, without this fix, when we move a slice, created in `Frame 2`, around in `Frame 1`, it wouldn't work visually (as depicted in the bug report) but the underlying slice _engine_ creates tens/hundreds of keys in the slice because of how iterators for keyframes in designed.  By sorting the key frame for a slice after an insertion, it allows the littlest frames come first and make the [iterator getter](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/doc/keyframes.h#L166) work [perfectly](https://github.com/aseprite/aseprite/blob/main/src/doc/keyframes_tests.cpp).

When one saves the file, the slice to be saved saves incorrect [countKeys](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/app/file/ase_format.cpp#L1044) and only [one key](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/app/file/ase_format.cpp#L1052-L1057). The reader, on the other hand, tries to [read `countKeys`](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/dio/aseprite_decoder.cpp#L929) keys for the [slice](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/dio/aseprite_decoder.cpp#L205) and end up reading past the allotted space for slices -- thus, it encounters an [invalid chunk](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/dio/aseprite_decoder.cpp#L226).

**EDIT**: This is added as a late fix (thus second commit). Every time we do an [insertion](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/doc/slice.cpp#L64), this function creates a `new` object. If we have an object with the same frame number in the key frame, we [replace the object](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/doc/keyframes.h#L123) with the new value [without ever deleting](https://github.com/aseprite/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/doc/keyframes.h#L26) the old one! If we drag an existing slice around a few thousand times, that's a few thousand times we leaked some object.

### DIFF
--- a/src/doc/keyframes.h
+++ b/src/doc/keyframes.h
@@ -24,10 +24,14 @@ namespace doc {
       frame_t frame() const { return m_frame; }
       T* value() const { return m_value; }
       void setFrame(const frame_t frame) { m_frame = frame; }
-      void setValue(T* value) { m_value = value; }
+      void setValue(T* value) {
+        if (m_value)
+          delete m_value;
+        m_value = value;
+      }
     private:
-      frame_t m_frame;
-      T* m_value;
+      frame_t m_frame = -1;
+      T* m_value = nullptr;
     };
 
     typedef std::vector<Key> List;
--- a/src/doc/keyframes.h
+++ b/src/doc/keyframes.h
@@ -10,6 +10,7 @@
 
 #include "doc/frame.h"
 
+#include <algorithm>
 #include <vector>
 
 namespace doc {
@@ -124,6 +125,9 @@ namespace doc {
       else {
         ++it;
         m_keys.insert(it, Key(frame, value));
+        std::sort(begin(), end(), [](const auto& a, const auto& b) {
+          return a.frame() < b.frame();
+        });
       }
     }
 


iamOgunyinka:
@Gasparoken can you please help test this PR?

Gasparoken:
I was reviewing your code and works ok!🎉

While I was reviewing this fix, I discovered other issues:

1. The undo command doesn’t delete the m_keys elements, it just resize each slice key (it seems undo commands work, but isn't).
How to realize it:

- a.  New sprite with 5 frames
- b. New slice in frame 1
- c. Modify slice size on frame 3 (at this point we have keys on frame 1 and 3, i.e frames 1,2 looks the same, frames 3,4,5 looks the same)
- d. Do an undo, now all the frames seems the same (but the key on frame 1 and key on frame 3 still there)
- e. Go to frame 2 and change slice size 
- f. The expect behavior is: slices on frames 2, 3, 4, 5 should look the same slice rectangle. But the current behavior is (unexpected): the only different slice rectangle is on frames 2. Slice rectangles on 1, 3, 4, 5 look the same. This is because we have 3 keys now: frame 1, 2 and 3.

2. If we create a slice on frame 3 (in a sprite of 5 frames), save the file, close it, re-open it. The key isn’t in the frame 3 anymore, the key was created by the aseprite decoder on the frame 1. So, if we change the slice rectangle on frame 1, we will see the rectangle slice on frame 3 was changed too, this fact implies no slice key on frame 3.

I'll create new issues for all this if they don't already exist.

dacap:
Thanks @iamOgunyinka, I'm closing this PR now as I've used a different patch (for the leak using [unique_ptrs in Keyframes](https://github.com/aseprite/aseprite/commit/2100c45de20efe810dbfa353d6abff7a5f8fbe30), and for the specific keyframes issue [inserting in order](https://github.com/aseprite/aseprite/commit/2bb93247aafdb0190139e00c5d991a5a5346c17f) instead of an after `std::sort`). Anyway thanks for investigating about it, it was really helpful.

PR #3151, opened 2022-01-28
Fix moving rectangular marquee box while selecting color(fix #3143)

### LINKED ISSUE #3143: Alt clicking(EYEDROPPER) after selecting then moving something with arrow keys, moved the selection instead of selecting the color
Describe your bug report or feature request here:

Pretty much what it says in the title, I don't know if this is maybe some feature I am unaware of how to use or an actually bug so here I am. Order of operations > working on a file with enabled 18x18 grid, doubleclicking a gridspace to select it, moving it with shift+arrowkey to the right, then trying to use the eyedropper tool on it, moves the the selection vaguely in the southeastern direction(often off the canvas) then selects the mask as main color, instead of the expected functionality of not moving it at all then just selecting the color.

If moved by mouse then used eyedropper with ALT + LPM it will snap the selection to gridspace, if moved by mouse then used eyedropper by selecting the eyedropper tool by pressing I or clicking on the layout button it will instead move the image a few pixels in the direction based on where you clicked from the center of selection (Which seems doubly weird here since as I assume alt click and pressing I or the ui button are supposed to be the same functionally)

* Aseprite version: 1.2.32-x64 Steam
* System: Windows 10

iamOgunyinka:
Title says it all. This PR fixes a bug that moves the rectangular marquee box from its default location when we `Alt` to select a color from within or outside of the box.

### DIFF
--- a/src/app/ui/editor/moving_pixels_state.cpp
+++ b/src/app/ui/editor/moving_pixels_state.cpp
@@ -391,6 +391,9 @@ bool MovingPixelsState::onMouseMove(Editor* editor, MouseMessage* msg)
 void MovingPixelsState::onCommitMouseMove(Editor* editor,
                                           const gfx::PointF& spritePos)
 {
+  if (!m_pixelsMovement->isDragging())
+    return;
+
   m_pixelsMovement->setFastMode(true);
 
   // Get the customization for the pixels movement (snap to grid, angle snap, etc.).


Gasparoken:
Thank you Josh. I tested your solution, and works ok. I ran the `tests` and worked ok too. I think this should be merged asap.

dacap:
Merged in aa188d641b9b1f5ce366a2a4df0645aeea14f623

PR #3152, opened 2022-01-29
TIFF Reading Support

DevTechS:
Hello, this is my first contribution ever on Github, so please be nice. I am using TinyTiff to read a TIFF file. It should be pretty simple to add write functionality, but I wanted to make sure reading worked before adding too much complexity. I learned a lot while making this, so even if its not used, I'm really glad I worked on it. Thank you!

### DIFF
--- a/src/app/file/tiff_format.cpp
+++ b/src/app/file/tiff_format.cpp
@@ -22,6 +22,13 @@
 #include "doc/doc.h"
 #include "ui/window.h"
 
+#include "tinytiff_defs.h"
+#include "tinytiff_export.h"
+#include "tinytiff_tools.hxx"
+#include "tinytiffreader.hxx"
+#include "tinytiffwriter.h"
+
+
 namespace app {
 
 using namespace base;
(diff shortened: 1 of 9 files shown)


src/dio/detect_format.cpp
```diff
@@ -21,6 +21,7 @@
 #define GIF_89_STAMP     "GIF89a"
 #define PNG_MAGIC_DWORD1 0x474E5089
 #define PNG_MAGIC_DWORD2 0x0A1A0A0D
+//#define TIFF_MAGIC_DWORD 0x     //TODO implement byte check
```
ryandesign, line 24: [According to Wikipedia](https://en.wikipedia.org/wiki/List_of_file_signatures) there are two possible magic dwords (the first for little-endian images, the second for big-endian images):
```suggestion
#define TIFF_MAGIC_DWORD1 0x49492A00
#define TIFF_MAGIC_DWORD2 0x4D4D002A
```
And then you'll need something like this to actually detect the format based on those:
```diff
--- src/dio/detect_format.cpp.orig	2022-04-01 11:56:58.000000000 -0500
+++ src/dio/detect_format.cpp	2022-04-01 11:58:51.000000000 -0500
@@ -48,29 +48,35 @@
    (buf[offset+3] == ((dword & 0xff000000) >> 24)))
 
   if (n >= 2) {
-    if (n >= 6) {
-      if (n >= 8) {
-        if (n >= 12) {
-          if (std::strncmp((const char*)buf, WEBP_STAMP_1, 4) == 0 ||
-              std::strncmp((const char*)buf+8, WEBP_STAMP_2, 4) == 0)
-            return FileFormat::WEBP_ANIMATION;
+    if (n >= 4) {
+      if (n >= 6) {
+        if (n >= 8) {
+          if (n >= 12) {
+            if (std::strncmp((const char*)buf, WEBP_STAMP_1, 4) == 0 ||
+                std::strncmp((const char*)buf+8, WEBP_STAMP_2, 4) == 0)
+              return FileFormat::WEBP_ANIMATION;
+          }
+
+          if (IS_MAGIC_DWORD(0, PNG_MAGIC_DWORD1) &&
+              IS_MAGIC_DWORD(4, PNG_MAGIC_DWORD2))
+            return FileFormat::PNG_IMAGE;
         }
 
-        if (IS_MAGIC_DWORD(0, PNG_MAGIC_DWORD1) &&
-            IS_MAGIC_DWORD(4, PNG_MAGIC_DWORD2))
-          return FileFormat::PNG_IMAGE;
+        if (std::strncmp((const char*)buf, GIF_87_STAMP, 6) == 0 ||
+            std::strncmp((const char*)buf, GIF_89_STAMP, 6) == 0)
+          return FileFormat::GIF_ANIMATION;
+
+        if (IS_MAGIC_WORD(4, ASE_MAGIC_NUMBER))
+          return FileFormat::ASE_ANIMATION;
+
+        if (IS_MAGIC_WORD(4, FLI_MAGIC_NUMBER) ||
+            IS_MAGIC_WORD(4, FLC_MAGIC_NUMBER))
+          return FileFormat::FLIC_ANIMATION;
       }
 
-      if (std::strncmp((const char*)buf, GIF_87_STAMP, 6) == 0 ||
-          std::strncmp((const char*)buf, GIF_89_STAMP, 6) == 0)
-        return FileFormat::GIF_ANIMATION;
-
-      if (IS_MAGIC_WORD(4, ASE_MAGIC_NUMBER))
-        return FileFormat::ASE_ANIMATION;
-
-      if (IS_MAGIC_WORD(4, FLI_MAGIC_NUMBER) ||
-          IS_MAGIC_WORD(4, FLC_MAGIC_NUMBER))
-        return FileFormat::FLIC_ANIMATION;
+      if (IS_MAGIC_DWORD(0, TIFF_MAGIC_DWORD1) ||
+          IS_MAGIC_DWORD(0, TIFF_MAGIC_DWORD2))
+        return FileFormat::TIFF_IMAGE;
     }
 
     if (IS_MAGIC_WORD(0, BMP_MAGIC_NUMBER))
```

third_party/CMakeLists.txt
```diff
@@ -136,6 +136,7 @@ add_subdirectory(libarchive)
 target_include_directories(archive_static INTERFACE
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/libarchive/libarchive>)
 
+
```
ryandesign, line 139: The addition of this blank line is an example of a superfluous change that you should remove.

src/app/file/tiff_format.cpp
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    Image* image = fop->sequenceImage(pixelFormat, imageWidth, imageHeight);
+    if (!image) {
+      fop->setError("file_sequence_image %dx%d\n", imageWidth, imageHeight);
+      return false;
+    }
+
+    for (int y = 0; y < imageHeight; y++) {
+      //uint8_t* src_address = rows_pointer[y];
```
ryandesign, line 131: I suppose this commented-out line could be removed.
```suggestion
```

src/app/file/tiff_format.cpp
```diff
@@ -0,0 +1,179 @@
...
+FileFormat* CreateTiffFormat()
+{
+  return new TiffFormat;
+}
+
+bool TiffFormat::onLoad(FileOp* fop)
+{
+  PixelFormat pixelFormat;
+  int imageWidth = 100;
+  int imageHeight = 100;
```
ryandesign, line 79: Do these variables need to be initialized to 100? Seems arbitrary. The variables will be filled with the image dimensions after the image is opened, and it doesn't look like the code tries to use the variables if the image couldn't be opened.
```suggestion
  int imageWidth;
  int imageHeight;
```
In fact you could move the declaration of these variables to where they're set to the real values, since you already declare other variables there.

src/app/file/tiff_format.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+  //FileHandle handle(open_file_with_exception(fop->filename(), "rb"));
+  //FILE* file = handle.get();
+
+  TinyTIFFReaderFile* tiffr = NULL;
+  tiffr = TinyTIFFReader_open(fop->filename().c_str());
```
ryandesign, line 85: Does this need to be initialized to NULL? You're setting it to its real value on the very next line.
```suggestion
  TinyTIFFReaderFile* tiffr = TinyTIFFReader_open(fop->filename().c_str());
```

ryandesign (commented):
Just today I was trying unsuccessfully to work with a TIFF in Aseprite so thanks for working on this.

I'm not very familiar with the Aseprite code and I haven't tested the changes I'm proposing so take my suggestions with care. And I'm not an Aseprite developer; they may have a different take on things.

I see there are three commits in this PR, and I see some evidence of you adding debugging code and then removing it later. That history probably does not want to be preserved in the Aseprite main branch. To make reviewing the PR more straightforward, my recommendation would be to squash your commits down to one and then look at the diff closely to make sure every change you're proposing is necessary and relevant to this PR. I see some blank lines and comments added to unrelated files that I might recommend undoing.

Is there a particular reason why you implemented this with tinytiff (which I hadn't heard of before and which is not yet in the package manager I use) as opposed to the ubiquitous [libtiff](http://www.simplesystems.org/libtiff/)?

dacap:
Hi @DevTechS, thanks for this PR. I'm seeing that TinyTIFF is licensed under LGPL-3.0, so we cannot use it (as we link all libraries statically).

Thanks @ryandesign for the comments and reviews. Several of them valid. And I think we would prefer to use libtiff too: https://gitlab.com/libtiff/libtiff for this kind of work.

ryandesign:
I forgot to add that I looked into tinytiff a bit more, and it looks like its design goals aren't a good fit for aseprite. It was designed specifically for its author's unusual requirement of processing tiffs from his microscope that had hundreds of thousands of frames each; the tiffs users are likely to load into aseprite will almost certainly be single-frame images so tinytiff's performance advantage for multi-frame tiffs isn't relevant here. Meanwhile, tinytiff achieves some of its tininess by not implementing some common tiff features like image compression. Users will get tiffs from a variety of sources and may not be able to control whether or not they were stored in compressed format when they were created; if they were, they can't be read by tinytiff. So even if tinytiff's license were not a problem, I would suggest using the de facto standard of libtiff instead.

Rinalusi44:
> Halo, ini adalah kontribusi pertama saya di Github, jadi mohon berbaik hati. Saya menggunakan TinyTiff untuk membaca file TIFF. Seharusnya cukup mudah untuk menambahkan fungsionalitas menulis, tapi saya ingin memastikan membaca berfungsi sebelum menambahkan terlalu banyak kerumitan. Saya belajar banyak saat membuat ini, jadi meskipun tidak terpakai, saya sangat senang bisa mengerjakannya. Terima kasih!
YEs

PR #3160, opened 2022-02-03
Fix occasional crashes in despeckle/median filter (fix #2903)

### LINKED ISSUE #2903: Crash after modifying width and height in Despecke FX
This crash is frequent, but it doesn't always happen.
- New Sprite 100x100 RGBA
- draw something that takes up the whole canvas
- Edit > FX > Despeckle (median filter)
- Change to 100 width and height.
Crash (sometimes it doesn't crash)

I attach a crash report and a print screen of Xcode debugging
[aseprite_2021-08-19-145501_Gaspars-MacBook-Pro.crash.zip](https://github.com/aseprite/aseprite/files/7017251/aseprite_2021-08-19-145501_Gaspars-MacBook-Pro.crash.zip)
Screen Shot 2021-08-19 at 15 46 24

* Aseprite version: 1.3-beta5
* System: macOS
* Reported on support ticket 2575 and 2602 too

iamOgunyinka:
A data race in the `DespeckleWindow::onSizeChange` causes seldom crashes when the size of the filter is changed while a filter preview is currently running. The call to the filter's `setSize` causes [a change in the size](https://github.com/iamOgunyinka/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/filters/median_filter.cpp#L116) of the `m_channel` data [being used in the background](https://github.com/iamOgunyinka/aseprite/blob/9e23d31d84a5bfdccc32aea065ffb60b02b57856/src/filters/median_filter.cpp#L131-L178) from another thread.

This PR explicitly request that the background thread be stopped before the size of the filter is changed. This may cause a little but noticeable lapse in the despeckle window but it brings stability to its behavior.

### DIFF
--- a/src/app/commands/filters/cmd_despeckle.cpp
+++ b/src/app/commands/filters/cmd_despeckle.cpp
@@ -81,6 +81,11 @@ private:
     newSize.w = base::clamp(newSize.w, 1, 100);
     newSize.h = base::clamp(newSize.h, 1, 100);
 
+    // if we had a previous filter preview running in the background, we
+    // explicitly request it be stopped. Otherwise, changing the size of
+    // the filter would cause a race condition on the filter's `m_channel`
+    requestPreviewStop();
+
     m_filter.setSize(newSize.w, newSize.h);
     restartPreview();
   }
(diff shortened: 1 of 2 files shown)


dacap:
Nice catch @iamOgunyinka, I'll merge this one and try to use a similar approach for other filters which I think might have similar issues.

dacap:
Merged 811b8933209b349af6fabc5a74088f24bb8431db

PR #3165, opened 2022-02-06
Add BMP v5 support.

paulguy:
Adds alpha channel support to BMP export.


paulguy:
Just noticed the file conflict.  I'm way too sick to fix it at the moment but when I get better it should be easy to fix, it just might be a few weeks.  Just leaving this here so this doesn't get closed as abandoned or something.

paulguy:
I've had a look at it and it looks like this has been obsoleted by the current code which more completely implements the BMP format overall so I can probably close this now.

PR #3175, opened 2022-02-15
When eyedropper tool picks a pixel of mask color, the FG color changes to the entry 0 (fix #2813)

### LINKED ISSUE #2813: When eyedropper tool picks a pixel of mask color, the FG color changes to the entry 0

Gasparoken:
Updated PR of https://github.com/aseprite/aseprite/pull/2816


src/app/ui/palette_view.cpp
```diff
@@ -971,8 +971,12 @@ int PaletteView::findExactIndex(const app::Color& color) const
 {
   switch (color.getType()) {
 
-    case Color::MaskType:
-      return (current_editor ? current_editor->sprite()->transparentColor(): -1);
+    case Color::MaskType: {
+      if (color.getType() == Color::IndexType)
```
dacap, line 975: It looks like this condition will be always `false` because the switch-case already said that `color.getType()` is `MaskType`.

Gasparoken: Yeah, you are right. I'll force a push soon.
The initial intention was to know if the mask color was selected on Indexed Color Mode. Perhaps I can add a type `Color::TransparentIndex` to identify this case. Anyway I will test it again.

Gasparoken:
@dacap this is ready to review.

PR #3180, opened 2022-02-17
Fix ColorSpace error in PNG files (fix #3170)

### LINKED ISSUE #3170: Error Saving .jpg converted From Gray to RGB Color as .png
Hello,

This is a bug report. I encountered it in version 1.3-beta11-x64 on Windows 10.

Steps to reproduce:

1. Open a .jpg image that is in grayscale color mode.
2. Either
a. Run a script with the code `app.command.ChangePixelFormat { format = "rgb" }`.
b. Go to `Sprite > Color Mode > RGB Color`.
3. Go to `File > Export`.
4. Save as a .png.
5. Observe error message: "libpng: profile 'Custom Profile': 'GRAY': Gray color space not permitted on RGB PNG."

grayCap

Test image was from [here](http://planetpixelemporium.com/earth.html).

Thank you.
Jeremy

iamOgunyinka:
[Only for PNG files] This PR is quite a suggestion and may be supported by a flag/option in the preference window. When a picture with a custom color profile is converted to a different color mode and is exported, the exporting format may not support the custom profile and thus fails. For example, in cases where we have an image with an iCCP of gray "profile", if we convert the sprite to RGB and try exporting it as a PNG file, it may fail as the PNG encoder doesn't recognize the profile.

Instead of failing, this PR allows warns the user of the problem but still go-ahead and save the exported image.

### DIFF
--- a/src/app/file/png_format.cpp
+++ b/src/app/file/png_format.cpp
@@ -608,8 +608,24 @@ bool PngFormat::onSave(FileOp* fop)
   }
 
   if (fop->preserveColorProfile() &&
-      fop->document()->sprite()->colorSpace())
-    saveColorSpace(png, info, fop->document()->sprite()->colorSpace().get());
+      fop->document()->sprite()->colorSpace()) {
+    auto newCallback = +[](png_structp png, png_const_charp err) {
+      throw std::runtime_error(err);
+    };
+
+    // set a temporary callback that reports the color space error as warning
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @iamOgunyinka for this. We need a different approach to solve this issue at the moment (without setting the error handling function). Closing this PR as we'll try to open another one in the future to fix this.

PR #3193, opened 2022-03-04
Fix multi layer movement in tilemap mode (fix #3144)

### LINKED ISSUE #3144: Multiple tileset layers  selection move is broken
Hi all 👋 

I just found out that moving a selection in tileset mode having multiple tileset layers selected leads to loss of information. Only the last selected layers is moved, all the others are cleared (in the selection).

See an example here:
Screen Recording 2022-01-21 at 13 38 42

* Aseprite version: v1.3-beta10-dev compiled from the release archive
* System: macOS 12.1

Gasparoken:
Prior this fix, a multi layer mask movement (with mixed layer types: regular layer and tilemap layers with different grids) resulted in loss of drawing areas.

The hearth of this fix relies in a correct align of selection mask and transform data according the layer's grid, and also, forcing the 'site' TilemapMode/TilesetMode before each reproduceAllTransformationsWithInnerCmds() iteration.

Scale transform can be some things to improve in later commits. Not fully tested yet.


dacap:
I didn't review this yet, maybe next week.

Anyway it would be nice if we can start working (or thinking) in some Lua API to transform the selection (`app.transform`?). It was [requested before](https://github.com/aseprite/api/issues/15) and it might be useful for testing purposes in the [tests](https://github.com/aseprite/tests/tree/main/scripts) repo. This is not to be included in this PR.

The main problem for a Lua API to transform the selection is being able to do the same an `app::Editor` can do to the selection without UI (e.g. when `!ENABLE_UI` or when we run in `-batch` mode). Here we can think of some possibilities:

1. Split the transformation logic from the Editor to some UI-agnostic module (too much refactoring)
2. Being able to use an `app::Editor` "in the shadows" without a window/UI at all (this may be the easiest approach)

Anyway we can think about this after this PR is merged. No hurries.

PR #3200, opened 2022-03-14
No longer add -stdlib=libc++ in CMakeLists.txt

ryandesign:
Adding `-stdlib=libc++` was needed when 10.7 was being targeted because on 10.8 and earlier libstdc++ is the default. Now that the target is 10.9 specifying `-stdlib=libc++` is unnecessary since libc++ is the default on 10.9 and later.

-----


dacap:
Thanks @ryandesign, I've just cherry-picked your change be7c26da5eb29dd50705babc4277cb8ebce9af91

Issue #3202, opened 2022-03-15
Pivot point moves itself


turesiht:
When selecting the area and then clicking inside selection or moving it, the pivot point moves away for no reason

Below are imgur links to screenshots in the followng order:
1) Selected the area
2) Moved it and pivot point moved itself too
3) Rotated the are to show how it behaves itself with the pivot point off

https://imgur.com/gmZ88e5
https://imgur.com/rNrv0Qy
https://imgur.com/WvG0JBs

Aseprite Version: v1.3-beta13 on Steam

OS (sorry, I can only copy this in russian):
Выпуск	Windows 10 Домашняя для одного языка
Версия	21H1
Дата установки	‎02.‎02.‎2022
Сборка ОС	19043.1586
Взаимодействие	Windows Feature Experience Pack 120.2212.4170.0

turesiht:
After testing around a bit I found that when first rotating the selected area and then moving it results in jumping of that area around

Gasparoken:
Thanks for this report @turesiht. I was able to reproduce it. It'll solved soon.

dacap:
I think it's related: https://steamcommunity.com/app/431730/discussions/2/4658416565943773518/

dacap:
Also reported here: https://steamcommunity.com/app/431730/discussions/2/4658416565944248871/

@Gasparoken tell me how it goes the progress about it, I can test it too.

PR #3205, opened 2022-03-15
Fix multi layer movement in tilemap mode (fix #3144)

### LINKED ISSUE #3144: Multiple tileset layers  selection move is broken

Gasparoken:
Before this fix, a multi-layer mask move (with mixed layer types: normal layer and tilemap layers with different grids) caused loss of drawing areas.
The heart of this solution is to correctly align the 'selection mask' and 'transform data' according to the layer's grid, and also, forcing 'site' TilemapMode/TilesetMode before each reproduceAllTransformationsWithInnerCmds() iteration.

Additionally:
Fix scaling a tilemap produces unexpected cel position and unwanted transform size.
Before this fix, it can be reproduced:
. Make a selection in Pixels Mode on a tilemap layer
. The selection bounds doesn't match with the tilemap grid AND some tiles are partially selected.
. Switch to Tilemap Mode
. Scale it with a Handle
. Result: unexpected cel position and unwanted transform size.

fix #3144

### DIFF
--- a/src/app/ui/editor/pixels_movement.cpp
+++ b/src/app/ui/editor/pixels_movement.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2021  Igara Studio S.A.
+// Copyright (C) 2019-2022  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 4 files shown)


dacap:
As a side note this was merged in bc050e8f1 and then reverted c08dc4490 because introduced the regression #3202

PR #3216, opened 2022-03-20
Support saving without showing a dialog & adding to Recent Files

VelocityRa:
My use case for these two changes is improving the "Automatic Snapshot" feature here [https://sprngr.itch.io/aseprite-record](https://t.co/qGqx88tG9A) for recording timelapses.
Specifically making it 1) not display a distracting 'saving' alert window on every snapshot and 2) not make the snapshots appear on the Recent Files list.

Wasn't sure if I should make `sprite:saveAs`/`saveCopyAs` have named params instead since that would break the API afaik. Instead I exposed the 'add to recent files' param as an optional second bool. The other one (whether to display the alert) is set to always not display (when called via scripting - I used the existing `useUI` bool).

TODO: (if merged) Update docs on https://github.com/aseprite/api.

---

I tried to contact maintainers on `#aseprite-dev` on Discord and Twitter so as to ask for feedback and not waste my time with a pre-emptive PR, but received no response.

---

### DIFF
--- a/src/app/commands/cmd_save_file.cpp
+++ b/src/app/commands/cmd_save_file.cpp
@@ -47,13 +47,13 @@ namespace app {
 
 class SaveFileJob : public Job, public IFileOpProgress {
 public:
-  SaveFileJob(FileOp* fop)
-    : Job("Saving file")
+  SaveFileJob(FileOp* fop, const bool showAlertWindow = true)
+    : Job("Saving file", showAlertWindow)
     , m_fop(fop)
   {
   }
 
-  void showProgressWindow() {
+  void run() {
     startJob();
 
     if (isCanceled()) {
@@ -112,7 +112,7 @@ void SaveFileBaseCommand::onLoadParams(const Params& params)
     m_adjustFramesByTag = false;
   }
 
-  std::string useUI = params.get("useUI");
+  const std::string useUI = params.get("useUI");
   m_useUI = (useUI.empty() || (useUI == "true"));
 
   m_ignoreEmpty = params.get_as<bool>("ignoreEmpty");
@@ -172,8 +172,7 @@ std::string SaveFileBaseCommand::saveAsDialog(
 
   if (saveInBackground) {
     saveDocumentInBackground(
-      context, document,
-      filename, markAsSaved);
+      context, document, filename, markAsSaved, m_useUI);
 
     // Reset the "saveCopy" document preferences of the new document
     // (here "document" contains the new filename), because these
@@ -198,7 +197,8 @@ void SaveFileBaseCommand::saveDocumentInBackground(
   const Context* context,
   Doc* document,
   const std::string& filename,
-  const bool markAsSaved)
+  const bool markAsSaved,
+  const bool showAlertWindow)
 {
   if (!m_aniDir.empty()) {
     switch (convert_string_to_anidir(m_aniDir)) {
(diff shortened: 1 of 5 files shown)


VelocityRa:
Bump?
It's been 9 days...

VelocityRa:
Everything about how the repo is set up (detailed building instructions, nice code/tree documentation, etc) suggests that you actually _want_ people contributing.

Not to mention that you require contributors to give you their _home address_ for the privilege of being able to contribute (talking about the CLA thing). And then you ignore their PRs.

VelocityRa:
I don't intend to be rude; hopefully you understand my frustration.

dacap:
Hi @VelocityRa, sorry but sometimes priorities are focused in other things and I cannot review the PRs. I know that in a perfect world PRs will be reviewed immediately and merged if everything is fine. I also know that there are a lot of old PRs. But my time is limited and I'm already giving too much of my personal life in this project.

dacap:
I have just added some information about the current CLA process https://github.com/aseprite/sourcecode/issues/83

I'll review this PR just right now and merge it, as it LGTM

src/app/commands/cmd_save_file.cpp
```diff
@@ -112,9 +112,12 @@ void SaveFileBaseCommand::onLoadParams(const Params& params)
     m_adjustFramesByTag = false;
   }
 
-  std::string useUI = params.get("useUI");
+  const std::string useUI = params.get("useUI");
   m_useUI = (useUI.empty() || (useUI == "true"));
 
+  const std::string addToRecentFiles = params.get("addToRecentFiles");
+  m_addToRecentFiles = (addToRecentFiles == "true");
```
dacap, line 119: I think I would prefer a shorter param name like "addToRecent", and check if it's empty (in that case we add to recent, as it's the default behavior):
```suggestion
  const std::string addToRecent = params.get("addToRecent");
  m_addToRecent = (addToRecent.empty() || (addToRecent == "true"));
```
In a future we have to change the parameters of all commands to NewParams, but that is a refactor for the future.

src/app/script/sprite_class.cpp
```diff
@@ -218,6 +218,11 @@ int Sprite_saveAs_base(lua_State* L, std::string& absFn)
     Params params;
     params.set("filename", absFn.c_str());
     params.set("useUI", "false");
+
+    if (!lua_isnone(L, 2))
+      params.set("addToRecentFiles", lua_toboolean(L, -1) ? "true" : "false");
```
dacap, line 223:
```suggestion
      params.set("addToRecent", lua_toboolean(L, -1) ? "true" : "false");
```

dacap (changes requested):
Some minor changes requested to keep the default behavior (addToRecent) when the parameter is not specified.

dacap:
Oh sorry, didn't check the comment about the new parameter in the Lua API. I would prefer named parameters in this case (as we don't know what the boolean means from the client-side).

Basically we should be able to call Sprite:saveAs(string) or Sprite:saveAs{table} ([app.alert](https://github.com/aseprite/api/blob/main/api/app.md#appalert) does something similar)

dacap:
This was implemented with 0cc1161d6494bdb0f06b6a3c470055ec5f952939 and 3ea0437e1dd92b939ff103e5e775397f345cfa52

PR #3364, opened 2022-06-07
Add '{title}' functionality to the Export dialog

Gasparoken:
Now, {title} keyword is an alias of the sprite name in the filename field on the Export dialog. Partial solution to issue 3363.

### DIFF
--- a/src/app/file/file.cpp
+++ b/src/app/file/file.cpp
@@ -31,6 +31,7 @@
 #include "app/ui/status_bar.h"
 #include "base/fs.h"
 #include "base/mutex.h"
+#include "base/replace_string.h"
 #include "base/scoped_lock.h"
 #include "base/string.h"
 #include "dio/detect_format.h"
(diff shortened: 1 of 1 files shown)


src/app/file/file.cpp
```diff
@@ -529,6 +530,10 @@ FileOp* FileOp::createSaveDocumentOperation(const Context* context,
     Sprite* spr = fop->m_document->sprite();
     frame_t outputFrame = 0;
 
+    std::string newfn = base::get_file_name(fn);
+    base::replace_string(newfn, "{title}", base::get_file_title(fop->m_document->name()));
```
dacap, line 534: The `filename_formatter()` function (called below) already should do this, so the source of this problem is from the Export command (`SaveFileCopyAsCommand::onExecute`) which calls `createSaveDocumentOperation()` with `filename`=the string that the user inserted in the output filename. Maybe, when the user put some text like `{title}` (or any other {...}) the filename should be the original sprite name, and its `SaveFileBaseCommand::m_filenameFormat` should be equal to the inserted text, which now specifies a format, and not the final filename).

Issue #3394, opened 2022-06-24
Octree mapping review and improvements


Gasparoken:
__Background:__

Octree mapping is used to generate a well-known color organization: by knowing the RGB number a color represents, we know where to look in the color map without having to iteratively search one by one. The Octree mapping is better than a simple listing of consecutive indices since it would take a lot of memory to achieve the same thing without data loss.

Octree mapping is also used to accomplish another function: creating a suggested color palette by the sprite itself. In this case octree mapping is used as an intermediate step to achieve the final color palette.

__Why is needed to keep on memory a color map in INDEXED color mode?__

When pasting a new "source" RGBA image (for example) into a "destination" INDEXED sprite. A quantization process must be run on the entire “source” image:
Iteratively each pixel of "source" RGBA image is converted to an index according to its color by looking at the octree mapping. Currently, if the color does NOT belong to the palette of the “destination” sprite, the mapping will return an index that was generated from approximations in doc::OctreeNode::fillOrphansNodes function based on the "destination" palette, this is thought this way for speed performance reasons.
However, it would be good to venture NOT to fill the empty spaces of the mapping with fillOrphansNodes, and see if it is convenient to add each new color of the “source” sprite to the octree mapping. This will slow down the conversion but it will get a better quantification of the “source” image.

__Tasks:__

- [ ] 1. Improve color quantification (change the color distance criteria inside findBestfit2() function by other proven color distances  like [LinearRGB, CIELAB, CIEXYZ](https://community.aseprite.org/t/octree-color-indexed-conversion-testing/9682)), see about adding different quantization methods (not present in the UI yet). Related: https://github.com/aseprite/aseprite/issues/2787, 

- [ ] 2. Discard the doc::OctreeNode::fillOrphansNodes implementation of filling empty branches of the octree. Related: https://github.com/aseprite/aseprite/issues/3337.

- [ ] 3. Implement in UI the selection of color quantification methods (Preferences > Experimental > Quantization method: Linear RGB, CIELAB, CIEXYZ). Related: https://github.com/aseprite/aseprite/issues/2787

- [ ] 4. Check that the conversion preview works correctly during the Sprite > Color Mode > More Options… dialog when an RGB>>INDEXED conversion is wanted.

PR #3395, opened 2022-06-24
Add other color comparison criteria to findBestfit

Gasparoken:
Implementation of item 1 of https://github.com/aseprite/aseprite/issues/3394

### DIFF
--- a/src/doc/palette.cpp
+++ b/src/doc/palette.cpp
@@ -20,10 +20,19 @@
 
 #include <algorithm>
 #include <limits>
+#include <cmath>
 
 namespace doc {
 
 using namespace gfx;
+
+enum class FitCriteria {
+  OLD,
+  RGB,
+  linearizedRGB,
+  CIEXYZ,
+  CIELAB
+};
 
 Palette::Palette()
   : Palette(0, 256)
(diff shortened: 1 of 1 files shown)


src/doc/palette.cpp
```diff
@@ -400,6 +408,117 @@ int Palette::findBestfit2(int r, int g, int b, int a) const
   return bestfit;
 }
 
+// Auxiliary function for rgbToOtherSpace()
+double f(double t)
```
dacap, line 412: Add static keyword:
```suggestion
static double f(double t)
```

src/doc/palette.cpp
```diff
@@ -400,6 +408,117 @@ int Palette::findBestfit2(int r, int g, int b, int a) const
...
+{
+  if (t > 0.00885645171)
+    return pow(t, 0.3333333333333333);
```
dacap, line 415: Use `std::` namespace for standard functions. And include `<cmath>`:
```suggestion
    return std::pow(t, 0.3333333333333333);
```

src/doc/palette.cpp
```diff
@@ -400,6 +408,117 @@ int Palette::findBestfit2(int r, int g, int b, int a) const
...
+  else
+    return (t / 0.12841855 + 0.137931034);
+}
+
+// Auxiliary function for findBestfit3()
+void rgbToOtherSpace(double& r, double& g, double& b, FitCriteria fc)
```
dacap, line 421:
```suggestion
static void rgbToOtherSpace(double& r, double& g, double& b, FitCriteria fc)
```

src/doc/palette.cpp
```diff
@@ -400,6 +408,117 @@ int Palette::findBestfit2(int r, int g, int b, int a) const
...
+    Rl = pow((r + 0.055) / 1.055, 2.4);
+  if (g <= 0.04045)
+    Gl = g / 12.92;
+  else
+    Gl = pow((g + 0.055) / 1.055, 2.4);
+  if (b <= 0.04045)
+    Bl = b / 12.92;
+  else
+    Bl = pow((b + 0.055) / 1.055, 2.4);
+  if (fc == linearizedRGB) {
```
dacap, line 442: I think we can use a switch-case here:
```suggestion
  switch (fc) {
    case FitCriteria::linearizedRGB:
      ...
      break;
  }  
```

src/doc/palette.cpp
```diff
@@ -25,6 +25,14 @@ namespace doc {
 
 using namespace gfx;
 
+enum FitCriteria{
```
dacap, line 28: Prefer a `enum class` for global enums (enums outside classes):
```suggestion
enum class FitCriteria {
```

src/doc/palette.h
```diff
@@ -102,6 +102,7 @@ namespace doc {
     bool findExactMatch(color_t color) const;
     int findBestfit(int r, int g, int b, int a, int mask_index) const;
     int findBestfit2(int r, int g, int b, int a) const;
+    int findBestfit3(int r, int g, int b, int a, int mask_index) const;
```
dacap, line 105: Do we need findBestfit()? Or can we just change the implementation of findBestfit() directly with the new findBest3 code?

dacap (changes requested):

Gasparoken:
Changes just pushed.

dacap:
Cherry-picked https://github.com/aseprite/aseprite/commit/54443ad20d0f0c018704d5f201ca34d39443d61e

PR #3400, opened 2022-06-27
Clear the exported images with its mask color instead of 0 (fix #3391)

### LINKED ISSUE #3391: Export Sprite Sheet with Spacing will fill borders with index 0 instead of the real mask color index
This bug appears when:

1. The file being exported is indexed mode (e.g. a GIF file)
1. The sprite is transparent (to background/opaque layer)
1. The transparent color in the file is not index 0 (and index 0 is an opaque color)
1. The File > Export Sprite Sheet action will use index 0 to clear borders instead of the transparent index (26).

Reported here: https://community.aseprite.org/t/exporting-spritesheet-transparent-spacing/14774
Example file in: https://igarastudio.zendesk.com/agent/tickets/3678

martincapello:


dacap:
Merged 306e13dfad8a0b6b4bee5c07c3b04c2626ca1153

Issue #3207, opened 2022-03-16
Conversion RGB to INDEXED color mode always picks index 0 as transparent color


Gasparoken:
Even if mask color is present in the palette:

<img width="616" alt="Screen Shot 2022-03-16 at 10 36 37" src="https://user-images.githubusercontent.com/32165475/158654277-d7cda97c-64f6-41eb-936e-42b81d453f49.png">

In case of mask color isn't present the default behavior is:
<img width="608" alt="Screen Shot 2022-03-16 at 14 51 56" src="https://user-images.githubusercontent.com/32165475/158655483-0e1d8d7e-869b-4789-bf10-13feef3fe514.png">

Proposal behavior of INDEXED conversion:
- The editor must review if the mask color is present in the palette:
...      **Yes**: transparent color index will be defined.
...      **No**: add the mask color at the beginning of the palette and define transparent color as index 0.

I think this option is better because the drawing is not modified in any case and is more intuitive (for me). It's an automatic way to do '2' on this comment: https://github.com/aseprite/aseprite/issues/2178#issuecomment-541745936

Please give your impression 👍 or 👎 to this alternative behavior.

* Aseprite version: 1.2.34.1
* System: MacOS

Gasparoken:
Proposal changes just pushed.

Pending to add the mask color at the beginning of the palette only if the palette has color count <256, as default: define transparent color as index 0.

PR #3414, opened 2022-07-05
Removed fillOrphanNodes, now new colors are added to the octree color arragement.

Gasparoken:
This gives more accuracy in the color picking criteria on new images pasted into an INDEXED sprite.

Also, added findMaskColor to fix the behavior reported: aseprite/aseprite#3207

Both issues are related when RGBMAP is created. The 'mask color' and 'mask index' must be defined correctly to include/exclude during the table/octree map color search.

To do: When converting sprite to INDEXED, it's needed to add 'mask color' to palette when number of colors is < 256 and transparent color is not in palette (also recommended in aseprite/aseprite#3207).


dacap:
Cherry-picked https://github.com/aseprite/aseprite/commit/2785a9fef71ff8aa45aee086b9d30f8a8805f67f

PR #3425, opened 2022-07-13
Remember new tilemap layer base index & grid width/height (fix #2735)

### LINKED ISSUE #2735: v1.3 beta1 - Remember new tilemap layer base index
In v1.3 beta1 I need to change the tile base index from 1 to 0 every time I create a new tilemap layer.

It'd be nice if either the new tilemap layer dialog remembered the last used value or there was a preferences setting to override the default value.

Gasparoken:


Gasparoken:
I would like to add the option

`<option id="grid" type="doc::Grid" default="doc::Grid()" />`

instead of
```
<option id="grid_width" type="int" default="16" />
<option id="grid_height" type="int" default="16" />
```
inside `<section id="tileset">` in `pref.xml` but I couldn't because of 'int<->doc::Grid' conversion errors that I haven't been able to determine yet.

src/app/commands/cmd_new_layer.cpp
```diff
@@ -206,6 +207,10 @@ void NewLayerCommand::onExecute(Context* context)
   tilesetInfo.baseIndex = 1;
 
 #ifdef ENABLE_UI
+  auto& pref = Preferences::instance();
+  tilesetInfo.grid = doc::Grid(gfx::Size(pref.tileset.gridWidth(),
+                                         pref.tileset.gridHeight()));
```
dacap, line 212: I think we shouldn't save the grid width/height in preferences, as the current grid settings are used anyway (see `tilesetInfo.grid = context->activeSite().grid();` above) and I think we prefer that one as the default.

src/app/commands/cmd_new_layer.cpp
```diff
@@ -227,6 +232,11 @@ void NewLayerCommand::onExecute(Context* context)
     if (window.closer() != window.ok())
       return;
 
+    pref.tileset.gridWidth(tilesetSelector->getInfo().grid.tileSize().w);
+    pref.tileset.gridHeight(tilesetSelector->getInfo().grid.tileSize().h);
+    pref.tileset.baseIndex(tilesetSelector->getInfo().baseIndex);
+    pref.tileset.save();
```
dacap, line 238: There is no need to call save() immediately here (`Preferences::save()` is called in `App::~App()` anyway and is the real function that saves to disk -or should be-)
```suggestion
```

dacap:
I'm going to recycle this PR (amend the commit) to save the tile index only to fix #2735, to be merged soon

dacap:
Merged https://github.com/aseprite/aseprite/commit/12e59b887f6d8e387a1d9d01662822a44efae87a :+1:

PR #3427, opened 2022-07-15
Fix uninstalling (or updating) an extensions that has subdirectories results in an error (fix #3385)

### LINKED ISSUE #3385: Uninstalling (or updating) an extensions that has subdirectories results in an error
Uninstalling (or updating, I'm guessing it's an uninstall+install combo action) an extension that has subdirectories in its directory results in an error shown below.

image

It leaves the extension directory with empty subdirectories, there are zero files left.

* Aseprite version: 1.2.35-x64
* System: Windows 11

Gasparoken:
Prior to this fix, any additional external files or directories within the Extension's directory tree, break the subdirectory removal iteration, resulting in empty subdirectories and a console error.

### DIFF
--- a/src/app/extensions.cpp
+++ b/src/app/extensions.cpp
@@ -41,6 +41,8 @@
 #include <sstream>
 #include <string>
 
+#include "base/log.h"
+
 namespace app {
 
 namespace {
@@ -368,7 +370,6 @@ void Extension::uninstall(const DeletePluginPref delPref)
 
   // Remove all files inside the extension path
   uninstallFiles(m_path, delPref);
-  ASSERT(!base::is_directory(m_path) || delPref == DeletePluginPref::kNo);
 
   m_isEnabled = false;
   m_isInstalled = false;
@@ -422,7 +423,13 @@ void Extension::uninstallFiles(const std::string& path,
 
   for (const auto& dir : installedDirs) {
     TRACE("EXT: Deleting directory '%s'\n", dir.c_str());
-    base::remove_directory(dir);
+    try {
+      base::remove_directory(dir);
+    }
+    catch (const std::exception& ex) {
+      LOG(ERROR, "RECO: Extension subdirectory cannot be removed, it's not empty.\n"
+                 "      Error: %s\n", ex.what());
+    }
   }
 
   // Delete __info.json file if it does exist (e.g. maybe the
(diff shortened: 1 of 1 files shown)


Gasparoken:
Sorry, I need to do a modification.

martincapello (commented):
Left a couple of minor comments.
I noted that you are missing the #include "base/log.h" as the last include.

Also I feel that it would be better to let the user know that the extension's directory couldn't be deleted for some reason. So it actually makes sense to me to let the exception message be displayed in the console. Or we could improve the message to give some hints on how to fix the problem. @dacap what do you think about this?

dacap:
As [I wasn't able to reproduce the problem](https://github.com/aseprite/aseprite/issues/3385#issuecomment-1205608163), I left this PR open, not sure if we should merge it. Maybe we can close it and left it for the future in case some specific steps can be reproduce it (and probably show a dialog to warn the user what unexpected files are going to be deleted).

src/app/extensions.cpp
```diff
@@ -422,7 +421,14 @@ void Extension::uninstallFiles(const std::string& path,
 
   for (const auto& dir : installedDirs) {
     TRACE("EXT: Deleting directory '%s'\n", dir.c_str());
-    base::remove_directory(dir);
+    try {
+      base::remove_directory(dir);
+    }
+    catch (const std::exception& ex) {
+      (void)ex;
```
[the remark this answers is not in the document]
dacap: This is true. Generally `(void)varname` is used to avoid a "unused variable" warning from the compiler, but the `ex` variable is being used with the LOG() functions.

Gasparoken:
I was able to reproduce the problem, and so was @martincapello. If you want to reproduce the bug, I can send an extension.

dacap:
If the extension can be published, I would prefer to upload the .zip to the issue https://github.com/aseprite/aseprite/issues/3385 (e.g. in a new comment)

Gasparoken:
Done. @dacap you can merge it now.

src/app/extensions.cpp
```diff
@@ -25,6 +25,7 @@
 #include "base/file_handle.h"
 #include "base/fs.h"
 #include "base/fstream_path.h"
+#include "base/log.h"
```
martincapello, line 28: I know it looks good here...but **I think** that it has to be the last include of all the includes...so, I would move this under the <string> include.

Gasparoken: Nice catch!

Gasparoken:
Done. @dacap can do the final review now, or just merge it.

src/app/extensions.cpp
dacap: :+1: I forgot about the `#undef ERROR` in `base/log.h`, and in several files `base/log.h` is not included as the last header file. Anyway in the future we might rename the `ERROR` macro to avoid clashes with other libraries (or we could just use other logging library, like https://github.com/gabime/spdlog and remove our functions, but it's something for the future). I'll merge this.

dacap:
Merged 877ae9eeecb90b0452326a48fcee5026ce5972c7

PR #3434, opened 2022-07-20
Invalidate Preview Window when pasting Frame (fix #2215)

### LINKED ISSUE #2215: When a frame is pasted into the timeline, the sprite editor is not invalidated
Aseprite v1.2.16.2-x64, Windows 10.

To reproduce: Open a sprite with multiple frames. Select a single frame in the timeline, and copy it using Ctrl+C. Select a frame elsewhere in the timeline which is visually different to the first frame. Paste, either using Ctrl+V or using the menu item "Edit > Paste".

Expected behaviour: The sprite editor should immediately redraw itself to reflect the fact that the active frame has changed.

Observed behaviour: The sprite editor is not invalidated and redrawn. If the user takes an action which invalidates the sprite editor (such as scrolling it off-screen then back on-screen, or changing the zoom level), the displayed graphics will suddenly change.

Joefish:

### DIFF
--- a/src/app/util/clipboard.cpp
+++ b/src/app/util/clipboard.cpp
@@ -28,6 +28,8 @@
 #include "app/ui/color_bar.h"
 #include "app/ui/editor/editor.h"
 #include "app/ui/skin/skin_theme.h"
+#include "app/ui/main_window.h"
+#include "app/ui/preview_editor.h"
 #include "app/ui/timeline/timeline.h"
 #include "app/ui_context.h"
 #include "app/util/clipboard.h"
@@ -469,8 +471,10 @@ void paste(Context* ctx, const bool interactive)
             // This is the app::copy_range (not clipboard::copy_range()).
             if (srcRange.layers() == dstRange.layers())
               app::copy_range(srcDoc, srcRange, dstRange, kDocRangeBefore);
-            if (current_editor)
+            if (current_editor) {
               current_editor->invalidate(); // TODO check if this is necessary
+              App::instance()->mainWindow()->getPreviewEditor()->invalidate();
+            }
             return;
           }
 
@@ -516,8 +520,10 @@ void paste(Context* ctx, const bool interactive)
           }
 
           tx.commit();
-          if (current_editor)
+          if (current_editor) {
             current_editor->invalidate(); // TODO check if this is necessary
+            App::instance()->mainWindow()->getPreviewEditor()->invalidate();
+          }
           break;
         }
 


Joefish:
While the original Issue #2215 that the frame editor was not updated when pasting a cel, was resolved, the preview window still did not update correctly. I know throwing another widget invalidate in there is not the best solution. I am still new to the code base so it took me a bit to understand how the messaging and observer system work in this case.
PreviewEditorWindow inherits from EditorObserver, and although it has onBefore/AfterFrameChanged, it didn't fire when the Cel was pasted and rendered in the editor. The Document observer interface has a onCelCopied function that could be used but adding a DocObserver to the PreviewEditor Class would be needed or extending the Editor Class to forward the notify.
A working example would be like this,
`
diff --git a/src/app/ui/preview_editor.cpp b/src/app/ui/preview_editor.cpp
index 9c89a6fa5..4c63d96cd 100644
--- a/src/app/ui/preview_editor.cpp
+++ b/src/app/ui/preview_editor.cpp
@@ -379,6 +379,7 @@ void PreviewEditorWindow::updateUsingEditor(Editor* editor)
     miniEditor->setState(EditorStatePtr(new NavigateState));
     miniEditor->setAnimationSpeedMultiplier(m_aniSpeed);
     miniEditor->add_observer(this);
+    document->add_observer(this);
     layout();
·
     if (!autoScroll)
@@ -513,4 +514,9 @@ void PreviewEditorWindow::adjustPlayingTag()
                    Preferences::instance().preview.playAll());
 }
·
+void PreviewEditorWindow::onCelCopied(DocEvent &ev)
+{
+  invalidate();
+}
+
 } // namespace app
diff --git a/src/app/ui/preview_editor.h b/src/app/ui/preview_editor.h
index ec7d20c7e..ea0c6fdf3 100644
--- a/src/app/ui/preview_editor.h
+++ b/src/app/ui/preview_editor.h
@@ -21,6 +21,7 @@ namespace app {
·
   class PreviewEditorWindow : public ui::Window
                             , public EditorObserver
+                            , public DocObserver
                             , public DocViewPreviewDelegate {
   public:
     PreviewEditorWindow();
@@ -51,6 +52,8 @@ namespace app {
     void onClose(ui::CloseEvent& ev) override;
     void onWindowResize() override;
·
+    void onCelCopied(DocEvent& ev) override;
+
   private:
     void uncheckCenterButton();
     bool hasDocument() const;
`

But as I said I wanted to keep the changes to a minimum and imo another observer class just for this would be overkill.

dacap:
Thanks for your PR @Joefish :+1: I've just tested and while it fixes the problem with the preview editor, I saw that it doesn't fix the problem if we have multiple editors, e.g. (in second 0:05 the cel is pasted):

https://user-images.githubusercontent.com/39654/202469955-8d7338be-decb-4f99-b027-5edd84a6a488.mp4

As you saw, an `onCelCopied` solution might be better, but I've found other bigger problems when frames are copied (whole frames). I'll review the behavior of cels, frames, and layers, and fix them.

dacap:
I've used Doc::notifyGeneralUpdate to update all visible Editor/DocView: https://github.com/aseprite/aseprite/commit/d8c1b19c80d7bbaf93ea0b196ce1296bfee800ef

Also there were a missing redraw when copying/pasting frames (not cels) in the same doc (the `DocRange::kFrames` case). This was fixed too in the commit. Thanks for your help @Joefish to detect and fix this issue :+1:

Issue #2554, opened 2020-10-31
What does the tooltip_text color definition applies to?


yu-tang:
I am looking into a theme/skin feature now. Specifically, `theme.xml` in default theme.
I picked a tooltip up and change its color:

        <color id="tooltip_text" value="#000000" />
        <color id="tooltip_face" value="#ffff7d" />

to:

        <color id="tooltip_text" value="#ff0000" />
        <color id="tooltip_face" value="#0000ff" />

Tooltip face (i.e. background color of text) has changed as expected. But the text foreground color has not.
I tinkered with it and found the `textbox_text` color definition applies to a tooltip text color.
Yes, now I have the worst tooltip which is consists blue and red color!

Here is an unexplained one.
What does the `tooltip_text` color definition applies to?

Aseprite v1.2.25

thkwznk:
The same issue can be observed in version 1.2.35-x64 (on Windows 11), the `tooltip_text` doesn't seem to be applied to the tooltips text.

PR #3441, opened 2022-07-27
Fix [regression] textBox not rendering text (doesn't update the bounds area of the text box)

Gasparoken:
Regression in dd0c29620974e56d9e62653804b996740a89e88a (Use tooltip_text style & color to paint tooltip windows #2554).
Prior this fix, text box do not update its bounds area because it has not a style assigned.

### DIFF
--- a/data/extensions/aseprite-theme/theme.xml
+++ b/data/extensions/aseprite-theme/theme.xml
@@ -503,6 +503,10 @@
             <background color="tooltip_face" />
             <text color="tooltip_text" align="left" />
         </style>
+        <style id="textbox_text" align="left top wordwrap" border="4">
+            <background color="textbox_face" />
+            <text color="textbox_text" align="left" />
+        </style>
         <style id="label" padding="1">
             <text color="text" align="left" />
             <text color="disabled" align="left" state="disabled" />
--- a/src/app/ui/skin/skin_theme.cpp
+++ b/src/app/ui/skin/skin_theme.cpp
@@ -913,9 +913,8 @@ void SkinTheme::initWidget(Widget* widget)
       break;
 
     case kTextBoxWidget:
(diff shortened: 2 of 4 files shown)


martincapello (approved):
Looks good to me! Just tested this fix and it seem to work just fine.

dacap (changes requested):
Please use lower case for ids as all other styles: `textbox_text`

Gasparoken:
Thank you @dacap.  Done.

Gasparoken:
This PR will solve issue #3442

brickviking:
I tried the suggested fix for this over in the beta branch too, it seems that aseprite/data/extensions/aseprite-theme/dark/theme.xml is also affected. I'm guessing the same lines can go in.

Gasparoken:
The last forced push also resolves the incorrect wrapped text in the Console.

Gasparoken:
Hi @brickviking. I've tried other themes (without 'textbox_text' style lines on 'theme.xml') and can't reproduce the original error. The Beta branch will be updated eventually.

brickviking:
Yup, agreed. When I ran Help->About with dark theme enabled, I got the text, though it was white background, dark foreground, presumably because it hadn't been assigned with `<background color="textbox_face" />` and `<text color="textbox_text" align="left" />`. Once I put in the requisite lines, the text appeared the right colours for the theme.

dacap:
Thanks everybody! Merged in https://github.com/aseprite/aseprite/commit/55885ee266fbb7d01894cf0a456aaceb8cc807c7 and then in beta in https://github.com/aseprite/aseprite/commit/05bf34f084b6f372a2cab2d38204192085840923 (adding the new style to the dark xml to avoid a white console) :+1:

PR #3446, opened 2022-07-28
Fix gif opening causes color reindexing when duplicate colors in color table (fix #3426)

### LINKED ISSUE #3426: Gif opening causes color reindexing when duplicate colors in color table
When Aseprite opens a gif which has a single global color table, it will re-index pixels in some of the frames when they share RGB values with other indices in the color table.

black_test_aseprite

In this gif, the palette is 256 copies of black, but the 2 frames are indexed to color 240 in the table.  When Aseprite opens it, frame 1 is all indexed as 240, but in frame 2, all the pixels are re-indexed to 1 since it's also black.

Expected/preferred behavior: When Aseprite opens a gif with only a global color table, the colors are copied from the color table to the current palette, and the indices across all the frames are left alone and copied in as-is.

* Aseprite version: v1.2.34.1-x64, installer
* System: Windows 10 Home

Gasparoken:


Gasparoken:
Passed standard tests, also tested with our gif database. Improved similarity between GIFs loaded/saved in Aseprite.

src/app/file/gif_format.cpp
```diff
@@ -561,18 +561,31 @@ class GifDecoder {
...
+        found = usedEntries.size();
+      else {
+        for (int i=0; i<ncolors; ++i) {
+          if (!usedEntries[i])
+            continue;
+
+          if (localCMap && i < localCMap->ColorCount &&
+              rgba(localCMap->Colors[i].Red,
+                   localCMap->Colors[i].Green,
+                   localCMap->Colors[i].Blue, 256) == palette->getEntry(i)) {
```
dacap, line 576: Didn't test this yet, but seeing the code, it looks like this 256 should be 255.
```suggestion
                   localCMap->Colors[i].Blue, 255) == palette->getEntry(i)) {
```

dacap:
Just in case, the issue was reported for main, and this patch is for beta. Anyway I would prefer to apply this only in beta (as it will be the new decoder/encoder anyway).

Gasparoken:
Thank you David for the review.

> Just in case, the issue was reported for main, and this patch is for beta. Anyway I would prefer to apply this only in beta (as it will be the new decoder/encoder anyway).

You are right, and this is not casual. I did it in beta since many important modifications of the gif format were made in this branch in the last years.

dacap (changes requested):
Just in case, the 256 is still there (in both places), not sure if this PR is ready to be merged.

Gasparoken:
Sorry for that. Idk why 256 is still there. I remember changing the value🤔. Perhaps a mix-up from working with VSCODE and XCODE at the same time (I'm guessing I may have been open to the same file and hit 'save' on a previous view of the file, Idk). I'll fix it soon.

Gasparoken:
Done.

martincapello (approved):
Looks good to me!

dacap:
I'll merge this PR :+1: 

Anyway I've detected other issue but I think it's unrelated to this.

dacap:
Just merged https://github.com/aseprite/aseprite/commit/fbf62407d3c0a719a531f0a61c22d0a443a981a3 and created other issue related to the gif encoder (but unrelated to this patch)

PR #3467, opened 2022-08-08
Fix reference layer scaled incorrectly after Sprite > Sprite Size (fix #3461)

### LINKED ISSUE #3461: Reference layer scaled incorrectly after Sprite > Sprite Size
Using the give sprite as example: [ref.zip](https://github.com/aseprite/aseprite/files/9268489/ref.zip)
 And then Sprite > Sprite Size with width=32, heigth=32 will put the reference layer in an incorrect place (I think the scaling is correct, but the position not).

Gasparoken:


dacap:
I'll approach this in a different way testings for resize_cel_image() and Image:pivot() (Lua) because I think the call to resize_cel_image() in SpriteSizeCommand might be wrong:
```diff
diff --git a/src/app/commands/cmd_sprite_size.cpp b/src/app/commands/cmd_sprite_size.cpp
index 1adac4622..8e1ae7478 100644
--- a/src/app/commands/cmd_sprite_size.cpp
+++ b/src/app/commands/cmd_sprite_size.cpp
@@ -105,9 +105,7 @@ protected:
       resize_cel_image(
         tx(), cel, scale,
         m_resize_method,
-        cel->layer()->isReference() ?
-          -cel->boundsF().origin():
-          gfx::PointF(-cel->bounds().origin()));
+        gfx::PointF(0, 0));
 
       jobProgress((float)progress / cels_count);
       ++progress;
```

dacap:
The solution is right :+1: I've just merged with minor changes: https://github.com/aseprite/aseprite/commit/19b18d2a63b3e807a3200e79f76876864d2877ab

The main issue here is that Image:resize() (when the image is associated to a cel) and the C++ resize_cel_image() function receive a pivot that is relative to the cel origin (so Point(0, 0) is the cel's origin, not the sprite origin).

dacap:
I'm leaving this little script to test the resize of the active cel using different pivot points:

```lua
local dlg = Dialog("Resize Cel x2")
dlg:entry{ id="x", label="Pivot X", text="0" }
dlg:entry{ id="y", label="Pivot Y", text="0" }
dlg:button{ text="Resize", onclick=function()
              app.transaction(function()
                  local data = dlg.data
                  local s = app.activeSprite
                  local i = app.activeImage
                  i:resize{ size=Size(i.width*2, i.height*2), pivot=Point(data.x, data.y) }
                  s.width = 2*s.width;
                  s.height = 2*s.height;
                  app.refresh()
              end)
             end }
dlg:show{ wait=false }
```

Issue #1252, opened 2016-09-12
Create a File > Export... submenu


dacap:
There are several features that could be grouped in just one submenu.

* Export one frame (#1080)
* Export animation as a GIF (same as `File > Save Copy As...` with `.gif` extension)
* Publish to social network (#1220)
* Export Sprite Sheet
etc.

Ashedragon:
Does this include the ability to export a GIF animation of specific tags, in the same way exporting a sprite sheet functions?

dacap:
Yes, some options from the File > Export menu should use #1080, see this comment https://github.com/aseprite/aseprite/issues/1080#issuecomment-244763783

littletinman:
I would very much like the ability to export a marquee selection. This is a make or break feature for me.

oceanhahn:
The workaround I use for that right now is either to Copy Merged to a new file and export that, or use New Sprite From Selection and save that (which transfers colour data like palette and mode state but neglects layers other than current).

Being able to save a selection directly to a new file would be nice, though, especially if it could save through layers and frames! :D

Issue #1982, opened 2019-01-14
Extrude Tileset option


lrasomattos:
Hi! I just got notified that Extrude Sprite Sheet option (for bleeding sprites) was added like asked in:
#1890 

Unfortunately it still doesn't cover the case for editing and making Tile Sheets / Tilesets. So here I would like to request that feature, as it would for sure be of great value for me and many people who use Aseprite for creating tile sheets.

I imagine an option would be cool, on export or save as, to extrude, one or more pixels "per tile"/"per grid", not only "per sprite".

dacap:
Hi @lrasomattos, so instead of extruding each frame, we should extrude each grid tile, right? Sounds like a nice option. I think it can be useful to combine with other option (e.g. adding padding to grid tiles, etc.).

With this option I guess that each grid tile and each frame of the animation counts like a different item to extrude in the final sprite sheet.

lrasomattos:
Yes!!

Remembering that both extrudes/bleed, in between tiles AND in the margins would be really important for seamless tiles.

Also I believe it's much better as "export feature" instead of an "edit feature", but could work as both. Unless you can de-extrude as well, the main idea is to be able to work on tilesets and easily export them extruded, not to keep them extruded which makes it much harder to edit.

About how it would work with animations, I guess what you said is ok. I imagine people won't be making animations when working with tiles, I don't see how it could be useful now... Or if they did, might prefer each frame in a different file.. (Like using one .ase file for many tilesets).

Also, I don't know how you handle this normally, but if the preference on how to export the extruded files could be saved per file it would be perfect!

By the way, thanks for considering!

PR #3473, opened 2022-08-11
Implement more Export/Import options

dacap:
This is a work-in-progress. We want to implement #1252, moving all export/import options into File > Export/Import submenus, and adding more options:

https://twitter.com/davidcapello/status/1557098807917264898

Submenus:

image

Export a specific area (selection, slices, etc.):
image


dacap:
We've added a new File > Export > Export Tileset option (and -export-tileset CLI option):

image

When it's used it will open the same "Export Sprite Sheet" dialog but with the "Tilesets" option selected in this new Source field:

image

We've also added a new "Sprite Grid" (or -split-grid CLI option) to export each grid cell/tile as an individual sprite (so we can apply extrude/borders to each grid cell, just like requested in #1982).

PR #3476, opened 2022-08-14
Initial path in the file dialog is set to the path of the currently open file

sumibi-yakitori:
This kind of behavior is more convenient.
For example, the initial path in the file dialog of Visual Studio Code and other programs behaves like this PR.


dacap:
Thanks for this PR @sumibi-yakitori, sorry for the late response. Probably we could merge this change, and then add an option to disable it by default (I saw that Affinity Photo uses the latest folder instead of the active document folder to open new files).

PR #3481, opened 2022-08-22
Request frame duration when opening sequences

lampysprites:
Currently, Aseprite defaults to 10fps for new files, and when opening animation sequences (`file-001.png`, `file-002.png`, ...).

While I started using it for the latter often, it became a bit irritating to set framerate each time opening the files. Mostly due to the fact that it's easy to miss and begin export at default value - takes a while for large gifs.

This commit puts duration field into the sequence popup. Alternatively, it could've been a setting for the default value, which is less flexible but affects new files as well. Or both. Input needed.

I tested that this change does not affect loaded gifs' timing, but tbh entire image loading is a large chunk, not sure what else can be affected.

image

------------------------

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -1033,6 +1033,7 @@ double_high = Double-high Pixels (1:2)
 title = Notice
 description = Do you want to load the following files as an animation?
 repeat = Do the same for other files
+duration = Duration
 agree = &Agree
 skip = &Skip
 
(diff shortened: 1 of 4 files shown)


dacap:
Hi @lampysprites, thanks for your new contribution <3 I think it's a great idea, I've just merged with minor changes (mainly removing tabs from the open_sequence.xml file): https://github.com/aseprite/aseprite/commit/c67902dee372e4e6d824a8a1e2b7f1486faff349

Then I've added a minor change in other commit: https://github.com/aseprite/aseprite/commit/6da4000e6c949fe5e7b032ffcb9a749fefac6d49

Issue #518, opened 2014-11-04
Make all elements on screen dockable/movable/resizable


dacap:
The color-bar, color-picker (#492), timeline, tabs (#118), minieditor (#404), and tool-bar should be dockable windows that can be moved to any other position on the screen.

Review #250, [t/933](https://community.aseprite.org/t/preview-on-another-window/933), [t/1065](https://community.aseprite.org/t/layer-selection-more-rapid/1065), [/5472](https://community.aseprite.org/t/customizing-layout-of-panels/5472)

Tasks:
- [x] Multiple native windows are required to drag-and-drop widgets to separated windows (#139)
- [x] Initial `Dock` widget to put widgets on each side, center, create subdocks ([preview](https://twitter.com/davidcapello/status/1440068577109766154))
- [x] Restore old layout configuration (color bar split pos, timeline pos, etc.) and migrate to new Dock layout
- [x] Load/saving Dock layout
- [x] Create & customize current layoout (drag-and-drop widgets, etc.)

Following work will be done in #5327

SleepyEleven:
This feature will be the bee's knees when it is implemented. Can't wait.

dacap:
Some possibilities requested by other people:

![cxgp0qpwqaa_lkg](https://cloud.githubusercontent.com/assets/39654/20245118/782ddeda-a978-11e6-845d-ee8a485e451f.jpg)
[Source](https://twitter.com/mckack/status/797605447939321857)

dacap:
Related to #250

Omar-AbdulAzeez:
something to consider for the future (if not already considered) is also making sprite tabs floating and dockable... if the palette, timeline, preview, (other sprite dependent elements) are already docked to that tab then they should undock with it with the exception being tools i.e.. only 1 tools element should exist at all times with the ability to hide it altogether from menubar > view... basically you can just leave it docked to the main window until specifically grabbed. Also don't forget to make copies of the menubar as well when undocking sprite tabs. if the sprite dependent elements are undocked they should be sprite tab specific... i.e.. should only dock back to the same sprite tab they came from and opening a new tab will still give it it's own docked elements even if there are undocked elements elsewhere... the sprite specific part could be an option toggle (FOR EACH ELEMENT SEPERATELY!) since it can be beneficial to just have those elements update themselves when focusing on another sprite (might create issues with selected cel and fg and bg color for each sprite tho idk)

[...]

PR #3485, opened 2022-08-23
Dock

dacap:
This is a work-in-progress. We want to add multiple ways to customize the layout/workspace https://github.com/aseprite/aseprite/issues/518


dacap:
The initial iteration of this feature is quite ready to be tested. Still some minor things to be fixed (and a lot of room of improvement to dock widgets in other parts, and sub-divide widgets like the ColorBar), but it's a good start for a first iteration.

data/strings/en.ini
```diff
@@ -1232,6 +1242,22 @@ name = Name:
...
 default_new_layer_name = New Layer
 
+[new_layout]
+title = New Workspace Layout
+base = Base:
+name = Name:
+modified = {} (Modified)
+deleting_layout = Deleting Layout
+deleting_layout_confirmation = Are you sure you want to delete the layout '{}'?
+restoring_layout = "Restoring Layout"
```
ckaiser, line 1252:
```suggestion
restoring_layout = Restoring Layout
```

data/strings/en.ini
```diff
@@ -1232,6 +1242,22 @@ name = Name:
...
+restoring_layout_confirmation = Are you sure you want to restore the {} layout?"
```
ckaiser, line 1253:
```suggestion
restoring_layout_confirmation = Are you sure you want to restore the '{}' layout?
```

data/widgets/new_layout.xml
```diff
@@ -0,0 +1,25 @@
+<!-- Aseprite -->
+<!-- Copyright (C) 2022 by Igara Studio S.A. -->
+<gui>
+<window id="new_layout" text="@.title">
+  <vbox>
+    <grid columns="2">
+      <label text="@.base" />
+      <combobox id="base" expansive="true">
```
ckaiser, line 8: I feel like this dialogue should probably have some kind of explainer/help text either in a tooltip or in a help button or something, just to make sure the concept of a "base" layout is understood since it's a pretty programmer-y thing.

ckaiser (commented):
Gave it a test and found an issue, when modifying the default theme, closing the app and re-opening it again, the modifications aren't applied, only when I open the dropdown and re-select "Default*" do my layout changes show up. The rest is looking good!

Had some minor string stuff in the comments and usability thoughts:

New icon looks good! I'm wondering if it should have some kind of toggle state beyond just the dropdown appearing?

Another thing that I was thinking about was if it'd be wise to automatically hide the dropdown & docking handles, because I can definitely see a user clicking on it and then just leaving it on for the rest of the session. It's not a big deal since it doesn't persist, but yeah maybe opening/closing a file, losing focus or switching tabs could hide it?

dacap:
> Gave it a test and found an issue, when modifying the default theme, closing the app and re-opening it again, the modifications aren't applied, only when I open the dropdown and re-select "Default*" do my layout changes show up.

True. I was able to reproduce it :+1:

> New icon looks good! I'm wondering if it should have some kind of toggle state beyond just the dropdown appearing?

Not sure about this, just gave a try to something like this:

https://github.com/user-attachments/assets/ba557838-40e4-46b3-aa60-9a45249378e3

> Another thing that I was thinking about was if it'd be wise to automatically hide the dropdown & docking handles, because I can definitely see a user clicking on it and then just leaving it on for the rest of the session. It's not a big deal since it doesn't persist, but yeah maybe opening/closing a file, losing focus or switching tabs could hide it?

I'm not sure about this, although I think these dock handles can be annoying, hiding them automatically after some other events might be a little confusing. Probably something to wait for some feedback first.

data/widgets/new_layout.xml
dacap: I'm not sure if the "base" field is even necessary. From the user point of view, creating a new layout will be based on the "actual look" of the UI (the actual layout). Not sure what effect has the "base" layout to the new user layout. Could we just remove this field?

ckaiser: Yeah now that I think about it, it's a relic from when you had to make a new layout to modify the default ones - better off without it, simpler.

dacap:
I've just updated `beta` branch and rebased `dock` to continue working on this.

PR #3492, opened 2022-08-30
Use layer index to differentiate between layers with the same name

### LINKED ISSUE #2656: Export Sprite Sheet cannot differentiate between Layers with matching names
If multiple Layers have the same name, the Export Sprite Sheet pane only recognizes one of them regardless of which of the Layers (of the same name) is selected from the "Layers" pulldown box.
Layers with matching names can be in the same or different Layer Groups and the bug will still occur.

It seems as though layers are keyed to the string used to name them, regardless of Grouping, rather than the specific identity they have in the Layers selection pulldown box.

..using ASEprite 1.2.25-x64 on Win10.

martincapello:
Fix #2656

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -485,6 +485,7 @@
       <option id="filename" type="std::string" />
       <option id="resize_scale" type="double" default="1" />
       <option id="layer" type="std::string" />
+      <option id="layer_index" type="int" default ="-1" />
       <option id="frame_tag" type="std::string" />
       <option id="ani_dir" type="doc::AniDir" default="doc::AniDir::FORWARD" />
       <option id="apply_pixel_ratio" type="bool" default="false" />
@@ -512,6 +513,7 @@
       <option id="ignore_empty" type="bool" default="false" />
       <option id="open_generated" type="bool" default="false" />
       <option id="layer" type="std::string" />
+      <option id="layer_index" type="int" default ="-1" />
       <option id="frame_tag" type="std::string" />
       <option id="split_layers" type="bool" default="false" />
       <option id="split_tags" type="bool" default="false" />
--- a/src/app/commands/cmd_export_sprite_sheet.cpp
+++ b/src/app/commands/cmd_export_sprite_sheet.cpp
@@ -76,6 +76,7 @@ struct ExportSpriteSheetParams : public NewParams {
   Param<bool> mergeDuplicates { this, false, "mergeDuplicates" };
   Param<bool> openGenerated { this, false, "openGenerated" };
   Param<std::string> layer { this, std::string(), "layer" };
+  Param<int> layerIndex { this, -1, "layerIndex" };
   Param<std::string> tag { this, std::string(), "tag" };
   Param<bool> splitLayers { this, false, "splitLayers" };
   Param<bool> splitTags { this, false, "splitTags" };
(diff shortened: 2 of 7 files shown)


dacap:
It looks like this change https://github.com/aseprite/aseprite/pull/3492/commits/6a748db63abcd99d6a77beb91b731711e42d4469 includes changes to all end of lines (from Unix-like to Windows-like EOL). Could you please fix that?

dacap:
And then 2229b5be9 does the reverse (?)

martincapello:
How about now?

dacap:
I'll merge the commit with a minor change in the message (adding a missing reference to #2656). And then I'll add some extra changes (because the "- 2" is hard-coded and I would prefer constant defined in layer_frame_comboboxes.h)

dacap:
Also two things: 1) I have to check if the index is from top to bottom or bottom to top, but maybe 0 should be the background (bottom), and going up from there. 2) from what it looks, if we specified the layerIndex, we have to specify the layer name too in ExportSpriteSheetParams.

I'll try to fix these issues.

dacap:
@martincapello I have pushed some changes into this branch: https://github.com/dacap/aseprite/tree/fix-export-layers-same-name2 but I'm not going to merge this yet as ExportSpriteSheet doesn't remember the last layer exported correctly (to do so we'd need an extra value in pref.xml: https://github.com/aseprite/aseprite/blob/277e24c799a8be63fbab9962c94b163de59c0310/data/pref.xml#L514 )

If you want to continue from fix-export-layers-same-name2 you can do so. We'll merge this for a future version.

martincapello:
> I'll merge the commit with a minor change in the message (adding a missing reference to #2656). And then I'll add some extra changes (because the "- 2" is hard-coded and I would prefer constant defined in layer_frame_comboboxes.h)

Makes sense. Also, do you think we should have a constant for the -1 returned by the layerIndex() functions?

> @martincapello I have pushed some changes into this branch: https://github.com/dacap/aseprite/tree/fix-export-layers-same-name2 but I'm not going to merge this yet as ExportSpriteSheet doesn't remember the last layer exported correctly (to do so we'd need an extra value in pref.xml:

I wasn't sure about adding this extra value to pref.xml, I will do it then.

> If you want to continue from fix-export-layers-same-name2 you can do so. We'll merge this for a future version.

Okay, I will continue from that branch. Thanks!

martincapello:
@dacap Just pushed some changes to let the exporting dialogs remember the last selected layer index.

data/pref.xml
```diff
@@ -485,6 +485,7 @@
       <option id="filename" type="std::string" />
       <option id="resize_scale" type="double" default="1" />
       <option id="layer" type="std::string" />
+      <option id="layerIndex" type="int" default ="-1" />
```
dacap, line 488: Please check that we are using other format in the pref.xml file:
```suggestion
      <option id="layer_index" type="int" default ="-1" />
```

martincapello: I think that this is not the first time I made this mistake...🤦
Will fix it, thanks!

src/app/commands/cmd_export_sprite_sheet.cpp
```diff
@@ -212,13 +219,16 @@ Doc* generate_sprite_sheet_from_params(
...
 
   SelectedLayers selLayers;
   if (layerName != kSelectedLayers) {
     // TODO add a getLayerByName
+    int i = sprite->allLayersCount();
     for (const Layer* layer : sprite->allLayers()) {
-      if (layer->name() == layerName) {
+      i--;
+      if (layer->name() == layerName && layerIndex == -1 ||
+          layer->name() == layerName && layerIndex == i) {
```
dacap, line 231: @martincapello just saw a warning (`warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]`) about all these conditions, it's already merged but it would be nice to fix it:
```c++
      if (layer->name() == layerName && layerIndex == -1 ||
          layer->name() == layerName && layerIndex == i) {
```
It's equal to:
```c++
      if ((layer->name() == layerName) && (layerIndex == -1 || layerIndex == i)) {
```
The same for the other conditions, must be reviewed (these are 3 times in this PR).

PR #3494, opened 2022-08-31
[lua] Add app.selectedSlices field

lampysprites:
Added `app.selectedSlices` field to the lua api which gets/sets slice tool selection. Although it's possible with some workarounds to allow user to select this slice or that (combobox or selection overlap), I think that's more straightforward?

Regarding the naming, I'm not sure whether `activeSlices` would be a correct name, as the active slice in the editor seems to be (implicitly?) the alphabetically first one in the selection. For the same reason I do not see a point in retrieveing/setting it separately.

Api usage is same as selection ranges:

```lua
for i,slice in ipairs(app.selectedSlices) do 
  print(slice.name)
end

app.selectedSlices = { slice1, slice2 }
```

Further work, if needed: it seems that the selection persists even when slice tool is inactive, so it can be added to site/sitechange as well.

--------------------------

### DIFF
--- a/src/app/script/app_object.cpp
+++ b/src/app/script/app_object.cpp
@@ -17,6 +17,7 @@
 #include "app/doc_access.h"
 #include "app/i18n/strings.h"
 #include "app/loop_tag.h"
+#include "app/modules/editors.h"
 #include "app/modules/palettes.h"
 #include "app/pref/preferences.h"
 #include "app/script/api_version.h"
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @lampysprites, I think we'll need some extra time to think about the API here. Probably the selected slices could be in `app.range.slices`, but not yet sure.

dacap:
I've just merged this in: https://github.com/aseprite/aseprite/commit/4d65b5c92535f90901334c678dd1259376d68755
but then moving the code to app.range.slices: https://github.com/aseprite/aseprite/commit/6dce23540aa6595f0652213e36c9f1ec6beccd6b
with a minor change to support assigning a Slices object to app.range.slices: https://github.com/aseprite/aseprite/commit/47aae7fc07e30c2c5aeb06258f6aa95f3270485f

PR #3509, opened 2022-09-07
Fix the `group` attribute for Slice Properties

thkwznk:


dacap:
Thanks @thkwznk <3 just cherry picked this patch

PR #3514, opened 2022-09-09
Add SUSE dependencies

justanobdy:


dacap:
Generally we're not accepting more Linux distro details (as it would be almost infinite) but I'll make an exception, thanks for this patch  @justanobdy, maybe in the future we could create different guides (one specific for Linux as it is the most complex/more details require). Merged: f7154aa0d

PR #3518, opened 2022-09-13
Small fixes to improve debugging

martincapello:


Gasparoken:
Thanks @martincapello for your PR. Looks like a simple patch that is OK for me. The only tweak I would make on e6c3122a3bf484af3b045407345c9b3e69cd89b6 is to put ASSERTs before the assignment to keep the style of the code.

9134dc3d7ed25ffec4ec6be28d674732c26714b5 is needed to fix the cells height on the dynamics dialogs on MacOS, and fix the ASSERT in debug mode on Windows. Ok for this!

Just for the records: as we chatted, we can discover other problems in the future due to change of the base::clamp to std:clamp in 71d885d2a072751df1f7e6d34db270c92f792f31.
Mainly because the undefined behavior of std:clamp(value, min, max) function when min > max (as is specified in the [reference documentation](https://en.cppreference.com/w/cpp/algorithm/clamp)).
For example: 
std::clamp(5, 10, 1) = 10
base::clamp(5, 10, 1) = 1

martincapello:
@Gasparoken FYI I force pushed the branch, so your comment points to commit hashes that doesn't exist any more.

Gasparoken (approved):

Issue #2676, opened 2021-03-23
Dynamic's Gradient Row Checkmark is Positioned Wrong


JJCUBER:
### Minor UI Bug Report
For the `Dynamics` settings, the checkmark corresponding to the `Gradient` row is offset vertically compared to the checkmark for the other two rows (`Size` and `Angle`):
![image](https://user-images.githubusercontent.com/34446698/112190101-10bf4000-8bdb-11eb-8c30-e6589450e528.png)
![image](https://user-images.githubusercontent.com/34446698/112190126-174db780-8bdb-11eb-91eb-77882225cc7d.png)
(**Last row of the grid at the top of both images**)

### Aseprite and System version
* Aseprite version: **v1.2.25 dev**
* System: **Windows 10**

PR #3519, opened 2022-09-14
Small UI button set painting fixes

martincapello:
This PR contains 2 fixes to the way ButtonSet items are painted.
One fixes the vertical position of the text, and the other fixes the vertical position of the icon (#2676).

### DIFF
--- a/src/app/ui/button_set.cpp
+++ b/src/app/ui/button_set.cpp
@@ -96,6 +96,7 @@ void ButtonSet::Item::onPaint(ui::PaintEvent& ev)
   if (m_icon || isLastRow) {
     textRc.y -= 2*guiscale();
     iconRc.y -= 1*guiscale();
+    if (isLastRow && info.row > 0) iconRc.y -= 2*guiscale();
   }
 
   if (!gfx::is_transparent(bgColor()))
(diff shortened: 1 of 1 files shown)


Gasparoken:
This is not as simple as it seems. This fix fixes the Dynamics popup dialog, but changes other buttons. I think there is something deeper to fix (like the general text/icon alignment in the generic button widget). I'll do some more research tomorrow.
Possible causes of misalignment: buttons border treatment (mainly those with decorative shading added -up to 3 extra pixels on the bottom-), sizes in odd/even pixels of widgets/icons/text, rounding of integers when divided For 2.

Gasparoken:
Finally, I think this solution is fine. The text/icon alignment in the last row is consistent with the remaining buttons that are not in the last row. General text/icon vertical alignment in ButtonSet widgets may be covered in another issue/PR.

Gasparoken (approved):

dacap:
Although some other buttons have changed, I think the new look for the Dynamics is better than before, so I'll merge this as it is. I might add some new properties in the theme to adjust the height of the buttons after main and beta are merged :+1:

dacap:
Merged https://github.com/aseprite/aseprite/commit/6e57546aa986f68e502c21fc61642c539f762732 and https://github.com/aseprite/aseprite/commit/fecfbb1145ffa2636e3e063d80eadf16cc0c6baf

PR #3521, opened 2022-09-15
Fix index mode spray not handling masked color as expected (fix #3063)

### LINKED ISSUE #3063: Index mode spray does not handle masked color as expected

martincapello:
This PR updates an if condition in BaseInk::prepareInk() method to let the spray tool behave like the pencil tool when the mask color is selected in an indexed mode sprite.

### DIFF
--- a/src/app/tools/inks.h
+++ b/src/app/tools/inks.h
@@ -120,12 +120,7 @@ public:
 
           // Opacity is set to 255 when InkType=Simple in ToolLoopBase()
           if (loop->getOpacity() == 255 &&
-              // The trace policy is "overlap" when the dynamics has
-              // a gradient between FG <-> BG
-              //
-              // TODO this trace policy is configured in
-              //      ToolLoopBase() ctor, is there a better place?
-              loop->getTracePolicy() != TracePolicy::Overlap) {
+              loop->getDynamics().gradient == DynamicSensor::Static) {
             color_t color = loop->getPrimaryColor();
 
             switch (loop->sprite()->pixelFormat()) {


src/app/tools/inks.h
```diff
@@ -120,12 +120,7 @@ class PaintInk : public BaseInk {
 
           // Opacity is set to 255 when InkType=Simple in ToolLoopBase()
           if (loop->getOpacity() == 255 &&
-              // The trace policy is "overlap" when the dynamics has
-              // a gradient between FG <-> BG
-              //
-              // TODO this trace policy is configured in
-              //      ToolLoopBase() ctor, is there a better place?
-              loop->getTracePolicy() != TracePolicy::Overlap) {
+              loop->getDynamics().gradient == DynamicSensor::Static) {
```
Gasparoken, line 123: This code works fine, solves the problem, and it can stay like this.
On the other hand, I was testing and looking for the reason behind the existence of the
`loop->getTracePolicy() != TracePolicy::Overlap` included inside the 'if' condition in the first place.
If I remove `loop->getDynamics().gradient == DynamicSensor::Static` , the editor works fine too.
Later I'll try more cases to assess why it works without the gradient condition.

martincapello: I struggled quite a bit trying to understand the same, but I didn't come to any conclusion...

Gasparoken: I'll dig into the code a bit more to see if I can come to any conclusions.

dacap:
Just as a side note: Please use more descriptive commit messages (not only "fix # of issue") :pray:

src/app/tools/inks.h
Gasparoken: Interesting, I found this:
- If we omit `loop->getDynamics().gradient == DynamicSensor::Static`, all the dynamics will change the opaque/transparent ink processing. For example, for gradient dynamics we need always TransparentInkProcessing, but with the absence of `loop->getDynamics().gradient == DynamicSensor::Static` it may be opaque in some cases.
- Evaluating this PR, I found the gradient dynamics (velocity) + primary color = opaque color + secondary color = mask + Simple Ink or Alpha Compositing results in no gradient because some issue in point_shapes.h (I'll do a new PR for this)

So this PR is fine for me. The only thing to replace is the commit title to
"Fix index mode spray does not handle masked color as expected (fix #3063)"

Gasparoken (approved):

dacap:
Merged https://github.com/aseprite/aseprite/commit/8d5f28199835c0c6236f66a280309cc63775693f :+1:

Issue #2349, opened 2020-04-22
Rewrite the ToolLoop


dacap:
[The "ToolLoop"](https://dev.aseprite.org/2015/06/06/the-tool-loop/) needs a complete rewrite. As we've added more options, the logic got really complex and features/logic is scattered in all over the place (`ActiveToolManager::activeInk()`, `PaintInk::prepareInk()`, `ToolLoopBase`, `ToolLoopManager`, etc.).

PR #3525, opened 2022-09-16
Fix dynamics + gradient + no dithering + fg or bg is mask color results in an opaque stroke.

Gasparoken:
Before this fix, dynamic + gradient (pressure or velocity) + no dithering + fg or bg = mask color results in an opaque stroke (in RGBA and grayscale color mode).

### DIFF
--- a/src/app/tools/point_shapes.h
+++ b/src/app/tools/point_shapes.h
@@ -98,20 +98,39 @@ public:
         color_t b = m_primaryColor;
         const float t = pt.gradient;
         const float ti = 1.0f - pt.gradient;
+
+        auto rgbaGradient = [t, ti](color_t a, color_t b) -> color_t {
+          if (rgba_geta(a) == 0)
+            return doc::rgba(rgba_getr(b),
+                             rgba_getg(b),
+                             rgba_getb(b),
+                             int(t*rgba_geta(b)));
+          else if (rgba_geta(b) == 0)
+            return doc::rgba(rgba_getr(a),
+                             rgba_getg(a),
+                             rgba_getb(a),
+                             int(ti*rgba_geta(a)));
+          else
+            return doc::rgba(int(ti*rgba_getr(a) + t*rgba_getr(b)),
+                             int(ti*rgba_getg(a) + t*rgba_getg(b)),
+                             int(ti*rgba_getb(a) + t*rgba_getb(b)),
+                             int(ti*rgba_geta(a) + t*rgba_geta(b)));
+        };
+
         switch (loop->sprite()->pixelFormat()) {
           case IMAGE_RGB:
-            if (rgba_geta(a) == 0) a = b;
-            else if (rgba_geta(b) == 0) b = a;
-            a = doc::rgba(int(ti*rgba_getr(a) + t*rgba_getr(b)),
-                          int(ti*rgba_getg(a) + t*rgba_getg(b)),
-                          int(ti*rgba_getb(a) + t*rgba_getb(b)),
-                          int(ti*rgba_geta(a) + t*rgba_geta(b)));
+            a = rgbaGradient(a, b);
             break;
           case IMAGE_GRAYSCALE:
-            if (graya_geta(a) == 0) a = b;
-            else if (graya_geta(b) == 0) b = a;
-            a = doc::graya(int(ti*graya_getv(a) + t*graya_getv(b)),
-                           int(ti*graya_geta(a) + t*graya_geta(b)));
+            if (graya_geta(a) == 0)
+              a = doc::graya(graya_getv(b),
+                             int(t*graya_geta(b)));
+            else if (graya_geta(b) == 0)
+              a = doc::graya(graya_getv(a),
+                             int(ti*graya_geta(a)));
+            else
+              a = doc::graya(int(ti*graya_getv(a) + t*graya_getv(b)),
+                             int(ti*graya_geta(a) + t*graya_geta(b)));
             break;
           case IMAGE_INDEXED: {
             int maskIndex = (loop->getLayer()->isBackground() ? -1: loop->sprite()->transparentColor());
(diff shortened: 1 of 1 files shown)


src/app/tools/point_shapes.h
```diff
@@ -100,18 +100,32 @@ class BrushPointShape : public PointShape {
...
+              a = doc::rgba(rgba_getr(b),
+                            rgba_getg(b),
+                            rgba_getb(b),
+                            int(t*rgba_geta(b)));
+            else if (rgba_geta(b) == 0)
+              a = doc::rgba(rgba_getr(a),
+                            rgba_getg(a),
+                            rgba_getb(a),
+                            int(ti*rgba_geta(a)));
+            else
```
martincapello, line 113: Why is this needed? I tried this PR and then tried again without these lines...I couldn't say the difference...

Gasparoken: If you comment these lines, you'll get a mix between a transparent black and some opaque color (I use Alpha Compositing Ink)
Screen Shot 2022-09-19 at 14 11 02

With this new lines you'll get a mix between the desired opaque color and the same but full transparent (I use Alpha Compositing Ink):
Screen Shot 2022-09-19 at 14 18 39

Anyway, by explaining this, I now understand what David intended with the original code. I'll try to simplify the code and then force a new commit.

martincapello: I see now, thanks!

Gasparoken: I found
```
if (rgba_geta(a) == 0)
  a = doc::rgba(rgba_getr(b),
                rgba_getg(b),
                rgba_getb(b),
                int(t*rgba_geta(b)));
```
it's fine, because another more general approach will require more math to cancel members.

martincapello (approved):

src/app/tools/point_shapes.h
```diff
@@ -98,20 +98,39 @@ class BrushPointShape : public PointShape {
...
+            a = doc::rgba(rgba_getr(a),
+                          rgba_getg(a),
+                          rgba_getb(a),
+                          int(ti*rgba_geta(a)));
+          else
             a = doc::rgba(int(ti*rgba_getr(a) + t*rgba_getr(b)),
                           int(ti*rgba_getg(a) + t*rgba_getg(b)),
                           int(ti*rgba_getb(a) + t*rgba_getb(b)),
                           int(ti*rgba_geta(a) + t*rgba_geta(b)));
+        };
```
dacap, line 118: I'd recommend you to pass a and b as arguments (in the case of indexed, we are using the index b, and not the rgba value of b, because values are captured at this point, and not in the call site, [check this demo](https://godbolt.org/z/vMab7zdbY)). And return the new color:
```suggestion
        auto rgbaGradient = [t, ti](color_t a, color_t b) -> color_t {
          if (rgba_geta(a) == 0)
            return doc::rgba(rgba_getr(b),
                             rgba_getg(b),
                             rgba_getb(b),
                             int(t*rgba_geta(b)));
          else if (rgba_geta(b) == 0)
            return doc::rgba(rgba_getr(a),
                             rgba_getg(a),
                             rgba_getb(a),
                             int(ti*rgba_geta(a)));
          else
            return doc::rgba(int(ti*rgba_getr(a) + t*rgba_getr(b)),
                             int(ti*rgba_getg(a) + t*rgba_getg(b)),
                             int(ti*rgba_getb(a) + t*rgba_getb(b)),
                             int(ti*rgba_geta(a) + t*rgba_geta(b)));
        };
```
Other possibility would be to create a new function it self (passing t and ti too) to avoid confusion with a and b in the parent scope.

src/app/tools/point_shapes.h
```diff
@@ -98,20 +98,39 @@ class BrushPointShape : public PointShape {
...
+
+        switch (loop->sprite()->pixelFormat()) {
+          case IMAGE_RGB:
+            rgbaGradient();
```
dacap, line 122:
```suggestion
            a = rgbaGradient(a, b);
```

src/app/tools/point_shapes.h
```diff
@@ -121,12 +140,7 @@ class BrushPointShape : public PointShape {
             if (b == maskIndex) b = 0;
             else b = loop->getPalette()->getEntry(b);
             // Same as in RGBA gradient
-            if (rgba_geta(a) == 0) a = b;
-            else if (rgba_geta(b) == 0) b = a;
-            a = doc::rgba(int(ti*rgba_getr(a) + t*rgba_getr(b)),
-                          int(ti*rgba_getg(a) + t*rgba_getg(b)),
-                          int(ti*rgba_getb(a) + t*rgba_getb(b)),
-                          int(ti*rgba_geta(a) + t*rgba_geta(b)));
+            rgbaGradient();
```
dacap, line 143: This is the important place, here we are really passing the new updated value of b variable:
```suggestion
            a = rgbaGradient(a, b);
```

dacap (changes requested):
Some minor changes requested.

Some things I've noticed:
1. After merging this we should test this with Pixel Perfect mode in `beta` branch, because in the `main` it looks like the final results is not displayed until we fully re-draw the editor (e.g. pressing Ctrl or Command key will refresh the editor and display the canvas in the real-final state).
2. Not sure how we could improve the indexed mode when the mask color is selected in fg or bg color, at the moment it looks like it choose the closest non-mask color to black (or similar to mask color if the alpha of the mask color is not a=0). Anyway something to think about in the future.

src/app/tools/point_shapes.h
Gasparoken: 🤯Very good observation!

Gasparoken:
What do you think @dacap? Is it better for us to close this PR and create a new PR on the beta branch as a base?

dacap:
> Is it better for us to close this PR and create a new PR on the beta branch as a base?

Whatever you find that's easier to do, as we're going to merge beta on main in these days.

Gasparoken:
Sorry, I forgot to push my last commit.

dacap:
Testing this I've noticed a different behavior depending if the mask is the background or is the foreground:

Screenshot from 2022-09-23 09-12-28

Screenshot from 2022-09-23 09-12-38

Gasparoken:
I'm investigating it.

Gasparoken:
Related #2349.
When the primary color is full transparent, the ink behavior is changed here: https://github.com/aseprite/aseprite/blob/877ae9eeecb90b0452326a48fcee5026ce5972c7/src/app/tools/active_tool.cpp#L100-L105
Then, the next `switch case` will decide by an incorrect `InkProcessing` according `PaintInk` type here:
https://github.com/aseprite/aseprite/blob/877ae9eeecb90b0452326a48fcee5026ce5972c7/src/app/tools/inks.h#L116-L123
I already have a solution, but I'll test it some more.

Gasparoken:
https://github.com/aseprite/aseprite/pull/3525#issuecomment-1256134727 is solved with this other PR: https://github.com/aseprite/aseprite/pull/3538

dacap:
I'll merge this, because I think #3541 needs to be rebased to the latest main + this PR #3525

dacap:
Just merged 022c0c38d, I highly recommend to rebase #3541 to latest main commit (because there are conflicts)

PR #3529, opened 2022-09-20
Fix Advanced Mode disables Tilemap tool (fix #3478)

### LINKED ISSUE #3478: Advanced Mode disables Tilemap tool
Create and select a tilemap layer. Select a tile to paint with and be ready to place it. Enter advanced mode (where the UI hides). The brush will revert from Tile mode to pixel mode, and you can no longer place or interact with tiles properly. The Dropper tool only works on colors instead of tiles. Any drawing will edit tiles' pixels instead of painting with tiles. All you can do is move the tile layer itself. The current Space+Tab shortcut does nothing to fix this either. 

* Aseprite version: Aseprite v1.3-beta20-x64, Steam beta
* System: 
-- Edition:		Windows 10 Pro
-- Version:		21H1
-- Installed on:	‎6/‎19/‎2021
-- OS build:	19043.1889

martincapello:


martincapello:
Just realized that TilesetMode ColorBar::tilesetMode() depends on the m_tilesHBox visibility too, so I'm going to test how the tileset mode behaves in advanced mode and see if we need to make the same adjustment there.

martincapello:
Okay, seems that it is not necessary to modify the tilesetMode() method for now.

Gasparoken:
Your fix looks good.

> Just realized that TilesetMode ColorBar::tilesetMode() depends on the m_tilesHBox visibility too

I highly recommend doing the same modification here: `TilesetMode ColorBar::tilesetMode()`

I don't see why `m_tilesHBox.isVisible() &&` is there. I guess it was to define some default behavior in CLI/Script when there is no sprite and/or editor, but I'm not sure.

martincapello:
Okay I will make the same change in the tilesetMode() method. Thanks!

Gasparoken (approved):
It's ok for me.

dacap:
The isVisible() came from the first commit of the tilemap editor: https://github.com/aseprite/aseprite/commit/26139c4ae2aa3d7304c9daf7a002344b28c94427#diff-91de2a5d4d910690c6d907615c148656b7391253751d5e3fa030d0008b5db4abR459
and https://github.com/aseprite/aseprite/commit/26139c4ae2aa3d7304c9daf7a002344b28c94427#diff-91de2a5d4d910690c6d907615c148656b7391253751d5e3fa030d0008b5db4abR493 so probably a bug (or probably something necessary in that time when the palette color and the tiles weren't visible at the same time). There is an alternative mode where we can right-click the Edit Tiles button and the palette is hidden or the tiles are hidden. But I think that doesn't affect this fix.

PR #3535, opened 2022-09-23
Fix bmp file and bitmap size are incorrectly set during save command (fix #3534)

### LINKED ISSUE #3534: `bmp_file_header` fields are incorrectly set on bmp export

* Aseprite 1.3-beta21-x64
* Windows 10

Thought I would make this a separate bug report. Contributors at the Godot Engine project have looked into the bmp files not properly importing, and found that something about the bmps is in fact incorrecly set.

>The fields bmp_file_header.bmp_file_size and bmp_info_header.bmp_size_image are incorrectly set

Here is the specific post, seems like something that might need investigating further:
https://github.com/godotengine/godot/issues/66238#issuecomment-1256442592

Gasparoken:
At the moment this fix can only be tested by opening in Godot a BMP saved in Aseprite (RGBA color mode, just one transparent layer or more)

### DIFF
--- a/src/app/file/bmp_format.cpp
+++ b/src/app/file/bmp_format.cpp
@@ -1213,7 +1213,7 @@ bool BmpFormat::onSave(FileOp *fop)
               + biSizeImage);                        // image data
   }
   else {
-    biSizeImage = (w*3 + filler) * h;
+    biSizeImage = (w*bpp/8 + filler) * h;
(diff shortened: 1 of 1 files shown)


Gasparoken:
Please just merge it. There is not much to review here.

dacap:
Just merged :+1:

PR #3537, opened 2022-09-26
Fix {duration} file format always gives an incorrect number (fix #2998)

### LINKED ISSUE #2998: {duration} file format always gives the number 8015567
Hi! I'm new to Aseprite! I was trying to use the {duration} filename formatter when exporting a sequence of pngs and it doesn't work. it just always gives the number 8015567.

I'm trying to export through the GUI, didn't try any of the CLI stuff.

* Aseprite version: v1.2.29-x64
* System: Manjaro Linux, updated but don't know the current version.

Gasparoken:


martincapello (approved):
LGTM!

src/app/filename_formatter.cpp
```diff
@@ -112,7 +113,8 @@ std::string filename_formatter(
   base::replace_string(output, "{layer}", info.layerName());
   base::replace_string(output, "{group}", info.groupName());
   base::replace_string(output, "{slice}", info.sliceName());
-  base::replace_string(output, "{duration}", std::to_string(info.duration()));
+  if (info.duration() && filename.find("{duration}", 0))
+    base::replace_string(output, "{duration}", std::to_string(info.duration()));
```
dacap, line 117: Is this required? Because `base::replace_string()` replaces `{duration}` only if it's found, and check that the same `if ()` is not required for other properties.

Gasparoken: It seems you are right about `&& filename.find("{duration}", 0)`, but `if (info.duration())` is mandatory. In a few minutes I'll check the code without the `&& filename.find("{duration}", 0)`

dacap: I think `if (info.duration())` is not needed. The important thing is that the duration is set on every FilenameInfo usage (as in the [first change](https://github.com/aseprite/aseprite/pull/3537/files#diff-4ce39943a3e0fdfc1c60b12d207229d3ac5ca4cd22142478970da3518d3d2258R680)).

We can add some default values in FilenameInfo fields just in case, e.g.: In https://github.com/aseprite/aseprite/blob/00be9afcea9abf1f109c161f3b5d0002a9f1d371/src/app/filename_formatter.h#L83
Add default values:
```c++
    int m_frame = 0;
    int m_tagFrame = 0;
    int m_duration = 0;
```

But we should check that on every FilenameInfo instance (when possible) the duration field is set correctly (whatever the duration value is).

Gasparoken: Let me investigate, I think that approach crossed my mind but I discarded it for some reason (which I don't remember now).

Gasparoken: Thank you @dacap for letting me think again this PR 😄.
I did that `if`, because in `cli_processor.cpp` the `filename_formatter()` function is called without a frame iteration and then `fnInfo.duration == 0`. All of this results in incorrect processing of `filename_formatter()`
https://github.com/aseprite/aseprite/blob/25fbe786f8353a2ddb57de3bcc5db00066cc9ca6/src/app/cli/cli_processor.cpp#L821-L845

Anyway, I realized that `if (replaceFrame)` will be useful for solving the issue. I already tested.
https://github.com/aseprite/aseprite/blob/59e01493cb6118852b0684fd532f58fa79dd6987/src/app/filename_formatter.cpp#L115-L123

src/app/filename_formatter.cpp
```diff
@@ -112,14 +113,14 @@ std::string filename_formatter(
...
   base::replace_string(output, "{slice}", info.sliceName());
-  base::replace_string(output, "{duration}", std::to_string(info.duration()));
 
   if (replaceFrame) {
     base::replace_string(output, "{tag}", info.innerTagName());
     base::replace_string(output, "{innertag}", info.innerTagName());
     base::replace_string(output, "{outertag}", info.outerTagName());
     replace_frame("{frame", info.frame(), output);
     replace_frame("{tagframe", info.tagFrame(), output);
+    replace_frame("{duration", info.duration(), output);
```
dacap, line 123: Sorry didn't see it correctly, please don't use replace_frame for duration (it has a specific logic for frame numbers):
```suggestion
    base::replace_string(output, "{duration}", std::to_string(info.duration()));
```

Gasparoken: Ok! Thank you.

Gasparoken:
Done. I'll do a test for this soon.

dacap:
Merged 16362cc51 :+1:

PR #3538, opened 2022-09-26
Fix different brush behavior using dynamics+gradient depending if the mask is the fg or bg color

Gasparoken:


Gasparoken:
This completes PR https://github.com/aseprite/aseprite/pull/3525. I did this separate PR in case we need another approach to solve this.

PR #3541, opened 2022-09-28
Fix different brush behavior using dynamics+gradient depending if the mask is the fg or bg color

Gasparoken:


Gasparoken:
The big problem is the singularity of the behavior of `Simple Ink` when `fgColor` is the mask color:
https://github.com/aseprite/aseprite/blob/877ae9eeecb90b0452326a48fcee5026ce5972c7/src/app/tools/active_tool.cpp#L100-L105

The lack of this information transmission within the `Ink` object when we have to choose the right ink processor is a problem: We cannot discriminate if m_type == `Copy Ink` was from `Copy Color + Alpha` ink or is was changed from an original `Simple` ink.
Also, I had to adapt the INDEX behavior when using Dynamics + Gradient to get an expected result when using a palette with alpha content.

src/app/tools/active_tool.cpp
```diff
@@ -116,8 +119,10 @@ Ink* ActiveToolManager::adjustToolInkDependingOnSelectedInkType(
         id = tools::WellKnownInks::Shading;
         break;
     }
-    if (id)
+    if (id) {
       ink = m_toolbox->getInkById(id);
+      ink->setForcedCopyInk(simpleTurnsCopyColor);
```
martincapello, line 124: I made some comments about how we could refactor the implementation of this method...but now I'm not sure that we should actually have this method (along with the wasForcedCopyInk() method). This is because here we are modifying the state of an ink object of the m_toolbox->m_inks map. This map is initialized in the ToolBox constructor, and it seems to me that the idea is that each ink should be immutable...
If we were returning a new instance of an Ink then it shouldn't be a problem...but then I don't know about the  implications of this change.

martincapello (commented):
I left some comments, but I'm still trying to understand the logic a bit deeper.

PR #3545, opened 2022-10-03
fixed frame numbers truncating even when having sufficient space #3350

### LINKED ISSUE #3350: bug: frame numbers
Hi

When making an animation that has more then 99 frames, the frame number 100 is not displayed as "100" but as "00".
The frame "101" is displayed as "01", frame "102" is displayed as "02" ...

See screenshot. 

afbeelding

This is kind of confusing because of following reasons:

- you can confuse frame "1" with frame "01", which is here frame "101". So you must constantly think: what frame is this now because you can't read it directly from the frame number (in case when there are more then 99 frames)?

- if you have over 200 frames, then there will be TWO "01" frames, one for "101" and one for "201".  Also TWO "02" frames, one for "102" and one for "202", ...
Then the confusion will be totally present because there is no way to know what frame is used by "01"!
So you must scroll constantly through the frames too see at what position in the animation you are and what frame it is. That's exactly what the frame number should indicate...

afbeelding

So I suggest frames should display the FULL frame number. This will make it a lot more clear when working with more then 99 frames.

Windows 10
Aseprite 1.3 beta14

jackschedel:

This fixes #3350 
All it does is prevents the frame numbers from being truncated if the timeline is zoomed in enough to where the digits will not clip out of the tile.

### DIFF
--- a/src/app/ui/timeline/timeline.cpp
+++ b/src/app/ui/timeline/timeline.cpp
@@ -2033,11 +2033,23 @@ void Timeline::drawHeaderFrame(ui::Graphics* g, frame_t frame)
     return;
 
   // Draw the header for the layers.
-  const int n = (docPref().timeline.firstFrame()+frame);
-  std::string text = base::convert_to<std::string, int>(n % 100);
-  if (n >= 100 && (n % 100) < 10)
-    text.insert(0, 1, '0');
+  const int n = (docPref().timeline.firstFrame() + frame);
 
+  std::string text;
+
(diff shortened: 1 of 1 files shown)


dacap:
Hi @jackschedel, thanks for this PR. Seeing this change at a glance, the only thing I suspect might be problematic is that it doesn't calculate the required width of the text in pixels vs the available width in pixels which depends on the active theme & font.

SiriasAxel:
Será que me puedas ayudar

PR #3565, opened 2022-10-20
Fix buttonset items' layout calculations and painting (fix #3554)

### LINKED ISSUE #3554: Regression in button set icons in Canvas size dialog and Keyboard Shortcuts > Drag Value
Regression introduced in #3519  

New aspect:

image

Old aspect:

image

Perhaps this can be improved by moving the icons down one pixel.

martincapello:
Fixes several issues when displaying buttonsets. This required a few changes in the sheet.png file.
To verify the changes see:
- Context bar buttons.
- Canvas size buttons.
- Debugger buttons.
- Outline buttons.
- Dynamics buttons.
- Symmetry buttons.
- Pivot point options.

Fix #3554


martincapello:
@Gasparoken Don't review this yet, because I will add some styles to make the buttonset a bit customizable. I will reassign you when done.

dacap:
Just giving a try to this PR, the brush size button is not aligned with the center (I didn't see other buttons):

image image

Previously:

image image

martincapello:
I can see the difference in the 1pixel case, but I don't see any difference with the 16px brush. Also, I think that any even brush size is displayed the same as before. For the odd cases I will investigate a bit, but I would suggest comparing how all the other buttonset items look now.

martincapello:
I will try to explain the issue using the canvas size buttons. This is how it looks with these PR:

image

In the image you can see the dimensions used to calculate the position of the icon inside each button. As you may notice, the last row of buttons look different than the others, despite using the same area for positioning elements. This is due to the following overlapping when painting:
image

In the above image I displaced a bit to the right one of the buttons in the last row to see how they overlap.

Then, for the Brush Type button in the context bar I use the same criteria to calculate the position of the icon:
image

Without this PR, the height used to calculate the position of the icon in the last row is greater than the other rows, then you actually end up having some misalignment issues like this:
image

A possible solution would be to modify a bit how the button set items are displayed to something like:
image

But, giving it a second thought, this may not work as well. Because the brush type button height would end up being 15px, and 15px / 2 = 7px = 14 px / 2. Currently the height used to calculate the position of the icon is the client bound height, which for the last row of buttons has 3 extra pixels, so currently is 16px for the brush type button.

martincapello:
> Sorry, the problem appears only with odd sizes:
> 
> image

Currently
image

So, as I said in my previous comment, it is displaced just 1 pixel, but neither case is aligned perfectly to the center.

martincapello:
That's right. Without this patch, the last column of buttons in a buttonset are one pixel smaller in its width. Look at the 3 buttons after the lock in the original image you posted, the last one is has a width 1 pixel smaller than the other two, which by the way have the same width that when using this patch.

dacap:
Some mockups of possible solutions:

image

Still not sure which one is better. I have to design a lot of details yet.

martincapello:
The one at the top looks better IMO...I think that I could try to make it look like that with a minor change. I will try it and let you know.

martincapello:
Okay, I was working in the minor change I mentioned, but then I had an idea to improve this fix, so I'm on it.

martincapello:
I have more work in progress, these just are some commits I had ready to push.

dacap:
Hi @martincapello, seeing the PR, the commits continued from 003f5c75d2a9dcd4a0be41d1d9c0587672ac00f5, are these changes necessary? mainly the changes of the .png file + making some icon bounds bigger.

martincapello:
This is something I will figure out soon as a result of the work I still have in progress. There is some possibility that some changes to the png and parts are not needed anymore. But still working on this.

martincapello:
Just fixed one commit and removed another. If I found out that the changes in png and parts are not needed I will remove those commits as well.

martincapello:
I'm facing the following issue: Currently the "hot focused" items of a button set that are not in the last row are showed with a different height (1 pixel less):
Screen Shot 2022-12-13 at 09 13 56
This make it look good, but is actually quite inconvenient to mimic with styles/layers because currently we cannot alter the size of a layer, let alone altering the size of just the non-last rows.
However, this kind of workaround to the look and feel have a drawback, the focused elements in the last row have a focus rectangle that is bigger (by 1 pixel) than the other rows.

This is how it would look using styles:
Screen Shot 2022-12-13 at 09 07 52
Screen Shot 2022-12-13 at 09 23 39
As you can see, the last row looks great, but the one above it doesn't.

So, I would like to know how do you think we should fix/workaround this before doing what I believe is the best option.

By the way, I believe that the best option is to have 1 pixel more of separation between rows in button sets, then it would end up looking like this:
Screen Shot 2022-12-13 at 09 32 14
Screen Shot 2022-12-13 at 09 32 47

Another alternative is to modify the "hot focused" slice and icon a bit from this:
Screen Shot 2022-12-13 at 10 00 25

to this:
Screen Shot 2022-12-13 at 09 58 09

So the button set would look like this:
Screen Shot 2022-12-13 at 09 52 12
Screen Shot 2022-12-13 at 09 54 15
However, there is a caveat with this approach...the icon is displaced one pixel upwards when the button is "hot focused", so when it lost focus the icon goes one pixel down to its original position. IMO it doesn't look bad at all, but I think I could try to compensate this by setting a y-offset for the icon layer for the appropriate states if you think this looks better than the other option.

dacap:
I saw some changes in the .png file/sheet in these commits. I'm not sure if they are necessary but I'm going to push some changes in the sheet too. (also if the sprite sheet of the light theme is modified, the dark one must be modified too).

dacap:
Pushed these commits in a new branch https://github.com/aseprite/aseprite/tree/buttonset-issues rebasing from main.

I've squashed 76d2f6177 + 791f225ea.

Some issues I've found so far:

**1. Buttonsets with text only are not working anymore:**

Before:
Screenshot from 2023-02-24 19-58-25

Now:
Screenshot from 2023-02-24 19-56-29

**2. Editable palette button looks strange in the new version**

Before:
Screenshot from 2023-02-24 19-59-46

Now:
Screenshot from 2023-02-24 20-01-07

**3. All these changes will need to be copied to the dark theme (probably when we find a complete solution in the theme xml/png)**

We'll review all these changes on Monday and decide how to continue for the first v1.3.0-rc1.

martincapello:
I'm aware of number 2 and 3. Number 1 I think it is not about buttonsets being text only. There are other text only buttonsets that are displayed correctly (as in the outline dialog for instance). I think it happended to me in other cases, will take a look on monday.

martincapello:
If I recall correctly, it is because the theme is not initialized for that buttonset. We will see on monday

PR #3582, opened 2022-11-01
Show saved undo state in Edit > Undo History window (fix #3578)

### LINKED ISSUE #3578: Checkpoint in the Undo History
Hello, my sincere apologies for the bothering, i wanted to recommend something i tought it would be very useful for the Undo History.

I have been working those days (and as many other do) for longer periods and the constant save always got my head a bit messy because when i didn't liked a change i did in the sprite i didn't know which step in the Undo History was the one i saved last.

And if the `Undo History` showed where the last save occured, like this:

```
Initial State
|_Rectangular Tool
|_Clear
|_Paste
|_Deselect(Last Saved Here)
```

And if you undid the file to its `Initial State` and redid something it would show:

```
Initial State
|_Pencil Tool
--Last Saved Step Undone--
```

Sorry for the bothering and hope this is useful for someone.

dacap:
Tries to implement #3578 showing a "disk" icon in the undo state that matches the sprite in the disk on the Edit > Undo History window.


src/app/commands/cmd_undo_history.cpp
```diff
@@ -94,6 +99,25 @@ class UndoHistoryWindow : public app::gen::UndoHistory,
...
 
+    void updateSavedState() {
+      const auto oldIsAssociatedToFile = m_isAssociatedToFile;
+      const auto oldSavedState = m_savedState;
+
+      m_isAssociatedToFile =
+        (m_doc && m_undoHistory ?
+         m_doc->isAssociatedToFile() && !m_undoHistory->isSavedStateIsLost() :
+         false);
+      // Calculate the stated that is where the sprite was saved.
```
Gasparoken, line 110: 'Calculate the state' instead 'Calculate the stated'

PR #3591, opened 2022-11-11
Ignore .vscode

martincapello:
Add .vscode to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 *.res
 .DS_Store
 .vs
+.vscode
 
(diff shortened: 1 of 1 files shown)


dacap:
We can merge this. Anyway just in case, adding files to `.git/info/exclude` is also possible (to ignore files locally) and doesn't require new entries in `.gitignore` (e.g. I have a lot of files exclude locally related to my own text editor, etc.)

martincapello:
Will exclude it locally.

PR #3630, opened 2022-12-15
Add user data to tiles (fix #3626)

### LINKED ISSUE #3626: Add user data to tiles
We should be able to associated user data/properties to tiles (at least programmatically, the UI will be for the future/other issue).

Add UserData to tiles inside doc::Tileset converting the array of tiles (Image) to a new type (Tile = Image + UserData) or keeping two arrays synchronized (vector<Image> + vector<UserData>).

Gasparoken:
Also added 'WithUserData' inheritance to Tilesets.
Fix aseprite/aseprite#3626

### DIFF
--- a/src/app/cmd/add_tile.cpp
+++ b/src/app/cmd/add_tile.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2020  Igara Studio S.A.
+// Copyright (C) 2019-2022  Igara Studio S.A.
 //
 // This program is distributed under the terms of
 // the End-User License Agreement for Aseprite.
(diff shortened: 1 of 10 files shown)


dacap:
I've merged these changes in two commits with some modifications:

1. Fixes copying Tileset user data (on sprite resize, expand cel, etc.), which includes several bug fixes (ea17eae524f11b0386c89f674c641be7c647e8ed)
2. The new user data for tiles (6d6dfb96a9034df2e128303312a689be2ba2f058)

Some notes about the modifications I've made:

* New `Tileset::MakeCopyWithoutImages` to copy name/userData when cloning tilesets
* In `ExpandCelCanvas`, the tileset user data is copied in `ExpandCelCanvas::getDestTileset()` instead of `ExpandCelCanvas::copySourceTilestToDestTileset()` using the new `Tileset::MakeCopyWithoutImages`
* `Tileset::setUserData()` is kept intact (no `Tileset::setTilesetData`), to avoid confusion modifying a base method (`WithUserData::setUserData`) and we use `Tileset::setTileData/getTileData` to access user data of tiles
* Renamed `Tileset::m_tilesUserData` to `Tileset::m_datas` just to keep the variable name shorter.
* `cmd::AddTile` is not a `doc::WithUserData`/`doc::Object`, these `doc::Object` are used for `doc` objects with `doc::ObjectId`, which are stored in the `doc::objects` map accessible through `doc::get_object()`
  * Removed `ObjectType::Tile` as it's not needed for `cmd::AddTile`
  * Just added `cmd::AddTile::m_userData` to store the tile user data

PR #3636, opened 2022-12-19
Add set of properties to doc::UserData

### LINKED ISSUE #3631: Add set of properties to doc::UserData
Related to https://github.com/aseprite/aseprite/issues/3625

Add a set of sets of typed-properties (including points) to doc::UserData depending on extension ID (empty ID are user-defined properties, just like the current user data field).
Something like

    std::map<std::string, std::map<std::string, std::any>> m_properties;

or

    std::map<std::string, std::map<std::string, Variant>> m_properties;

?
Where Variant = std::variant<bool, int, std::string, gfx::Point, etc….> (these types will be limited to the .aseprite format, and probably will not be possible to extend it after the initial set of types)

martincapello:
Fix #3631


docs/ase-file-specs.md
```diff
@@ -338,13 +340,51 @@ there is an User Data Chunk at the first frame after the Palette Chunk.
...
+                  4 = Has properties
     + If flags have bit 1
       STRING    Text
     + If flags have bit 2
       BYTE      Color Red (0-255)
       BYTE      Color Green (0-255)
       BYTE      Color Blue (0-255)
       BYTE      Color Alpha (0-255)
+    + If flags have bit 4
+      DWORD     Number of properties maps
```
dacap, line 352: Add a DWORD to know the size to skip all the properties data.

docs/ase-file-specs.md
```diff
@@ -338,13 +340,51 @@ there is an User Data Chunk at the first frame after the Palette Chunk.
...
+          + If property type is fixmath::fixed (0x0009)
+            FIXED
+          + If property type is std::string (0x000A)
+            STRING
+          + If property type is std::vector (0x0100)
+            DWORD     Number of elements
+            WORD      Element's type
+            BYTE[]    As many values as the number of elements indicates
+                      Structure depends on the element's type
+          + If property type is UserData::Properties (0x0101)
```
dacap, line 384: Why this kind of data is needed? are not we in the properties map already?

docs/ase-file-specs.md
```diff
@@ -338,13 +340,51 @@ there is an User Data Chunk at the first frame after the Palette Chunk.
...
       BYTE      Color Blue (0-255)
       BYTE      Color Alpha (0-255)
+    + If flags have bit 4
+      DWORD     Number of properties maps
+      + For each properties map:
+        DWORD     Properties maps key (Zero or extension's Entry ID (see External Files Chunk))
+        DWORD     Number of properties
+        + For each property:
+          STRING    Property name
+          WORD      Property type
```
dacap, line 358: We need POINT, SIZE, RECT types.

docs/ase-file-specs.md
```diff
@@ -338,13 +340,51 @@ there is an User Data Chunk at the first frame after the Palette Chunk.
...
+            WORD
+          + If property type is int (0x0005)
+            LONG
+          + If property type is unsigned int (0x0006)
+            DWORD
+          + If property type is long long (0x0007)
+            LONG64
+          + If property type is unsigned long long (0x0008)
+            QWORD
+          + If property type is fixmath::fixed (0x0009)
```
dacap, line 375: `fixmath::fixed` is not a public data (not needed to specify this internal type in this doc). We just reference fixed point numbers as FIXED.

docs/ase-file-specs.md
dacap: Didn't think about the possibility to save maps in one property. Probably the serialization could be simplified in such a way that there are two new types of properties (one of them a virtual property): 1) Open map, 2) Close map

So all the properties between Open Map <-> Close Map belong to the key specified in Open Map.

We have to talk with @Gasparoken about so the Lua API gives this possibility.

src/doc/user_data.h
```diff
@@ -10,12 +10,18 @@
...
 #include <string>
+#include <variant>
+#include <vector>
 
 namespace doc {
 
   class UserData {
   public:
+    typedef std::variant<bool, int, std::string> PrimitiveVariant;
+    typedef std::variant<PrimitiveVariant, std::vector<PrimitiveVariant>> Variant;
```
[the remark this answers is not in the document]
dacap: This is not possible because std::vector<> needs a complete type/with well-known size at compile time. And generally std:: types weren't created to be derived (but that's other issue). Something like std::vector<VariantStruct*> would be possible.

Anyway I'm not sure about this. Probably to keep this (and the serialization) simple, we should just support one level of properties (without maps/objects).

dacap:
For the specs, maybe we can rewrite it to something like this:
```
    + If flags have bit 4
      DWORD     Properties data size (to skip to the next flag if needed)
      DWORD     Number of properties maps
      + For each properties map:
        DWORD     Properties maps key
                    == 0 means user properties
                    != 0 means an extension Entry ID (see External Files Chunk)
        DWORD     Number of properties
        + For each property:
          STRING    Property name
          WORD      Property type
          BYTE      If type=0x0001 (bool)
          BYTE      If type=0x0002 (int8)
          BYTE      If type=0x0003 (uint8)
          SHORT     If type=0x0004 (int16)
          WORD      If type=0x0005 (uint16)
          LONG      If type=0x0006 (int32)
          DWORD     If type=0x0007 (uint32)
          ...
```
(I have added int8 type, probably the bool type is not needed and we can just use the uint8)

martincapello:
I actually like how the "If"s are indented right now (and keeps them consistent with how they are used in other chunks in the same file). So I will make the other changes and then if you feel strongly about changing the "If"s I can change them.

src/app/file/ase_format.cpp
```diff
@@ -1325,6 +1378,142 @@ static void ase_file_write_tileset_chunk(FILE* f, FileOp* fop,
...
+    const std::string& extensionKey = propertiesMap.first;
+    uint32_t extensionId = extensionKey == "" ? 0 : ext_files.to_id.at(extensionKey);
+    fputl(extensionId, f);
+    size+=4;
+
+    size+=ase_file_write_properties(f, properties);
+  }
+  long endPos = ftell(f);
+  fseek(f, startPos, SEEK_SET);
+  fputl(size, f);
```
dacap, line 1513: Check that the size could be calculated with a difference between two ftell() (see how it's done with ase_file_write_start_chunk and ase_file_write_close_chunk)

src/doc/user_data.h
```diff
@@ -9,23 +9,57 @@
...
 
     size_t size() const { return m_text.size(); }
     bool isEmpty() const {
-      return m_text.empty() && !doc::rgba_geta(m_color);
+      return m_text.empty() && !doc::rgba_geta(m_color) && m_propertiesMaps.empty();
     }
 
     const std::string& text() const { return m_text; }
     color_t color() const { return m_color; }
+    const PropertiesMaps& propertiesMaps() const { return m_propertiesMaps; }
```
dacap, line 60: This can be called just `properties` (no need to use Maps). Same for the type `Properties`.

dacap: Oh sorry, I see that the Maps version is the map of maps. My mistake.

docs/ase-file-specs.md
```diff
@@ -251,13 +262,13 @@ Color profile for RGB or grayscale values.
...
-reference external palettes or tilesets.
+A list of external files linked with this file can be found in the first frame. It might be used to
+reference external palettes, tilesets, or extensions that make use of extended properties.
 
     DWORD       Number of entries
     BYTE[8]     Reserved (set to zero)
     + For each entry
-      DWORD     Entry ID (this ID is referenced by tilesets or palettes)
+      DWORD     Entry ID (this ID is referenced by tilesets, palettes, or extended properties)
       BYTE[8]   Reserved (set to zero)
```
dacap, line 272: Probably we'll need to use one of these reserved bytes to specify the kind of external file (the idea of this chunk was to have a way to pre-load external files, but in this case for extensions it's for other usage, so we'll need a way to make a difference)

martincapello: I was appending an "extension:" prefix to differentiate them. But sure, I can use one of those reserved bytes if you think it would be better. Just let me know if you prefer to use a reserved byte.

dacap: We can use one BYTE for the type (0 external palette, 1 external tileset, 2 extension name for properties)

martincapello:
We might have to change -DCMAKE_OSX_DEPLOYMENT_TARGET to 10.13 instead of 10.9. Right? Seems that the macos build if failing because of this

docs/ase-file-specs.md
martincapello: 👍

dacap:
We compile official releases targeting macOS 10.9 (it's the minimum macOS version we support), the CMAKE_OSX_DEPLOYMENT_TARGET is set to 10.9 so the CI can detect problems that will be detected in the release in other way.

The C++ runtime on macOS 10.9 is not fully standard compliant, so sometimes we have to find some other alternatives (e.g. [with std::optional](https://github.com/aseprite/laf/commit/81622fcbb9e4a0edc14a02250c387bd6fa878708)).

martincapello:
> We compile official releases targeting macOS 10.9 (it's the minimum macOS version we support), the CMAKE_OSX_DEPLOYMENT_TARGET is set to 10.9 so the CI can detect problems that will be detected in the release in other way.
> 
> The C++ runtime on macOS 10.9 is not fully standard compliant, so sometimes we have to find some other alternatives (e.g. [with std::optional](https://github.com/aseprite/laf/commit/81622fcbb9e4a0edc14a02250c387bd6fa878708)).

I see, will investigate an alternative. Thanks!

martincapello:
Closing in favor of https://github.com/aseprite/aseprite/pull/3645

PR #3641, opened 2022-12-27
Add a map of custom properties to be used by extensions

### LINKED ISSUE #3631: Add set of properties to doc::UserData

martincapello:
Fix #3631

### DIFF
--- a/src/doc/user_data.h
+++ b/src/doc/user_data.h
@@ -9,23 +9,57 @@
 #pragma once
 
 #include "doc/color.h"
+#include "fixmath/fixmath.h"
+#include "gfx/point.h"
+#include "gfx/size.h"
+#include "gfx/rect.h"
 
+#include <map>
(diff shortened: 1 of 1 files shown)


dacap:
Cherry-picked https://github.com/aseprite/aseprite/commit/856839030607739669beca3bc99277b5a5d9c500

PR #3648, opened 2022-12-30
Lua API for user/extensions properties

dacap:
Add a Lua API to access user data properties (https://github.com/aseprite/api/issues/88)

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -1,5 +1,5 @@
 # Aseprite
-# Copyright (C) 2018-2022  Igara Studio S.A.
+# Copyright (C) 2018-2023  Igara Studio S.A.
 # Copyright (C) 2001-2018  David Capello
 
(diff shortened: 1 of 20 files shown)


dacap:
Some weeks ago we talked and found a possible syntax to access properties (user properties and extension properties), e.g.:

```lua
local spr = Sprite(4, 4)
spr.properties.a = 10
spr.properties("ext1").a = 20
assert(spr.properties.a == 10)
assert(spr.properties("ext1").a == 20)
```

The only problem with this is that we cannot assign a whole table to the extension properties:

```lua
spr.properties = { a=10 }          -- This is possible
spr.properties("ext1") = { a=20 }  -- This isn't possible
```

A solution would be 1) find an alternative syntax, 2) or add a special way to do this like:

```lua
spr.properties = { a=10 }
spr.properties("ext1", { a=20 })
```

An alternative syntax would be something like `spr.extensionProperties["ext1"] = { a=10 }`.

PR #3665, opened 2023-01-09
Add add/erase tiles from a tileset using Lua (fix #3663)

### LINKED ISSUE #3663: A way to add/erase tiles from a tileset using Lua

Gasparoken:
First approach to add/erase Tiles in a Tileset.

# Tileset:addTile()
`destinationTileset:addTile(sourceImage)`

Inserts an image copy of the given sourceImage into destinationTileset. This method generates undo information, so you could use it as an individual operation or in a [transaction](https://github.com/aseprite/api/blob/main/api/app.md#apptransaction).
If the sourceImage is duplicated, this function do nothing.

# Tileset:eraseTile()
`destinationTileset:eraseTile(image)`
`destinationTileset:eraseTile(tileIndex)`

Removes a tile from a tileset which image matches with some existent tile. Also, the argument can be an index (from 1 to tileset.size - 1) instead of an image. This method generates undo information, so you could use it as an individual operation or in a [transaction](https://github.com/aseprite/api/blob/main/api/app.md#apptransaction).

fix aseprite/aseprite#3663

### DIFF
--- a/src/app/script/tileset_class.cpp
+++ b/src/app/script/tileset_class.cpp
@@ -8,6 +8,8 @@
 #include "config.h"
 #endif
 
+#include "app/cmd/add_tile.h"
+#include "app/cmd/remove_tile.h"
 #include "app/script/docobj.h"
 #include "app/script/engine.h"
 #include "app/script/luacpp.h"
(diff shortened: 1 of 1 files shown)


dacap:
I think we can create a similar API to the existent methods in Sprite metatable:
```lua
local tile = Sprite:newTile(tileset[, tileIndex]) -- returns a Tile metatable
Sprite:deleteTile(tile)  -- tile is a Tile metatable
```
If newTile doesn't receive a tile index, the new tile is added at the end (as an empty image). We then can modify the image using `tile.image`.

Gasparoken:
Whether the index in newTile() corresponds to an existing Tile. Should this function overwrite the tile? Warn us through error that it is overwriting? Or put the mosaic at the end of the tileset?

dacap:
There shouldn't be warning, the addTile() should insert a new tile in the given index (moving all other tiles to the next item). E.g.

```
before: 0, 1, 2, 3, 4
...addTile(ts, 2)...
after: 0, 1, [2], 3, 4, 5
```
The `[2]` is the new tile, and the old 2, 3, 4 are now 3, 4, 5.

PR #3669, opened 2023-01-10
Add new/delete tiles from a tileset using Lua (fix #3663)

### LINKED ISSUE #3663: A way to add/erase tiles from a tileset using Lua

Gasparoken:
New approach according https://github.com/aseprite/aseprite/pull/3665#issuecomment-1377188572

## Sprite:newTile()
```lua
local tile = sprite:newTile(tileset [, tile_index])
```
Inserts an empty tile into the given `tileset` at a given `tile_index`. If `tile_index` is not provided, the new tile is added to the end of the tileset. This method generates undo information, so you could use it as an individual operation or in a [transaction](https://github.com/aseprite/api/blob/main/api/app.md#apptransaction).

## Sprite:deleteTile()
```lua
Sprite:deleteTile(tile)
```
Removes a tile from a tileset. This method generates undo information, so you could use it as an individual operation or in a [transaction](https://github.com/aseprite/api/blob/main/api/app.md#apptransaction).

## Tile()
```lua
local tile = Tile(tileset, tile_index)
```
A tile object just holds an index and a reference to a tileset.

fix aseprite/aseprite#3663

### DIFF
--- a/src/app/script/sprite_class.cpp
+++ b/src/app/script/sprite_class.cpp
@@ -12,6 +12,7 @@
 #include "app/app.h"
 #include "app/cmd/add_layer.h"
 #include "app/cmd/add_slice.h"
+#include "app/cmd/add_tile.h"
 #include "app/cmd/assign_color_profile.h"
 #include "app/cmd/clear_cel.h"
 #include "app/cmd/convert_color_profile.h"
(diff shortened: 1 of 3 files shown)


Gasparoken:
Tests pending

dacap:
I'll merge this later without the Tile() constructor (as it's accessible with tileset:tile(i))

dacap:
Merged here: https://github.com/aseprite/aseprite/commit/822357869a1567ff02df9ee75a7a8aff8ac7e42e
This will require an update to the API docs ([next-version branch](https://github.com/aseprite/api/tree/next-version)).

PR #3673, opened 2023-01-12
Add warning to Sprite:newTile() and Sprite:deleteTile()

Gasparoken:
Now, if the tileset does not belong to the sprite object, the lua engine throws an error. Also added a warning when tile_index = 0 on Sprite:deleteTile().

### DIFF
--- a/src/app/script/sprite_class.cpp
+++ b/src/app/script/sprite_class.cpp
@@ -688,9 +688,10 @@ int Sprite_newTile(lua_State* L)
 {
   auto sprite = get_docobj<Sprite>(L, 1);
   auto ts = get_docobj<Tileset>(L, 2);
-  if (!ts) {
+  if (!ts)
     return luaL_error(L, "empty argument not allowed and must be a Tileset object");
-  }
+  if (ts->sprite() != sprite)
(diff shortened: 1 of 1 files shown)


src/app/script/sprite_class.cpp
```diff
@@ -710,12 +711,19 @@ int Sprite_deleteTile(lua_State* L)
...
     ti = lua_tointeger(L, 3);
-  else
+  else {
     ts = get_tile_index_from_arg(L, 2, ti);
+    temp = " on the TileObj";
+  }
   if (!ts)
     return luaL_error(L, "Sprite:deleteTile() needs a Tileset or Tile as first argument");
+  if (ts->sprite() != sprite)
+    return luaL_error(L, "the tileset%s belongs to another sprite", temp.c_str());
```
dacap, line 724: Probably we can simplify the error message just to avoid creating a std::string, e.g.
```suggestion
    return luaL_error(L, "the tileset belongs to another sprite");
```

Gasparoken: No problem. Done!

dacap:
Merged e0aaa76e9c081c4bad869756f75fe348457e986f

PR #3679, opened 2023-01-17
Fix vector serialization and reduce size of integer types

### LINKED ISSUE #3677: Possible bug/lost all work saving a vector of properties
`ase_file_write_property_value()` doesn't check that all values of the vector are of the same type, and that might create a problem/lost work.

A possible new test for userdata_codec.lua:

```lua
-- Test bug saving vector with different types inside
do
  local spr = Sprite(32, 32)
  spr.properties.a = { 3, "hi" }

  spr:saveAs("_test_userdata_codec_3.aseprite")
  spr:close()

  spr = Sprite{ fromFile="_test_userdata_codec_3.aseprite" }
  assert(#spr.properties.a == 2)
  assert(spr.properties.a[1] == 3)
  assert(spr.properties.a[2] == "hi")
end
```

There are several things:

- [ ] Who check the variant type? I have added a couple of ASSERTs about this: [1](https://github.com/aseprite/aseprite/blob/e0aaa76e9c081c4bad869756f75fe348457e986f/src/app/file/ase_format.cpp#LL1500) [2](https://github.com/aseprite/aseprite/blob/e0aaa76e9c081c4bad869756f75fe348457e986f/src/app/script/values.cpp#L416) [3](https://github.com/aseprite/aseprite/blob/e0aaa76e9c081c4bad869756f75fe348457e986f/src/app/script/values.cpp#L512)

- [ ] How to reduce the size of integer representation, e.g. most of the time we don't need to save a 64-bit integer (we can use 8-bit int for small values, etc.) https://github.com/aseprite/aseprite/blob/e0aaa76e9c081c4bad869756f75fe348457e986f/src/app/file/ase_format.cpp#LL1435C16-L1435C16

For vectors we should check all values and store the smaller int type to store all same types.

martincapello:
- [x] Fixes vector serialization to support elements of differents types all mixed in the same collection of elements. For this I have introduced two modes vector serialization:
  - Elems type == actual type. Used when all the elements have the same type. This stores the element's type only once.
  - Elems type == 0: Used when all the types don't have the same type. This stores the type for each element.
- [x] Reduces the size of the integers stored in .aseprite files as much as possible.

Fix #3677

### DIFF
--- a/docs/ase-file-specs.md
+++ b/docs/ase-file-specs.md
@@ -413,9 +413,16 @@ there is an User Data Chunk at the first frame after the Palette Chunk.
             RECT
           + If type==0x0011 (vector)
             DWORD     Number of elements
-            WORD      Element's type
-            BYTE[]    As many values as the number of elements indicates
-                      Structure depends on the element's type
+            WORD      Element's type.
+            + If Element's type == 0 (all elements are not of the same type)
+              For each element:
+                WORD      Element's type
+                BYTE[]    Element's value. Structure depends on the
+                          element's type
+            + Else (all elements are of the same type)
+              For each element:
+                BYTE[]    Element's value. Structure depends on the
+                          element's type
           + If type==0x0012 (nested properties map)
             DWORD     Number of properties
             BYTE[]    Nested properties data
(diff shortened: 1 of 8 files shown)


martincapello:
I'm working on the integer size reduction now.

dacap:
Ok, I'll take a look to this later, I liked the idea to use type=0 in vector to avoid creating two kind of vectors type.

dacap:
I think we must use the type=0, because we cannot change the already stored properties (as .aseprite files are already being saved with the current spec).

dacap:
Merged 8b547adfb + 47a1c407c + 062d1d922 + a7b5ab24b + 70a388177 + 8f5e8256c + 33a6a1c97 + cbe62e09e

PR #3698, opened 2023-02-19
Create SECURITY.md

DimensionalSpace:

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,23 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 1.0   | :white_check_mark: |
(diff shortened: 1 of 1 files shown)


DimensionalSpace:
Might want to change 1.0 to 1.2

dacap:
Hi @DimensionalSpace, we're not sure if we should handle security issues with public information or sending a private email to support@aseprite.org. We've to investigate a little how other projects on GitHub handle this and the security policy.

DimensionalSpace:
OK so what should I do?

dacap:
We've to investigate an alternative way to report vulnerabilities, something similar to [curl](https://github.com/curl/curl/security/policy). I'm going to close this PR because it needs to define the process from our side.

PR #3705, opened 2023-02-23
Add std:: to nullptr_t to avoid compilation error in linux

martincapello:


martincapello:
Just saw this fix is already made. Closing.

PR #3716, opened 2023-03-02
Removed alpha threshold in octree map algorithm (fix #3715)

### LINKED ISSUE #3715: Palette From Sprite Bug
Palette from sprite does not see colors with alpha less than 16 on RGB to palette index mapping: Default (Octree).

* Aseprite version: v1.3-rc1-dev
* System: Windows, 10 Pro 21H2, OS Build: 19044.1766

Gasparoken:
fix #3715

### DIFF
--- a/src/doc/octree_map.cpp
+++ b/src/doc/octree_map.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (c) 2020-2022 Igara Studio S.A.
+// Copyright (c) 2020-2023 Igara Studio S.A.
 //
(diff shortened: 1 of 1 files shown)


dacap:
I think if `imageIsRGBA` is true we should at least filter out colors with alpha == 0.

Gasparoken:
alpha == 0 are filtered now.

dacap:
Merged https://github.com/aseprite/aseprite/commit/31ca1b6350053b1a96341e065438b390f2b78906

PR #3720, opened 2023-03-03
Fix crash on undo command after creating new tilemap layer with custom properties added via plugin.

Gasparoken:
There may be something deeper to this, for now this solution avoids a major crash when it comes to plugin custom properties.


martincapello (approved):

dacap:
I'll try to find another solution, because `doc::get<doc::WithUserData>(m_objId)` shouldn't fail in any case.

dacap:
Fixed with b43f2a34280b64c95a8941bb43ba7b87a77e9c3b

PR #3756, opened 2023-03-15
Fix onion skin loop tag

### LINKED ISSUE #3739: Onion Skin Loop through tag frames not working correctly
The onion skin option "Loop through tag frames" has stopped working recently. While having the setup to onion skin the last and next frames, when at the start of a tagged group of frames, it will show the last frame out of the tag, and when at the end of a tagged group of frames, it will show the first two frames

https://user-images.githubusercontent.com/43629991/223775023-844c28a0-1b85-49fa-8703-a900b9bfda90.mp4

* Aseprite version: Aseprite v1.3-rc1-x64 Steam [Beta]
* System: Windows 10 Pro 10.0.19044 Build 19044 x64

martincapello:
Fix #3739
Also introduces a fix to make TwoLoopsInCascadeReverse test work as expected.

### DIFF
--- a/src/doc/playback.cpp
+++ b/src/doc/playback.cpp
@@ -119,10 +119,7 @@ frame_t Playback::nextFrame(frame_t frameDelta)
       handleMoveFrame(step);
     handleEnterFrame(step, false);
 
-    if (frameDelta > 0)
-      --frameDelta;
-    else if (frameDelta < 0)
-      ++frameDelta;
+    frameDelta -= step;
   }
 
   PLAY_TRACE("  } =", m_frame,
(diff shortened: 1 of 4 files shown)


src/doc/playback.cpp
```diff
@@ -119,10 +119,7 @@ frame_t Playback::nextFrame(frame_t frameDelta)
       handleMoveFrame(step);
     handleEnterFrame(step, false);
 
-    if (frameDelta > 0)
-      --frameDelta;
-    else if (frameDelta < 0)
-      ++frameDelta;
+    frameDelta -= step;
```
dacap, line 122: I think in our tests and code `frameDelta` is always +1 or -1, but what happen if it's +2 or -2? (it looks like the while could be infinite). Is it necessary to replace the original code?

dacap: Oh sorry, it looks fine, in some way so something like frame += frameDelta, I think this is fine.

PR #3770, opened 2023-03-22
[lua] Add blend modes to Image:drawImage()

Gasparoken:
Also added tests.

This commit includes a default `NORMAL` blend mode (instead `SRC`). `NORMAL` blend mode is more intuitive in image composition.

To get the old `drawImage()` behavior you should add `opacity` and `blend mode` as final arguments of function:
- `Image:drawImage(image) `   -->   `Image:drawImage(image, Point(0, 0), 255, BlendMode.SRC)`
- `Image:drawImage(image, position)` --> `Image:drawImage(image, position, 255, BlendMode.SRC)`

### DIFF
--- a/src/app/script/image_class.cpp
+++ b/src/app/script/image_class.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2022  Igara Studio S.A.
+// Copyright (C) 2018-2023  Igara Studio S.A.
 // Copyright (C) 2015-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -15,6 +15,7 @@
 #include "app/context.h"
 #include "app/doc.h"
 #include "app/file/file.h"
+#include "app/script/blend_mode.h"
 #include "app/script/docobj.h"
 #include "app/script/engine.h"
 #include "app/script/luacpp.h"
@@ -40,6 +41,8 @@ namespace app {
 namespace script {
 
 namespace {
+
+static ImageBufferPtr buf; // TODO non-thread safe
 
 struct ImageObj {
   doc::ObjectId imageId = 0;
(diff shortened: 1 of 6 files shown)


dacap:
Looks good so far, the only thing I'd avoid is overloading the copy member function (`ImageImpl::copy` should be kept intact/without overloading), and the new `doc::copy_image()` should be called `doc::blend_image` and move the implementation there in `primitives.cpp` (instead of having the implementation inside `ImageImpl::copy`, etc.).

A side-effect of this PR as it's, every place where a simple copy/blit between images was used, now does a blend using blender functions (even when just a copy is needed, the SRC blend function is used).

Gasparoken:
> ... should be called doc::blend_image and move the implementation there in primitives.cpp (instead of having the implementation inside ImageImpl::copy, etc.).

I trying to do so, but the `address_t` type and pixel address operations are private types/functions of `image_impl.h` (not accesible from `primitives.cpp`). Could I keep a  `doc::blend()` function on `image_impl.h` to be used inside `doc::blend_image()` on `primitives.cpp` (in the same way `doc::copy()` on `image_impl.h` is used inside `doc::copy_image` on `primitives.cpp`)?

dacap:
@Gasparoken you need to do something similar to (e.g.) [mask_image()](https://github.com/aseprite/aseprite/blob/1f7f3677f274b76a5b2a9f772a334fa482d23dfa/src/app/util/cel_ops.cpp#L75) and [mask_image_templ()](https://github.com/aseprite/aseprite/blob/1f7f3677f274b76a5b2a9f772a334fa482d23dfa/src/app/util/cel_ops.cpp#L59), using LockImageBits to access pixel data and a template function to generalize the algorithm.

Gasparoken:
Ready for review.

dacap:
Please check that I'm the author of the commit and you the committer, not sure how I ended up there :sweat_smile:

src/app/script/image_class.cpp
```diff
@@ -247,23 +250,41 @@ int Image_drawImage(lua_State* L)
   auto obj = get_obj<ImageObj>(L, 1);
   auto sprite = get_obj<ImageObj>(L, 2);
   gfx::Point pos = convert_args_into_point(L, 3);
+
+  int opacity = 255;
+  if (lua_isinteger(L, 4))
+    opacity = std::max(0, std::min(255, int(lua_tointeger(L, 4))));
```
dacap, line 256:
```suggestion
    opacity = std::clamp(int(lua_tointeger(L, 4)), 0, 255);
```

src/doc/layer_tilemap.cpp
```diff
@@ -21,7 +21,10 @@ LayerTilemap::LayerTilemap(Sprite* sprite, const tileset_index tsi)
   , m_tileset(sprite->tilesets()->get(tsi))
   , m_tilesetIndex(tsi)
 {
-  ASSERT(m_tileset);
+  // m_tileset can be temporarily nullptr when tsi is 0, which means
+  // that the layer was read from doc::read_layer() and the tsi is 0
+  // until we can read that field.
+  ASSERT(m_tileset || tsi == 0);
```
dacap, line 27: I think these lines are from other commit (?)

Gasparoken:
Ready for review.

dacap:
Merged 637632eafbfb052ea3b6e18b6764c75e88c00e9e with minor changes

PR #3775, opened 2023-03-27
Move window relayout before inserting the window to the manager

### LINKED ISSUE #3746: Calling `print` function within the `onpaint` method causes and additional dialog to be rendered
Demo code to replicate the issue:

```lua
if app.apiVersion < 21 then
    print("API version is lower than 21")
    return
end

local dialog = Dialog("Canvas Resize Test")

dialog --
:canvas{
    width = 200,
    height = 200,
    onpaint = function(ev)
        local context = ev.context

        print(context.width, context.height)
    end
} 

dialog:show()
```

Preview:

canvas-print-issue

* Aseprite version: 1.3-rc1-x64, Steam
* System: Windows 11

martincapello:
Fix #3746

### DIFF
--- a/src/app/ui/preview_editor.cpp
+++ b/src/app/ui/preview_editor.cpp
@@ -183,6 +183,7 @@ PreviewEditorWindow::PreviewEditorWindow()
   , m_refFrame(0)
   , m_aniSpeed(1.0)
   , m_relatedEditor(nullptr)
+  , m_opening(false)
 {
   setAutoRemap(false);
   setWantFocus(false);
@@ -343,6 +344,8 @@ Editor* PreviewEditorWindow::previewEditor() const
 
 void PreviewEditorWindow::updateUsingEditor(Editor* editor)
 {
+  if (m_opening) return;
+
   if (!m_isEnabled || !editor) {
     hideWindow();
     m_relatedEditor = nullptr;
@@ -357,8 +360,11 @@ void PreviewEditorWindow::updateUsingEditor(Editor* editor)
   Doc* document = editor->document();
   Editor* miniEditor = (m_docView ? m_docView->editor(): nullptr);
 
-  if (!isVisible())
+  if (!isVisible()) {
+    m_opening = true;
     openWindow();
+    m_opening = false;
+  }
 
   // Document preferences used to store the preferred zoom/scroll point
   auto& docPref = Preferences::instance().document(document);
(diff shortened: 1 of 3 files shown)


martincapello:
Seems that this doesn't fix the problem...I realized that there are sometimes that the issue doesn't happen when using the menu items (this can happen several times in a row, making it hard to tell if it is actually fixed). Now I'm using a keyboard shortcut to run the script to reproduce the issue, which seems to be much more effective in doing so.

martincapello:
Hopefully this is the fix 🤞

martincapello:
The fix stayed the same, I just had to adjust some code in the preview editor, because the issue was an infinite loop as follows:
1) `PreviewEditorWindow::updateUsingEditor()` is called from `setPreviewEnabled()` when the command to show the Preview is pressed by the user.
2) `Window::openWindow()` is called from `PreviewEditorWindow::updateUsingEditor`. 
3) `Window::openWindow()` calls `Manager::getDefault()->_openWindow()` when the window has no parent.
4) `Widget::layout()` is called from `Manager::getDefault()->_openWindow()`.
5) Eventually the call stack (started at `Widget::layout`) reaches`PreviewEditorWindow::onWindowResize()`.
6) `PreviewEditorWindow::updateUsingEditor()` is called from `PreviewEditorWindow::onWindowResize()`, jump to step 2 and repeat until crash.

Why this happened with the manager's fix? Because the window's parent (checked in step 3) is assigned when the manager calls its `insertChild()` method, which now happens **after** the window's relayout. For this reason I moved the openWindow call (along with the hideWindow call for consistency) to other place.

dacap:
I've tested the latest change and I've found a regression:

1. Create a new file
2. Open the preview
3. Use Ctrl+Tab to switch to the Home tab: the Preview window dissapear (this is expected)
4. Then press Ctrl+Tab again: the Preview window doesn't appear (in the current main the Preview window appears automatically)

As a side note, if some change is finally merged, it would be nice to have some comments/details in the code and the commit description about why the change is necessary (so in the future we don't move the code to the original location). At the moment the commit description says what the diff shows.

martincapello:
I put back the openWindow call to where it was. Added a new flag to indicate that the preview editor is being opened so I can use it to break the recursion.

dacap:
Merged 8841f3da32b4332674825e3de8db0f37b5fe4183

PR #3786, opened 2023-03-31
Hide small preview window when going full screen preview

### LINKED ISSUE #3762: [Feature Request/Bug Fix] Closing Preview Window in Full-Screen Preview Mode.
My sincere apologies for the bothering once again, but i came today to suggest something that felt... Weird for me i don't know for how much.

When going into Full-Screen Preview Mode (F8), everything's alright, perfect even, but the lil Preview Window is still there (if you had it on during your session) and the problem is that the Preview Window captures the mouse.

So even if you can move freely with the mouse, the Preview Window can and WILL capture its input.

Also, isn't a bit redundant to have a Preview Window inside another? haha

Yet again, my sincere apologies for the bothering, and hope this is useful.

Aseprite Version: Latest(v1.3-rc1-dev)
OS: Windows 10 - Latest Update

martincapello:
Fix #3762

### DIFF
--- a/src/app/commands/cmd_fullscreen_preview.cpp
+++ b/src/app/commands/cmd_fullscreen_preview.cpp
@@ -11,6 +11,7 @@
 
 #include "ui/ui.h"
 
+#include "app/app.h"
 #include "app/commands/command.h"
 #include "app/commands/commands.h"
 #include "app/context.h"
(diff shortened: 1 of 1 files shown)


dacap:
Merged in a7e155a391984a6772c0ba8a57259b994e5d6855 with minor changes (replaced `inline` with `static` for non-member functions, almost never is required `inline`, only for non-member functions in header files)

PR #3790, opened 2023-04-04
Fix backward compatibility with Image:drawImage(image, x, y) was broken (fix #3788)

### LINKED ISSUE #3788: Backward compatibility with Image:drawImage(image, x, y) was broken
As we've [added the new opacity and blendMode args to Image:drawImage()](https://github.com/aseprite/aseprite/commit/637632eafbfb052ea3b6e18b6764c75e88c00e9e), it broke a possible way to call Image:drawImage(image, x, y) because [convert_args_into_point()](https://github.com/aseprite/aseprite/commit/637632eafbfb052ea3b6e18b6764c75e88c00e9e#diff-ccc6426892aafbcb558911d4cd5c5b517e004859852ccdfe7950065887f57959R252) can convert a Point or two consecutive integer values.

Some extensions fixed this using the call with the Point: https://github.com/thkwznk/aseprite-scripts/commit/da35917b38278b53d4d241dfe0b607d5fe924b84 but I think it would be possible to keep the backward compatibility if we detect that the second argument is or not a Point.

Gasparoken:
fix #3788

### DIFF
--- a/src/app/script/image_class.cpp
+++ b/src/app/script/image_class.cpp
@@ -251,14 +251,23 @@ int Image_drawImage(lua_State* L)
   auto sprite = get_obj<ImageObj>(L, 2);
   gfx::Point pos = convert_args_into_point(L, 3);
 
+  // Arguments index fix to support the following cases:
+  //   - Image:drawImage(image, x, y, opacity, blendMode)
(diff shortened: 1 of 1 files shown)


dacap:
Merged 2b3372df0e45c25986f1c7ce6c31150fd9b58eaa

PR #3792, opened 2023-04-04
Improve message error when loading a sequence of png files whose color modes don't match

### LINKED ISSUE #3751: Improve error when loading a sequence of png files w/different color modes
Saw on Twitter today this error:

https://github.com/aseprite/aseprite/blob/23557a190b4f5ab46c9b3ddb19146a7dcfb9dd82/src/app/file/png_format.cpp#L283

It can be reproduced if we load a sequence of PNG files (e.g. `1.png` and `2.png`), but `1.png` is RGB and `2.png` is Indexed and we load the sequence.

```
file_sequence_image 200x100
Error loading frame 2 from file "2.png"
```

martincapello:
Fix #3751

### DIFF
--- a/src/app/file/file.cpp
+++ b/src/app/file/file.cpp
@@ -1309,8 +1309,10 @@ ImageRef FileOp::sequenceImage(PixelFormat pixelFormat, int w, int h)
   else {
     sprite = m_document->sprite();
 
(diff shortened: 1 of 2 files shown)


dacap:
Merged e577852eb45da4b3fa402023fae99760007343fe

PR #3794, opened 2023-04-05
Add enable panning and zooming with an opened filter window (fix #2563)

### LINKED ISSUE #2563: Enable panning and zooming with an opened window
Enable panning and zooming (by pressing<kbd>mousewheel</kbd>) while replace color window is open.

This would be useful when previewing color changes before submitting them.

Gif below shows how you can pan and zoom while there no window, but cannot when window is open.
kST3cpbb3j

* Aseprite version: V.1.2.25-x64, Steam
* System: Windows

Gasparoken:
Replace Color, Color Curve, Outline, Convolution Matrix, etc are Filter Windows.

fix #2563

This fix enables:
- Panning and zooming.

Additional features for the future:
- Possibility to change the active Cel.

### DIFF
--- a/docs/ase-file-specs.md
+++ b/docs/ase-file-specs.md
@@ -43,6 +43,7 @@ ASE files use Intel (little-endian) byte order.
 * `TILE`: **Tilemaps**: Each tile can be a 8-bit (`BYTE`), 16-bit
   (`WORD`), or 32-bit (`DWORD`) value and there are masks related to
   the meaning of each bit.
+* `UUID`: A Universally Unique Identifier stored as `BYTE[16]`.
 
 ## Introduction
 
@@ -446,6 +447,8 @@ The data of this chunk is as follows:
             DWORD     Number of properties
             BYTE[]    Nested properties data
                       Structure is the same as indicated in this loop
+          + If type==0x0013
+            UUID
 
 ### Slice Chunk (0x2022)
 
--- a/src/app/commands/filters/filter_window.cpp
+++ b/src/app/commands/filters/filter_window.cpp
@@ -17,7 +17,10 @@
 #include "app/ini_file.h"
 #include "app/modules/gui.h"
 #include "app/pref/preferences.h"
+#include "app/tools/tool_box.h"
 #include "app/ui/editor/editor.h"
+#include "app/ui/context_bar.h"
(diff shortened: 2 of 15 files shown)


dacap:
@martincapello could you please test this?

martincapello:
It does enable panning and zooming while the filter window is opened. The only thing I noted is that the filter preview is applied to the current visible portion of the sprite, then if you preview the filter and then pan the viewport you will see that part of the sprite was not affected by the preview so you have to uncheck and check again the preview checkbox to refresh the viewport's image in its new position.

martincapello:
Now it works like a charm! I just found a minor issue (not sure if it is an actual issue)...when some of the filter windows are open and the preview windows is also open, the preview window cannot be moved by dragging its title bar, however resizing is allowed, as well as clicking the buttons on the title bar. I think we could merge this and create a separate issue for the preview window glitches.

Gasparoken:
Thank you!
About:
> ... when some of the filter windows are open and the preview windows is also open, the preview window cannot be moved by dragging its title bar, however resizing is allowed, as well as clicking the buttons on the title bar. I think we could merge this and create a separate issue for the preview window glitches.

Yes, it can be taken as another separate issue.

dacap:
I don't know why this PR contains 6 commits, I'll review 181a0884417a0da4d96e24ab1e710c9e73fe22b2 only cherry-picking it.

dacap:
> Sorry, I'm not sure why it happened.

Don't worry about this, I was able to cherry pick the commit. For the next time I would recommend you to avoid deleting/recreating the branch (checkout can be tricky and do things you don't expect). Instead you can try to do a [git rebase](https://git-scm.com/docs/git-rebase) with the latest `main` commit (anyway I would recommend you to backup the branch, e.g. create a local branch in the original commit before you rebase, because it's a complex operation that can fail with conflicts to solve, etc.). You could give a try to rebase with some local repo/tests before using it in production.

About the fix, it works great :+1: I'll merge it as it is. Then I have noticed some performance issues on Linux because it generates a lot of mouse messages for scrolling, so I'm planning to add some extra code to restart the preview after some timer (to consume/ignore several mouse movements).

dacap:
Merged in 2369f36322bcfdebe4f16d87d5f3b4f3551cbfc7

PR #3800, opened 2023-04-10
[lua] Add option to clear particular regions of an image via Image:clear function (fix #3799)

### LINKED ISSUE #3799: Allow for clearing a part of the image from Lua API
I think a straightforward option to clear a part of the image is missing from the Image API, something similar to `Image:clear(rectangle, [color])` would be very useful.

* Aseprite version: N/A
* System: N/A

Gasparoken:
New use:
Image::clear(rectangle, [color])

fix #3799

### DIFF
--- a/src/app/script/image_class.cpp
+++ b/src/app/script/image_class.cpp
@@ -220,6 +220,16 @@ int Image_clear(lua_State* L)
   auto obj = get_obj<ImageObj>(L, 1);
   auto img = obj->image(L);
   doc::color_t color;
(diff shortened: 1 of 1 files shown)


Gasparoken:
If this fix is OK, I'll update the API docs.

Gasparoken:
Not ready for review.

Gasparoken:
Ready for review.

dacap:
Merged in d4d18c99be9e8c252ae08632908c17c6b356f92a with a modified implementation a test case to check that the used color when it's not specified is the mask color (this was failing w/the proposed impl).

PR #3805, opened 2023-04-11
Include build type in ccache key

janisozaur:
This should lead to better cache utilisation, smaller caches


dacap:
Thanks @janisozaur! Sorry for the late response, I've just cherry-picked your change here: https://github.com/aseprite/aseprite/commit/af19d818069aaea6b4fe1bd130b09d7c5748c054

PR #3813, opened 2023-04-12
Add canvas autoScaling feature (fix #3748)

### LINKED ISSUE #3748: UI Elements Scaling doesn't apply to the canvas
The Preferences option `UI Element Scaling` doesn't apply to the canvas, leaving it to the script developer to multiply all values/sizes by the scaling value. Without it, elements drawn on the canvas will be misaligned after changing the scale of UI elements.

* Aseprite version: 1.3-rc1-x64, Steam

martincapello:
Fix #3748 by applying the following changes:
 - Add a new autoScaling property to the dialog's canvas so we can scale mouse positions and the surface used to draw its content.
 - Update app.script.Theme to hold the current ui scaling and take charge of calculating the styleMetrics and returning the dimensions downscaled by the ui scaling factor, so it will return the same value independently of the ui scaling. 
 - Add a new m_uiscale member to GraphicsContext to hold the current ui::guiscale() and use it to automatically scale drawing/measuring operations as needed.

### DIFF
--- a/src/app/script/app_theme_object.cpp
+++ b/src/app/script/app_theme_object.cpp
@@ -18,18 +18,75 @@ namespace script {
 
 namespace {
 
-struct Theme { };
-struct ThemeDimension { };
+class Theme {
+public:
+  Theme(int uiscale) : m_uiscale(uiscale) {}
+
+  gfx::Border styleMetrics(const std::string& id) const {
+    auto theme = skin::SkinTheme::instance();
+    if (!theme) return gfx::Border(0);
+
+    ui::Style* style = theme->getStyleById(id);
+    if (!style) return gfx::Border(0);
+
+    ui::Widget widget(ui::kGenericWidget);
+    auto border = theme->calcBorder(&widget, style);
+    if (m_uiscale > 1)
+      border /= m_uiscale;
+
+    return border;
+  }
+
+  int getDimensionById(const std::string& id) const {
+    int value = skin::SkinTheme::instance()->getDimensionById(id);
+    if (m_uiscale > 1)
+      value /= m_uiscale;
+
+    return value;
+  }
+
+private:
+  int m_uiscale;
+};
+
+struct ThemeDimension {
+  Theme* theme;
+
+  ThemeDimension(Theme* theme) : theme(theme) { }
+
+  int getById(const std::string& id) {
+    return theme->getDimensionById(id);
+  }
+};
+
 struct ThemeColor { };
+
+#ifdef ENABLE_UI
+void push_border_as_table(lua_State* L, const gfx::Border& border)
+{
+  lua_newtable(L);
+  {
+    lua_newtable(L);
+    setfield_integer(L, "left", border.left());
+    setfield_integer(L, "top", border.top());
+    setfield_integer(L, "right", border.right());
+    setfield_integer(L, "bottom", border.bottom());
+  }
+  lua_setfield(L, -2, "border");
+}
+#endif
 
 int ThemeDimension_index(lua_State* L)
 {
+  [[maybe_unused]]
+  auto themeDimension = get_obj<ThemeDimension>(L, 1);
+
   const char* id = lua_tostring(L, 2);
   if (!id)
     return luaL_error(L, "id in app.theme.dimension.id must be a string");
 
 #ifdef ENABLE_UI
-  const int value = skin::SkinTheme::instance()->getDimensionById(id);
+  const int value = themeDimension->getById(id);
   lua_pushinteger(L, value);
 #else
   lua_pushinteger(L, 0);
@@ -55,38 +112,15 @@ int ThemeColor_index(lua_State* L)
 int Theme_styleMetrics(lua_State* L)
 {
   [[maybe_unused]]
-  auto t = get_obj<Theme>(L, 1);
+  auto theme = get_obj<Theme>(L, 1);
 
   const char* id = lua_tostring(L, 2);
   if (!id)
     return 0;
 
 #ifdef ENABLE_UI
-  auto theme = skin::SkinTheme::instance();
-  if (!theme)
-    return 0;
-
-  ui::Style* style = theme->getStyleById(id);
-  if (!style)
-    return 0;
-
-  ui::Widget widget(ui::kGenericWidget);
-  gfx::Border border = theme->calcBorder(&widget, style);
-
-  lua_newtable(L);
-  {
-    lua_newtable(L);
-    lua_pushinteger(L, border.left());
-    lua_setfield(L, -2, "left");
-    lua_pushinteger(L, border.top());
-    lua_setfield(L, -2, "top");
-    lua_pushinteger(L, border.right());
-    lua_setfield(L, -2, "right");
-    lua_pushinteger(L, border.bottom());
-    lua_setfield(L, -2, "bottom");
-  }
-  lua_setfield(L, -2, "border");
-
+  gfx::Border border = theme->styleMetrics(id);
+  push_border_as_table(L, border);
   return 1;
 #else  // ENABLE_UI
   return 0;
@@ -95,7 +129,8 @@ int Theme_styleMetrics(lua_State* L)
 
 int Theme_get_dimension(lua_State* L)
 {
-  push_obj<ThemeDimension>(L, ThemeDimension());
+  auto theme = get_obj<Theme>(L, 1);
+  push_new<ThemeDimension>(L, theme);
   return 1;
 }
 
@@ -140,9 +175,9 @@ void register_theme_classes(lua_State* L)
   REG_CLASS_PROPERTIES(L, Theme);
 }
 
-void push_app_theme(lua_State* L)
+void push_app_theme(lua_State* L, int uiscale)
 {
-  push_obj<Theme>(L, Theme());
+  push_new<Theme>(L, uiscale);
 }
 
 } // namespace script
--- a/src/app/script/canvas_widget.cpp
+++ b/src/app/script/canvas_widget.cpp
@@ -38,7 +38,7 @@ void Canvas::stopKeyEventPropagation()
   s_stopKeyEventPropagation = true;
 }
 
-Canvas::Canvas() : ui::Widget(Type())
+Canvas::Canvas() : ui::Widget(Type()), m_autoScaling(true)
 {
 }
 
@@ -52,8 +52,14 @@ void Canvas::callPaint()
   m_surface->drawRect(m_surface->bounds(), p);
 
   // Draw only on resize (onPaint we draw the cached m_surface)
-  GraphicsContext gc(m_surface);
-  gc.font(AddRef(font()));
+  GraphicsContext gc(m_surface, m_autoScaling ? ui::guiscale() : 1);
+  if (m_autoScaling) {
+    auto theme = skin::SkinTheme::get(this);
+    gc.font(AddRef(theme->getUnscaledFont(font())));
+  }
+  else
+    gc.font(AddRef(font()));
+
   Paint(gc);
 }
 
@@ -67,6 +73,16 @@ void Canvas::onInitTheme(ui::InitThemeEvent& ev)
   else
     bg = gfx::rgba(0, 0, 0);
   setBgColor(bg);
+}
+
+template <typename T>
+static T makeScaled(T* msg, const gfx::Point& offset)
+{
+  static_assert(std::is_base_of<ui::Message, T>::value, "type parameter must derive from ui::Message");
+  auto result = T(msg->type(), *msg, ((msg->position() - offset) / ui::guiscale()) + offset);
+  result.setRecipient(static_cast<ui::Message*>(msg)->recipient());
+  result.setDisplay(static_cast<ui::Message*>(msg)->display());
+  return result;
 }
 
 bool Canvas::onProcessMessage(ui::Message* msg)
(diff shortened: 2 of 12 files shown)


src/app/script/app_theme_object.cpp
```diff
@@ -54,39 +104,15 @@ int ThemeColor_index(lua_State* L)
 
 int Theme_styleMetrics(lua_State* L)
 {
-  [[maybe_unused]]
```
dacap, line 57: Don't remove `[[maybe_unused]]` (the same for all the following cases).

src/app/script/app_theme_object.cpp
```diff
@@ -18,18 +18,68 @@ namespace script {
...
+    return value;
+  }
+
+private:
+  int m_uiscale;
+};
+struct ThemeDimension {
+  Theme* theme;
+
+  int getById(const std::string& id) {
```
dacap, line 54: Add const
```suggestion
  int getById(const std::string& id) const {
```

src/app/script/app_theme_object.cpp
```diff
@@ -140,9 +167,9 @@ void register_theme_classes(lua_State* L)
   REG_CLASS_PROPERTIES(L, Theme);
 }
 
-void push_app_theme(lua_State* L)
+void push_app_theme(lua_State* L, int uiscale)
 {
-  push_obj<Theme>(L, Theme());
+  push_obj<Theme>(L, Theme(uiscale));
```
dacap, line 172:
```suggestion
  push_new<Theme>(L, uiscale);
```

src/app/script/canvas_widget.cpp
```diff
@@ -69,6 +69,26 @@ void Canvas::onInitTheme(ui::InitThemeEvent& ev)
...
+                                     mouseMsg->pressure());
+  scaledMouseMsg->setRecipient(mouseMsg->recipient());
+  scaledMouseMsg->setCommonAncestor(mouseMsg->commonAncestor());
+  scaledMouseMsg->setDisplay(mouseMsg->display());
+  scaledMouseMsg->setPropagateToChildren(mouseMsg->propagateToChildren());
+  scaledMouseMsg->setPropagateToParent(mouseMsg->propagateToParent());
+  scaledMouseMsg->setFromFilter(mouseMsg->fromFilter());
+  return scaledMouseMsg;
+}
+
```
dacap, line 91: There is a ctor to copy a MouseMessage with a different position (also I've changed the return type to a value instead of a pointer):
```suggestion
static ui::MouseMessage makeScaled(ui::MouseMessage* mouseMsg, const gfx::Point& offset)
{
  return ui::MouseMessage(
    mouseMsg->type(), *mouseMsg,
    ((mouseMsg->position() - offset) / ui::guiscale()) + offset);
}
```

src/app/script/canvas_widget.cpp
```diff
@@ -99,6 +119,11 @@ bool Canvas::onProcessMessage(ui::Message* msg)
 
     case ui::kMouseMoveMessage: {
       auto mouseMsg = static_cast<ui::MouseMessage*>(msg);
+      std::unique_ptr<ui::MouseMessage> scaledMouseMsg;
+      if (m_autoScaling) {
+        scaledMouseMsg = makeScaled(mouseMsg, bounds().origin());
+        mouseMsg = scaledMouseMsg.get();
+      }
       MouseMove(mouseMsg);
```
dacap, line 127: We can change this to (with the new makeScaled()):
```suggestion
      auto mouseMsg = *static_cast<ui::MouseMessage*>(msg);
      if (m_autoScaling)
        mouseMsg = makeScaled(&mouseMsg, bounds().origin());
      MouseMove(&mouseMsg);
```
I think it's better to use values directly (instead of shared pointers here).

src/app/script/canvas_widget.cpp
```diff
@@ -150,14 +195,19 @@ void Canvas::onResize(ui::ResizeEvent& ev)
...
 
     if (!m_surface ||
         m_surface->width() != w ||
         m_surface->height() != h) {
-      m_surface = os::instance()->makeSurface(w, h);
 
+      if (m_autoScaling) {
+        w = w/ui::guiscale() + (w % ui::guiscale());
+        h = h/ui::guiscale() + (h % ui::guiscale());
+      }
```
dacap, line 208: The w/h should be modified before the if() condition.

dacap (changes requested):
In general it looks great. I've made some minor comments. Here some bigger things:

1. I think `ui::kTouchMagnifyMessage` on Canvas widget needs a similar adjustment in its position as mouse messages (you can try this on macOS trying to zoom with the touchpad).

2. I don't like the fact that this autoScaling will be the default behavior and the most important drawing functions (GraphicsContext::fillText, GraphicsContext::drawImage, GraphicsContext::drawThemeRect) must re-scale images. One possible solution would be to keep an non-scaled version of the font and the theme skin in `app::skin::SkinTheme` so we can use them directly.

src/app/script/graphics_context.h
```diff
@@ -118,8 +119,14 @@ class GraphicsContext {
     m_surface->clipPath(m_path);
   }
 
+  int uiscale() {
```
dacap, line 122:
```suggestion
  int uiscale() const {
```

src/app/script/app_theme_object.cpp
martincapello: This is the only attribute I removed, so I'm not sure what do you mean with "all the following cases". Anyway thank you for clarifying in another comment why this attribute is needed.

src/app/script/canvas_widget.cpp
martincapello: Good catch! Will fix this. Thanks!

martincapello:
> 1. I think `ui::kTouchMagnifyMessage` on Canvas widget needs a similar adjustment in its position as mouse messages (you can try this on macOS trying to zoom with the touchpad).

Yes I think you are right.

> 2. I don't like the fact that this autoScaling will be the default behavior and the most important drawing functions (GraphicsContext::fillText, GraphicsContext::drawImage, GraphicsContext::drawThemeRect) must re-scale images. One possible solution would be to keep an non-scaled version of the font and the theme skin in `app::skin::SkinTheme` so we can use them directly.

Sure, I think it would be definitively better if the memory overhead is negligible. At least it should be better than the CPU overhead I guess.

src/app/script/canvas_widget.cpp
martincapello: @dacap Okay I have just tested this and the issue here is that the recipient is not copied (this crashes the program when the mouse message is received). Should I add it to the constructor code or should I assign the recipient after creating the copy (in makeScaled function)?

dacap: Copy the receipt and display, don't add them in the ctor. Search for "mouseMsg2" in the source code, in other occasions this is done (and sometimes the receipt or the display changes).

martincapello:
I have pushed the suggested changes. Hope the bigger changes are close to what you had in mind.

src/app/ui/skin/skin_theme.h
```diff
@@ -65,6 +65,9 @@ namespace app {
       os::Font* getDefaultFont() const override { return m_defaultFont.get(); }
       os::Font* getWidgetFont(const ui::Widget* widget) const override;
       os::Font* getMiniFont() const { return m_miniFont.get(); }
+      os::Font* getUnscaledFont(os::Font* font) const {
+        return m_unscaledFonts.at(font) ? m_unscaledFonts.at(font).get() : font;
```
martincapello, line 69: Update this to avoid throwing exceptions.

dacap:
I'm giving a try to the [Custom Widgets](https://github.com/aseprite/Aseprite-Script-Examples/blob/main/Custom%20Widgets.lua) example and it needs this change:

```diff
diff --git a/Custom Widgets.lua b/Custom Widgets.lua
index c938615..bfd5a3a 100644
--- a/Custom Widgets.lua	
+++ b/Custom Widgets.lua	
@@ -79,8 +79,9 @@ local customWidgets = {
 
 dlg:canvas{
     id = "canvas",
-    width = 160,
-    height = 75,
+    width = 160 * app.uiScale,
+    height = 75 * app.uiScale,
     onpaint = function(ev)
         local ctx = ev.context
 ```

Not sure if the width/height fields should be auto-scaled by the app.uiScale too when `autoScaling = true` (probably yes).

**Comment by martincapello** [comment]
Just added automatic size adjustment for canvas when autoScaling is true.

**Comment by dacap** [comment]
Merged in 0958573cacae5f559b0f3cf67e25a5472e380787 with some changes (see below). We should add some information about this in the [api docs](https://github.com/aseprite/api) (gc.theme, autoScaling param, etc.)

```diff
diff --git a/src/app/script/app_theme_object.cpp b/src/app/script/app_theme_object.cpp
index a38c87310..264627eae 100644
--- a/src/app/script/app_theme_object.cpp
+++ b/src/app/script/app_theme_object.cpp
@@ -20,7 +20,7 @@ namespace {
 
 class Theme {
 public:
-  Theme(int uiscale) : m_uiscale(uiscale) {}
+  Theme(int uiscale) : m_uiscale(uiscale) { }
 
   gfx::Border styleMetrics(const std::string& id) const {
     auto theme = skin::SkinTheme::instance();
@@ -41,7 +41,6 @@ public:
     int value = skin::SkinTheme::instance()->getDimensionById(id);
     if (m_uiscale > 1)
       value /= m_uiscale;
-
     return value;
   }
 
@@ -54,7 +53,7 @@ struct ThemeDimension {
 
   ThemeDimension(Theme* theme) : theme(theme) { }
 
-  int getById(const std::string& id) {
+  int getById(const std::string& id) const {
     return theme->getDimensionById(id);
   }
 };
diff --git a/src/app/script/canvas_widget.cpp b/src/app/script/canvas_widget.cpp
index 8211ae329..e6f1f1881 100644
--- a/src/app/script/canvas_widget.cpp
+++ b/src/app/script/canvas_widget.cpp
@@ -38,7 +38,7 @@ void Canvas::stopKeyEventPropagation()
   s_stopKeyEventPropagation = true;
 }
 
-Canvas::Canvas() : ui::Widget(Type()), m_autoScaling(true)
+Canvas::Canvas() : ui::Widget(Type())
 {
 }
 
diff --git a/src/app/script/canvas_widget.h b/src/app/script/canvas_widget.h
index 63f352ea9..fbaf21bbd 100644
--- a/src/app/script/canvas_widget.h
+++ b/src/app/script/canvas_widget.h
@@ -39,7 +39,7 @@ public:
     m_autoScaling = v;
   }
 
-  bool isAutoScaling() const{
+  bool isAutoScaling() const {
     return m_autoScaling;
   }
 
@@ -69,7 +69,7 @@ private:
   // Flag used to indicate that the canvas will scale all the drawing operations
   // according to the UI scale's preferences setting. So the user doesn't have to
   // take care about the current scale when writing scripts.
-  bool m_autoScaling;
+  bool m_autoScaling = true;
 };
 
 } // namespace script
diff --git a/src/app/script/graphics_context.cpp b/src/app/script/graphics_context.cpp
index ce9621df7..0dfe6d350 100644
--- a/src/app/script/graphics_context.cpp
+++ b/src/app/script/graphics_context.cpp
@@ -84,7 +84,8 @@ void GraphicsContext::drawImage(const doc::Image* img,
 void GraphicsContext::drawThemeImage(const std::string& partId, const gfx::Point& pt)
 {
   if (auto theme = skin::SkinTheme::instance()) {
-    skin::SkinPartPtr part = m_uiscale > 1 ? theme->getUnscaledPartById(partId) : theme->getPartById(partId);
+    skin::SkinPartPtr part = (m_uiscale > 1 ? theme->getUnscaledPartById(partId):
+                                              theme->getPartById(partId));
     if (part && part->bitmap(0)) {
       auto bmp = part->bitmap(0);
       m_surface->drawRgbaSurface(bmp, pt.x, pt.y);
@@ -95,7 +96,8 @@ void GraphicsContext::drawThemeImage(const std::string& partId, const gfx::Point
 void GraphicsContext::drawThemeRect(const std::string& partId, const gfx::Rect& rc)
 {
   if (auto theme = skin::SkinTheme::instance()) {
-    skin::SkinPartPtr part = m_uiscale > 1 ? theme->getUnscaledPartById(partId) : theme->getPartById(partId);
+    skin::SkinPartPtr part = (m_uiscale > 1 ? theme->getUnscaledPartById(partId):
+                                              theme->getPartById(partId));
     if (part && part->bitmap(0)) {
       ui::Graphics g(nullptr, m_surface, 0, 0);
 
diff --git a/src/app/ui/skin/skin_theme.cpp b/src/app/ui/skin/skin_theme.cpp
index 29933d3af..99150aa57 100644
--- a/src/app/ui/skin/skin_theme.cpp
+++ b/src/app/ui/skin/skin_theme.cpp
@@ -388,6 +388,10 @@ void SkinTheme::loadSheet()
   if (!newSheet)
     throw base::Exception("Error loading %s file", sheet_filename.c_str());
 
+  // TODO Change os::Surface::applyScale() to return a new surface,
+  //      avoid loading two times the same file (even more, if there
+  //      is no scale to apply, m_unscaledSheet must reference the
+  //      same m_sheet).
   m_unscaledSheet = os::instance()->loadRgbaSurface(sheet_filename.c_str());
 
   // Replace the sprite sheet
diff --git a/src/app/ui/skin/skin_theme.h b/src/app/ui/skin/skin_theme.h
index f8a26351e..2791d8193 100644
--- a/src/app/ui/skin/skin_theme.h
+++ b/src/app/ui/skin/skin_theme.h
@@ -67,7 +67,10 @@ namespace app {
       os::Font* getMiniFont() const { return m_miniFont.get(); }
       os::Font* getUnscaledFont(os::Font* font) const {
         auto it = m_unscaledFonts.find(font);
-        return it != m_unscaledFonts.end() ? it->second.get() : font;
+        if (it != m_unscaledFonts.end())
+          return it->second.get();
+        else
+          return font;
       }
 
       ui::Cursor* getStandardCursor(ui::CursorType type) override;
```

martincapello:
Okay, I have created an [issue](https://github.com/aseprite/api/issues/101) to track those api docs additions.

PR #3829, opened 2023-04-25
Add warning when loading unsupported property type and mark the file as read-only

### LINKED ISSUE #3811: Open files as read-only
It would be nice to have an option to load files as read-only, or load then as read-only by default when something strange happens in the loading process, mainly when loading .aseprite files, e.g. some unsupported information/chunk inside the file was discarded, so we don't want to overwrite the content by default.

### LINKED ISSUE #3812: Add warning when we load unsupported properties
Add forward compatibility to possible future properties showing a warning sign when an unknown property type is found (in this way the whole user data/properties chunk could be discarded, but the user knows that something strange happened and a new version of Aseprite is needed).

Possibly related to #3811 to avoid losing user-defined/plugin-defined properties.

martincapello:
Fix #3812, fix #3811.

This PR shows the following message when an incompatibility issue is found when loading a .aseprite file:
Screen Shot 2023-04-25 at 11 45 25

EDIT: The last message in the image above was modified after uploading it. So it now says: "Document will be marked as read-only. Please upgrade Aseprite to the latest version to be able to save this document without losing data."

Note: I have tested this PR by commenting out the `case` block for type 18 (USER_DATA_PROPERTY_TYPE_PROPERTIES) in readPropertyValue method in aseprite_decoder.cpp, and using a file that has some user data custom properties.

Then it shows the sprite and it adds [Read-Only] to the window's title bar as follows:
Screen Shot 2023-04-25 at 11 45 42

Then if the user makes some changes and tries to save the sprite, it shows:
Screen Shot 2023-04-25 at 11 46 04

If the user tries to close the sprite after some modification was made, it first shows the regular "save before closing" warning dialog:
warning

But then when the user presses the "Save" button it shows the following message in the console and opens the warning again:
Screen Shot 2023-04-25 at 13 08 57

So, the only options the user actually have are "Don't save" or "Cancel".

If the user uses the "Save As..." dialog, the file can be saved in the same path with a different name or in another path with any name.

I'm happy to make any change to this if needed.

### DIFF
--- a/src/app/file/file.cpp
+++ b/src/app/file/file.cpp
@@ -920,6 +920,11 @@ void FileOp::operate(IFileOpProgress* progress)
         setError("Error loading data file: %s\n", ex.what());
       }
     }
+
+    if (hasIncompatibilityError()) {
+      setError(fmt::format(Strings::alerts_load_file_with_incompatibilities(),
+               m_incompatibilityError).c_str());
+    }
   }
   // Save //////////////////////////////////////////////////////////////////////
   else if (m_type == FileOpSave &&
@@ -935,6 +940,11 @@ void FileOp::operate(IFileOpProgress* progress)
                     get_app_download_url()).c_str());
     }
 #endif
+
+    if (m_document && m_document->isReadOnly()) {
+      setError(fmt::format(Strings::alerts_cannot_overwrite_readonly()).c_str());
+      return;
+    }
 
     // Save a sequence
     if (isSequence()) {
@@ -1090,12 +1100,14 @@ FileOp::~FileOp()
   delete m_seq.palette;
 }
 
-void FileOp::createDocument(Sprite* spr)
+void FileOp::createDocument(Sprite* spr, bool readOnly)
 {
   // spr can be NULL if the sprite is set in onPostLoad() then
 
   ASSERT(m_document == NULL);
   m_document = new Doc(spr);
+  if (readOnly)
+    m_document->markAsReadOnly();
 }
 
 void FileOp::postLoad()
@@ -1368,6 +1380,18 @@ void FileOp::setError(const char *format, ...)
     if (!m_error.empty() && m_error.back() != '\n')
       m_error.push_back('\n');
     m_error += buf_error;
+  }
+}
+
+void FileOp::setIncompatibilityError(const std::string& msg)
+{
+  // Concatenate the new error
+  {
+    std::lock_guard lock(m_mutex);
+    // Add a newline char automatically if it's needed
+    if (!m_incompatibilityError.empty() && m_incompatibilityError.back() != '\n')
+      m_incompatibilityError.push_back('\n');
+    m_incompatibilityError += msg;
   }
 }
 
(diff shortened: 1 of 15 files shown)


src/app/file/file.cpp
```diff
@@ -920,6 +920,11 @@ void FileOp::operate(IFileOpProgress* progress)
         setError("Error loading data file: %s\n", ex.what());
       }
     }
+
+    if (hasIncompatibilityError()) {
+      setError("Incompatibility error found: %s.\nDocument will be marked as read-only." \
+               " Please upgrade Aseprite to the latest version to be able to save this document without losing data.", m_incompatibilityError.c_str());
+    }
```
dacap, line 927: Change the text to:
```suggestion
    if (hasIncompatibilityError()) {
      setError("Incompatibility error found:\n\n  %s.\n\n"
               "This file will be opened as read-only."
               " Please upgrade Aseprite to the latest version to be able to save changes without losing data.",
               m_incompatibilityError.c_str());
```
And probably we should add these strings in the en.ini strings.

src/app/file/file.cpp
```diff
@@ -936,6 +941,11 @@ void FileOp::operate(IFileOpProgress* progress)
     }
 #endif
 
+    if (m_document && m_document->isReadOnly()) {
+      setError("Cannot save/overwrite a document marked as read-only\n");
```
dacap, line 945: Change to (also this could be in strings/en.ini file)
```suggestion
      setError("Cannot save/overwrite a read-only sprite. Use File > Save As option.\n");
```

dacap (changes requested):
It looks good, some minor changes. But there is a major problem (I guess): the sprite is opened as read-only but we can modify it. I think the sprite should be truly read-only (we shouldn't be able to edit it).

martincapello:
With the latest changes, in addition to the messages' modifications:
- If the user tries to make some change using some tool it shows a tip message saying "Read-only sprite cannot be modified".
- The "arrow + forbidden sign" mouse cursor is set for the editor when a read-only sprite is active, and menú commands that modify the sprite are disabled.
- Commands executed through Transaction, now throw the following exception for read-only documents:
Screen Shot 2023-05-04 at 11 47 03

dacap:
Looks good :+1: I'll try to merge this today + adding some extra changes: 1) A disabled File > Save is confusing (even more, the user could think that Ctrl+S is working but it doesn't say nothing that the file is not being saved, fails silently), 2) Some read-only messages/alert don't say how to solve the situation (File > Save As).

I'll try to think how to present these messages.

dacap:
Merged in 3f581a5dfa1328fba1c0033d0515ef2b8993695d with some changes after that:

* Minor formatting changes 387b46d72ba7891f00ee00d0c4f69117bcae2a9d
* More possible incompatible objects in .aseprite decoder 32ed6622d45b748a0340877e3658ff1eafa05299
* Avoid crash in Timeline when no layers are loaded 10cdad9c588b9eacd5a8d6608d8284ad1562fde9 (because now we could load a sprite without compatible layers)
* A centralized warning message c32b9b07a8630d6af9b9b1f8b5b871920a0bcaec (the one in https://github.com/aseprite/aseprite/issues/3811#issuecomment-1544129432 )

With this latest change the file is loaded `impossibleToBackToSavedState` so the program shows that the file is not equal to the state in the disk:

image

The bad side about this is that when we want to close the file Aseprite shows a warning message to save changes.

PR #3854, opened 2023-05-11
[lua] Add Image:flip()

Gasparoken:
Added:

- Image:flip()
- Constants: FlipType.HORIZONTAL and FlipType.VERTICAL
- Tests

Pending: API documentation.

### DIFF
--- a/src/app/script/engine.cpp
+++ b/src/app/script/engine.cpp
@@ -17,6 +17,7 @@
 #include "app/doc_range.h"
 #include "app/pref/preferences.h"
 #include "app/script/blend_mode.h"
+#include "app/script/flip_type.h"
 #include "app/script/luacpp.h"
 #include "app/script/require.h"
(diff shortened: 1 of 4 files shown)


dacap:
I'll try to merge this change, but there is a problem: the Image_flip() assumes that the image is inside a sprite and adds undo information to the current sprite when we flip an image that might not be part of the sprite (other methods check if the image is owned by a sprite or not checking the cel pointer). The test does the same, it fails with a `"No active document to execute a transaction"` if the `app.sprite == nil`.

I'll fix that and merge this for rc4.

dacap:
Merged in f3ed22e1a6313a3ac4285c3580960f5de3d6f785 + a98ad46a55d1d9eeb186ed4f5fd1756610e93300 + 637d71a276c45fe9671473a6bfc0c64e4d653972

PR #3871, opened 2023-05-15
Fix crash in app::tools::BaseInk::prepareForPointShape

martincapello:
Fix https://github.com/aseprite/aseprite/issues/3338


dacap:
Merged in c902c563e355d9732c49fcf14cca61f583173f58

No need for ; after a member function definition, and no need to define a virtual ~() = default destructor.

PR #3872, opened 2023-05-15
Fix timeline is not repainted/invalidated when we switch onion skin option (F3 key) (fix #3806)

### LINKED ISSUE #3806: Timeline is not repainted/invalidated when we switch onion skin option (F3 key)
If we enable onion skin (F3) key the indicators in the Timeline don't appear until it's invalidated:

E.g.

image

Instead of seeing this:

image

Gasparoken:
fix #3806


Gasparoken:
I'm going to do a new PR for this.

PR #3880, opened 2023-05-17
Test UUID user data property serialization

martincapello:
This PR adds a Lua test to check that undoing and redoing setting a UUID user data property works as expected.


dacap:
Debugging the test it doesn't use the code of properties serialization because the property is saved as it's (std::variant) in the undo information. So it doesn't fail without d5406cbc554411327e5a4438f490ee137ba3c5ee anyway. 

Well, the fix should work to recover files, not sure how to test this (there is no API yet to save/access recovery data). I'll left this PR open for the future.

Issue #3904, opened 2023-05-31
New (new) timeline


dacap:
We're going to work on several changes for the timeline that might introduced new bugs, and we'll try to use the `beta` branch (Beta Steam channel for customers) to release v1.3.1-beta1 (the beta will exist for short periods of times, it will not be as the old beta that took years to finally be merged with `main`).

This is prioritized from several features from https://github.com/aseprite/Attachment-System/issues/9, but some features are long-waited from users:

- [ ] As a first step we would like to introduce the possibility to collapse tags  (#920, https://community.aseprite.org/t/request-collapse-tags/3349)
- [ ] Then we should be able to show only one tag in the timeline (https://github.com/aseprite/Attachment-System/issues/55)
- [ ] And stack/preview several tags in any order (tags might have an "main animation" tag so we can differentiate/arrange "big tags/main animations" easily)
- [ ] We can then review more issues from https://github.com/aseprite/Attachment-System/issues/9
- [ ] Finally we could review some features from the old #293 (hide specific frames, #244?)

To do this we'll do some refactors and introduce a new `view` module/library to convert the Timeline/DocRange as more abstract (non-UI) concepts. The Timeline must be an abstract view of the layers/frames that could be handled/tested programmatically from scripts (the same for the range, and drag-and-drop operations).

Issue #920, opened 2015-12-30
Expandable/collapsable loop tags bar


oceanhahn:
I have a sprite with a few overlapping tags in the animation/layers pane, but if two loop tags begin on the same frame (or even on a nearby frame), the tags can overlap, making the occluded one inaccessible without (as far as I know) deleting the superposed loop. Which I don't want to do.

If we could have some way of either selecting a loop from, maybe a context menu of overlapping ones (generated on-the-spot when multiple tag labels overlap) or have the space loop labels appear in grow and shrink automatically when tags <i>would</i> overlap (example below), that would be dandy!

I made this mock-up of what I was thinking, if it's of any help:
![aseprite tabs](https://cloud.githubusercontent.com/assets/11544843/12056338/b5eb025e-af35-11e5-8659-05979015a8c9.png)

The upper section is the existing form; the lower part is the same data presented in a format I'd imagined based on the interface as it is now. There's also a new button there, that collapses the tags area into a single row (as it currently exists). Perhaps when labels overlap, only the first letter is shown, and the label is offset according to length (with the labels for longer loops occurring last, so that they're above at least some of their frames)? 
It might also be convenient if the loop selection brackets were coloured after the loop labels!

Thanks for considering it!

dacap:
Related to #768

jpnk:
@oceanhahn i'm not sure about collapsing the tag area - tags/animations are of the same importance as layers. they are the third dimension, after layers and frames.
maybe if the tag area grows too big let the user collapse it into a selector (a dropdown list?). i need to reflect on this some more.

oceanhahn:
The basic purpose I'd thought for collapsing the tag/loop area is so that, if the total height of the tags became too deep (that is, it intruded too greatly upon the editor space), the user could press a button to roll them up (and, later, press it again to restore them). Basically, it just hustles them out of the way when they aren't needed and brings them back with a single click, in much the same way that [Tab] hides and reveals the entire pane.

I'm not super-sure how useful it might be considering that, to be meaningfully cumbersome, there would probably have be be at least three stacked sets of layers (fewer, I suppose, if ASEprite is running in a smallish window or on a small or low-res screen), but it might be nice to be able to switch between a full profile of loop labels and a compressed one during viewing.

jpnk:
that's an interesting idea about [Tab]. \<Modifier>+[Tab] could be used to toggle the state of the tag area.

[...]

PR #3905, opened 2023-05-31
Implement new timeline with collapsed/hidden tags

dacap:
First progress for #3904 and #920 

One commit ("Add support for variable frame width in Timeline") comes from #3583 and we're not sure if it will be finally required.


dacap:
I'll change the target branch to beta.

PR #3910, opened 2023-06-02
Fix unnecessary propagation of onProcessMessage across all open editors in kKeyDownMessage cases

### LINKED ISSUE #3884: Brush preview issues entering on Drag Value
Some times in certain conditions the edge preview of the erase tool is printed on the viewport just pressing Control+Option. This problem is just visual, no sprite modifications are done.

Steps to reproduce:
- New Indexed or Grayscale Sprite 200x200 (RGBA works fine)
- 100% zoom
- Pick the Erase tool, brush type: circle, size 20
- Press 'Alt' ('option' on macOS = quick Eyedropper tool) and release
- Do not move the cursor, press 'control' and hold + 'Alt' and hold (entering to drag value to modify brush size).
- Result: printing of the edge preview of eraser.

https://github.com/aseprite/aseprite/assets/32165475/d6b9e494-29f4-4d1d-9577-c6b9c39413d2

Similar problem with the pencil tool, but instead of printing the preview border, the brush preview doesn't show:

https://github.com/aseprite/aseprite/assets/32165475/c55fe7c6-2b32-423e-814b-23afa2d65fc6

* Aseprite version: 1.3rc4 dev
* System: macOS Ventura

Gasparoken:
The pressed key should work only in the active editor.

Also this fix solves ‘Brush preview issues entering on Drag Value' (fix #3884)

### DIFF
--- a/src/app/ui/editor/editor.cpp
+++ b/src/app/ui/editor/editor.cpp
@@ -2078,13 +2078,13 @@ bool Editor::onProcessMessage(Message* msg)
       }
 #endif  // ENABLE_DEVMODE
 
-      if (m_sprite) {
+      if (m_sprite && (isActive() || hasMouse())) {
         EditorStatePtr holdState(m_state);
         bool used = m_state->onKeyDown(this, static_cast<KeyMessage*>(msg));
 
-        updateToolLoopModifiersIndicators();
-        updateAutoCelGuides(msg);
         if (hasMouse()) {
+          updateToolLoopModifiersIndicators();
+          updateAutoCelGuides(msg);
           updateQuicktool();
           setCursor(mousePosInDisplay());
         }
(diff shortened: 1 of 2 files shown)


dacap:
Testing this patch, if we put two editors side by side the one that is not active will not show the brush preview. I'm not sure if the condition is "isActive()" or more like "(isActive() || hasMouse())", maybe it requires some extra testing for these cases.

dacap:
Assigning @martincapello as reviewer just in case if some alternative solution or problems are found.

Some things I guess we can review:

1) When we should call `updateToolLoopModifiersIndicators();` and `updateAutoCelGuides(msg);` (when the editor is active, or has the mouse, or both)
2) A similar change must be applied to KeyUp

I'd not focus only on fixing the bug, but trying to find/simplify some code/unnecessary calls.

Gasparoken:
@martincapello I did some change to avoid unnecessary calls as @dacap mentioned. Could you review it one more time?
There may be other cases that can be optimized in the same way. But I'd like to address it as a separate issue (at low priority) in the future.

martincapello:
Looks good to me! 🎉

dacap:
Merged in a628bbb58bc0980a8ac02cb7a3b2e967341da2d5 (changed the message).

Just a detail: I highly recommend to write in the first line of the commit what issue fixes the commit (instead of the code changes/details that were made which can be commented in the description + the code itself).

PR #3921, opened 2023-06-09
Fix crash when loading an extension with invalid dithering matrices and the console must be shown (fix #3914)

### LINKED ISSUE #3914: Crash when loading an extension with invalid dithering matrices (?) and the console must be shown
Sentry Issue: [ASEPRITE-14T](https://igara-studio.sentry.io/issues/4221597323/?referrer=github_integration)

```
OS Version: Windows 10.0.19045 (2965)
Report Version: 104

Crashed Thread: 23012

Application Specific Information:
Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0x280

Thread 23012 Crashed:
0   aseprite.exe                    0x7ff614f4189e      mtx_do_lock (mutex.cpp:88)
1   aseprite.exe                    0x7ff6148f5f81      [inlined] std::_Mutex_base::lock (mutex:50)
2   aseprite.exe                    0x7ff6148f5f81      [inlined] std::lock_guard<T>::{ctor} (mutex:427)
3   aseprite.exe                    0x7ff6148f5f81      [inlined] obs::safe_list<T>::push_back (safe_list.h:293)
4   aseprite.exe                    0x7ff6148f5f81      obs::signal<T>::add_slot (signal.h:98)
5   aseprite.exe                    0x7ff6148f5394      [inlined] obs::signal<T>::connect (signal.h:104)
6   aseprite.exe                    0x7ff6148f5394      app::Console::ConsoleWindow::ConsoleWindow (console.cpp:49)
7   aseprite.exe                    0x7ff6148f5178      app::Console::Console (console.cpp:225)
8   aseprite.exe                    0x7ff61499980c      app::load_document (file.cpp:219)
9   aseprite.exe                    0x7ff614b0c1b3      app::load_dithering_matrix_from_sprite (load_matrix.cpp:26)
10  aseprite.exe                    0x7ff61490e31b      app::Extension::DitheringMatrixInfo::matrix (extensions.cpp:223)
11  aseprite.exe                    0x7ff614c3b1bd      app::DitheringSelector::regenerate (dithering_selector.cpp:235)
12  aseprite.exe                    0x7ff614c3a3af      app::DitheringSelector::DitheringSelector (dithering_selector.cpp:190)
13  aseprite.exe                    0x7ff614b4c147      app::ContextBar::ContextBar (context_bar.cpp:1762)
14  aseprite.exe                    0x7ff614974e88      app::MainWindow::MainWindow (main_window.cpp:109)
15  aseprite.exe                    0x7ff6148698d7      app::App::initialize (app.cpp:365)
16  aseprite.exe                    0x7ff61486238a      app_main (main.cpp:143)
17  aseprite.exe                    0x7ff614ffe635      wWinMain (main.cpp:40)
18  aseprite.exe                    0x7ff614f72b29      [inlined] invoke_main (exe_common.inl:118)
19  aseprite.exe                    0x7ff614f72b29      __scrt_common_main_seh (exe_common.inl:288)
20  KERNEL32.DLL                    0x7ffabee87613      BaseThreadInitThunk
21  ntdll.dll                       0x7ffac09c26a0      RtlUserThreadStart
```

Gasparoken:


Gasparoken:
Ready for review!

Gasparoken:
Ready for review.

martincapello (approved):

src/app/extensions.cpp
```diff
@@ -219,8 +219,9 @@ Extension::DitheringMatrixInfo::DitheringMatrixInfo(const std::string& path,
 const render::DitheringMatrix& Extension::DitheringMatrixInfo::matrix() const
 {
   if (!m_loaded) {
+    if (!load_dithering_matrix_from_sprite(m_path, m_matrix))
+      throw std::runtime_error("Invalid matrix name");
```
dacap, line 223: I think we have two options:

1. `load_dithering_matrix_from_sprite()` could throw the exception directly (instead of returning a bool), or 
2. adding a new bool `DitheringMatrixInfo::loadMatrix()` or `render::DitheringMatrix::isValidMatrix()` to avoid the exception at all (I think I prefer this one)

In this way we can create each `DitherItem` only for valid matrices (ignoring invalid matrices). With this probably we'd need in `render/dithering_matrix.h` something like:

```c++
  class DitheringMatrix {
  public:
    ...
    static DitheringMatrix InvalidMatrix() { return DitheringMatrix(0, 0); }

    bool isValid() const { return m_rows > 0 && m_cols > 0; }
    ...
```

dacap (changes requested):
Some comments to handle the error in a little different way.

src/app/app.cpp
```diff
@@ -363,6 +363,7 @@ int App::initialize(const AppOptions& options)
 
     // Create the main window.
     m_mainWindow.reset(new MainWindow);
+    m_mainWindow->afterInit();
```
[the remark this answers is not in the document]
Gasparoken: During the process of creating the MainWindow, if any error occurs, whatever it is, the Console is called and the constructor creates a connection to the "Close" command of the MainWindow:
```cpp
      // When the main window is closed, we must close the console (in
      // otherwise, the main message loop will continue to run for the duration of the
      // console as well).
      m_mainWindowClosedConn =
        Application::instance()->mainwindow()->Close.connect(
          [this]{ closeWindow(nullptr); });
```
The call to `App::instance()->mainWindow()` results in a crash since no such instance exists yet.
Alternative:
- Add an `if (App::instance()->mainWindow())` condition before creating the connection, and get rid of `afterInit()`. But this produces the following behaviour:
1. The error appears in the aseprite's console.
2. If you want to close Aseprite, it doesn't close, the app waits for the console to close.

What do you think?

src/app/extensions.cpp
dacap: From the `load_dithering_matrix_from_sprite` impl, it looks like the only way that it returns `false` is because the filename doesn't exist, if we keep the exception at this point, I'd change the error message to something like `fmt::format("The dithering matrix file {} doesn't exist", m_path)`.

src/app/ui/main_window.cpp
```diff
@@ -90,8 +90,66 @@ MainWindow::MainWindow()
   , m_devConsoleView(nullptr)
 #endif
 {
+}
+
+MainWindow::~MainWindow()
```
dacap, line 95: Probably there is no need to move the ~MainWindow code to this place to keep diff to a minimum.

src/app/commands/cmd_change_pixel_format.cpp
```diff
@@ -534,7 +534,9 @@ void ChangePixelFormatCommand::onLoadParams(const Params& params)
     else {
       render::DitheringMatrix ditMatrix;
       if (!load_dithering_matrix_from_sprite(matrix, ditMatrix))
-        throw std::runtime_error("Invalid matrix name");
+        throw std::runtime_error(
+          fmt::format("The dithering matrix file {} doesn't exist",
+                      matrix));
```
dacap, line 539: Probably I'll move the throw `std::runtime_error` inside `load_dithering_matrix_from_sprite` to avoid duplicating this exact same code here and `Extension::DitheringMatrixInfo::matrix()`.

src/app/ui/main_window.cpp
```diff
@@ -90,8 +90,66 @@ MainWindow::MainWindow()
...
+// all UI components that can trigger the console to report
+// any unexpected errors/warnings are initialized.
+// Prior to this split, Aseprite could crash due to errors during launch.
+// Along the MainWindow creation if any error occurs, whatever it is,
+// Console was called, its constructor tried to create a connection to
+// the MainWindow "Close" command, but resulted in a crash since
+// such an instance did not yet exist.
+// Refer to issue https://github.com/aseprite/aseprite/issues/3914
+// "Fix crash when loading an extension with invalid dithering
+// matrices and the console must be shown"
```
dacap, line 146: Generally we don't add the issue in the comment (as git blame gives this information), but if you want to keep a link to the issue, it can be just:
```
// Refer to https://github.com/aseprite/aseprite/issues/3914
```
(without copying the title)

Anyway probably I'll reword the comment before the merge.

dacap (changes requested):

src/app/app.cpp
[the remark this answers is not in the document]
Gasparoken: Thanks for the advice! I implemented something like this but got unexpected results when Aseprite starts with an invalid dithering matrix:
- Sometimes Aseprite starts fine, other times it gives an error

Screenshot 2023-06-26 at 13 08 06
Screenshot 2023-06-26 at 13 09 49

I can't quite see why this effect occurs. It would take quite a bit more time to get to deeper. The thing is, this doesn't happen if I just move to `initialize()` just the creation and manipulation of the `ContextBar`.

src/app/extensions.cpp
Gasparoken: I kept the exception and took your recommendation on the error message.

src/app/ui/main_window.cpp
Gasparoken: I just saw these last comments, I'll take care of making these changes.

src/app/app.cpp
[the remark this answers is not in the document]
Gasparoken: I'll try to figure this out today.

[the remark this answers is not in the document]
Gasparoken: Ready for review.

dacap:
Merged f7bc91892679ea14ea482f2b56226b6461309323

PR #3923, opened 2023-06-10
Update INSTALL.md for Apple Silicon without Xcode.app installed

shawwn:
This PR updates [INSTALL.md](https://github.com/aseprite/aseprite/blob/main/INSTALL.md) so that the build instructions work even for people without Xcode.app installed. Xcode isn't a requirement; the command-line tools are enough, and a surprising number of people refuse to install Xcode. So I've tried to avoid installing it for as long as possible on my new M2 to simulate this stubbornness, precisely so that I can go around and fix issues like this in other projects. :)


dacap:
Hi @shawwn, thanks for this PR. Anyway I think I prefer to keep the `CMAKE_OSX_SYSROOT` parameter in the instructions because it shows that you need a macOS SDK to compile the program (it doesn't matter if it's located on Xcode or some other place, but the SDK must be available).

PR #3927, opened 2023-06-12
Scripting: window size

lampysprites:
Proposed changes to scripting API:
- `app.width`, `app.height` properties to access size of the client area of the window.
  It was sort of possible to retrieve those as `2*x+width` of a freshly created dialog, but not very intuitive or efficient.
- `"resize"` type of `app.events`, which is triggered after windows resize. Callback receives a table with `width` and `height` keys.

It will allow to keep dialogs centered, or positioned relative to right/bottom edge of the window; or limit size to show a scrollbar when they're taller that the window itself.

------------------------------------------------------

### DIFF
--- a/src/app/script/app_object.cpp
+++ b/src/app/script/app_object.cpp
@@ -37,6 +37,7 @@
 #include "app/ui/editor/editor.h"
 #include "app/ui/editor/tool_loop_impl.h"
 #include "app/ui/timeline/timeline.h"
+#include "app/ui/main_window.h"
 #include "app/ui_context.h"
 #include "base/fs.h"
 #include "base/replace_string.h"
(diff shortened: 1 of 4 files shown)


dacap:
Thanks for this PR @lampysprites! I have to review it, but probably I'll move `app.width/height` to `app.window.bounds/width/height` or something similar (`app.mainWindow...`?). As I'm thinking that we should be able to access the list of all windows and identify them (Window.MAIN_WINDOW/PLUGIN/FOREGROUND_COLOR, etc.)

`app.width` and `app.height` look too generic but might work for the simplest case (the main window current size). In the future we might have multiple windows (e.g. if we drag a tab outside the main window) so I'm not sure what `app.width/height` should be in those cases.

lampysprites:
Np I'll change that real quick then. For future proofing it can be wrapped
in a table like palettes[1] now, perhaps?

dacap:
Merged in 30b858d89e6ef41507f9475a6ccca0e127420a6c + 7ee866643b4d9c5ec93698b04cd79a523a6a6aaa + a2d8a080f52428041c6e58f99c9abcedf19c1dd9

PR #3928, opened 2023-06-12
Delete unused tilesets after deleting tilemaps 

### LINKED ISSUE #3876: Delete unused tilesets after deleting tilemaps
There is a bug where several tilesets are accumulate when tilemap layers are deleted. E.g. if we create 3 tilemaps:

image

And then remove them, and try to create a new tilemap layer, we can see 3 tilesets that should have been removed:

image

martincapello:
Fix #3876

This PR also include two unrelated commits that fix a couple of things:
40e44da202aacc298b31544851bc97f739c3e16a
4a82570d4276d97d2b39b95f1665dc9bd7b4a315

If you think that it would be better to put these commits in another PR, just let me know please.

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -55,6 +55,13 @@ Automatic Remap
 ||&OK||&Cancel"
 END
 cannot_delete_all_layers = Error<<You cannot delete all layers.||&OK
+deleting_tilemaps_will_delete_tilesets = <<<END
+Warning
+<<Deleting the following layers will delete their tilesets:
+<<'{0}'
+<<Do you want to continue anyway?
+||&OK||&Cancel
+END
 cannot_file_overwrite_on_export = <<<END
 Overwrite Warning
 <<You cannot Export with the same name (overwrite the original file).
@@ -1698,6 +1705,7 @@ file_format_doesnt_support_alert = Show warning when saving a file with unsuppor
 export_animation_in_sequence_alert = Show warning when saving an animation as a sequence of static images
 overwrite_files_on_export_alert = Show warning when overwriting files on File > Export
 overwrite_files_on_export_sprite_sheet_alert = Show warning when overwriting files on Export Sprite Sheet
+delete_tilemap_delete_unused_tileset_alert = Show warning when deleting a tilemap will delete its unused tileset
 image_format_alerts = Show options when saving files:
 advanced_mode_alert = Show alert when we enter to Advanced Mode
 invalid_fg_bg_color_alert = Show alert when drawing with index out of palette bounds
--- a/src/app/commands/cmd_remove_layer.cpp
+++ b/src/app/commands/cmd_remove_layer.cpp
@@ -30,6 +30,24 @@
 
 namespace app {
 
+// If the UI is available this function shows an alert informing the user that they
+// cannot remove all the layers from the sprite (this is true when topLevelLayersToDelete
+// value is equals to the number of layers in the sprite's root group) and returns true.
+// If the UI is not available, this function just returns true if topLevelLayersToDelete is
+// equals to the number of layers in the sprite's root group.
+static bool deleting_all_layers(Context* ctx, Sprite* sprite, int topLevelLayersToDelete)
+{
+  const bool deletingAll = topLevelLayersToDelete == sprite->root()->layersCount();
+
+#ifdef ENABLE_UI
+  if (ctx->isUIAvailable() && deletingAll) {
+    ui::Alert::show(Strings::alerts_cannot_delete_all_layers());
+  }
+#endif
+
+  return deletingAll;
+}
+
 // Calculates the list of unused tileset indexes (returned in tsiToDelete parameter)
 // once the layers specified are removed.
 // Also, if the UI is available, shows a warning message about the deletion of unused
@@ -130,8 +148,7 @@ void RemoveLayerCommand::onExecute(Context* context)
           ++deletedTopLevelLayers;
       }
 
-      if (deletedTopLevelLayers == sprite->root()->layersCount()) {
-        ui::Alert::show(Strings::alerts_cannot_delete_all_layers());
+      if (deleting_all_layers(context, sprite, deletedTopLevelLayers)) {
         return;
       }
 
@@ -144,8 +161,7 @@ void RemoveLayerCommand::onExecute(Context* context)
       }
     }
     else {
-      if (sprite->root()->layersCount() == 1) {
-        ui::Alert::show(Strings::alerts_cannot_delete_all_layers());
+      if (deleting_all_layers(context, sprite, 1)) {
         return;
       }
 
--- a/src/app/commands/cmd_remove_layer.cpp
+++ b/src/app/commands/cmd_remove_layer.cpp
@@ -11,19 +11,76 @@
 
 #include "app/app.h"
 #include "app/commands/command.h"
+#include "app/cmd/remove_tileset.h"
 #include "app/context_access.h"
 #include "app/doc_api.h"
 #include "app/i18n/strings.h"
 #include "app/modules/gui.h"
 #include "app/tx.h"
+#include "app/pref/preferences.h"
+#include "app/ui/optional_alert.h"
 #include "app/ui/status_bar.h"
 #include "doc/layer.h"
+#include "doc/layer_tilemap.h"
 #include "doc/sprite.h"
+#include "doc/tilesets.h"
 #include "fmt/format.h"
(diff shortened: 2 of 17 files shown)


data/strings/en.ini
```diff
@@ -55,6 +55,18 @@ Automatic Remap
...
+<<Deleting '{0}' layer will delete its associated tileset.
+<<Do you want to continue anyway?
+||&OK||&Cancel
+END
+deleting_tilemaps_will_delete_tilesets = <<<END
+Warning
+<<Deleting '{0}' layers will delete their associated tilesets.
+<<Do you want to continue anyway?
+||&OK||&Cancel
+END
```
dacap, line 69: I'm not sure about this alert, probably a good idea to add it. If we keep it, two things:
1. Add this alert as optional (search for `OptionalAlert`)
2. Could we merge these two messages in one (with one generic string for both cases)

src/app/commands/cmd_remove_layer.cpp
```diff
@@ -11,20 +11,89 @@
...
+static bool deletingAllLayers(Context* ctx, Sprite* sprite, int topLevelLayersToDelete)
+{
+#ifdef ENABLE_UI
+  if (ctx->isUIAvailable() && topLevelLayersToDelete == sprite->root()->layersCount()) {
+    ui::Alert::show(Strings::alerts_cannot_delete_all_layers());
+  }
+#endif
+
+  return topLevelLayersToDelete == sprite->root()->layersCount();
+}
```
dacap, line 45: Minor change
```suggestion
{
  const bool cancelDelete = (topLevelLayersToDelete == sprite->root()->layersCount());

#ifdef ENABLE_UI
  if (ctx->isUIAvailable() && cancelDelete) {
    ui::Alert::show(Strings::alerts_cannot_delete_all_layers());
  }
#endif

  return cancelDelete;
}
```

src/app/commands/cmd_remove_layer.cpp
```diff
@@ -11,20 +11,89 @@
...
+
+// Calculates the list of unused tileset indexes (returned in tsiToDelete parameter)
+// once the layers specified are removed.
+// Also, if the UI is available, shows a warning message about the deletion of unused
+// tilesets.
+// This function returns true in any of the following:
+// - There won't be deletion of tilesets, this means tsiToDelete is empty.
+// - The user accepts continuing despite the warning.
+// - There is no UI available.
+static bool continueDeletingUnusedTilesets(Context* ctx, Sprite* sprite, const LayerList layers, std::set<tileset_index, std::greater<tileset_index>>& tsiToDelete)
```
dacap, line 55: Break long lines.

tests/scripts/layers.lua
```diff
@@ -213,3 +213,21 @@ do
...
+  app.command.RemoveLayer()
+  assert(#s.layers == 1)
+
+  -- Try deleting all layers when there are more than one.
+  s = Sprite(2, 2)
+  local a = s.layers[1]
+  local b = s:newLayer()
+  local c = s:newLayer()
+  app.range.layers = { a, b, c }
+  assert(#s.layers == 3)
```
dacap, line 232: I think a RemoveLayer is missing?
```suggestion
  assert(#s.layers == 3)
  app.command.RemoveLayer()
  assert(#s.layers == 3)
```

src/doc/tileset.cpp
```diff
@@ -406,4 +409,15 @@ TilesetHashTable& Tileset::hashTable()
   return m_hash;
 }
 
+int Tileset::tilemapsCount() {
```
dacap, line 412: If possible
```suggestion
int Tileset::tilemapsCount() const {
```

src/doc/tileset.h
```diff
@@ -131,6 +131,9 @@ namespace doc {
     // have to regenerate the empty tile with that new mask color.
     void notifyRegenerateEmptyTile();
 
+    // Returns the number of tilemap layers that are referencing this tileset.
+    int tilemapsCount();
```
dacap, line 135: If possible
```suggestion
    int tilemapsCount() const;
```

src/doc/selected_layers.cpp
```diff
@@ -99,6 +99,24 @@ LayerList SelectedLayers::toBrowsableLayerList() const
...
+  if (empty())
+    return output;
+
+  ASSERT(*begin());
+  ASSERT((*begin())->sprite());
+
+  for (Layer* layer : (*begin())->sprite()->allTilemaps()) {
+    if (contains(layer))
+      output.push_back(layer);
+  }
```
dacap, line 115: I think we could iterate the selected layers and ask if they are tilemap (`isTilemap`) instead of asking for `contains()` of all tilemaps in the sprite.

```suggestion
  for (auto layer : *this) {
    if (layer->isTilemap())
      output.push_back(layer);
  }
```

src/doc/selected_layers.cpp
```diff
@@ -99,6 +99,24 @@ LayerList SelectedLayers::toBrowsableLayerList() const
   return output;
 }
 
+LayerList SelectedLayers::toAllTilemaps() const
```
dacap, line 102: Probably I'd change the name to
```suggestion
LayerList SelectedLayers::onlyTilemaps() const
```

tests/scripts/tilemap.lua
```diff
@@ -1117,3 +1117,113 @@ do
...
+  app.range.layers = { tilemapLay2 }
+  app.command.RemoveLayer{ }
+  assert(#spr.layers == 3)
+  assert(#spr.tilesets == 2)
+  assert(spr.tilesets[1] == tilemapLay1.tileset)
+  assert(spr.tilesets[2] == tilemapLay3.tileset)
+
+  -- Undo tilemap removal and check that it goes back to
+  -- previous state.
+  app.command.Undo()
```
dacap, line 1166: `app.undo()` is preferred, but it's not mandatory
```suggestion
  app.undo()
```

tests/scripts/tilemap.lua
```diff
@@ -1117,3 +1117,113 @@ do
...
+  local tilemapLay3 = spr.layers[4]
+  assert(tilemapLay3.isImage)
+  assert(tilemapLay3.isTilemap)
+  assert(#spr.tilesets == 3)
+  assert(spr.tilesets[3] == tilemapLay3.tileset)
+
+  -- Remove tilemap 2 and check that a tilemap was removed and
+  -- tilesets of remaining tilemaps are correct.
+  app.range.layers = { tilemapLay2 }
+  app.command.RemoveLayer{ }
```
dacap, line 1158:
```suggestion
  app.command.RemoveLayer()
```
Same for all `RemoveLayer()` calls.

dacap (changes requested):
Looks great so far! and it's great to have some new tests for dthis. I've added some minor comments about this PR, also give a little test and probably a bug with the new alert (when a tilemap is children of a group the alert for its tileset is not displayed)?

image

data/strings/en.ini
martincapello: Sure, will make it optional

src/doc/selected_layers.cpp
martincapello: I used this name because there are toBrowsableLayerList and toAllLayersList methods already, so I tried to keep it consistent...let me know if I should rename it to onlyTilemaps anyway

dacap: Seeing that it returns a `LayerList` (and not a filtered/new `SelectedLayers`) it can be called `toAllTilemaps()` (I think all 3 names are confusing anyway, probably something to rename in the future).

src/app/commands/cmd_remove_layer.cpp
martincapello: I agree about the change, but I would like to name the bool "deletingAll" instead of "cancelDelete" if it makes sense to you

tests/scripts/layers.lua
martincapello: Good catch!

src/app/commands/cmd_remove_layer.cpp
dacap: Yes, it can be `deletingAll` 👍  just one thing I forgot: we prefer `static_functions_with_this_naming_convention` to avoid confusing them with member functions (methods) and var names.

dacap:
Merged rebased on `main` :+1: 

There is a possible future option/extra button for the warning dialog but there is no need to add it now, which is deleting the tilemaps but keeping the tilesets. But we'll see if it's required by the users later.

Issue #3900, opened 2023-05-30
Using simple math for color input


Gasparoken:
Example:
<img width="450" alt="Screenshot 2023-05-30 at 09 02 54" src="https://github.com/aseprite/aseprite/assets/32165475/ea41dd46-3d9e-4c80-a7cd-8e5d844e3b7b">

LuckyTigerr:
Hello, i want to try make this issue, can you please give more information about what need to do and about file location?

Gasparoken:
Hi @LuckyTigerr, thank you in advance!
If you push a PR to the Aseprite project, we will ask for your consent regarding our [Igara Studio Contributor License Agreements](https://github.com/igarastudio/cla).
You can start in `color_sliders.h` in the `struct item` declaration. The `ui::Entry` correspond to the 'entry' on the old way:
```cpp
    struct Item {
      bool show = false;
      ui::Label* label = nullptr;
      ui::Box* box = nullptr;
      ui::Slider* absSlider = nullptr;
      ui::Slider* relSlider = nullptr;
      ui::Entry* entry = nullptr;
    };
```
Some reference can be taken in the following commit (when the input expressions were implemented): https://github.com/aseprite/aseprite/commit/f595ceea7f60a14d53c0fa086cb6c45cb738d323

Akylzhan:
Hey @LuckyTigerr, are you still working on it? If not, I would like to do it

PR #3930, opened 2023-06-21
Replace ColorEntry with ExprEntry in Color Sliders

Akylzhan:
Adds https://github.com/aseprite/aseprite/issues/3900

### DIFF
--- a/src/app/ui/color_sliders.cpp
+++ b/src/app/ui/color_sliders.cpp
@@ -14,6 +14,7 @@
 #include "app/color_utils.h"
 #include "app/modules/gfx.h"
 #include "app/ui/color_sliders.h"
+#include "app/ui/expr_entry.h"
 #include "app/ui/skin/skin_slider_property.h"
 #include "app/ui/skin/skin_theme.h"
 #include "base/scoped_value.h"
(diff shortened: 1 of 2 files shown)


src/app/ui/color_sliders.cpp
```diff
@@ -373,7 +373,7 @@ void ColorSliders::addSlider(const Channel channel,
   item.box       = new HBox();
   item.absSlider = new Slider(absMin, absMax, 0);
   item.relSlider = new Slider(relMin, relMax, 0);
-  item.entry     = new ColorEntry(item.absSlider, item.relSlider);
+  item.entry     = new ExprEntry();
```
dacap, line 376: With this change we lost the behavior of `ColorEntry` (we cannot use up/down arrow keys to modify the value).

src/app/ui/color_sliders.h
```diff
@@ -81,7 +82,7 @@ namespace app {
       ui::Box* box = nullptr;
       ui::Slider* absSlider = nullptr;
       ui::Slider* relSlider = nullptr;
-      ui::Entry* entry = nullptr;
+      app::ExprEntry* entry = nullptr;
```
dacap, line 85: This change is not required.

dacap (changes requested):
Hi @Akylzhan, thanks for your PR. Probably we require a different approach for this. Probably just changing:

```c++
class ColorEntry : public ExprEntry
```

Akylzhan:
@dacap Yes, you are right. Did not think about that. Changed ColorEntry's base to ExprEntry

dacap:
Thanks @Akylzhan, I've just merged this change in https://github.com/aseprite/aseprite/commit/3e171d0f6aebeff50138bce6fb98c987d1006723 with a missing setText("0") so the entry fields start with 0 in the *Edit > Adjustments > Hue/Saturation* dialog.

PR #3931, opened 2023-06-23
Show, duplicate and delete tilesets in Sprite Properties dialog

### LINKED ISSUE #3875: When a sprite contains tilemap layers, the sprite properties dialog should show a section of "tilesets"
We need a way to reorganize tilesets (delete tilesets/duplicate them) in the Sprite Properties dialog.

martincapello:
Fix #3875

And also adds the possibility of changing the tileset associated with a tilemap layer.

### DIFF
--- a/data/widgets/sprite_properties.xml
+++ b/data/widgets/sprite_properties.xml
@@ -43,6 +43,14 @@
         </hbox>
       </hbox>
     </grid>
+
+    <vbox expansive="true" id="tilesets_placeholder">
+      <separator text="@.tilesets" horizontal="true" />
+      <view id="tilesets_view" expansive="true">
+        <listbox id="tilesets"></listbox>
+      </view>
+    </vbox>
+
     <separator horizontal="true" />
     <hbox>
       <boxfiller />
--- a/src/app/cmd/set_layer_tileset.cpp
+++ b/src/app/cmd/set_layer_tileset.cpp
@@ -10,6 +10,7 @@
 
 #include "app/cmd/set_layer_tileset.h"
 
+#include "app/app.h"
 #include "app/doc.h"
 #include "doc/layer_tilemap.h"
 #include "doc/sprite.h"
@@ -37,6 +38,13 @@ void SetLayerTileset::onUndo()
   auto layer = static_cast<LayerTilemap*>(this->layer());
   layer->setTilesetIndex(m_oldTsi);
   layer->incrementVersion();
+}
+
+void SetLayerTileset::onFireNotifications()
+{
+  auto layer = static_cast<LayerTilemap*>(this->layer());
+  Doc* doc = static_cast<Doc*>(layer->sprite()->document());
+  doc->notifyTilesetChanged(layer->tileset());
 }
 
 } // namespace cmd
--- a/src/app/commands/cmd_sprite_properties.cpp
+++ b/src/app/commands/cmd_sprite_properties.cpp
@@ -11,6 +11,8 @@
 
 #include "app/cmd/assign_color_profile.h"
 #include "app/cmd/convert_color_profile.h"
+#include "app/cmd/add_tileset.h"
+#include "app/cmd/remove_tileset.h"
 #include "app/cmd/set_pixel_ratio.h"
 #include "app/cmd/set_user_data.h"
 #include "app/color.h"
@@ -20,15 +22,18 @@
 #include "app/i18n/strings.h"
 #include "app/modules/gui.h"
 #include "app/pref/preferences.h"
+#include "app/util/tileset_utils.h"
 #include "app/tx.h"
 #include "app/ui/color_button.h"
 #include "app/ui/user_data_view.h"
+#include "app/ui/skin/skin_theme.h"
 #include "app/util/pixel_ratio.h"
 #include "base/mem_utils.h"
 #include "doc/image.h"
 #include "doc/palette.h"
 #include "doc/sprite.h"
 #include "doc/user_data.h"
+#include "doc/tilesets.h"
 #include "fmt/format.h"
 #include "os/color_space.h"
 #include "os/system.h"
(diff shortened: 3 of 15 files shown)


dacap:
Just in case, I know this is in draft yet, just seeing that this branch is including (rebased) existing commits from `main`.

martincapello:
Yes, I'm aware of that, this is just because this is a branch from my `delete-unused-tilesets` branch because I needed some changes there that are not merged yet. After that branch is merged I will fix the commits in this one.

src/ui/listbox.cpp
```diff
@@ -249,9 +249,11 @@ bool ListBox::onProcessMessage(Message* msg)
           if (dynamic_cast<ui::Separator*>(picked))
             picked = nullptr;
 
-          // If the picked widget is a child of the list, select it
-          if (picked && hasChild(picked))
-            selectChild(picked, msg);
+          // If the picked widget is a descendant of the list, select the item containing it.
+          if (picked && hasDescendant(picked)) {
```
dacap, line 253: Probably this can use the `hasAnscestor()` (which doesn't have to iterate through all children of children, and it's already used in other places):
```suggestion
          if (picked && picked->hasAncestor(this)) {
```

data/widgets/sprite_properties.xml
```diff
@@ -43,6 +43,16 @@
...
     </grid>
+
+    <grid columns="3">
+      <separator text="@.tilesets" horizontal="true" cell_hspan="3" />
+
+      <hbox></hbox>
+      <view expansive="true" cell_align="horizontal" cell_hspan="2" minheight="63">
+        <listbox id="tilesets"></listbox>
+      </view>
+    </grid>
```
dacap, line 54: I think this could be just a `<vbox>` (added `id="tilesets_placeholder"` as we could do a `tilesetsPlaceholder()->setVisible(false)` if the sprite doesn't have tilesets):
```suggestion
    <vbox expansive="true" id="tilesets_placeholder">
      <separator text="@.tilesets" horizontal="true" />
      <view expansive="true" minheight="63">
        <listbox id="tilesets"></listbox>
      </view>
    </vbox>
```
In this way the listbox expands correctly if the window is bigger:
image

src/app/cmd/set_layer_tileset.cpp
```diff
@@ -39,5 +40,10 @@ void SetLayerTileset::onUndo()
   layer->incrementVersion();
 }
 
+void SetLayerTileset::onFireNotifications()
+{
+  App::instance()->PaletteChange();
+}
```
dacap, line 46: We don't access `App` instance directly from Cmd notifcations:
```suggestion
{
  doc::Sprite* sprite = this->sprite();
  Doc* doc = static_cast<Doc*>(sprite->document());
  doc->notifyPaletteChanged();
}
```
This exact same code is from `SetPalette::onFireNotifications()`.

src/app/commands/cmd_sprite_properties.cpp
```diff
@@ -40,6 +44,95 @@ namespace app {
 
 using namespace ui;
 
+class TilesetListItem : public ui::ListItem {
+public:
+  TilesetListItem(const doc::Tileset* tileset, doc::tileset_index tsi)
+    : ListItem(app::tileset_label(tileset, tsi)),
+    m_buttons(new HBox()) {
```
dacap, line 51: There is an extra indentation in the ctor block:
```suggestion
  TilesetListItem(const doc::Tileset* tileset, doc::tileset_index tsi)
    : ListItem(app::tileset_label(tileset, tsi))
    , m_buttons(new HBox())
  {
```

src/app/commands/cmd_sprite_properties.cpp
```diff
@@ -40,6 +44,95 @@ namespace app {
...
+      m_buttons->setTransparent(true);
+      m_buttons->setExpansive(true);
+      m_buttons->setVisible(false);
+
+      auto filler = new BoxFiller();
+      filler->setTransparent(true);
+      m_buttons->addChild(filler);
+
+      auto duplicateBtn = new Button(Strings::sprite_properties_duplicate_tileset());
+      duplicateBtn->setTransparent(true);
```
dacap, line 61: Add the "miniButton" style to buttons (same as in KeyItem in cmd_keyboard_shortcuts.cpp):
```
duplicateBtn->setStyle(theme->styles.miniButton());
```

src/app/commands/cmd_sprite_properties.cpp
```diff
@@ -40,6 +44,95 @@ namespace app {
...
+          if (listbox->getSelectedChild()) {
+            static_cast<TilesetListItem*>(listbox->getSelectedChild())->m_buttons->setVisible(false);
+          }
+          listbox->selectChild(this);
+          m_buttons->setVisible(true);
+        }
+        break;
+    }
+    return ui::ListItem::onProcessMessage(msg);
+  }
```
dacap, line 116: Change with this impl which makes the item behave similar to items in Keyboard Shortcuts diaog:
```suggestion
  bool onProcessMessage(Message* msg) override
  {
    switch (msg->type()) {
      case kMouseLeaveMessage:
        m_buttons->setVisible(false);
        invalidate();
        break;
      case kMouseEnterMessage:
        m_buttons->setVisible(true);
        invalidate();
        break;
    }
    return ui::ListItem::onProcessMessage(msg);
  }
```

src/app/commands/cmd_sprite_properties.cpp
```diff
@@ -40,6 +44,95 @@ namespace app {
...
+  void onVisible(bool visible) override
+  {
+    if (!visible) {
+      // If we don't invalidate here list items could leave old state on screen
+      // when moving the selected item.
+      // TODO: Should we add an invalidate() call on Widget::setVisible(false)? This
+      // would make this onVisible unnecessary.
+      invalidate();
+    }
+  }
```
dacap, line 127: This can be removed with the new `onProcessMessage()`
```suggestion
```

src/app/commands/cmd_sprite_properties.cpp
```diff
@@ -40,6 +44,95 @@ namespace app {
...
+
+  obs::signal<void(TilesetListItem *)> TilesetDeleted;
+  obs::signal<void(const Tileset*)> TilesetDuplicated;
+
+private:
+  ui::HBox* m_buttons;
```
dacap, line 133: This could be changed to:
```suggestion
  ui::HBox m_buttons;
```
Changing `m_buttons->` with `m_buttons.` and `addChild(&m_buttons);`.

Other possibility would be to create/destroy buttons on kMouseEnter/LeaveMessage (KeyItem does this). But not 100% required.

dacap (changes requested):
It works great! but some minor changes in the UI.

src/app/cmd/set_layer_tileset.cpp
martincapello: I have tried that when I was working on this and it didn't work. Because I need that PaletteView::onAppPaletteChange gets called, which in turn calls the paletteChange method of the adapter. AFAIK PaletteView::onAppPaletteChange is called only when App::instance()->PaletteChange() is called. Am I missing something?

Btw, just to double check, I tried again with the suggested code and checked that it makes Aseprite crash after changing the tileset of a tilemap and trying to pick a tile in the tile selector.

martincapello:
Just pushed all the suggested changes but one.

src/app/cmd/set_layer_tileset.cpp
dacap: Didn't test the change, probably we need another kind of notification notifyTilesetChanged (which makes more sense because we're changing the tileset, not the palette). Anyway I'm not sure why this was originally required.

martincapello: I see. Then we should take note about updating this function in palettes.cpp, right?:
```c++
// Changes the current system palette and triggers the
// App::PaletteChange signal.
//
// If "_palette" is nullptr the default palette is set.
bool set_current_palette(const Palette *_palette, bool forced)
{
  const Palette* palette = (_palette ? _palette: ase_default_palette);
  bool ret = false;

  // Have changes
  if (forced ||
      palette->countDiff(ase_current_palette, NULL, NULL) > 0) {
    // Copy current palette
    palette->copyColorsTo(ase_current_palette);

    // Call slots in signals
    App::instance()->PaletteChange();  // <--- this call would update the tiles palette too?

    ret = true;
  }

  return ret;
}
```

dacap: The `set_current_palette` function (and the `App::PaletteChange` signal) is an old function from the Allegro era, were the graphics mode could be in 8-bit (and there were an actual "current palette" set in the screen). Now this function (and almost everything from `modules/` [should be removed](https://github.com/aseprite/aseprite/blob/main/src/app/modules/README.md)/replaced with other kind of notifications).

Right now it's a kind of workaround for some situations but probably `ContextObserver::onActiveSiteChange` will be the correct place to listen if the current palette changes. Anyway this is refactor for the future.

I'm not sure why `set_current_palette` should update the tileset view. What exactly are you trying to achieve and where? If the `PaletteView+TilesetViewAdapter` must be updated because the current tileset was changed, I think `ColorBar::onTilesetChanged` already does that (probably you can generate a `Doc::notifyTilesetChanged()`).

martincapello: Never mind...to my surprise (and maybe yours 😄) there was already a notifyTilesetChanged that you introduced a long time ago, and it seems to work as well. But now I have to fix a bug I found while trying some things.

dacap:
We can hide the whole Tilesets section/placeholder if there are no tilemaps/tilesets in the sprite:

image

And there is this little bug when we shrink the dialog (the bottom of the view):

image

I know it's related to the `minheight="63"` in the `<view>` widget, I'm not sure if there is a better way to handle this (probably setting the initial height of the sprite properties dialog to `std::max(bounds().h, sizeHint().h)` or something like that and removing the minheight)

Another detail: the commit https://github.com/aseprite/aseprite/pull/3931/commits/8fd6ddbbf76f8f9d11b8a3a09db9b679d4717041 contains a line in the its description that just says "fixes", you can delete it.

PR #3932, opened 2023-06-24
Added Mingw support

kaissouDev:
hello! i'm making this pull request to add support for the mingw compiler. 
I'm adding this so that aseprite is accessible to more people because msvc can take up a lot of space on a computer;
 whereas mingw is much lighter ... that's why I'm making this pull request :)


kaissouDev:
How do I fix byteswap under mingw?

dacap:
Hi @kaissouDev, there are no plans to give support to mingw compiler mainly because the official skia version cannot be compiled with mingw (at least, as far as I know). There is this package I've just found: https://packages.msys2.org/base/mingw-w64-skia which looks like contains a serie of patches for a specific skia version, it might be a good start (anyway we are not interested on maintain a large set of patches for skia).

kaissouDev:
ok, but I have a question: is the cityhash library compatible with mingw?

PR #3940, opened 2023-06-28
Fix direction reset after scrolling the editor while playing an animation with "ping-pong" direction

### LINKED ISSUE #3903: Using “Loop Section” with “Ping Pong” and moving camera causes it to forget which direction it’s going
https://community.aseprite.org/t/using-loop-section-with-ping-pong-and-moving-camera-causes-it-to-forget-which-direction-its-going/18963

martincapello:
Fix #3903


dacap:
Merged b0b2fb378ea071d7b8f0c4bf10896dd5c575c72e

PR #3943, opened 2023-06-30
Fix export sprite sheet to RGB is not exporting the palette (fix #3881)

### LINKED ISSUE #3881: Export sprite sheet to RGB is not exporting the palette?
Source: https://community.aseprite.org/t/colour-pallete-turning-black-when-exporting-animation-into-sprite-sheet/17757

Gasparoken:


Gasparoken:
I need to review:
```cpp
void Palette::addNonRepeatedColors(const Palette* palette)
```
Although this function is only used for one RGB color mode. I have to evaluate if it would add ASSERT() or the other color modes.

src/doc/palette.cpp
```diff
@@ -175,6 +175,24 @@ int Palette::countDiff(const Palette* other, int* from, int* to) const
   return diff;
 }
 
+void Palette::addNonRepeatedColors(const Palette* palette)
+{
+  ASSERT(palette);
+  for (int i=0; i < palette->size(); i++) {
```
martincapello, line 181: Should we also check for null palette and return in that case?

martincapello (commented):
It woks great, just left a couple of comments.

Gasparoken:
Ready for a new review.

martincapello (approved):

Gasparoken:
I think it's ready to merge @dacap.

src/app/doc_exporter.cpp
```diff
@@ -1236,8 +1242,10 @@ Doc* DocExporter::createEmptyTexture(const Samples& samples,
       maxColors,
       m_docBuf));
 
-  if (palette)
-    sprite->setPalette(palette, false);
+  if (palette.size() == 0)
+    // Old behavior
+    palette.resize(256);
+  sprite->setPalette(&palette, false);
```
dacap, line 1248: Is this needed? or can we do this:
```suggestion
  if (palette.size() > 0)
    sprite->setPalette(&palette, false);
```

Gasparoken:
Your advices were applied and a condition "format does not support palette" was added to avoid palette addition without sense. I can add some other condition like "palette color count <256 or 512 or..." What do you think?

dacap:
> a condition "format does not support palette" was added to avoid palette addition without sense. I can add some other condition like "palette color count <256 or 512 or..." 

The `format_supports_palette()` is great :+1: Probably we could add an option in `addNonRepeatedColors()` to avoid adding more than N colors so we avoid creating palettes bigger than 256 colors.

Gasparoken:
Ready for a new review.

dacap:
Probably I'm doing something wrong, but I've just tested this and it looks like the bug is still there. Example file: [123.zip](https://github.com/aseprite/aseprite/files/13297823/123.zip)

image

Gasparoken:
Ready for review.

src/app/doc_exporter.cpp
```diff
@@ -1180,10 +1180,13 @@ Doc* DocExporter::createEmptyTexture(const Samples& samples,
...
 {
   ColorMode colorMode = ColorMode::INDEXED;
-  Palette* palette = nullptr;
+  Palette palette(0, 0);
   int maxColors = 256;
   gfx::ColorSpaceRef colorSpace;
   color_t transparentColor = 0;
+  const bool textureSupportsPalette =
+    format_supports_palette(m_textureFilename) ||
+    m_textureFilename == ""; // if m_textureFile == "" aseprite format is taken as default.
```
dacap, line 1189: Actually the comment is not correct, if the filename is empty, the `FileFormat::UNKNOWN` is used in `format_supports_palette()`.

My suggestion (add `(...)` between the expression, don't use `== ""` to compare empty strings, and fix the comment)
```suggestion
    (m_textureFilename.empty() || // This is the preview and we create the palette anyway
     format_supports_palette(m_textureFilename)); 
```

dacap:
I'll merge this as it is, there is still a bug but I think there something else:

export-sprite-sheet

PR #3947, opened 2023-07-05
Limit windows positions and introduction of `autoscrollbars` flag

### LINKED ISSUE #3839: Problems with scripts (and the the program itself) creating big windows
If a script creates a window that is big enough in such a way that its title bar is not visible, the window cannot be moved and it's super confusing to the user. E.g.

* https://community.aseprite.org/t/extension-palette-helper/14429/16?u=dacap
* https://community.aseprite.org/t/extension-palette-helper/14429/22?u=dacap

We should check the visibility of any window that is created, the title bar must be visible (maybe a scrollbar?). Probably a problem only in the one-window interface (check if it's an issue for #3556)

martincapello:
This PR fixes #3839 by implementing the following:
- Limit windows positions to avoid hiding title bars when using single-window UI.
- Add a new `autoscrollbars` parameter to the Dialog:show function of Aseprite's Lua API.
- ~~Introduction of the Tab widget into the Lua API.~~
- ~~Introduction of the View widget into the Lua API.~~

EDIT: I will work on the crossed out items in another pull request to reduce the scope of this one.
EDIT2: Created https://github.com/aseprite/aseprite/issues/3991.

### DIFF
--- a/src/ui/window.cpp
+++ b/src/ui/window.cpp
@@ -518,17 +518,15 @@ bool Window::onProcessMessage(Message* msg)
 
         // Reposition/resize
         if (m_hitTest == HitTestCaption) {
-          int x = clickedWindowPos->x + (mousePos.x - clickedMousePos.x);
-          int y = clickedWindowPos->y + (mousePos.y - clickedMousePos.y);
-          moveWindow(gfx::Rect(x, y,
-                               bounds().w,
-                               bounds().h), true);
+          gfx::Point pos = clickedWindowPos->origin() + mousePos - clickedMousePos;
+          gfx::Rect rect(pos, bounds().size());
+          // Limit window position to avoid hiding the title bar
+          limitPosition(rect);
+
+          moveWindow(rect, true);
         }
         else {
-          int x, y, w, h;
-
-          w = clickedWindowPos->w;
-          h = clickedWindowPos->h;
+          gfx::Size size = clickedWindowPos->size();
 
           bool hitLeft = (m_hitTest == HitTestBorderNW ||
                           m_hitTest == HitTestBorderW ||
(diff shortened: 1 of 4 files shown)


src/ui/window.cpp
```diff
@@ -759,6 +765,32 @@ void Window::limitSize(int* w, int* h)
   *h = std::max(*h, border().height());
 }
 
+void Window::limitPosition(int* x, int* y, int* w, int* h)
```
dacap, line 768: Prefer returning a gfx::Point or gfx::Rect, and if w and h can be accessed via `bounds()`, remove these params (anyway just making these comments without further information, I saw that this PR will introduced several things later)

martincapello: My first thought was to follow how the limitSize function is defined, but I will see if I can refactor it.

martincapello: Oh, and this private function is used just in the commit that contains it, there is no plan to use it later.

martincapello: I believe I would just leave this as is, because this function just wants to limit the values of the parameters, which are the future values for the window's position and size. So I cannot use bounds(), and returning a Point or Rect would be inconvenient IMO.

dacap: Avoid using raw pointers and try to use gfx::Rect or gfx::Point if it's possible. There are still some functions using raw int* pointers in the code, are from the C era and should be refactored, I will highly recommend you to avoid using raw pointers (they are too error prone, needs a lot of `if (nonNull)` checks, and make the code a lot harder to read).

dacap: Just as an example, you can use something like this:
```c++
          gfx::Point pos = clickedWindowPos->origin() + mousePos - clickedMousePos;

          // Limit window position to avoid hiding the title bar
          limitPosition(pos);
```
And `limitPosition(gfx::Point& pos)`.

Something that I don't like from the original code, is that there is no check for `clickedWindowPos != nullptr` (not even an `assert`). And it came from [the very first import to Git code](https://github.com/aseprite/aseprite/blob/57e978b102f3ae0924d47b5f6b5f7d1305933363/jinete/src/jwindow.c#L346) (C code, year 2007).

martincapello: Oh I didn't want to modify the original code, or the least possible at least...now I'm going to update it, thanks.

PR #3951, opened 2023-07-06
Fix en strings typo

TakWolf:

Use diff util: https://github.com/aseprite-quest/aseprite-unified-languages-extension/blob/master/utils/aseini.py

```python
file_path = 'en.ini'
Aseini.load(file_path).save(file_path)
```
There are also some typo in 1.2.40, But I can't find the branch.


TakWolf:
By the way, I noticed slight differences between version 1.3 and version 1.2.

Is there a way to safely compat with both 1.3 and 1.2 in one language.ini

dacap:
Hi @TakWolf, we have to receive CLA signatures with real full names (not just nick names).

About the changes, I'll review, didn't notice the duplicated assign/convert_color_profile introduced in 974c13dd33e85dfd5de470ea80bc267834a6a7b6. Something to fix.

dacap:
> Is there a way to safely compat with both 1.3 and 1.2 in one language.ini

I'm not sure. I would recommend you to drop support for 1.2.40 (we are already removing that distribution from the website and soon from Steam, so v1.3 will be the only one to support).

Thanks for signing the CLA, I'll merge this now. Just as a side note, I've introduced this change: c527d68c648bc244a3f8952dd0c8ddf645968fbf (we keep the short Assign/Convert words for the Sprite Properties dialog).

PR #3967, opened 2023-07-13
Fix undo after inserting tiles in a tilemap containing tiles indexes that were removed from its tileset

martincapello:
This PR fixes this bug I found:

https://github.com/aseprite/aseprite/assets/5520828/bc2e731c-4939-416c-8298-2c08dcecd05b

Steps to reproduce it:
1) Create a new sprite and add a tilemap layer.
2) Draw some random stuff to generate some tiles.
3) Reduce tileset size.
4) Pick a tile and put it on top of one of the empty indices in the tilemap. This will shrink the tilemap, removing the tiles that are pointing to indexes that are no longer present in the tilemap.
5) Try to undo and you will see that you cannot go back to the initial tilemap you had created.

### DIFF
--- a/src/doc/algorithm/shrink_bounds.cpp
+++ b/src/doc/algorithm/shrink_bounds.cpp
@@ -262,10 +262,7 @@ bool shrink_bounds_tilemap(const Image* image,
     shrink = true;
     for (v=bounds.y; v<bounds.y+bounds.h; ++v) {
       const tile_t tile = get_pixel_fast<TilemapTraits>(image, u, v);
-      const tile_t tileIndex = tile_geti(tile);
-      const ImageRef tileImg = tileset->get(tileIndex);
-
-      if (tileImg && !is_plain_image(tileImg.get(), refpixel)) {
+      if (tile != refpixel) {
         shrink = false;
         break;
       }
@@ -281,10 +278,7 @@ bool shrink_bounds_tilemap(const Image* image,
     shrink = true;
     for (v=bounds.y; v<bounds.y+bounds.h; ++v) {
       const tile_t tile = get_pixel_fast<TilemapTraits>(image, u, v);
-      const tile_t tileIndex = tile_geti(tile);
-      const ImageRef tileImg = tileset->get(tileIndex);
-
-      if (tileImg && !is_plain_image(tileImg.get(), refpixel)) {
+      if (tile != refpixel) {
         shrink = false;
         break;
       }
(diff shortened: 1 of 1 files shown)


src/doc/algorithm/shrink_bounds.cpp
```diff
@@ -262,10 +262,7 @@ bool shrink_bounds_tilemap(const Image* image,
     shrink = true;
     for (v=bounds.y; v<bounds.y+bounds.h; ++v) {
       const tile_t tile = get_pixel_fast<TilemapTraits>(image, u, v);
-      const tile_t tileIndex = tile_geti(tile);
-      const ImageRef tileImg = tileset->get(tileIndex);
-
-      if (tileImg && !is_plain_image(tileImg.get(), refpixel)) {
+      if (tile != refpixel) {
```
dacap, line 265: Here we are comparing a tile with a pixel color (refpixel can be != 0 in indexed mode when the transparent color != 0)

Test this change with an indexed sprite and transparent mode = 1, e.g.:

image

dacap (changes requested):
Something I'm working on and seeing in the .aseprite decoder is that there is a specific code to remove these "tiles out of bounds":
https://github.com/aseprite/aseprite/blob/00b75a76a8e087c7377fb97b2a55b15b5cd29ed8/src/dio/aseprite_decoder.cpp#L983-L989

Introduced in ed8e2c8392b01fbe844bdab11a912a6f298c2d1d to fix #2877, as I'm touching this code, I'll review the decoder to (probably) keep this data (or ask what to do about it).

I'm still not sure if we should preserve the hidden tilemap data if the tileset is reduced.

dacap:
> I'm still not sure if we should preserve the hidden tilemap data if the tileset is reduced.

The indexes out of bounds will be preserved (a WIP in my [flip branch](https://github.com/dacap/aseprite/commit/77cfe6026b9a64a394a18aaded250e42b8193e16#diff-794b452ed9483be047cfc55a9949f383bc8b2b6b0e5aafc1eb803686ba5f8fa4R995-R1006)).

src/doc/algorithm/shrink_bounds.cpp
martincapello: I've looking into this and tested the case you mention, and it doesn't seem to be an issue for tilemaps (please correct me if I'm wrong). As far as I have been able to see, this change affects shrink_bounds_tilemap function, which is called only when the cel's image is from a tilemap layer. Then it doesn't matter if the sprite is indexed, because refpixel takes its value from the cel's image mask color, which still is 0 for the tilemap. 

The only odd thing I saw (but seems to be unrelated to this change, because it also happens on the latest rc) is that the tile with index 0 is painted with the idx-0 color in the tiles palette, instead of the  idx-1. And also after drawing something to generate tiles, I saw that it generates an "empty tile" (actually a tile that is painted with the idx-1 color, which is the current transparent color) to use it instead of the 0:
 Screen Shot 2023-07-19 at 10 54 27

In the other hand, when the transparent color is kept as 0, we got the expected result:
Screen Shot 2023-07-19 at 10 56 11

PR #3978, opened 2023-07-21
Fix glitch with Selection Tool + Alt (fix #3976)

### LINKED ISSUE #3976: Glitch with Selection Tool + Alt (Beta 1.3 rc5)
When using the rectangular selection tool there's a few keyboard shortcuts to modify the selection. Holding shift allows you to add more rectangles, and holding alt allows you to crop rectangles from the selection.
However, in the latest version of the beta alt seems to break as soon as your mouse touches the UI, and the selection starts rotating randomly.

I have attatched a video showing the problem. Note that the rotations are not intentional / undesirable.

Thank you for the amazing work you do! I have been using Aseprite for many years and it has become my favorite tool for pixel art.

**Aseprite version: Steam Beta 1.3 rc5
System: Windows 11 x64**

https://github.com/aseprite/aseprite/assets/108590573/a3e6a3cf-9a77-4886-9d75-33a898afd481

Gasparoken:
Prior to this fix, when the mouse exited and re-entered the editor area, the Subtract Selection quick command would inadvertently change to 'rotate shape'.

### DIFF
--- a/src/app/ui/editor/editor.cpp
+++ b/src/app/ui/editor/editor.cpp
@@ -1917,8 +1917,16 @@ bool Editor::onProcessMessage(Message* msg)
 
     case kMouseEnterMessage:
       m_brushPreview.hide();
-      updateToolLoopModifiersIndicators();
-      updateQuicktool();
+
+      // Do not update tool loop modifiers when the mouse exits and/re-enters
+      // the editor area while we are inside the same tool loop (hasCapture()).
+      // E.g. This avoids starting to rotate a rectangular marquee (Alt key
+      // pressed) if we have the subtract mode on (Shift+Alt) and touch the
+      // editor edge (MouseLeave/Enter)
+      if (!hasCapture()) {
+        updateToolLoopModifiersIndicators();
+        updateQuicktool();
+      }
       break;
 
     case kMouseLeaveMessage:


src/app/ui/editor/editor.cpp
```diff
@@ -1917,8 +1917,19 @@ bool Editor::onProcessMessage(Message* msg)
...
-      updateQuicktool();
+
+      // Not update tool loop modifiers when the mouse exits and
+      // re-enters the editor area while the mouse has an
+      // active 'click'.
+      // If it isn't done, when the mouse exits and re-enters the
+      // editor area, the Subtract Selection quick command
+      // (for example) will change to 'rotate shape'.
+      // More details:
+      // https://github.com/aseprite/aseprite/issues/3976
```
dacap, line 1928:
```suggestion
      // Do not update tool loop modifiers when the mouse exits and/re-enters
      // the editor area while we are inside the same tool loop (hasCapture()).
      // E.g. This avoids starting to rotate a rectangular marquee (Alt key
      // pressed) if we have the subtract mode on (Shift+Alt) and touch the
      // editor edge (MouseLeave/Enter)
```

dacap (changes requested):
It looks good :+1: 

Preferable the fix would go to `Editor::updateToolLoopModifiersIndicators()` (as it and `updateQuicktool` already take care of the `hasCapture()` case), but it's better to keep this simple as it is.

I'd modify the comment (removing the link to the issue as it'll be in the commit message and can be accessed through git blame) and write this commit message:

```
Don't rotate rectangular selection when we're in "subtract mode" and touch editor edges (fix #3976)

As Shift+Alt enables the subtract mode, if we touch an editor edges
and receive a MouseEnter message, we cannot update the selection
modifiers with the pressed Alt key because that will start rotating the
rectangular marquee automatically. We've to start rotating only if we
release the Alt key and then press it again (not by just moving the
mouse).

Recent regression introduced in bd91a6430fdc04c3bc4d35b5d0c11de80c763198
when the ordering of MouseEnter/Move/Leave message order was fixed.
```

Gasparoken:
Ready to merge.

PR #3981, opened 2023-07-24
Fix Magic Wand should "refer to visible layers" anyway when the active layer is hidden (fix #3939)

### LINKED ISSUE #3939: Magic Wand should "refer to visible layers" anyway when the active layer is hidden
Reported here:

https://community.aseprite.org/t/selecting-things-with-the-magic-wand-when-layer-focus-is-in-a-hidden-layer/19260

Gasparoken:
fix #3939

### DIFF
--- a/src/app/ui/editor/tool_loop_impl.cpp
+++ b/src/app/ui/editor/tool_loop_impl.cpp
@@ -829,11 +829,15 @@ tools::ToolLoop* create_tool_loop(
       return nullptr;
     }
     else if (!layer->isVisibleHierarchy()) {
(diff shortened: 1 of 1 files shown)


dacap:
From https://community.aseprite.org/t/selecting-things-with-the-magic-wand-when-layer-focus-is-in-a-hidden-layer/19260/2?u=dacap I thought that maybe if the layer is hidden the magic wand should always refer to visible layers (even if the "refer only active layer" option is checked), but probably that might be confusing too.

I'll merge this patch as it is.

PR #3986, opened 2023-07-28
Add save dynamics options between sessions (fix #3933)

### LINKED ISSUE #3933: Save pen pressure between sessions
As a first fix it would be nice to keep the pen pressure options between sessions:

https://community.aseprite.org/t/default-the-pen-pressure/19244

Gasparoken:
First approach and observations:

Selected DitheringMatrix preference is saved as an integer (selected index in the comboBox).

At the moment, dynamic options are configured:
- maxSize/Angle == max size/angle of the active tool
- The rest of the dynamic options are shared between all the tools.

Suggestions:
- Make all dynamic options global, or
- Make sets of dynamic options for each tool.

Starting a new sprite with dynamic options active + no warning about this can be confusing until the artist realizes dynamic options are active. Alternatives:
- Warn the artist that dynamic options are active.
- Start new sessions with Static sensors on Size/Angle/Gradient.

Suggestion: add "Dynamics ON/OFF" CheckBox.

fix #3933

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -459,6 +459,21 @@
       <option id="refer_to" type="FillReferTo" default="FillReferTo::ACTIVE_LAYER" />
       <option id="pixel_connectivity" type="PixelConnectivity" default="PixelConnectivity::FOUR_CONNECTED" />
     </section>
+    <section id="dynamics">
+      <option id="stabilizer" type="bool" default="false" />
+      <option id="stabilizer_factor" type="int" default="0" />
+      <option id="size" type="app::tools::DynamicSensor" default="app::tools::DynamicSensor::Static" />
+      <option id="angle" type="app::tools::DynamicSensor" default="app::tools::DynamicSensor::Static" />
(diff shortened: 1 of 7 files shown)


dacap:
I saw that the Dynamics configuration is in <tool> but I'm not sure if the "Same in All Tools" makes a difference:

image

Gasparoken:
At the moment, the Dynamics options works as "Same in all Tools".

I see that the settings for other tools work independently, and Ink's "Same in all Tools" does not affect them. I mention this to establish that the Ink's "Same in all tools" should not be related to the "Dynamics" options.

I think it would be great to memorize the dynamics options per tool. And implement an additional "Same in all Tools" flag in the 'Dynamics' options window. I'm going to create an issue for this new feature.

dacap:
> And implement an additional "Same in all Tools" flag in the 'Dynamics' options window. I'm going to create an issue for this new feature.

that’s right 👍

PR #3992, opened 2023-08-01
Fix Dialog.bounds doesn't work as expected (fix #3898)

### LINKED ISSUE #3898: Scripting Problem: Dialog.bounds doesn't work as expected
```
local dlg = Dialog("Hello World!")
dlg.bounds = Rectangle(64,64,128,128)
dlg:show{wait=false}
```
A simple script test trying to change dlg bounds and here's the result:
image

I have been test on both stable version and beta version on steam.So it seems to be a little bug.

And here's another thing:
```
local dlg = Dialog("Hello World!")
dlg:label{text="Hello World!"}
dlg:show{wait=false}

```
These code will show a dialog in the middle of screen.
image

But with these:

```
local dlg = Dialog("Hello World!")
dlg:label{text="Hello World!"}
local bounds = dlg.bounds
dlg:show{wait=false}
```

Dialog goes to the left-top corner again.Even we were just trying to access the bounds without doing anythings.
image

* Aseprite version: v1.3-rc4-x64, steam beta test
* System: Win10

Gasparoken:
fix #3898

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -413,6 +413,7 @@ int Dialog_show(lua_State* L)
       if (!rc.isEmpty()) {
         conn = dlg->window.Open.connect([dlg, rc]{
           dlg->setWindowBounds(rc);
+          dlg->window.setAutoRemap(false);
         });
       }
     }
@@ -1836,9 +1837,10 @@ int Dialog_set_data(lua_State* L)
 int Dialog_get_bounds(lua_State* L)
 {
   auto dlg = get_obj<Dialog>(L, 1);
-  if (!dlg->window.isVisible())
+  if (!dlg->window.isVisible() && dlg->window.bounds().isEmpty()) {
     dlg->window.remapWindow();
-
+    dlg->window.centerWindow(dlg->parentDisplay());
+  }
   push_new<gfx::Rect>(L, dlg->getWindowBounds());
   return 1;
 }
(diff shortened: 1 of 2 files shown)


Gasparoken:
I'll see if I can add some tests for this.

src/app/script/dialog_class.cpp
```diff
@@ -1591,9 +1592,6 @@ int Dialog_set_data(lua_State* L)
 int Dialog_get_bounds(lua_State* L)
 {
   auto dlg = get_obj<Dialog>(L, 1);
-  if (!dlg->window.isVisible())
-    dlg->window.remapWindow();
-
```
martincapello, line 1596: I've been testing a bit with this bit of code instead:
```c++
  if (!dlg->window.isVisible() && dlg->window.bounds().isEmpty()) {
    dlg->window.remapWindow();
    dlg->window.centerWindow(dlg->parentDisplay());
  }
```
And it seems to fix the case when the script asks for a Dialog's bounds before showing it, avoiding returning empty bounds. If we decide to introduce this change, remember to update the test.

martincapello (commented):
It looks good! I've just suggested an alternative fix to avoid returning empty bounds when the script asks for Dialog's bounds before the show method is invoked.

martincapello (approved):
LGTM!

dacap:
I was finally able to test this on Windows, macOS, and Linux and the fix is working great  :+1:

Issue #4005, opened 2023-08-16
Add missing features to Tab widget


martincapello:
- [x] Include which is the selected tab when asking for Dialog.data.
- [x] Add possibility to change the position of the tabs selector to the bottom.
- [x] ~~Add selected flag to each tab?~~ Currently the selected attribute of the endtabs() parameter is used to set the initial selected tab and it seems to be enough.
- [x] Add onclick event handlers to each tab() and ontabchange event handler to endtabs().

PR #4008, opened 2023-08-18
Implement Tab's widget missing features

martincapello:
Fix https://github.com/aseprite/aseprite/issues/4005

NOTE: To implement these features it was required to refactor the existing code related to the tabs widgets creation.

### DIFF
--- a/src/app/script/engine.cpp
+++ b/src/app/script/engine.cpp
@@ -474,6 +474,9 @@ Engine::Engine()
   setfield_integer(L, "LEFT",   ui::LEFT);
   setfield_integer(L, "CENTER", ui::CENTER);
   setfield_integer(L, "RIGHT",  ui::RIGHT);
+  setfield_integer(L, "TOP",    ui::TOP);
+  setfield_integer(L, "BOTTOM", ui::BOTTOM);
+
   lua_pop(L, 1);
 
   // Register classes/prototypes
--- a/src/app/script/tabs_widget.cpp
+++ b/src/app/script/tabs_widget.cpp
@@ -1,0 +1,184 @@
+// Aseprite
+// Copyright (c) 2022-2023  Igara Studio S.A.
+//
+// This program is distributed under the terms of
+// the End-User License Agreement for Aseprite.
+
+#include "app/script/tabs_widget.h"
+
+#ifdef ENABLE_UI
+
+using namespace ui;
+
+namespace app {
+namespace script {
+
+// static
+ui::WidgetType Tabs::Type()
+{
+  static ui::WidgetType type = ui::kGenericWidget;
+  if (type == ui::kGenericWidget)
+    type = ui::register_widget_type();
+  return type;
+}
+
+Tabs::Tabs(int selectorFlags) : m_selectorFlags(selectorFlags)
+{
+  m_sepL.setExpansive(true);
+  m_sepR.setExpansive(true);
+  m_pages.setExpansive(true);
+  setType(Type());
+  layoutChilds();
+}
+
+Tab* Tabs::addTab(Grid* content)
+{
+  m_pages.addChild(content);
+
+  if (m_buttons.children().empty()) {
+    m_buttonsBox.addChild(&m_buttons);
+    m_buttons.ItemChange.connect([this](ButtonSet::Item* selItem) {
+      int oldSelectedTabIndex = m_selectedTab;
+      for (int i=0; i<m_pages.children().size(); ++i) {
+        auto tab = m_pages.children()[i];
+        bool isSelectedTab = selItem->text() ==  tab->text();
+        tab->setVisible(isSelectedTab);
+        if (isSelectedTab)
+          m_selectedTab = i;
+      }
+      layout();
+      if (oldSelectedTabIndex != m_selectedTab) {
+        TabChanged();
+      }
+    });
+  }
+  else {
+    m_buttons.setColumns(m_pages.children().size());
+  }
+  auto tab = new Tab();
+  tab->setText(content->text());
+  m_buttons.addItem(tab);
+  // Select the first tab by default.
+  if (m_buttons.children().size() == 1) {
+    selectTab(0);
+  }
+
+  m_buttons.initTheme();
+
+  return tab;
+}
+
+void Tabs::selectTab(int index)
+{
+  if (index >= 0 && index < m_pages.children().size()) {
+    for (int i=0; i<m_pages.children().size(); ++i) {
+      m_buttons.getItem(i)->setSelected(i == index);
+      m_pages.children()[i]->setVisible(i == index);
+    }
+    m_selectedTab = index;
+  }
+}
+
+void Tabs::setSelectorFlags(int selectorFlags)
+{
+  if (selectorFlags != m_selectorFlags) {
+    m_selectorFlags = selectorFlags;
+    layoutChilds();
+  }
+}
+
+int Tabs::tabIndexById(const std::string& id) const
+{
+  for (int i=0; i<m_pages.children().size(); ++i) {
+    if (m_pages.children()[i]->id() == id)
+      return i;
+  }
+  return -1;
+}
+
+int Tabs::tabIndexByText(const std::string& text) const
+{
+  for (int i=0; i<m_pages.children().size(); ++i) {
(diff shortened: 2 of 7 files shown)


dacap:
@martincapello if you have used a .lua script to test this, could you please share it here or in a [gist](https://gist.github.com/)?

src/app/script/engine.cpp
```diff
@@ -474,6 +474,9 @@ Engine::Engine()
   setfield_integer(L, "LEFT",   ui::LEFT);
   setfield_integer(L, "CENTER", ui::CENTER);
   setfield_integer(L, "RIGHT",  ui::RIGHT);
+  setfield_integer(L, "TOP",    ui::TOP);
+  setfield_integer(L, "BOTTOM", ui::BOTTOM);
```
dacap, line 478: I'm not a big fan of exposing these internal flags as they are, but it's the most simple way, we can keep this as is.

src/app/script/tabs_widget.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+void Tabs::setSelectorFlags(int selectorFlags)
+{
+  if (selectorFlags != m_selectorFlags) {
+    m_selectorFlags = selectorFlags;
+    layoutChilds();
+  }
+}
+
+int Tabs::tabIndexById(std::string id) const
```
dacap, line 94:
```suggestion
int Tabs::tabIndexById(const std::string& id) const
```

src/app/script/tabs_widget.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+std::string Tabs::tabId(int index) const
+{
+  return index >= 0 && index < m_pages->children().size() ? m_pages->children()[index]->id() : "";
+}
+
+std::string Tabs::tabText(int index) const
+{
+  return index >= 0 && index < m_pages->children().size() ? m_pages->children()[index]->text() : "";
+}
```
dacap, line 120: gcc and clang optimize this correctly, but msvc [prefers std::string()](https://godbolt.org/z/WvEToo5f7) instead [of ""](https://godbolt.org/z/eqovn8xGT) for an empty string. And break the length of these lines:
```suggestion
std::string Tabs::tabId(int index) const
{
  return index >= 0 && index < m_pages->children().size() ? m_pages->children()[index]->id() : std::string();
}

std::string Tabs::tabText(int index) const
{
  return index >= 0 && index < m_pages->children().size() ? m_pages->children()[index]->text() : std::string();
}
```

src/app/script/tabs_widget.h
```diff
@@ -0,0 +1,63 @@
...
+  void setSelectorFlags(int selectorFlags);
+
+  int tabIndexById(std::string text) const;
+  int tabIndexByText(std::string text) const;
+  std::string tabId(int index) const;
+  std::string tabText(int index) const;
+  int selectedTab() const;
+  int size() const { return m_pages ? m_pages->children().size() : 0; };
+
+  obs::signal<void(std::string deselTabId, std::string selTabId)> TabChanged;
```
dacap, line 38: Probably I would just remove the signal parameters to simplify and access the tabs widget to get the selected tab:
```suggestion
  obs::signal<void()> TabChanged;
```

src/app/script/tabs_widget.h
```diff
@@ -0,0 +1,63 @@
...
+
+private:
+  ui::Separator* separator() {
+    auto sep = new ui::Separator("", ui::HORIZONTAL);
+    sep->setExpansive(true);
+    return sep;
+  }
```
dacap, line 45: Can be static (or just an auxiliary/static function in tabs_widgets.cpp). In this way ui/separator.h can go to the .cpp too.

src/app/ui/button_set.h
```diff
@@ -27,6 +27,8 @@ namespace app {
       os::Surface* iconSurface() const override { return m_icon ? m_icon->bitmap(0) : nullptr; }
       skin::SkinPartPtr icon() const { return m_icon; }
       ButtonSet* buttonSet();
+
+      obs::signal<void(Item*)> Click;
```
dacap, line 31: Is this really needed? Could we use the ButtonSet ItemChange signal? Probably it's too complex, if that is the case maybe we can just implement/derive a ButtonSet Item for the tabs widget specifically (using `ButtonSet::addItem(Item*, ...)`)

src/ui/grid.cpp
```diff
@@ -129,6 +129,26 @@ void Grid::setStyle(Style* style)
   setGap(style->gap());
 }
 
+void Grid::setColumns(int columns)
+{
+  if (columns == m_colstrip.size()) return;
```
dacap, line 134:
```suggestion
  if (columns == m_colstrip.size())
    return;
```

src/ui/grid.cpp
```diff
@@ -129,6 +129,26 @@ void Grid::setStyle(Style* style)
...
+
+  int oldSize = m_colstrip.size();
+
+  m_colstrip.resize(columns);
+  for (std::size_t col=0; col<m_colstrip.size(); ++col) {
+    m_colstrip[col].size = 0;
+    m_colstrip[col].expand_count = 0;
+  }
```
dacap, line 142: No sure if we should reset the information of the columns after the resize or doing something similar as the rows (I didn't test this).

src/app/script/tabs_widget.cpp
```diff
@@ -0,0 +1,166 @@
...
+  return type;
+}
+
+Tabs::Tabs(int selectorFlags) : m_selector(new HBox()),
+                                m_sepL(separator()),
+                                m_sepR(separator()),
+                                m_sepTB(separator()),
+                                m_buttonsBox(new HBox()),
+                                m_pages(new VBox()),
+                                m_selectorFlags(selectorFlags)
```
dacap, line 31: You can avoid pointers of widgets if they are always created (using values like `HBox m_selector;`, etc.), e.g. https://github.com/aseprite/aseprite/blob/f4aa19ad8c12a97988383498dd39dd2dcc0e6d22/src/app/ui/color_bar.h#L208

dacap (changes requested):

src/app/script/tabs_widget.cpp
martincapello: I thought in using as you suggested, but since it wasn't clear to me what happens when the delete operator is called on such values (or pointers to values) I decided to avoid it. Could you clarify this point?

dacap: The `~Widget` destructor removes the widget itself from its parent, so this prevent a double deletion (as children will remove themselves from the parent). The order of execution when the Tabs is deleted is this:

1. ~Tabs is called
2. After that the destructor of each of its field is called (so ~HBox & ~Widget of each children is called, etc.)
3. Finally the ~Widget is called for the tabs widget

If a widget is still a children of Tabs (e.g. it was added with `addChild(new ...)`), it will be deleted automatically because ~Widget will delete the chlidren of the tabs widget.

Here an example: https://godbolt.org/z/rhGEdTfnr

martincapello: I see it now, thanks!

src/app/script/tabs_widget.h
martincapello: Just removed this, I don't use it anymore. Thank you!

src/app/ui/button_set.h
martincapello: I've considered using the ItemChange signal, but I couldn't figure out how to use it without creating a new signal for each item. Then, since I had to create a new signal anyway, I decided to connect it directly via  Dialog_connect_signal. Sure, no problem, I will create a derived class.

src/ui/grid.cpp
martincapello: I have just tried removing the for loop, and it seems to work anyway. So, I've removed it. Thanks!

martincapello:
Now is ready for review, besides introducing your suggestions, I have fixed a couple of things:
- Set same tab page size for all tabs using the max size of all the tab pages of the group. This avoids an annoying resize of the dialog when changing tabs.
- Fix expansiveness of the tab content. Now when you set the vexpand=true for an :endtabs{} it will work as expected.

PR #4012, opened 2023-08-23
Add memorization of tool-shared dynamics options (fix #4007)

### LINKED ISSUE #4007: Dynamics option fixes
- [ ] **Important** Don't save/set the properties of all tools when the "Save in all Tools" is checked, we could use a shared tool section (`[tool.shared.dynamics]`?) for that. Because unchecking the "same in all tools" should not keep the dynamics on all tools.
- [ ] Use the matrix name/ID (instead of combobox index) to save the selected matrix for dynamic dithering
- [ ] Reset dynamics option from `Preferences::resetToolPreferences()` (there is [a comment about this in pref.xml](https://github.com/aseprite/aseprite/blob/572cdf2b0ee9deb4a21acea72ddcf47a8601e94e/data/pref.xml#L448))

Gasparoken:
fix #4007

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -320,7 +320,7 @@
       <option id="discard_brush" type="bool" default="false" />
     </section>
     <section id="shared">
-      <option id="share_dynamics" type="bool" default="false" />
+      <option id="share_dynamics" type="bool" default="true" />
       <option id="share_ink" type="bool" default="false" />
       <option id="ink" type="app::tools::InkType" default="app::tools::InkType::DEFAULT" />
     </section>
@@ -472,7 +472,7 @@
       <option id="min_size" type="int" default="1" />
       <option id="min_angle" type="int" default="1" />
       <option id="color_from_to" type="app::tools::ColorFromTo" default="app::tools::ColorFromTo::BgToFg" />
-      <option id="matrix_index" type="int" default="0" />
+      <option id="matrix_name" type="std::string" />
       <option id="min_pressure_threshold" type="double" default="0.1" />
       <option id="max_pressure_threshold" type="double" default="0.9" />
       <option id="min_velocity_threshold" type="double" default="0.1" />
--- a/src/app/pref/preferences.cpp
+++ b/src/app/pref/preferences.cpp
@@ -147,23 +147,29 @@ bool Preferences::isSet(OptionBase& opt) const
 
 ToolPreferences& Preferences::tool(tools::Tool* tool)
 {
-  ASSERT(tool != NULL);
+  std::string id;
 
-  auto it = m_tools.find(tool->getId());
+  // If tool == nullptr it means that the tool's options are shared with all tools.
+  if (tool)
+    id = tool->getId();
+  else
+    id = "shared";
+
+  auto it = m_tools.find(id);
   if (it != m_tools.end()) {
     return *it->second;
   }
   else {
-    std::string section = std::string("tool.") + tool->getId();
+    std::string section = std::string("tool.") + id;
     ToolPreferences* toolPref = new ToolPreferences(section);
 
     // Default size for eraser, blur, etc.
-    if (tool->getInk(0)->isEraser() ||
-        tool->getInk(0)->isEffect()) {
+    if (tool && (tool->getInk(0)->isEraser() ||
+                 tool->getInk(0)->isEffect())) {
       toolPref->brush.size.setDefaultValue(8);
     }
 
-    m_tools[tool->getId()] = toolPref;
+    m_tools[id] = toolPref;
     toolPref->load();
     return *toolPref;
   }
@@ -223,6 +229,9 @@ void Preferences::resetToolPreferences(tools::Tool* tool)
   del_config_section((section + ".brush").c_str());
   del_config_section((section + ".spray").c_str());
   del_config_section((section + ".floodfill").c_str());
+  del_config_section((section + ".dynamics").c_str());
+  std::string shared = "tool.shared";
+  del_config_section(shared.c_str());
 }
 
 void Preferences::removeDocument(Doc* doc)
--- a/src/app/ui/dithering_selector.cpp
+++ b/src/app/ui/dithering_selector.cpp
@@ -198,7 +198,14 @@ void DitheringSelector::onInitTheme(ui::InitThemeEvent& ev)
     setSizeHint(getItem(0)->sizeHint());
 }
 
-void DitheringSelector::regenerate()
+void DitheringSelector::setSelectedItemByName(const std::string& name)
+{
+  int index = findItemIndex(name);
+  setSelectedItemIndex(index);
+  regenerate(index);
+}
+
+void DitheringSelector::regenerate(int selectedItemIndex)
 {
   deleteAllItems();
 
@@ -255,9 +262,9 @@ void DitheringSelector::regenerate()
       }
       break;
   }
-
-  setSelectedItemIndex(0);
-  setSizeHint(getItem(0)->sizeHint());
+  selectedItemIndex = std::clamp(selectedItemIndex, 0, std::max(0, getItemCount()-1));
+  setSelectedItemIndex(selectedItemIndex);
+  setSizeHint(getItem(selectedItemIndex)->sizeHint());
 }
 
 render::DitheringAlgorithm DitheringSelector::ditheringAlgorithm()
--- a/src/app/ui/dithering_selector.h
+++ b/src/app/ui/dithering_selector.h
@@ -27,13 +27,13 @@ namespace app {
 
     render::DitheringAlgorithm ditheringAlgorithm();
     render::DitheringMatrix ditheringMatrix();
+    void setSelectedItemByName(const std::string& name);
 
   protected:
     void onInitTheme(ui::InitThemeEvent& ev) override;
 
   private:
-    void regenerate();
-
+  void regenerate(int selectedItemIndex = 0);
     Type m_type;
     obs::scoped_connection m_extChanges;
   };
--- a/src/app/ui/dynamics_popup.cpp
+++ b/src/app/ui/dynamics_popup.cpp
@@ -289,9 +289,6 @@ DynamicsPopup::DynamicsPopup(Delegate* delegate)
       // Save the old dynamics options:
       saveDynamicsPref(!sameInAllTools);
       Preferences::instance().shared.shareDynamics(sameInAllTools);
-      // Load the new dynamics options:
-      loadDynamicsPref(sameInAllTools);
-      refreshVisibility();
     });
   m_dynamics->gradientPlaceholder()->addChild(m_ditheringSel);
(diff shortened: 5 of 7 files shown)


Gasparoken:
First approach to solve first point of #4007.
I'll try to implement this https://github.com/aseprite/aseprite/issues/4007#issuecomment-1684317298

dacap:
I think there is no need to duplicate the shared preferences, probably we could use the `Preferences::tool(nullptr)` to reference the shared options.

Gasparoken:
> I think there is no need to duplicate the shared preferences, probably we could use the `Preferences::tool(nullptr)` to reference the shared options.

I am working on it, today I'll push separate commits for each point of #4007.

Gasparoken:
Ready for review.

src/app/pref/preferences.cpp
```diff
@@ -139,7 +139,20 @@ bool Preferences::isSet(OptionBase& opt) const
...
+      return *it->second;
+    }
+    else {
+      std::string section = std::string("tool.shared");
+      ToolPreferences* toolPref = new ToolPreferences(section);
+      m_tools["shared"] = toolPref;
+      toolPref->load();
+      return *toolPref;
+    }
+  }
```
dacap, line 155: As this code is similar to the one that follows, probably we can create an `id` local variable and do something like this:

```suggestion
 std::string id;
 
  // If tool == nullptr it means that the tool's options are shared with all tools.
  if (tool)
    id = tool->getId();
  else
    id = "shared";
```

and use `id` instead of `tool->getId()` in the following code, and check for `tool` before using `tool->`.

src/app/ui/dithering_selector.cpp
```diff
@@ -198,7 +198,14 @@ void DitheringSelector::onInitTheme(ui::InitThemeEvent& ev)
     setSizeHint(getItem(0)->sizeHint());
 }
 
-void DitheringSelector::regenerate()
+void DitheringSelector::setSelectedItemByName(std::string name)
```
dacap, line 201:
```suggestion
void DitheringSelector::setSelectedItemByName(const std::string& name)
```

src/app/ui/dithering_selector.cpp
```diff
@@ -255,9 +262,9 @@ void DitheringSelector::regenerate()
       }
       break;
   }
-
-  setSelectedItemIndex(0);
-  setSizeHint(getItem(0)->sizeHint());
+  selectedItemIndex = std::min(std::max(0, selectedItemIndex), getItemCount()-1);
```
dacap, line 265:
```suggestion
  selectedItemIndex = base::clamp(selectedItemIndex, 0, std::max(0, getItemCount()-1));
```
We need to include "base/clamp.h" for this.

src/app/ui/dithering_selector.h
```diff
@@ -27,13 +27,13 @@ namespace app {
 
     render::DitheringAlgorithm ditheringAlgorithm();
     render::DitheringMatrix ditheringMatrix();
+    void setSelectedItemByName(std::string name);
```
dacap, line 30:
```suggestion
    void setSelectedItemByName(const std::string& name);
```

src/app/ui/dynamics_popup.cpp
```diff
@@ -289,15 +306,20 @@ void DynamicsPopup::setOptionsGridVisibility(bool state)
...
 }
 
-int DynamicsPopup::ditheringIndex() const {
+std::string DynamicsPopup::ditheringMatrixName() const {
   if (m_ditheringSel)
-    return m_ditheringSel->getSelectedItemIndex();
-  return 0;
+    return m_ditheringSel->getItemText(
+      m_ditheringSel->getSelectedItemIndex());
+  return Strings::dithering_selector_no_dithering();
```
dacap, line 313: Not sure if this should return an empty string in this case:
```suggestion
  return std::string();
```

src/ui/combobox.h
```diff
@@ -75,6 +75,7 @@ namespace ui {
 
     int getSelectedItemIndex() const;
     void setSelectedItemIndex(int itemIndex);
+    int getSelectedIndexByName(std::string itemName) const;
```
dacap, line 78:
```suggestion
    int getSelectedIndexByName(const std::string& itemName) const;
```

src/app/ui/dynamics_popup.cpp
```diff
@@ -50,6 +52,8 @@ enum {
 
 } // anonymous namespace
 
+static bool g_same_toggle = false;
+
```
dacap, line 56: Unused variable:
```suggestion
```

dacap (changes requested):

src/app/ui/dithering_selector.cpp
Gasparoken: I used `std::clamp` instead.

dacap: Oh yes, now we can use it 👍

Gasparoken:
Squashed, and ready for review.

dacap (changes requested):

src/ui/combobox.cpp
```diff
@@ -312,6 +312,14 @@ void ComboBox::setSelectedItemIndex(int itemIndex)
   }
 }
 
+int ComboBox::getSelectedIndexByName(const std::string& itemName) const
```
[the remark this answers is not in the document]
Gasparoken: Yes! You're right, I'll use `findItemIndex()` instead.

data/pref.xml
```diff
@@ -317,9 +317,9 @@
       <option id="discard_brush" type="bool" default="false" />
     </section>
     <section id="shared">
-      <option id="share_dynamics" type="bool" default="false" />
       <option id="share_ink" type="bool" default="false" />
       <option id="ink" type="app::tools::InkType" default="app::tools::InkType::DEFAULT" />
+      <option id="share_dynamics" type="bool" default="true" />
```
dacap, line 322: Not sure why this line changed to this place now, but you can just change `false` to `true`.

src/app/ui/dynamics_popup.cpp
```diff
@@ -289,15 +304,20 @@ void DynamicsPopup::setOptionsGridVisibility(bool state)
     expandWindow(sizeHint());
 }
 
-int DynamicsPopup::ditheringIndex() const {
+std::string DynamicsPopup::ditheringMatrixName() const {
```
dacap, line 307:
```suggestion
std::string DynamicsPopup::ditheringMatrixName() const
{
```

src/app/ui/dynamics_popup.cpp
```diff
@@ -289,15 +304,20 @@ void DynamicsPopup::setOptionsGridVisibility(bool state)
...
-  return 0;
+    return m_ditheringSel->getItemText(
+      m_ditheringSel->getSelectedItemIndex());
+  return std::string();
 }
 
-void DynamicsPopup::loadDynamicsPref() {
-  auto& dynaPref = Preferences::instance().tool(
-    App::instance()->activeTool()).dynamics;
+void DynamicsPopup::loadDynamicsPref(bool sameInAllTools) {
```
dacap, line 314:
```suggestion
void DynamicsPopup::loadDynamicsPref(bool sameInAllTools)
{
```

src/app/ui/dynamics_popup.cpp
```diff
@@ -324,10 +344,33 @@ void DynamicsPopup::loadDynamicsPref() {
            dynaPref.gradient() == tools::DynamicSensor::Velocity);
 
   if (m_ditheringSel)
-    m_ditheringSel->setSelectedItemIndex(dynaPref.matrixIndex());
+    m_ditheringSel->setSelectedItemByName(dynaPref.matrixName());
+}
+
+void DynamicsPopup::saveDynamicsPref(bool sameInAllTools) {
```
dacap, line 350:
```suggestion
void DynamicsPopup::saveDynamicsPref(bool sameInAllTools)
{
```

dacap (changes requested):
Just some minor changes, and remember to include fix #number of issue in the commit summary itself.

Gasparoken:
Ready for merge.

dacap:
Testing this, it works in a strange way (I think checking/unchecking "Same in all tools" option shouldn't modify the visible preferences of the active tool):

https://github.com/aseprite/aseprite/assets/39654/1ed563fe-703f-4e4d-82b5-6d4d1279d40c

The "Same in all tools" checkbox is switching between all the options between the active tool and the shared values. I think the checkbox shouldn't change the values.

This is 1) if I set some dynamics and the "Same in all tools" is unchecked, checking it should copy the active config as the shared values, and 2) If "Same in all tools" is checked and I uncheck it, it should copy the shared values to the active tool config (without touching other tools), in this way the visible options of the active tool will not be modified.

What do you think about this?

Gasparoken:
> This is 1) if I set some dynamics and the "Same in all tools" is unchecked, checking it should copy the active config as the shared values, and 2) If "Same in all tools" is checked and I uncheck it, it should copy the shared values to the active tool config (without touching other tools), in this way the visible options of the active tool will not be modified.
> 
> What do you think about this?

The current behavior is like a switch between the particular tool settings and shared settings as a separate memories.
I did it in that way to not destroy the settings of the tool: 
- This allows to use 2 configs as [shared set] <-> [tool's specific set] toggle button.
- This avoids to destroy the configuration of other tools accidentally by pressing "Same in all tools" + closing the Dynamics window.
- The drawback as you said: this can be a strange behavior for artists.

Maybe a different UI can improve the idea:
Sprite-0002


If this behavior is strange, so will the following case:
This implementation doesn't consider to save the `maxSize` and `maxAngle` as a `DynamicsOptions` since both settings are saved always as per-tool setting. So, the `DynamicsOptions` with "Same in all tools" marked won't have the same settings in all tools because each tool has its own `maxSize` and `maxAngle` setting.
Again, this is in this way to prioritize settings preservation, but I understand it could be confusing.
@dacap I'll need your opinion if I have to change the implementation to do some special with `maxSize` and `maxAngle`.

In conclusion of all this I think:
1. Should Brush Size/Brush Angle/Brush Type have a "Same in all tools" check box?
2. Implement something better in all ways: save the settings just as the Custom Brushes.
3. For the current commit, I'll modify the "Same in all tools" behavior like you said.

dacap:
> Should Brush Size/Brush Angle/Brush Type have a "Same in all tools" check box?

No right now (there is a plan to redesign the brush popup completely).

> Implement something better in all ways: save the settings just as the Custom Brushes.

No, this is too much for the moment and I think we have to still find a right design/UI for all these problems (redesigning the brush popup). It's something for the future.

> For the current commit, I'll modify the "Same in all tools" behavior like you said.

👍 

As a side note: I know that you squashed the commits in the past, but it would be great to reduce the commits in a future (e.g. the matrix name change could be a first PR/commit so it's not anymore in this PR and already merged). Right now you can continue all the work in this one, but for a future making smaller PRs/commits for each change will be a benefit in case that a commit must re-reviewed.

Gasparoken:
Ready.

PR #4023, opened 2023-08-29
Fix secondary color with custom brush paints the FgColor instead the BgColor (fix #4013)

### LINKED ISSUE #4013: Secondary color with Custom Brush paints the FgColor instead the BgColor
Pencil + Custom Brush: only the mask color as secondary color acts as eraser, otherwise FgColor is used.

Desired behavior:
The secondary color with Custom Brush should paint the BgColor and the set opacity should be applied.

* Aseprite version: 1.3rc6
* System: macOS

Gasparoken:
Before this fix to change the main color of the image brush, it was necessary to choose a new color from the palette. The secondary color can now also be used.

Also added some tests for image brushes:
Use of custom brush in all pixel formats VS all sprite pixel formats

fix #4013

### DIFF
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -1778,8 +1778,8 @@ void Manager::onInitTheme(InitThemeEvent& ev)
         gfx::Rect bounds = window->bounds();
         bounds *= newUIScale;
         bounds /= oldUIScale;
-        bounds.x = std::clamp(bounds.x, 0, displaySize.w - bounds.w);
-        bounds.y = std::clamp(bounds.y, 0, displaySize.h - bounds.h);
+        bounds.x = std::clamp(bounds.x, 0, std::max(0, displaySize.w - bounds.w));
+        bounds.y = std::clamp(bounds.y, 0, std::max(0, displaySize.h - bounds.h));
         window->setBounds(bounds);
       }
     }


Gasparoken:
This can be a first approach.

martincapello:
I've been testing this drawing an RGB pattern into an indexed sprite and the transparent color is painted black:
Screen Shot 2023-10-02 at 15 42 07

Is this expected?

dacap:
Just as a possible simpler solution would be to do something like the X key (switch colors) and use the modified brush when painting with the right button. To avoid modifying the ink processing.

Gasparoken:
> Just as a possible simpler solution would be to do something like the X key (switch colors) and use the modified brush when painting with the right button. To avoid modifying the ink processing.

Thanks @dacap. I tried a different approach.

martincapello:
Works great when the color mode of the sprite is the same as the color mode of the pattern (tested RGB and indexed only, didn't test grayscale). It doesn't work well otherwise. Not sure which level of priority does this have, but just wanted to let you know.

src/app/ui/editor/tool_loop_impl.cpp
```diff
@@ -143,9 +144,7 @@ class ToolLoopBase : public tools::ToolLoop {
                ToolLoopParams& params)
     : m_editor(editor)
     , m_tool(params.tool)
-    , m_brush(params.brush)
     , m_origBrush(params.brush)
-    , m_oldPatternOrigin(m_brush->patternOrigin())
```
martincapello, line 148: Could we keep these lines and remove the ones I indicate below?

src/app/ui/editor/tool_loop_impl.cpp
```diff
@@ -189,6 +188,20 @@ class ToolLoopBase : public tools::ToolLoop {
     ASSERT(m_ink);
     ASSERT(m_controller);
 
+    m_brush = params.brush;
```
martincapello, line 191: If we leave the lines removed above, we don't need this line.

src/app/ui/editor/tool_loop_impl.cpp
```diff
@@ -189,6 +188,20 @@ class ToolLoopBase : public tools::ToolLoop {
...
+      m_brush->setImageColor(
+        Brush::ImageColor::MainColor,
+        color_utils::color_for_target_mask(
+          (m_button == Left ? Preferences::instance().colorBar.fgColor() :
+                              Preferences::instance().colorBar.bgColor()),
+          ColorTarget(ColorTarget::TransparentLayer,
+                      m_brush->image()->pixelFormat(),
+                      -1)));
+    }
+    m_oldPatternOrigin = m_brush->patternOrigin();
```
martincapello, line 203: The same for this line, we can remove it if we leave the constructor as is.

martincapello (commented):
LGTM, I've just suggested some minor changes.

martincapello (approved):
🚀

dacap:
I'm seeing that this doesn't work as the X key to switch colors, I think that right click should paint with the inverted brush (the brush that is created after pressing the X key):

x

In the example ☝️ the result for right click should paint the top-right example (like pressing X and pressing left click but keeping the original brush intact).

Gasparoken:
> I think that right click should paint with the inverted brush (the brush that is created after pressing the X key):

The feature you cite would be appropriate if the image brush is only made of a couple of colors, but when the brush contains more than two colors:

- How it'll work? and
- Under which criteria will aseprite will choose primary and secondary color if the image brush doesn't contain the current fgColor and bgColor? I think it would be nice if it were selectable, but I think it's too much for now.

The initial intention of this PR was to have a way to paint with a image brush as the standard brushes, i.e. picking and painting the colors as regular brushes.
To return to the original colors, is needed to pick the brush slot again.
Nov-08-2023 14-56-41

dacap:
Finally merged in 12d81352647e96c8ac6d70e4a252c37ce5a29ade with a following refactor in 92edd5f70059a03af90e461156284a9388d66bdb, the main change/diff related to this PR is here:

https://github.com/aseprite/aseprite/commit/92edd5f70059a03af90e461156284a9388d66bdb#diff-65027930b4f13933e8a4fa47c33035857485afb34ff1222d407b5f8f64efb6a0R193-R217

If we right-click a new brush is created with new images so we don't modify the original left-click brush (whatever it is, stamp or with custom main/bg colors).

PR #4025, opened 2023-08-30
Fix in Save dialogs, current directory is not last visited (fix #3996)

### LINKED ISSUE #3996: Save path for never-previously-saved Workspaces defaults to the Desktop.
Whenever I create a new Document and save it for the first time, the default save path is my Desktop. I have never saved anything directly to the Desktop with ASEprite, and navigating to my usual save folders adds several (admittedly trivial) steps to my Save As experience.
This, I find undesirable.

Like, wow, substantially more undesirable than I would have imagined; I am really surprised at how extremely deeply undesirable I have been finding it, _hoo-whee mang_ it is a bad scene. :V

Other users with this version have shared discovery of this weird bug as well, and have been similarly frustrated.

Version 1.3-rc5-x64, nor any previous version to my memory, ever did this, and I find myself wondering why it might have started.

In all cases, please fix it, and thank you for looking into it! ;0; <3

..using ASEprite v.1.3-rc6-x64 on Win10.

Gasparoken:
This regression was added in 1.3rc5.
Every time you wanted to save a new sprite the dialog would start at the root path instead of the last used directory location.

fix #3996

### DIFF
--- a/src/app/ui/file_selector.cpp
+++ b/src/app/ui/file_selector.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2022  Igara Studio S.A.
+// Copyright (C) 2019-2023  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -411,13 +411,9 @@ bool FileSelector::show(
   fs->refresh();
 
   // We have to find where the user should begin to browse files
-  std::string start_folder_path;
-  if (initialPath.empty()) {
-    start_folder_path = get_initial_path_to_select_filename(initialPath);
-  }
-  else {
-    start_folder_path = base::get_file_path(initialPath);
-  }
+  std::string start_folder_path =
+    base::get_file_path(
+      get_initial_path_to_select_filename(initialPath));
 
   IFileItem* start_folder = fs->getFileItemFromPath(start_folder_path);
   if (!start_folder) {


Gasparoken:
This solution solves the problem, but I don't know the initial intention of the fix where the regression was inserted. So maybe there's something deeper.

src/app/ui/file_selector.cpp
```diff
@@ -411,13 +411,9 @@ bool FileSelector::show(
...
-  std::string start_folder_path;
-  if (initialPath.empty()) {
-    start_folder_path = get_initial_path_to_select_filename(initialPath);
-  }
-  else {
-    start_folder_path = base::get_file_path(initialPath);
-  }
+  std::string start_folder_path = base::get_file_path(initialPath);
+  if (start_folder_path.empty())
+    start_folder_path = get_current_dir_for_file_selector();
```
dacap, line 416: We can use `get_initial_path_to_select_filename()` to get the initial path+filename path, use the following code which has the same effect as the proposed fix:
```suggestion
  std::string start_folder_path =
    base::get_file_path(
      get_initial_path_to_select_filename(initialPath));
```

dacap (changes requested):
I was finally able to test this fix. Just a minor change in the code and you can change the commit message to something like this:
```
Fix initial folder for File > Save with the last one visited (fix #3996)

Regression added in 056073b3f1a7d54cdf9a113110f50947b9b4d525 in the
custom file selector. Every time you saved a new sprite the dialog
would start at the root path (e.g. Desktop) instead of the last used
directory.
```

Gasparoken:
Ready for merge

PR #4026, opened 2023-08-31
Fix respect snap to grid when we move the selection (fix #4021)

### LINKED ISSUE #4021: Respect snap to grid when we move the selection
From: https://community.aseprite.org/t/add-preference-for-dragged-selection-to-respect-snap-to-grid-setting/12884

If we move the selection and the Snap to Grid option is enabled, we should move the selection with snap to grid by default (and <kbd>Alt</kbd> modifier doing the inverse, disabling the snap to grid).

At the moment the only way to move the selection with snap to grid is pressing <kbd>Alt</kbd>.

Gasparoken:
fix #4021

### DIFF
--- a/src/app/ui/editor/moving_pixels_state.cpp
+++ b/src/app/ui/editor/moving_pixels_state.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2022  Igara Studio S.A.
+// Copyright (C) 2019-2023  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -441,7 +441,11 @@ void MovingPixelsState::onCommitMouseMove(Editor* editor,
   KeyAction action = m_editor->getCustomizationDelegate()
     ->getPressedKeyAction(keyContext);
 
-  if (int(action & KeyAction::SnapToGrid))
+  bool snapToGrid  = m_editor->docPref().grid.snap();
+  if (bool(action & KeyAction::SnapToGrid))
+    snapToGrid = !snapToGrid;
+
+  if (snapToGrid)
     moveModifier |= PixelsMovement::SnapToGridMovement;
 
   if (int(action & KeyAction::AngleSnap))


Gasparoken:
I see that the moving cel action (MovingCelState::onCommitMouseMove), does not respect the snap in any case.
I think this is not consistent. What do you think @dacap?
I can create a new issue for this.

dacap:
You can create a new issue about that.

src/app/ui/editor/moving_pixels_state.cpp
```diff
@@ -441,7 +441,11 @@ void MovingPixelsState::onCommitMouseMove(Editor* editor,
   KeyAction action = m_editor->getCustomizationDelegate()
     ->getPressedKeyAction(keyContext);
 
-  if (int(action & KeyAction::SnapToGrid))
+  bool snapKeyAction = bool(action & KeyAction::SnapToGrid);
+  if (Preferences::instance().document(editor->document()).grid.snap())
+    snapKeyAction = !snapKeyAction;
+
+  if (snapKeyAction)
```
dacap, line 448: You can use `Editor::docPref()`:
```suggestion
  bool snapKeyAction = bool(action & KeyAction::SnapToGrid);
  if (editor->docPref().grid.snap())
    snapKeyAction = !snapKeyAction;

  if (snapKeyAction)
```
As a minor detail, it looks like the logic is inverted, for example I'd have done something like:
```c++
  bool snapToGrid  = editor->docPref().grid.snap(); // Snap to grid depending on the user preference
  if (bool(action & KeyAction::SnapToGrid))
    snapToGrid = !snapToGrid; // Pressing the key inverts the action

  if (snapToGrid)
```
What do you think?

dacap: I know, what I said is that the logic is a little confusing to understand (why the Preference inverts the action, when the Alt key is the one that inverts the snap on/off). Both solutions should do the same, but what I think is that it makes more sense to use the preference "as the default", and the Alt modifier as just that, a modifier of the default (inverts the snap on/off depending on the preference).

PR #4028, opened 2023-08-31
Fix F6 shortcut to show the timeline thumbnail by default (fix #4020)

### LINKED ISSUE #4020: Fix F6 shortcut to show the timeline thumbnail by default
There is a "Toggle Timeline Thumbnails" command (<kbd>F6</kbd> key) to toggle the visibility of the timeline thumbnails, but it doesn't work until we change the timeline thumbnail size first, which makes little sense for this command. By default the command should set a size if the current thumbnail size is = 1.

image

From: https://community.aseprite.org/t/a-way-to-change-the-default-size-of-the-layer-thumbnail/19863/3?u=dacap

Gasparoken:
Before this fix, when the thumbnail size was 1, the "Toggle Timeline Thumbnails" command (F6 key) seemed to not work until we changed the timeline thumbnail size to 2 or more.

fix #4020

### DIFF
--- a/src/app/commands/cmd_toggle_timeline_thumbnails.cpp
+++ b/src/app/commands/cmd_toggle_timeline_thumbnails.cpp
@@ -1,4 +1,5 @@
 // Aseprite
+// Copyright (C) 2023  Igara Studio S.A.
 // Copyright (C) 2017  David Capello
 // Copyright (C) 2016  Carlo Caputo
 //
@@ -33,6 +34,11 @@ protected:
 
   void onExecute(Context* context) override {
     DocumentPreferences& docPref = Preferences::instance().document(context->activeDocument());
+
+    // Loading default zoom when activating thumbnail
+    if (docPref.thumbnails.zoom() <= 1 && !docPref.thumbnails.enabled())
+      docPref.thumbnails.zoom(2);
+
     docPref.thumbnails.enabled(!docPref.thumbnails.enabled());
   }
 };


Gasparoken:
I have seen that the timeline settings window does not update the zoom values of the thumbnails while the settings are open. It could be done with a 'Delegate' pass during window creation. Maybe the priority of this would be less than 'low'.
I can do some issue for this.

dacap:
Testing this fix we cannot set the thumbnail size to 1:

https://github.com/aseprite/aseprite/assets/39654/0fc9b8bd-c871-44cf-8bdc-6014323d4391

Anyway the thumbnail size = 1 doesn't make sense right now because it doesn't show the thumbnail. I'm thinking if we shouldn't fix this from the Timeline size, e.g. if the thumbnail is enabled, and the zoom=1, show the thumbnail (with a size that will be even smaller than zoom=2).

dacap:
> I have seen that the timeline settings window does not update the zoom values of the thumbnails while the settings are open. It could be done with a 'Delegate' pass during window creation. Maybe the priority of this would be less than 'low'.
I can do some issue for this.

You can create an issue about this 👍

Gasparoken:
Ready for review.
This solution works very well, but the definitive solution (most suitable code) will be the solution for: https://github.com/aseprite/aseprite/issues/4046 (via delegation)

dacap:
Sorry I was wrong here:

> Anyway the thumbnail size = 1 doesn't make sense right now because it doesn't show the thumbnail.

Because the "Overlay Size" is still useful:

Screenshot 2023-09-19 at 14 41 27

And saw that the timeline zoom = 1 is still useful to show no thumbnail in the timeline and enable the overlay anyway. I'll make some comments in this PR.

dacap:
Probably this should be fixed in `ToggleTimelineThumbnails` command only, to fix the F6 behavior, but without changing the behavior of the `ConfigureTimelinePopup` checkbox behavior.

Gasparoken:
Ready for review.
I tested it and this still has at least one problem: if we turn off Thumbnails in the popup, the Thumbnail Size and Overlay Size sliders collapse, then we assume the thumbnail size is set to 1 and no overlay, but the overlay is still activated.
IMO:
- Or, do not collapse the sliders (always leave them visible).
- Or, by disabling the Thumbnails we need also to disable the Overlay.

dacap:
I'll merge this 👍  about the Thumbnails UI, I think probably the best would be to keep the sliders always visible and remove the checkbox.

PR #4033, opened 2023-09-08
Avoid showing an alert when asking if the clipboard has a native bitmap

### LINKED ISSUE #4016: Cannot lock clipboard can generate stack overflow redrawing screen/showing alert about that we cannot lock clipboard
Sentry Issue: [ASEPRITE-1C9](https://igara-studio.sentry.io/issues/4395442794/?referrer=github_integration)

```
EXCEPTION_STACK_OVERFLOW / 0x7ff760348fb4: Fatal Error: EXCEPTION_STACK_OVERFLOW / 0x7ff760348fb4
  ?, in SkScan::FillRect
  ?, in SkDraw::drawRect
  ?, in SkBitmapDevice::drawRect
  ?, in SkCanvas::onDrawRect
  ?, in SkCanvas::drawRect
...
(252 additional frame(s) were not displayed)
```

martincapello:
Fix #4016

### DIFF
--- a/src/app/util/clipboard_native.cpp
+++ b/src/app/util/clipboard_native.cpp
@@ -84,6 +84,7 @@ void Clipboard::registerNativeFormats()
 
 bool Clipboard::hasNativeBitmap() const
 {
+  InhibitClipErrors ice;
   return clip::has(clip::image_format());
 }
 


src/app/util/clipboard_native.cpp
```diff
@@ -84,7 +84,7 @@ void Clipboard::registerNativeFormats()
 
 bool Clipboard::hasNativeBitmap() const
 {
-  return clip::has(clip::image_format());
+  return clip::has(clip::image_format(), OMIT_LOCKING_ERROR, 1, 0);
```
dacap, line 87: Cannot we do something like this? (to avoid changing clip API):
```suggestion
  auto old_handler = clip::get_error_handler();
  clip::set_error_handler(omit_locking_error_handler);
  bool result = clip::has(clip::image_format());
  clip::set_error_handler(old_handler);  
  return result;
```
And `omit_locking_error_handler` be a custom handler to ignore locking problems? Because this looks more like an issue that we could handle from the Aseprite side.

We could have an utility class in Aseprite like this too:
```c++
class IgnoreClipLockErrors {
  static clip::error_handler m_originalHandler;
public:
  static void ignoreLockErrorsHandler(clip::ErrorCode code) {
    ... call the original handler for other cases ...
  }
  IgnoreClipLockErrors()
    : m_originalHandler(clip::get_error_handle()) {
    clip::set_error_handle(ignoreLockErrorsHandler);
  }
  ~IgnoreClipLockErrors() {
    clip::set_error_handle(m_originalHandler);
  }
};
```

martincapello: Yes, we can. It was actually my first thought (I have mentioned it by mail, but maybe you missed it), but also I thought it wasn't the best approach...but if you are fine with it, either do I.
So, what do you prefer? The first option using the `omit_locking_error_handler` or the utility class?

dacap: The utility class.

martincapello: Sorry, but I don't know why or how I missed that there is an InhibitClipErrors class that I can use for this already...

dacap: That's great, I forgot about that class too 😭

PR #4034, opened 2023-09-08
Add constrain angle modifier during polygonal tools (fix #1933)

### LINKED ISSUE #1933: Constrain Angle on Shift/Control During Polygonal Lasso Selections
Normally, holding the Shift key while drawing a line with the Line Tool (or Shift+Control with the Brush Tool) constrains the drawing angle to the nearest 1:1, 1:2, or 2:1 angle or cardinal axis. Could this behaviour be integrated with the Polygonal Lasso Tool as well, that we may make selections (and deselctions!) at tidy angles more quickly and reliably?

Thanks for considering it! :D 

..using ASEprite 1.2.9-x86 on Win10.

Gasparoken:
First approach.

It may be better to add the algorithm used at each new point of a stroke, to allow mixed strokes to be drawn with or without constraint. I need an opinion on this.

The "Ctrl" key is better than "Shift", because "Shift" is already used for the Polygonal Lasso tool.

fix #1933

### DIFF
--- a/src/app/tools/controllers.h
+++ b/src/app/tools/controllers.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2022  Igara Studio S.A.
+// Copyright (C) 2019-2023  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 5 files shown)


martincapello:
I think it would be nice to let the user choose if they want to use the constrain on each pair of points. But that's just my opinion, so I'm not sure if it is actually useful.

Gasparoken:
> I think it would be nice to let the user choose if they want to use the constrain on each pair of points.

I agree.

> But that's just my opinion, so I'm not sure if it is actually useful.

Isometric drawings without "snap to grid" restriction will be a good addition 🙌
Oct-02-2023 18-33-40

I only see one problem with this, when I want to finish the polygon, the last point jumps to the cursor, so a perfect polygon is not easy to finish since I have to position the cursor very precisely.
Perhaps the cursor should jump to the last point added if shift is pressed.

PR #4038, opened 2023-09-12
[lua] Add 'tilemapMode' and 'tilemapMode' getters to Site object

Gasparoken:
If this is approved, I'll update the API documentation.

By the way, I found the lack of documentation for the following commands in the API, I will update it soon:

- app.command.ToggleTilesMode { }
- app.command.TilesetMode { }


Gasparoken:
I see these commands are available only when ENABLE_UI flag is true. I'll see how to adapt this to all cases. Perhaps will be better to manage both via Preferences.

PR #4040, opened 2023-09-13
Add "focus", "enabled", and "visible" common properties to tabs

martincapello:
This PR expands the Dialog:tab{} API call to be able to support the "focus", "enabled", and "visible" common properties.

Related #4005


dacap:
Probably we can create an auxiliary function (static / anonymous namespace) just to process these properties from `Dialog_tab` and `Dialog_add_widget`.

martincapello:
Sounds good, will do it, thanks!

PR #4043, opened 2023-09-15
Add indexed images support to GraphicsContext

### LINKED ISSUE #3960: GraphicsContext:drawImage() with blend mode doesn't work (or if we load an indexed image)
From: https://community.aseprite.org/t/unable-to-draw-to-transparent-image-to-dlg-canvas/19406/3

We cannot use the [GC blendMode](https://www.aseprite.org/api/graphicscontext#graphicscontextblendmode) correctly with the [GraphicsContext:drawImage(image, x, y)](https://www.aseprite.org/api/graphicscontext#graphicscontextdrawimage) function (the srcRect, dstRect version does work).

Also we should be able to draw indexed images with transparent pixels.

martincapello:
Fix #3960

### DIFF
--- a/src/app/script/engine.h
+++ b/src/app/script/engine.h
@@ -195,6 +195,7 @@ namespace app {
   doc::Image* may_get_image_from_arg(lua_State* L, int index);
   doc::Image* get_image_from_arg(lua_State* L, int index);
   doc::Cel* get_image_cel_from_arg(lua_State* L, int index);
+  doc::Tileset* get_image_tileset_from_arg(lua_State* L, int index);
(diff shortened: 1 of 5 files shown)


dacap:
Just merged 👍 

Tried to find a bug with https://github.com/aseprite/aseprite/pull/4043/commits/caef58d6b7639282e81b88c43487ca12d74391c8#diff-a53432dd005b328ee289e7cfff1492336fba218acc36abb3ee630ec945328abbR58  but I wasn't able. It's the only change that I'm a little worried about.

PR #4044, opened 2023-09-15
Fix pattern Fill does not "Align to Destination" (fix #2528)

### LINKED ISSUE #2528: Pattern Fill does not "Align to Destination"
When you have the Pattern Brush set to "Align to Destination" with the Pencil Tool, it aligns the pattern stamp to the first pixel clicked and iterates from there.

With the Fill Tool, this does not happen: **the pattern aligns to the workspace instead of the space to be filled.** If the space is moved, the fill retains its alignment. Take a look: 🔽
pattern align fill

The Fill tool ought to behave more like the Brush tool in regards to alignment when "Align to Destination" is active, yes?

..using ASEprite 1.2.25-x64 on Win10

Gasparoken:
fix #2528

### DIFF
--- a/src/app/tools/ink_processing.h
+++ b/src/app/tools/ink_processing.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2022  Igara Studio S.A.
+// Copyright (C) 2019-2023  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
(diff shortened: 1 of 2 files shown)


Warrioroflove671:
Thanks again for all your help and support. I really appreciate you all. One love and Jesus Lives 
[LICENSE.txt](https://github.com/aseprite/aseprite/files/14127345/LICENSE.txt)

PR #4054, opened 2023-09-20
Add snap to grid preferences (fix #723)

### LINKED ISSUE #723: Brush center for snap to grid
We should have an option to specify what is the center of the brush so we can align it with the grid corner.

Gasparoken:
This commit adds fields on Preferences > Grid about:
 - Snap To Grid (on/off)
 - Snap To <geometric location in a box grid> Also added PreferSnapTo::BoxCenter calculation in snap_to_grid.cpp.

fix #723


Gasparoken:
This is the first commit on fix #4054 that implements only UI elements. I'll soon push for the actual implementation.

Gasparoken:
This implementation align the brush's center with the grid box's center, instead to align the brush's center with the grid box's origin. Anyway, I should mention that I think this alignment can be useful in some cases, give it a try.

Gasparoken:
What do you think of this approach @dacap?

Issue #1949, opened 2018-12-11
Script in extensions


dacap:
There are plans to include scripts in extensions (#1403, #1928) and here I'll take note about some ideas that scripts **might** provide when are installed as extensions (i.e. scripts that run when Aseprite starts):

- [x] Add new commands
- [x] Save/load preferences
- [x] #4950
- [ ] Add new inks, e.g. `app.registerInk{ id=..., text=...  ondraw=function... }`  // Recent ideas about the design of this feature here: https://community.aseprite.org/t/registering-new-inks-in-extensions/14494
- [ ] Add new tools
- [ ] Add new special editor behaviors (like #1792)

All of this must be really considered because the complexity of scripts and APIs is huge and sometimes it doesn't worth to add such complexity when C++ is preferred.

dacap:
Requested here too: https://github.com/aseprite/api/issues/119

PR #4055, opened 2023-09-21
Added support for .chr format

ghost:
**Good Morning all**

I thought it might be fun to add .CHR format support, commonly used in retro consoles like the NES/Famicom and GameBoy, and more recently by the [uxn](https://100r.co/site/uxn.html) system.

The format is indexed, with a maximum of 4 colours (index 0 being invisible). The actual colour of the palette does not matter, as it is controlled by whatever hardware it is being used on.

The image is made of 16 x 16 tiles that are 8 x 8 pixels in size, giving a total image size 128 x 128. Really, the format can support any number of tiles, however as the file itself does not encode any information about this I decided to just use the standard for the NES.

I reckon this feature is probably outside of scope, but it's here in case anyone wants it!

Thanks,
Pat

**Links:**
- [NESDev Wiki](https://www.nesdev.org/wiki/PPU_pattern_tables)
- [XXIIVV Wiki](https://wiki.xxiivv.com/site/chr_format.html)

---

Example of the .CHR file taken from Super Mario Bros. on the NES
Example of the .CHR file taken from Super Mario Bros. on the NES

### DIFF
--- a/src/app/file/file.cpp
+++ b/src/app/file/file.cpp
@@ -686,18 +686,18 @@ FileOp* FileOp::createSaveDocumentOperation(const Context* context,
   // CHR allows no more than 4 colors, and the sprite dimensions
   // should be 128 x 128 to avoid rearanging.
   if(format->support(FILE_CHR_LIMITATIONS)) {
-	if(sprite->width() != 128 || sprite->height() != 128) {
-	  warnings += "<<- Dimensions other than 128 x 128";
-	}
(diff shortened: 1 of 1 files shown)


dacap:
Hi @hatcatpat, thanks for the PR, anyway I think this is too specific to NES/NES emulators and probably would be better as an extension.

There are plans to add support for [new file types in plugins](https://github.com/aseprite/aseprite/issues/1949)

PR #4057, opened 2023-09-21
Improve script engine's exception handling

martincapello:
This PR contains some changes on how we were treating unhandled exceptions:
1) Removes the try..catch clauses from the signal handler to let the engine's try..catch block to take care of any unhandled exception.
2) Outputs stack trace information for all unhandled exceptions. For instance, let's say some Lua script called an API function that for any reason threw an std::exception from its C++ implementation with the message "Some error occurred", then on the console the output would read:
   Before this change:
   ```
   Some error occurred
   ```
   After this change:
   ```
   /path/to/script.lua:99: Some error occurred
   stack traceback:
    /path/to/script.lua:99: in main chunk
   ```

3) Fixes how luaL_traceback and luaL_error were receiving their message parameters to show the error message at the top of the stack trace when trying to access a deleted object. 
   Before this change:
   ```
   /path/to/script.lua:99: stack traceback:
    /path/to/script.lua:99: in main chunk: Tried to access a deleted 'doc::Sprite'
   ```
   After this change:
   ```
   /path/to/script.lua:99: Tried to access a deleted 'doc::Sprite'
   stack traceback:
    /path/to/script.lua:99: in main chunk
   ```

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -287,20 +287,20 @@ void Dialog_connect_signal(lua_State* L,
       if (dlg->showRef == LUA_REFNIL)
         return;
 
-      try {
-        // Get the function "n" from the uservalue table of the dialog
-        lua_rawgeti(L, LUA_REGISTRYINDEX, dlg->showRef);
-        lua_getuservalue(L, -1);
-        lua_rawgeti(L, -1, n);
+      // Get the function "n" from the uservalue table of the dialog
+      lua_rawgeti(L, LUA_REGISTRYINDEX, dlg->showRef);
+      lua_getuservalue(L, -1);
+      lua_rawgeti(L, -1, n);
 
-        // Use the callback with a special table in the Lua stack to
-        // send it as parameter to the Lua function in the
-        // lua_pcall() (that table is like an "event data" parameter
-        // for the function).
-        lua_newtable(L);
-        callback(L, std::forward<Args>(args)...);
+      // Use the callback with a special table in the Lua stack to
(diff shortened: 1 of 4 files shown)


dacap:
It would be nice to include some practical example about this improvement in this PR (and probably some information in the commit summary/description about it / reference the same PR in the commit?)

dacap:
The review of this PR will take a while because I'm writing some Lua tests to reproduce each case.

dacap:
1.

```lua
dlg = Dialog("Dlg")
dlg:button{ text="Click", onclick=function()
  local spr = Sprite(32, 32)
  spr:close()
  print(spr.filename)
end }
dlg:show{ wait=false }
```

I need to test other cases, I'm not sure if it's 100% safe to throw an exception when a signal is called, but probably it is for scripts.

2. Still working on this.

3.

```lua
local spr = Sprite(32, 32)
print(spr.filename)
spr:close()
print(spr.filename)
```

dacap:
@martincapello As I saw that this reverted 7029e05dc4646df0eb599eccbfad2e9e6e2ce572, I wanted to give a try to the following snippet:

```lua
-- Close all sprites
while app.sprite do
  app.sprite:close()
end

dlg = Dialog("Dlg")
dlg:button{ text="Click",
  onclick=function()
   app.transaction(function()
       -- Do nothing
   end)
  end }
dlg:show{ wait=false }
```

And there is a regression, now this crashes.

martincapello:
I have added the try...catch block again, but now I've introduced a new function in the script engine that handles the std::exceptions, which is called from the engine and from the dialog class so both are handled the same.

Issue #3684, opened 2023-01-26
Drawing using a keyboard key


vivi3107:
1. assign "trigger left mouse button" to any free keboard key
2. try to draw a rectangle or an ellipse using the assigned key
3. this happens
![a](https://user-images.githubusercontent.com/44928480/214942071-f18daa5e-5d50-4f2f-91c5-9267937e4b52.png)

### Aseprite and System version

* Aseprite version: Aseprite 1.3 beta21
* System: Windows 10 22H2

Gasparoken:
Thanks for this report. I was able to reproduce it on 1.2.40 and 1.3beta21. It happens on Windows (MacOS is ok, pending linux test).

BananaChad:
on windows i think that this happens because windows thinks a key is being held down when typing and thus keeps re-clicking like a character does when holding down a key, if you change the repeat rate (as shown in image below) you can see that it changes the rate that the rectangles or ellipses appear.
![image](https://user-images.githubusercontent.com/105674111/215431780-4746ec49-97e4-496a-abf9-99a2c48bbe60.png)

### **Aseprite and System version**

- Aseprite version: Aseprite 1.3 beta21
- System: Windows 10 21H2

PR #4063, opened 2023-09-23
Fix drawing using a keyboard key

tetektoza:
This bug fixes  https://github.com/aseprite/aseprite/issues/3684. It was also reproducible on linux systems, and this small patch should fix it.

### DIFF
--- a/src/app/ui/editor/drawing_state.cpp
+++ b/src/app/ui/editor/drawing_state.cpp
@@ -154,6 +154,8 @@ bool DrawingState::onMouseDown(Editor* editor, MouseMessage* msg)
 
   if (!editor->hasCapture())
     editor->captureMouse();
+  else
+    return true;
 
   tools::Pointer pointer = pointer_from_msg(editor, msg,
                                             m_velocity.velocity());


dacap:
Thanks @tetektoza, I've just tested this and it breaks the following Aseprite functionality:

* Select the Line tool
* Start drawing with the left mouse button (press left mouse button)
* Then press the right mouse button (without releasing the left mouse button)
* It should cancel the line

With this PR the line is not canceled and it's immediately painted in the canvas.

tetektoza:
> Thanks @tetektoza, I've just tested this and it breaks the following Aseprite functionality:
> 
> * Select the Line tool
> * Start drawing with the left mouse button (press left mouse button)
> * Then press the right mouse button (without releasing the left mouse button)
> * It should cancel the line
> 
> With this PR the line is not canceled and it's immediately painted in the canvas.

I see, will work further on this in the couple of days. I've thought this might be a problem or not a good approach, but now I know the source of the problem so it might be easier. Should I reassign you to this PR if I'll get a better fix?

dacap:
> Should I reassign you to this PR when I'll get a better fix, or should I open a new PR?

You can update this PR (with a force-push updating your branch) and I'll take it (if you can reassign the PR, reassign it to me, but I'm not sure if that is possible for non-members of aseprite org).

tetektoza:
@dacap looking a bit further - it seems the problem, or maybe not a problem but the way that laf library implements keyboard events - it seems like if we press down a key - we can hold it and the event will come all the time. Mouse works the different way - it only processes one event at a time, so if we push left click - aseprite's code won't get an event spam as long as the button is being held. On the other hand if we hold keyboard button - an event will come all the time and we will essentially cast it down to LMB, so it will treat it as we're continously left clicking instead of treating it as the key being held down, that's why this bug is here.

My questions:
I could try looking into laf library and try to implement recognition between the key being held down and the key being pushed once (right now it's treating it as the same key event),

OR I could make a flag that recognizes if a key event was casted to mouse event, and then check if the last key event that was pressed is the same one that's being processed right now, if yes - we will not do the left click. Pretty simple fix I guess.

What do you think @dacap?

dacap:
Probably the information is already available, the `ui::KeyMessage` has a `repeat` field (that come from `os::Event::repeat`), which might be 0 (or 1, I don't remember right now) for the first "key down" event.

Maybe `AppEditor::onProcessMessage` can filter out all `kKeyDownMessage` messages with a `repeat` count > 0, but I'm not sure if that will do the job.

tetektoza:
@dacap - oh yeah, I've seen this variable inside the class although when I was debugging it was coming with 0 all the time. Might get a closer look on it later on, I will post more updates. Thanks.

By the way, do you guys have some Discord or some sort of that communicator? I'm planning to contribute more, and it's kinda more comfortable this way.

dacap:
> By the way, do you guys have some Discord or some sort of that communicator? I'm planning to contribute more, and it's kinda more comfortable this way.

There is a Discord but I'm not used to enter daily, but from time to time I enter to the server: https://discord.gg/Yb2CeX8
There is a #aseprite-dev channel.

tetektoza:
> Maybe `AppEditor::onProcessMessage` can filter out all `kKeyDownMessage` messages with a `repeat` count > 0, but I'm not sure if that will do the job.

@dacap, Just came back and checked - it is already done by the way, please check src/app/ui/doc_view.cpp:139, starting from this, method onProcessMessage - it is already being checked here. See? If repeat is equal to 0, it allows for the key event to be transformed to mouse event and to be queued once again, but if not - it just goes forward.

I am using Manjaro with GNOME, which is using X11 - and I've checked in laf/os/x11/window.cpp - procesX11Event method, and repeat is never set here, so this is *why I am receiving repeat count with the value of 0 all the time*.

I will probably prepare fix for that on the next days, the fix will land in laf library I suppose, you can correct me if I'm thinking wrong.

I wonder what's happening on Windows though, because on Windows repeat is being set on the event, so it shouldn't transform Key Event to Mouse Event based on what I've found in src/app/ui/doc_view.cpp:139.

Let me know what do you think and do you know if this issue is still reproducible on Windows?

tetektoza:
@dacap please check out https://github.com/aseprite/laf/pull/69, you can give your thoughts here.

I will probably need to implement something similar on Windows, but that's once I will come back home and will change systems.

dacap:
I think this PR is outdated and we can close it.

Issue #3551, opened 2022-10-10
Removing all frames results in a transparent background layer


KasumiArai:
Have more than one frame.
Select all frames with the timeline.
Delete the selected frames.

A transparent background layer is left.
![Aseprite transparent background layer](https://user-images.githubusercontent.com/8582123/194848274-66076b10-4c18-410e-84ea-69be5b1bd110.gif)

Aseprite v1.3-beta21-x64 on Steam
Windows 10

PR #4065, opened 2023-09-23
Fix removing all frames results in a transparent background layer

tetektoza:
This patch fixes https://github.com/aseprite/aseprite/issues/3551. Currently if we select all frames (as shown in
https://github.com/aseprite/aseprite/issues/3551), and we will remove all of them - it also removes cel inside the primary frame, resulting in a transparent background (again, as shown in issue#3551). This patch adds a skip for the primary (0th) frame inside the loop that loops over frames that will be deleted, so we will not take it under consideration during deletion of selected frames.


tetektoza:
Hey @dacap, have you seen this PR? Will we be able to merge it at some point?

dacap:
@tetektoza I'll take a look to this right now 👍

src/app/commands/cmd_remove_frame.cpp
```diff
@@ -59,6 +59,11 @@ void RemoveFrameCommand::onExecute(Context* context)
     if (site->inTimeline() &&
         !site->selectedFrames().empty()) {
       for (frame_t frame : site->selectedFrames().reversed()) {
+
+        // if we are on the first frame - don't do anything with it
+        if (frame == 0)
+          continue;
```
dacap, line 65: If we select all frames and delete them, they should be deleted, i.e. a frame without cel or a background layer cleared with the background color is just left (it's just like creating a new sprite with the given layers empty and just one frame).

With this change the first frame is kept (which is not expected).

From #3551 the bug is that the background layer is left without a cel (which is inconsistent with the background layer which must have one cel on each frame).

tetektoza: Hm, so if I understand correctly - the issue is actually not the cel being removed from the frame, but ONLY the cel being removed from the frame, not the frame itself?

So if I understand correctly - we should remove frame + cel inside of it, right?

dacap: > So if I understand correctly - we should remove frame + cel inside of it, right? So if we remove all frames, there should be no frame at all, right?

No. The issue is that for the background layer (and only for the background layer) the cel is being deleted instead of just cleared with the background color. Looking at a glance, probably the issue is that cmd::RemoveFrame() ctor is using cmd:RemoveCel instead of cmd::ClearCel.

tetektoza: Oh, yeah. I didn't see the person in this issue has background layer set up, my bad. Now I get it - I will hack a little bit and try to prepare a new solution for that.

tetektoza:
@dacap I've edited this patch - this should work now, hope I didn't miss something out. Now for the 0th frame - the background stays as it was, but anything on it gets cleared.

src/app/commands/cmd_remove_frame.cpp
tetektoza: @dacap I came back to this yesterday, I've pushed some changes but this is still not a full solution - after I've pushed it I finally understood why there's possibility to delete frame 0 - you can delete it and basically all frames will be moved to the left. So right now my patch disabled that, so I'm working on it further.

Clearing cel on frame 0 is not a solution either. At least it's not a full solution anyways. If we clear cel on frame 0, the only thing that will change is that if we will undo action - background is not returning. The difference between ClearCel and RemoveCel is that, that RemoveCel essentially copies all insides from the cel that was before and adds it once again upon "undo" operation. It is the same cel but it is readded from internal buffer instead of reusing it from the pointer. ClearCel clears the image only, it doesn't readd the cel, it uses the same image from the previous cel. It looks like something is removing the cel.

I will let you know on days how progress with this stays, I also found some places where there are redundant if checks for example, will make them as a separate commits.

tetektoza:
@dacap I dug a little deeper today, and you can check out my commit patch. It turns out that if we just clear the cel on 0th frame, it will get a new frame number which is -1 (this is done in LayerImage::moveCel), this is because each frame comes with a delta and it basically moves all of the frames one position to the left.

So if we have 0, 1, 2, 3, 4 frame, and we remove 0th frame, all frames will be moved to the left and it will look like this:
-1, 0, 1, 2. Cells are mapped between frames, so cel which was on frame 0 will be now on "frame -1".

Then, if we undo removeFrame on 0th frame, and we used ClearCel, then the frame number will be preserved. This causes the cel not being redrawn upon undo, because the cel has frame number -1 in it. 

My solution to that is to simply renew frame number in ClearImage "undo" operation if it has been changed, so it will give the same frame number as it had before. I don't think this has caused any regression, but there's one drawback - if someone removes a frame, and we use RemoveCel instead of ClearCel, what it does is that it simply copies the cel inside class member, and then we delete the original memory location of the cell. If user wants to undo that operation we simply allocate new Cel, we copy the buffer and add the cel to the list. ClearCel doesn't have that, so when we remove 0th frame now, it stays in the list of cels, it's just not being drawn.

So my question is: maybe we should differentiate somehow if we remove the frame in ClearCel, and if yes, maybe the memory cel of the buffer should be copied the same way as it is done in RemoveCel? This way we won't clog up m_cels list in LayerImage class?

src/app/cmd/clear_image.cpp
```diff
@@ -19,15 +19,16 @@ namespace cmd {
 
 using namespace doc;
 
-ClearImage::ClearImage(Image* image, color_t color)
-  : WithImage(image)
-  , m_color(color)
+ClearImage::ClearImage(Cel* cel, color_t color)
```
martincapello, line 22: I'm not sure if there is another solution (I will try to suggest something) but I don't think we should change the image parameter by a cel...it just doesn't feel right...ClearImage should clear an Image IMO.

dacap:
I didn't have time to check this PR yet, but please check [my comment](https://github.com/aseprite/aseprite/pull/4065#discussion_r1346535369), from what I see, probably changing this line:

https://github.com/aseprite/aseprite/blob/f6db44b81d36b03013ca8e1e262e3cce381f31f8/src/app/cmd/remove_frame.cpp#L32

with:

```c++
 m_seq.add(new cmd::ClearCel(cel)); 
```

Might fix the issue.

src/app/cmd/clear_image.cpp
tetektoza: Now thinking about it - maybe I could just save the frame and restore it in ClearCel since ClearCel calls ClearImage anyways?

martincapello: That sounds better, but I'm still wondering if there is an easier fix. Anyway David said that he will review this.

PR #2699, opened 2021-04-15
Use ccache on Linux in CI

janisozaur:
While ccache is also available on osx, there seems to be some problem when using it that I haven't root-caused yet

PR #4067, opened 2023-09-25
Use ccache on macOS

janisozaur:
Originally (#2699) ccache was only enable for Linux due to suspiciously low macOS yields.

I have addressed the likely issue upstream
(https://github.com/hendrikmuhs/ccache-action/pull/147) and ccache can be used on macOS now as well.


dacap:
Thanks again @janisozaur, great to hear that you've fixed this on ccache-action 👍  just merged this PR.

Issue #3219, opened 2022-03-21
New brush manager


dacap:
Issue created from #1387 and other comments:

- [ ] New brushes manager (https://github.com/aseprite/aseprite/issues/175) with [preset brushes and patterns](https://twitter.com/cyangmou/status/818191210732994566)
- [ ] Possibility to Import/Export brushes
- [ ] Named brushes
- [ ] Import GG brushes ([1bpp bmps](https://twitter.com/not_surt/status/818200706255437824))
- [ ] Brushes should be in extensions (https://github.com/aseprite/aseprite/issues/1403)

Basically the brush popup must be re-designed to simplify the management of brushes (better organization, simplify saving/loading brushes, naming brushes, switching between brushes and assigning keyboard shortcuts to brushes).

PR #4068, opened 2023-09-25
Show custom pattern brushes for Paint Bucket 

### LINKED ISSUE #4053: Show custom brushes/patterns for Paint Bucket
From #2528, it would make sense to show the button to show custom brushes to fill with patterns with the Paint Bucket tool.

martincapello:
This PR makes the brushes selector available when the user selects the paint bucket tool. This way the user can fill areas with patterns without first changing to another tool to select such brush.

Fix #4053

### DIFF
--- a/src/app/ui/context_bar.cpp
+++ b/src/app/ui/context_bar.cpp
@@ -2167,7 +2167,7 @@ void ContextBar::updateForTool(tools::Tool* tool)
   // Show/Hide fields
   m_zoomButtons->setVisible(needZoomButtons(tool));
   m_brushBack->setVisible(supportOpacity && hasImageBrush && !withDithering);
-  m_brushType->setVisible(supportOpacity && (!isFloodfill || (isFloodfill && hasImageBrush && !withDithering)));
+  m_brushType->setVisible(supportOpacity && (!isFloodfill || (isFloodfill && !withDithering)));
   m_brushSize->setVisible(supportOpacity && !isFloodfill && !hasImageBrush);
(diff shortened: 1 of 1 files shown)


Gasparoken (approved):
Reviewed and tested. It looks good!

src/app/ui/context_bar.cpp
```diff
@@ -2167,7 +2167,7 @@ void ContextBar::updateForTool(tools::Tool* tool)
   // Show/Hide fields
   m_zoomButtons->setVisible(needZoomButtons(tool));
   m_brushBack->setVisible(supportOpacity && hasImageBrush && !withDithering);
-  m_brushType->setVisible(supportOpacity && (!isFloodfill || (isFloodfill && hasImageBrush && !withDithering)));
+  m_brushType->setVisible(supportOpacity && (!isFloodfill || (isFloodfill && !withDithering)));
```
dacap, line 2170: To avoid confusion for the user, I'd leave only this change and keep the brush popup just as it looks when other tools are selected (and avoid hiding elements).

martincapello: Despite being a much easier fix I have the need of asking: What about disabling the brushes instead of hiding them?

dacap: I thought about the same but I don't know if it worth it. Just in case we need a way to go back to a plain brush to fill with solid colors (not sure if that was the case with this patch, the regular/solid brush types must be visible).

There are plans to redesign this popup in the future anyway (#3219) so probably doesn't worth the effort.

Issue #3649, opened 2023-01-02
Disable "Delete Layer" option when the sprite has only one layer


martincapello:
>Would it be possible if when there's just one layer, the "Delete" option is grayed out so the program doesn't scold you over it? A lot of other programs tend to have the Delete option unavailable if there's only one layer. I know it's a minor suggestion, but something to consider.

NOTE: Coming from https://igarastudio.zendesk.com/agent/tickets/4313

The-Vector:
@martincapello Hi Martin, 

I'm a university student and one of my assignments is to work on an open source project.
I found this issue and I think that I can fix it. 
Would it be possible to assign it to me?

Thanks in advance

martincapello:
Hey there, sure I'm assigning it to you. But first of all, please read the CONTRIBUTING.md document, and pay attention to this part: https://github.com/aseprite/aseprite/blob/main/CONTRIBUTING.md#forking--pull-requests
Basically it says that after you made your PR you will have to sign our CLA before we can merge it.

@dacap FYI

dacap:
Just in case @The-Vector we are not an "open source" project in the [OSI definition of "open source"](https://opensource.org/osd/), people prefer to say that we are "[source-available software](https://en.wikipedia.org/wiki/Source-available_software)", and we prefer that way to avoid any controversy.

Please check with your teacher of the assignment will be fulfilled with this kind of project.

PR #4070, opened 2023-09-26
Fix disable delete layer button when only one layer exists

The-Vector:
Fixes [#3649](https://github.com/aseprite/aseprite/issues/3649)

### DIFF
--- a/src/app/commands/cmd_remove_layer.cpp
+++ b/src/app/commands/cmd_remove_layer.cpp
@@ -116,9 +116,14 @@ RemoveLayerCommand::RemoveLayerCommand()
 
 bool RemoveLayerCommand::onEnabled(Context* context)
 {
-  return context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
-                             ContextFlags::HasActiveSprite |
-                             ContextFlags::HasActiveLayer);
+  if (!context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
+                 ContextFlags::HasActiveSprite |
+                 ContextFlags::HasActiveLayer))
+    return false;
+
+  const ContextReader reader(context);
+  const Sprite* sprite(reader.sprite());
+  return sprite && sprite->totalFrames() > 1;
 }
 
 void RemoveLayerCommand::onExecute(Context* context)
--- a/src/app/commands/cmd_remove_layer.cpp
+++ b/src/app/commands/cmd_remove_layer.cpp
@@ -123,7 +123,7 @@ bool RemoveLayerCommand::onEnabled(Context* context)
 
   const ContextReader reader(context);
   const Sprite* sprite(reader.sprite());
-  return sprite && sprite->totalFrames() > 1;
+  return sprite && sprite->allLayersCount() > 1;
 }
 
 void RemoveLayerCommand::onExecute(Context* context)
--- a/src/app/commands/cmd_remove_layer.cpp
+++ b/src/app/commands/cmd_remove_layer.cpp
@@ -117,13 +117,15 @@ RemoveLayerCommand::RemoveLayerCommand()
 bool RemoveLayerCommand::onEnabled(Context* context)
 {
   if (!context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
-                 ContextFlags::HasActiveSprite |
                  ContextFlags::HasActiveLayer))
     return false;
 
   const ContextReader reader(context);
   const Sprite* sprite(reader.sprite());
-  return sprite && sprite->allLayersCount() > 1;
+  const Layer* layer = reader.layer();
+
+  return sprite && (sprite->root()->layersCount() > 1 || layer->parent() != sprite->root());
+  
 }
 
 void RemoveLayerCommand::onExecute(Context* context)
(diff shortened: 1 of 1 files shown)


src/app/commands/cmd_remove_layer.cpp
```diff
@@ -116,9 +116,14 @@ RemoveLayerCommand::RemoveLayerCommand()
...
-                             ContextFlags::HasActiveSprite |
-                             ContextFlags::HasActiveLayer);
+  if (!context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
+                 ContextFlags::HasActiveSprite |
+                 ContextFlags::HasActiveLayer))
+    return false;
+
+  const ContextReader reader(context);
+  const Sprite* sprite(reader.sprite());
+  return sprite && sprite->totalFrames() > 1;
```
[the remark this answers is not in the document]
The-Vector: Ah, I don't know how I missed that. 
It should be fixed now.

src/app/commands/cmd_remove_layer.cpp
```diff
@@ -116,9 +116,14 @@ RemoveLayerCommand::RemoveLayerCommand()
...
-                             ContextFlags::HasActiveSprite |
-                             ContextFlags::HasActiveLayer);
+  if (!context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
+                 ContextFlags::HasActiveSprite |
+                 ContextFlags::HasActiveLayer))
+    return false;
+
+  const ContextReader reader(context);
+  const Sprite* sprite(reader.sprite());
+  return sprite && sprite->allLayersCount() > 1;
```
dacap, line 126: Hi @The-Vector, actually here you are counting the total number of layers (including layers in subgroups) which I think it's not the expected behavior, e.g. if you have a group with one child, that is allLayersCount=2 and enables the delete command for the group (which is the only layer at the root level).

You have to count the number of layers in the root (sprite->root()).

Also I'm seeing that the commit has 0quokka0@gmail.com as email but the CLA was signed with other email address, you should sign the CLA with the same name and email that you will use in the commit message.

dacap: @The-Vector was you able to see the above comments?

The-Vector: Yep, I filled out the CLA with the correct email, and I'm just testing my code so there's no other edge cases that I missed.

The-Vector:
I believe this should work now. 
The button is disabled when there is only one layer, and if you make a group with layers within it, you can delete the layers but not the group. 
Unless there are other edge cases which I missed?

src/app/commands/cmd_remove_layer.cpp
```diff
@@ -116,9 +116,16 @@ RemoveLayerCommand::RemoveLayerCommand()
 
 bool RemoveLayerCommand::onEnabled(Context* context)
 {
-  return context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
-                             ContextFlags::HasActiveSprite |
-                             ContextFlags::HasActiveLayer);
+  if (!context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
+                 ContextFlags::HasActiveLayer))
+    return false;
```
dacap, line 121: You can keep the same flags here:

```suggestion
 if (!context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
                          ContextFlags::HasActiveSprite |
                          ContextFlags::HasActiveLayer))
  return false;
```

src/app/commands/cmd_remove_layer.cpp
```diff
@@ -116,9 +116,16 @@ RemoveLayerCommand::RemoveLayerCommand()
...
+
+  const ContextReader reader(context);
+  const Sprite* sprite(reader.sprite());
+  const Layer* layer = reader.layer();
+
+  return sprite && !((sprite->root()->layersCount() == 1) && (layer->parent() == sprite->root()));
+
```
dacap, line 128: This expression is confusing and I think we can simplify it checking for the true cases (intead of the inverse):
```suggestion
  return sprite && layer &&
    // We can remove all layers from non-root groups
    ((layer->parent() != sprite->root()) ||
     // Check that we are not removing the last layer in the sprite
     (sprite->root()->layersCount() > 1));  
```
But I saw that you wrote it originally in this way in 56726d2e8b0806fab54812001903a561ee8925dc and then reverted with 5b02460299129c6c6101c4d8bca11fabe9b923c8. Logically they do the same (by [De Morgan's law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)). Can you explain why did you change this? Did this last commit really fix what it says it fix? (because it should work exactly the same as the previous one).

dacap (changes requested):
Given some comments with suggestions. I'm a little confused by the commits (not sure if the last commit does something in particular).

In any case I'm adding some suggestion to keep the code a little more readable + with some comments.

src/app/commands/cmd_remove_layer.cpp
The-Vector: I believe I changed it when testing my code and I saw that checking the true cases was still allowing you to delete a group if it was the root, but I checked just now and the code is identical and works the same, so I must have done something wrong in testing, sorry.
I'll add your suggestions and push the code with comments.

dacap:
Merged squashed in ea35725c85d97ceda9e9945e0d11960522bf7d02, thanks @The-Vector :+1:

PR #4073, opened 2023-09-27
Fix HSV/HSL values on ColorSliders

Gasparoken:
Before this fix, the HSV and HSV values displayed in the ColorSliders were truncated rather than rounded to the nearest integer.


dacap:
I remember that this PR was related to some issue or community post (or ticket?) but it wasn't referenced, could you please edit the PR description to include a link to it?

PR #4082, opened 2023-10-03
Rename all instances of Accelerator in classes to KeyShortcut

tetektoza:
Currently, Accelerator is a kind of misleading name since it holds shortcuts for all user defined keys. This patch introduces a name change for all instances of accelerator name in files.

STILL WORK IN PROGRESS -> need to consult maintainers about this change


tetektoza:
@dacap this change is inspired by my navigation through the code - I found that current name is kinda misleading, and also there was a "TODO" mark above Accelerator class, suggesting to rename this class to Shortcut, although I think KeyShortcut is more fitting.

What do you think about this? Mostly a cosmetic change.

dacap:
Thanks @tetektoza, at the moment I'd prefer to avoid making this huge refactor as it's a big change and we are trying to release a new version ASAP (and adding this might break things). Not sure about the "KeyShortcut(s)" name (or even "Shortcut(s)"). It's something I have to think a little more about it.

Issue #4084, opened 2023-10-04
Inconsistent layer number if we change from background layer to layer


tetektoza:
If we start Aseprite from ground up and open timeline - it says that the current layer is layer 1, but if we convert the layer to background and then convert it back to layer - it is no longer layer 1, but the enumeration starts from 0. Maybe it would be convenient to use the same layer name even if layer will get changed to background type?

* Aseprite version: latest compiled from main branch
* System: Linux, Manjaro

PR #4090, opened 2023-10-06
Fix: Inconsistent layer name if there's conversion from background to layer

tetektoza:
This patch ifxes https://github.com/aseprite/aseprite/issues/4084.

Currently if we set layer to background and then back to layer, last used name isn't preserved. This patch makes layer save last used layer name before it gets converted to background so it won't get lost.


tetektoza:
This works one way only right now, so if we convert from Layer -> Background, but let me know what you think about this solution, I can implement it in a similar way so it will save layer name for background as well.

dacap:
Thanks @tetektoza, but I think it's an overkill to add two new string fields to all `doc::Layer` objects to fix this UI problem.

Probably we should fix only the "Layer 0" name problem, and the solution would be to use some code similar to `NewLayerCommand::getUniqueLayerName()` (merging common code in some util functions).

tetektoza:
> Thanks @tetektoza, but I think it's an overkill to add two new string fields to all `doc::Layer` objects to fix this UI problem.
> 
> Probably we should fix only the "Layer 0" name problem, and the solution would be to use some code similar to `NewLayerCommand::getUniqueLayerName()` (merging common code in some util functions).

Hmm, and how about if we tried to do it with string_view somehow?

dacap:
> how about if we tried to do it with string_view somehow?

A `string_view` still needs a memory buffer to store the string (another issue is that we are not yet using string views, but it's unrelated to this). Not sure how a `string_view` would solve the issue here.

The idea would be to avoid storing extra strings in the `doc` library just to solve one problem that is generated in the UI level (`app` module).

dacap:
At the moment I think this bug is quite low priority, if a simple solution (without modifying the `doc` library) can be found we can think in merging this. Right now, as it is working, it's not quite bad and not a critical bug.

PR #4100, opened 2023-10-20
Add diagonal symmetry (fix #1171)

### LINKED ISSUE #1171: Diagonal symmetry
Just what the tittle says, it's pretty easy to make a matrix symmetrical (both 90° and 270°), and when the line is moved from the middle, simply reduce the working part of the matrix (without rotating the diagonal). Should be only aviable for NxN images(?)

Sorry if i said something stupid. Thanks for your attention.

Gasparoken:
Added symmetry UI buttons, 45 and -45 degrees axis. Now are possible activate all four symmetries, both diagonal symmetries, individual diagonal symmetry, and the old horizontal/vertical combinations.

fix #1171


Gasparoken:
Oct-20-2023 09-22-53

Gasparoken:
Pending:
- [ ] Image brush bug on 45 degrees reflections.
- [ ] Add constraints on Symmetry buttons (some configurations do not have sense, for example: Horizontal + 45 degree symmetry).
- [ ] Constraint on symmetry axis position origin when 'All' or dual diagonal symmetry is active (x, y have to be the same fractional part).

Gasparoken:
Pending:
- [ ] Need more testing when TileMode is ON.
- [ ] Is Compress Image still needed?
- [ ] Add constraint to half pixel steps when diagonal are ON (symmetry is not possible in some cases)
- [ ] Define what is done when an image brush is intersected by a symmetry axis.

Gasparoken:
When you can without rush, could you give me some feedback on this PR @martincapello?

martincapello (commented):
I have been testing this for a while and seems that it works great! About the code, the only thing that looked a bit odd to me is having RIGHT_DIAG_REFLEX and LEFT_DIAG_REFLEX as symmetry modes, because the user can only select 4 modes. So I reviewed the code in more detail and saw that this "modes" are actually used to flag the stroke points so they can be properly rendered when the different symmetry combinations are set. Then, might be nice to separate the modes that the user can select from the modes or flags that the stroke points can have, maybe using different enums...just a thought.

martincapello (approved):

PR #4108, opened 2023-10-26
Improve initialization logging

### LINKED ISSUE #3956: Add more/better logging in Aseprite initialization
There are some situations where Aseprite doesn't start (e.g. buggy custom theme, preferences, unknown problems, cannot initialize window/main surface, etc.) where it would be nice to report in some kind of log and add the attachment to a crash report ( https://docs.sentry.io/platforms/native/enriching-events/attachments/ ?)

Anyway we should improve the initial logging with -debug flag (it doesn't give enough information to fix some issues).

Related to #3952 and #3729

martincapello:
Fix #3956

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -457,6 +457,7 @@ endif()
 add_library(app-lib
   active_site_handler.cpp
   app.cpp
+  app_init.cpp
   check_update.cpp
   cli/app_options.cpp
   cli/cli_open_file.cpp
--- a/src/app/app.cpp
+++ b/src/app/app.cpp
@@ -58,11 +58,12 @@
 #include "base/exception.h"
 #include "base/fs.h"
 #include "base/split_string.h"
+#include "base/system_console.h"
+#include "base/thread.h"
 #include "doc/sprite.h"
 #include "fmt/format.h"
 #include "os/error.h"
 #include "os/surface.h"
-#include "os/system.h"
 #include "os/window.h"
 #include "render/render.h"
 #include "ui/intern.h"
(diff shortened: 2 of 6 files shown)


martincapello:
At first I had the need to initialize the logging module before sentry initialization to be able to log sentry status to log file. Then I saw that sentry was being initialized from main and the modules from App, so I thought that it would be a good idea if sentry is initialized from App, which will make it closer to the modules initialization, and maybe I could have a chance to initialize it after modules initialization. Then it made sense to me that other code in main was pretty good candidate for moving it to App initialization, so I went ahead and made the refactor. It is not strictly needed, but I think it improves readability/design of the code...yes, I know that moving code around is not as safe as not touching them...but...just look at how clean main ends up when applied this change:
```c++
// Aseprite entry point. (Called from "os" library.)
int app_main(int argc, char* argv[])
{
  try {
    app::AppOptions options(argc, const_cast<const char**>(argv));
    app::App app;
    const int code = app.initialize(options);

    app.run();

    // After starting the GUI, we'll always return 0, but in batch
    // mode we can return the error code.
    return (app.isGui() ? 0: code);
  }
  catch (std::exception& e) {
    std::cerr << e.what() << '\n';
    os::error_message(e.what());
    return 1;
  }
}
```

I would also love to modify the return value to something that app.run() could return...but I just didn't do it because it is returning something from app.initialize...

I know that all the complexity was moved elsewhere, but I think that it was moved to where it belongs. Just my opinion, if you strongly disagree, that's fine, I can revert all this.

martincapello:
Oh, also, I was about to test this changes on windows, that's the reason they are failing to compile right now.

dacap:
Please don't refactor the `app_main()` function.

dacap:
I'll close some PRs that will not be merged.

PR #4131, opened 2023-11-11
Fix numerical errors in gradient tool

Charlie-83:

This PR aims to fix numerical issues with the gradient tool. To reproduce the issue, do the following:

1. Create a new sprite with width 57 and height 1.
2. Select #00e000 as primary colour and #000000 as secondary.
3. Select the gradient tool and drag starting from (57, 0) and ending at (0, 0)

Expectation: the green value of each pixel is equal to four times its x value.
Reality: some pixels have a green value equal to 4x-1

This is caused by numerical errors giving a green value of, say, 63.9999999999999 which, when cast to int, becomes 63.

I have done two things to try to fix this and applied my changes to both the linear and radial gradients.

1. Remove the unnecessary dividing, then later multiplying, by 255 and keep the colour values as ints.
2. Add a small value (1e7) to the calculated values to avoid errors caused by the division in `double f = (q * w) / wmag`

While I'm sure that almost no one else cares about such small errors in the produced colours, they have been very annoying for me as I have been using aseprite to generate position maps (images where RGB correspond to XYZ coordinates in 3D space). The gradient tool should be very useful for generating sequences of values but isn't due to this issue.

### DIFF
--- a/src/render/gradient.cpp
+++ b/src/render/gradient.cpp
@@ -79,15 +79,15 @@ void render_rgba_linear_gradient(
     c1 = (c0 & doc::rgba_rgb_mask);
   }
 
-  const double r0 = double(doc::rgba_getr(c0)) / 255.0;
-  const double g0 = double(doc::rgba_getg(c0)) / 255.0;
-  const double b0 = double(doc::rgba_getb(c0)) / 255.0;
-  const double a0 = double(doc::rgba_geta(c0)) / 255.0;
+  const uint8_t r0 = doc::rgba_getr(c0);
+  const uint8_t g0 = doc::rgba_getg(c0);
+  const uint8_t b0 = doc::rgba_getb(c0);
+  const uint8_t a0 = doc::rgba_geta(c0);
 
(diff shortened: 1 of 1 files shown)


Charlie-83:
Marking as draft until I sign CLA. Otherwise, the PR is complete

Charlie-83:
Would anyone be able to review this at some point?

Charlie-83:
This is a very simple PR. Hopefully should be quick to review

dacap:
Hi @Charlie-83, I'll review this one today if I can.

dacap:
It works great @Charlie-83, I've just tested it, attaching an image here for future reference just in case:

diff

The first row is the gradient with the old version, and the second one with this patch 👍

dacap:
Merged in https://github.com/aseprite/aseprite/commit/e47448ca242b46f38d4fb7578476bf7574e854d1 adding a reference to this PR in the commit message (I'll close the PR but actually it's merged).

dacap:
Thanks for your contribution @Charlie-83!

PR #4156, opened 2023-11-20
Fix strings typo

TakWolf:
Fix some strings typo

------------------


dacap:
Thanks @TakWolf! I think we can merge first two commits, but I see some problems with the third one ( 73dc5222cbeb7e9dedcbfd3d6d74d4a7267fdd2a ), because we need to specify two `&&` in some cases, e.g.

With the patch:

image

Original:

image

Other regressions introduced:

image
image
image

TakWolf:
I withdrew the third commit. 
But I don't understand the principle.
What is the uniqueness of these four places？There are also cases where a single ‘&’ is used elsewhere.

dacap:
> What is the uniqueness of these four places？

In these places the `&` has a special meaning (to indicate which character acts like a shortcut/mnemonic to quickly access the action) and that character will be underscored (so a special text drawing function is used, one that processes the `&` chars in a special way). I'm not sure if it's working on all those places (e.g. in "Guides && Slices" shouldn't be the case), but probably it's something to fix in the future as it's low priority.

Issue #2241, opened 2020-01-07
Reload preferences when they are externally modified


dacap:
This is a long standing issue, we should watch the .ini file for changes, and if the file is changed by other instanced reload the preferences from that file (or at least asking for the changes in the other instance).

In some way, after closing Aseprite, we should be able to "merge" the preferences of the running instances with existing .ini (e.g. recent files). Not sure if this can be done with all preferences.

Anyway I guess that just reloading preferences when the .ini file is modified externally should avoid these "merge conflict" or "losing preferences" when we overwrite the file.

Arihant-Joshi:
I'll solve this in my free time as a first issue.
I don't see how we can externally modifying preferences can act as a trigger to Aseprite, since that action would be platform dependent.
We could use a timer which reloads preferences for each instance every "X" seconds.
We could also try and check the last modified date of aseprite.ini every "X" seconds, and only reload if its updated, but those operations again wouldn't be platform independent.

PR #4164, opened 2023-11-22
Save preferences periodically if they are modified (fix #4154)

### LINKED ISSUE #4154: Timed autosave for aseprite.ini
Presently, Recent Files, Recent Folders, and other `aseprite.ini` data is only saved on Quit, meaning that any recent data is lost if the software crashes or is otherwise forcibly terminated. Instead, perhaps ASEprite could save this data automatically every so often, as an option? It would not have to be frequent; hourly or even every three or four hours would be plenty, but this could also be a setting in the same way that regular file autosaves are.

I live in an area which is increasingly subject to power failures, and this would certainly help me remember what I was doing before It All Fell Apart.

Thanks for considering it! :D 

..using ASEprite v1.3-rc6-x64 on Win10.

dacap:
This is an experimental implementation to save preferences periodically. I'm not sure about this because might be a conflict between several instances of Aseprite running at the same time.

### DIFF
--- a/src/app/app.cpp
+++ b/src/app/app.cpp
@@ -429,6 +429,15 @@ namespace {
     CloseMainWindow(std::unique_ptr<MainWindow>& win) : m_win(win) { }
     ~CloseMainWindow() { m_win.reset(nullptr); }
   };
+  // Each 10 seconds we save the modified preferences
+  struct SavePreferencesPeriodically {
+    ui::Timer m_timer;
+    SavePreferencesPeriodically() : m_timer(10000) {
+      m_timer.Tick.connect([]{ Preferences::instance().save(); });
+    }
+    ~SavePreferencesPeriodically() { }
+    void start() { m_timer.start(); }
+  };
 #endif
 
   struct CloseAllDocs {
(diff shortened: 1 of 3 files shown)


ckaiser:
@dacap Could this (and maybe other file-conflict things) be solved with an approach based on a [filesystem event watcher](https://github.com/e-dant/watcher/)? I using the `watcher` library for another project and I think it'd be great for configuration stuff (and possibly sprite files themselves) - I can look into it for this issue.

dacap:
@ckaiser I think you are looking for a solution to #2241 (https://github.com/SpartanJ/efsw was the library I was thinking about at the moment). If you want to take #2241 you can go for it (probably the reloading of preferences should be show an alert and an option to avoid displaying the alert in the future).

Although #4154 is other issue, a conflict between two different running instances can cause the same problem, but the issue is related to power failures and losing preferences in those cases.

ckaiser:
IMO a combination of the two things (periodic saving and detecting when things change from another instance or user modification) is gonna be necessary to keep things synchronized, for example reloading things when receiving focus after a while, and saving preferences periodically when we're in use, so I see the issues as pretty related in any case.

PR #4192, opened 2023-12-01
Patch mnemonic processing to accept strings with a single '&' char

martincapello:
With this change we can use strings that contains single '&' chars when they are followed by a space. 
Useful when one has to use an existing en.ini string that doesn't have their '&' escaped because they were used in a context without mnemonics enabled.


dacap:
It's not a bad idea as the space bar cannot be used as mnemonic, anyway I'd like to know how the translation string is specified inside the `export_file.xml`, because it might be better to just add the string to the `[export_file]` section too.

dacap:
Closing this PR. We'll see how we can solve the mnemonics problems (e.g. duplicated strings for commands and menus) in the future.

Issue #3922, opened 2023-06-09
Error decoding pico-8 GIFs


dacap:
Problems decoding gif files created from pico-8.

Reported here https://twitter.com/calixjumio/status/1667223925292408833
Internal ticket: https://igarastudio.zendesk.com/agent/tickets/4880

I have created a GIF to with pico-8 to reproduce this bug:

![untitled_4](https://github.com/aseprite/aseprite/assets/39654/439a9ecf-5d3f-4f10-a02a-effa7553d1f7)

PR #4205, opened 2023-12-06
Fix saving a file as gif successively will increase the palette until it's quantized (fix #3528)

### LINKED ISSUE #3528: Saving a file as gif successively will increase the palette until it's quantized
On beta, if we open [this file](https://github.com/aseprite/aseprite/files/9609897/test.aseprite.zip):

image

And then we save it as a gif, reopen the gif, and save it again, the palette will change successively. When we save the gif, we use these options:

image

And here is each time we save a gif, reopen it, and save it again:

image

image

image

image

Gasparoken:
fix #3528

### DIFF
--- a/src/app/file/gif_format.cpp
+++ b/src/app/file/gif_format.cpp
@@ -57,6 +57,8 @@
 // GifBitSize can return 9 (it's a bug in giflib)
 #define GifBitSizeLimited(v) (std::min(GifBitSize(v), 8))
 
+#define BLACK 0xff000000
+
 namespace app {
 
 using namespace base;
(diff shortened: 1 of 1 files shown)


Gasparoken:
@martincapello points out to me that he cannot reproduce the error with the file attached to the issue and I confirmed the situation.
This issue was fixed undetected in 1.3rc5 (possibly fixing #3922).

This fix goes a little further, discarding the default colors (0xff000000 = opaque black) added during gif decoding when it is detected that the color palette is global.

I explain the current situation of version 1.3.2:
- When the color palette is global, the decoder understands that it must maintain the order of the colors and that they are all used, the number of colors is always a power of 2 (a priori it is not known which colors will be used and which no, all colors are only marked as 'used' in frame 0).
- Any aseprite document that has a color count of a quantity NOT a power of 2 (for example 33), if it is saved as a gif + the 'preserve order' option, after opening the newly saved gif, the number of colors in the palette will be 64 instead of 33.

The present fix detects what was the highest color index used or different from 0xff000000, then the final palette of the sprite is trimmed to obtain a palette equal to the original without extra colors.

During the development of the solution, I was using a file with 17 colors and saving only once to check the growth of the palette when opening the new gif. That is why I did not detect before that the problem of the issue was no longer problematic.

I think this solution can be discarded without consequences. That's why I close this PR.

dacap:
OK @Gasparoken 👍 I'm going to close the issue linking to your comment.

PR #4207, opened 2023-12-07
Fix some strings format to weblate

TakWolf:
```
Warning
<<File format ".{0}" doesn't support:
<<
{1}<<      # <----- This
<<You can use ".aseprite" format to keep all this information.
<<Do you want continue with ".{0}" anyway?
||&Yes||&No
```

to

```
Warning
<<File format ".{0}" doesn't support:
<<
{1}
<<
<<You can use ".aseprite" format to keep all this information.
<<Do you want continue with ".{0}" anyway?
||&Yes||&No
```


dacap:
Thanks for the PR @TakWolf going to merge it.

> By the way, where is this set of strings located in the software?

It's the command name of some special shortcuts (F4 is a historical shortcut from the Allegro Sprite Editor):

image

The code is here: https://github.com/aseprite/aseprite/blob/516a927a59d8b6dc65061da39bcbf3434965337a/src/app/commands/cmd_palette_editor.cpp#L91-L107

> Is it necessary to keep the prefix space?

Yes, but only because it's a little weird how we are building the final PaletteEditorCommand string (probably difficult to replicate on other languages).

dacap:
Right now it's not possible, as some whitespaces might be duplicated in the final result. Anyway we could post-process the result replacing `"  "` with `" "` and trimming trailing spaces. I can made the change as this might simplify all translations.

dacap:
I've made the change 👍

PR #4211, opened 2023-12-07
Add option to play subtags & repetitions when exporting animations

### LINKED ISSUE #4173: Add option to follow tags repeat field in GIF export
Related to #1252, we should be able to select which tags, in which order, n times to repeat/follow repeat counter, etc. when we export a GIF file: https://twitter.com/2dchaos/status/1557113836183896065

martincapello:
This PR fixes #4173 and also includes a commit with a fix for the "ping-pong reverse" direction when exporting animations.

The main changes are around a new FramesSequence class that is similar to SelectedFrames but allows us to keep ranges of frames in any direction (increasing or decreasing) in the same order as the frames were added.

### DIFF
--- a/src/app/ui/export_file_window.cpp
+++ b/src/app/ui/export_file_window.cpp
@@ -59,12 +59,13 @@ ExportFileWindow::ExportFileWindow(const Doc* doc)
   pixelRatio()->setSelected(m_docPref.saveCopy.applyPixelRatio());
   forTwitter()->setSelected(m_docPref.saveCopy.forTwitter());
   adjustResize()->setVisible(false);
-
+  playSubtags()->setSelected(m_docPref.saveCopy.playSubtags());
   // Here we don't call updateAniDir() because it's already filled and
   // set by the function fill_anidir_combobox(). So if the user
   // exported a tag with a specific AniDir, we want to keep the option
   // in the preference (instead of the tag's AniDir).
   //updateAniDir();
+  updatePlaySubtags();
 
   updateAdjustResizeButton();
 
@@ -82,7 +83,10 @@ ExportFileWindow::ExportFileWindow(const Doc* doc)
     });
 
   resize()->Change.connect([this]{ updateAdjustResizeButton(); });
-  frames()->Change.connect([this]{ updateAniDir(); });
+  frames()->Change.connect([this]{
+    updateAniDir();
+    updatePlaySubtags();
+  });
   forTwitter()->Click.connect([this]{ updateAdjustResizeButton(); });
   adjustResize()->Click.connect([this]{ onAdjustResize(); });
   ok()->Click.connect([this]{ onOK(); });
@@ -105,6 +109,7 @@ void ExportFileWindow::savePref()
   m_docPref.saveCopy.frameTag(framesValue());
   m_docPref.saveCopy.applyPixelRatio(applyPixelRatio());
   m_docPref.saveCopy.forTwitter(isForTwitter());
+  m_docPref.saveCopy.playSubtags(isPlaySubtags());
 }
 
 std::string ExportFileWindow::outputFilenameValue() const
@@ -143,6 +148,11 @@ std::string ExportFileWindow::framesValue() const
 doc::AniDir ExportFileWindow::aniDirValue() const
 {
   return (doc::AniDir)anidir()->getSelectedItemIndex();
+}
+
+bool ExportFileWindow::isPlaySubtags() const
+{
+  return playSubtags()->isSelected() && framesValue() != kSelectedFrames;
 }
 
 bool ExportFileWindow::applyPixelRatio() const
@@ -203,6 +213,13 @@ void ExportFileWindow::updateAniDir()
   }
   else
     anidir()->setSelectedItemIndex(int(doc::AniDir::FORWARD));
+}
+
+void ExportFileWindow::updatePlaySubtags()
+{
+  std::string framesValue = this->framesValue();
+  playSubtags()->setVisible(framesValue != kSelectedFrames);
+  layout();
 }
 
 void ExportFileWindow::updateAdjustResizeButton()
--- a/src/doc/frames_sequence.cpp
+++ b/src/doc/frames_sequence.cpp
@@ -1,0 +1,207 @@
+// Aseprite Document Library
+// Copyright (c) 2023 Igara Studio S.A.
+//
+// This file is released under the terms of the MIT license.
+// Read LICENSE.txt for more information.
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "doc/frames_sequence.h"
+
+#include "base/base.h"
+#include "base/debug.h"
+#include "base/serialization.h"
+
+#include <algorithm>
+#include <iostream>
+
+namespace doc {
+
+using namespace base::serialization;
+using namespace base::serialization::little_endian;
+
+FramesSequence::FramesSequence(const SelectedFrames& selectedFrames)
+{
+  // TODO: This might be optimized by copying each range directly, but
+  // for this I would need that SelectedFrames make this class its friend.
+  for(auto frame : selectedFrames) {
+    this->insert(frame);
+  }
+}
+
+std::size_t FramesSequence::size() const
+{
+  std::size_t size = 0;
+  for (auto& range : m_ranges)
+    size += ABS(range.toFrame - range.fromFrame) + 1;
+  return size;
+}
+
+void FramesSequence::clear()
+{
+  m_ranges.clear();
+}
+
+void FramesSequence::insert(frame_t frame)
+{
+  ASSERT(frame >= 0);
+
+  if (m_ranges.empty()) {
+    m_ranges.push_back(FrameRange(frame));
+    return;
+  }
+
+  auto& lastRange = m_ranges.back();
+  int rangeStep = (lastRange.fromFrame < lastRange.toFrame) -
+                  (lastRange.fromFrame > lastRange.toFrame);
+  int step = (lastRange.toFrame < frame) - (lastRange.toFrame > frame);
+  if (step &&
+      (rangeStep == step || !rangeStep) &&
+      frame-lastRange.toFrame == step)
+    lastRange.toFrame = frame;
+  else
+    m_ranges.push_back(FrameRange(frame));
+}
+
+void FramesSequence::insert(frame_t fromFrame, frame_t toFrame)
+{
+  int step = (fromFrame <= toFrame ? 1 : -1);
+  for (frame_t frame = fromFrame; frame != toFrame; frame+=step) {
+    insert(frame);
+  }
+  insert(toFrame);
+}
+
+FramesSequence FramesSequence::filter(frame_t fromFrame, frame_t toFrame) const
+{
+  FramesSequence f;
+
+  if (fromFrame > toFrame)
+    std::swap(fromFrame, toFrame);
+
+  for (const auto& range : m_ranges) {
+    FrameRange r(range);
+    const bool isForward = (r.fromFrame <= r.toFrame);
+
+    if (isForward) {
+      if (r.fromFrame < fromFrame) r.fromFrame = fromFrame;
+      if (r.toFrame > toFrame) r.toFrame = toFrame;
+    }
+    else {
+      if (r.fromFrame > toFrame) r.fromFrame = toFrame;
+      if (r.toFrame < fromFrame) r.toFrame = fromFrame;
+    }
+
+    if (( isForward && r.fromFrame <= r.toFrame) ||
+        (!isForward && r.fromFrame >= r.toFrame))
+      f.m_ranges.push_back(r);
+  }
+
+  return f;
+}
+
+bool FramesSequence::contains(frame_t frame) const
+{
+  return std::find_if(
+            m_ranges.begin(),
+            m_ranges.end(),
+            [frame](const FrameRange& r) -> bool {
+              return (r.fromFrame <= frame && frame <= r.toFrame) ||
+                     (r.fromFrame >= frame && frame >= r.toFrame);
+            }) != m_ranges.end();
+}
+
+frame_t FramesSequence::lowestFrame() const
+{
+  frame_t lof = this->firstFrame();
+  for (auto range : m_ranges) {
+    lof = std::min<frame_t>(range.fromFrame, std::min<frame_t>(range.toFrame, lof));
+  }
+  return lof;
+}
+
+void FramesSequence::displace(frame_t frameDelta)
+{
+  // Avoid setting negative numbers in frame ranges
+  auto lof = lowestFrame();
+  if (lof+frameDelta < 0)
+    frameDelta = -lof;
+
+  for (auto& range : m_ranges) {
+    range.fromFrame += frameDelta;
+    range.toFrame += frameDelta;
+
+    ASSERT(range.fromFrame >= 0);
+    ASSERT(range.toFrame >= 0);
+  }
+}
+
+FramesSequence FramesSequence::makeReverse() const
+{
+  FramesSequence newFrames;
+  for (const FrameRange& range : m_ranges)
+    newFrames.m_ranges.insert(
+      newFrames.m_ranges.begin(),
+      FrameRange(range.toFrame, range.fromFrame));
+  return newFrames;
+}
+
+FramesSequence FramesSequence::makePingPong() const
+{
+  FramesSequence newFrames = *this;
+  const int n = m_ranges.size();
+  int i = 0;
+  int j = m_ranges.size()-1;
+
+  for (const FrameRange& range : m_ranges) {
+    // Discard first or last range if it contains just one frame.
+    if (range.toFrame == range.fromFrame && (i == 0 || j == 0)) {
+      ++i;
+      --j;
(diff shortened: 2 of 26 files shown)


martincapello:
One thing that might worth considering is the fact that FramesSequence and SelectedFrames have several methods that behaves exactly the same way, so I think we could take this changes a step further and make SelectedFrames a subclass of FramesSequence without much hassle. Just let me know what do you think.

martincapello:
Let me know if you prefer all the commits squashed but 2ec2896471307f9c8b1e7355a6695a253c6ac92a

dacap:
I'm starting to test this change (didn't review it yet). There is no problem about the multiple commits 👍  

I've just tested this file: [tags4.aseprite.zip](https://github.com/aseprite/aseprite/files/13788249/tags4.aseprite.zip)

And the output doesn't include the last tag in reverse modes 4 times (it should show 10-9-8 four times):

tags4

But it shows only 10 at the end, you can compare with the regular editor playback.

martincapello:
Just fixed, it was an issue in the Playback class when using PlayAll mode. https://github.com/aseprite/aseprite/pull/4211/commits/c624f0a0586ede37df5e9c14b7a6c2fb0bc35884 fixes it.

src/doc/frames_sequence.cpp
```diff
@@ -0,0 +1,207 @@
...
+      (rangeStep == step || !rangeStep) &&
+      frame-lastRange.toFrame == step)
+    lastRange.toFrame = frame;
+  else
+    m_ranges.push_back(FrameRange(frame));
+}
+
+void FramesSequence::insert(frame_t fromFrame, frame_t toFrame)
+{
+  int step = fromFrame <= toFrame ? 1 : -1;
```
dacap, line 70: Please add `(...)` in these kind of operators.
```suggestion
  int step = (fromFrame <= toFrame ? 1 : -1);
```

src/doc/selected_frames.h
```diff
@@ -16,126 +16,136 @@
...
+      : m_frames(frames) {
+    }
 
+  private:
+    const T& m_frames;
+  };
+
+} // namespace frames
+
+  using namespace frames;
```
dacap, line 145: Creating a namespace and immediately using it has the same result that including everything in the parent namespace. In this way if we are in (or include) `doc` namespace we have `const_iterator` type available as a global type (which is not desirable).

One easy alternative (not the most elegant one but it works) is to change the namespace to a struct:

```c++
  struct frames {
    // ...
  }; // struct frames

  class SelectedFrames : public frames {
  public:
    // ...
```

src/doc/frames_sequence.h
```diff
@@ -0,0 +1,70 @@
...
+#include "doc/selected_frames.h"
+
+#include <iosfwd>
+#include <vector>
+
+namespace doc {
+
+  using namespace frames;
+
+  class FramesSequence {
```
dacap, line 21: Add a comment that this class is based in `SelectedFrames` and a `TODO` to for future work to be done to remove all the duplicated code in some way. A similar comment can go to the `SelectedFrames` class definition.

dacap (changes requested):
So far it looks great 👍  there are some minor changes and fixes to be done yet.

src/app/ui/export_file_window.cpp
```diff
@@ -198,13 +208,24 @@ void ExportFileWindow::updateAniDir()
...
     Tag* tag = calculate_selected_frames(
       UIContext::instance()->activeSite(), framesValue, selFrames);
-    if (tag)
-      anidir()->setSelectedItemIndex(int(tag->aniDir()));
+    if (tag) {
+      if (isPlaySubtags())
+        anidir()->setSelectedItemIndex(int(doc::AniDir::FORWARD));
+      else
+        anidir()->setSelectedItemIndex(int(tag->aniDir()));
+    }
```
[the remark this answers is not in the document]
martincapello: This was my reasoning: When the user checks the play subtags option then the direction of the animation is given by the tag properties (because I thought that the user is checking the option to actually follow what the tag says), then the "animation direction" combobox will change the whole animation direction.
Also I thought in this scenario: for instance, suppose I have a tag that has a "reverse" direction, then I select it and the UI sets the anidir to Reverse. Then when I check the "play subtags" option I would like that the animation respects what my tags say (the thing is that a tag can algo have subtags inside, so at the end playing all frames is the same as playing one tag that contains subtags), then if the animation direction is kept to reverse, it will reverse the reverse of the tag...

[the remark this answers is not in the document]
martincapello: But yes, you are right that it is a bug to my reasoning that the anidir is not being set to forward when checking the play subtags checkbox and the anidir was set in a different value.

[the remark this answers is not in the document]
martincapello: In summary, my intention was to always make Animation Direction work in such a way that it sets the direction of the final animation.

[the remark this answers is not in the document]
dacap: From an user point of view, I think they will never need to click/unclick the Play subtags checkbox. I forgot to mention to change the `<option id="play_subtags" type="bool" default="false" />` to `default="true"` because this is the expected behavior by users and it's the fix that they are requesting.

As the Play subtags should be checked by default (expected behavior for user that set the Repeat field of a tag, etc.), it makes sense that the anidir should show the tag anidir when we export specifically one tag (which is the current UI behavior of v1.3.2). The "forward" item to indicate "export the tag in its anidir but we'll not touch the final result keeping it going forward the whole result" is confusing because it's too internal to the implementation)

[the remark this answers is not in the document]
martincapello: > From an user point of view, I think they will never need to click/unclick the Play subtags checkbox. I forgot to mention to change the `<option id="play_subtags" type="bool" default="false" />` to `default="true"` because this is the expected behavior by users and it's the fix that they are requesting.

👍 

> As the Play subtags should be checked by default (expected behavior for user that set the Repeat field of a tag, etc.), it makes sense that the anidir should show the tag anidir when we export specifically one tag (which is the current UI behavior of v1.3.2). The "forward" item to indicate "export the tag in its anidir but we'll not touch the final result keeping it going forward the whole result" is confusing because it's too internal to the implementation)

Let's say we have something like this, where we have a tag (with reverse direction) that is parent of two other tags and extends to the whole animation ([tetstags.aseprite.zip](https://github.com/aseprite/aseprite/files/14104150/tetstags.aseprite.zip))
Then if the user selects the "parent tag" we should show the "Reverse" in the anidir combo, right? Then when exporting the animation it should follow the selected direction as the tag says.
But, in the other case, when the user selects "All frames", if the user selects "Forward", what should be exported? The same as if the user selected the parent tag with anidir "Reverse"?
Even without having child tags, just with one big tag (with reverse direction) that extends to the whole animation. What should be expected in each case?

NOTE: When testing this case, I found that there are some issues when exporting animations with nested tags.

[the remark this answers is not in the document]
martincapello: I believe that the best would be to not allow to select the animation direction when the play subtags is checked.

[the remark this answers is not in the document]
dacap: The current v1.3 behavior should answer your questions. You will see that when "All frames" is selected, the "Forward" option is selected to indicate that we're going to playback all frames/tags as they are (the export must be exactly as the playback). If we change the "All frames" direction the final result will change.

But for the case of tags, the anidir indicates the direction of the tag: if we choose the parent tag and the anidir says "reverse", we are going to export the playing the tag with its original anidir, but if we change the anidir, we change the anidir of the parent (exported) tag.

src/doc/selected_frames.h
martincapello: Oh I see. I think I will move the doc::frames namespace to an independent header file. Thank you!

martincapello:
I've introduced all the suggested changes along some additional fixes

src/doc/selected_frames.h
dacap: I think the problem is still persistent, the issue here is `using namespace doc::frames;`, not the header file name or where the namespace is defined.

martincapello: Why do you say it? Maybe I'm getting it wrong, or I'm testing it wrong.
To check if this works I'm doing the following:
1. Go to layer_frame_comboboxes.cpp file (actually any file that includes selected_frames.h)
2. In some method try to declare:
`doc::frames::Ranges r;` <-- compiles
`doc::Ranges r;` <-- doesn't compile

I'm using the Ranges type, but with the const_iterator class happens the same given appropriate ctor parameters.

Before moving the namespace to its own header, and more importantly (because this made it actually work), moving the `using namespace doc::frames` outside of the doc namespace, I was able to use doc::Ranges or doc::const_iterator.

dacap: Add this line in `sprite.cpp`:
```diff
diff --git a/src/doc/sprite.cpp b/src/doc/sprite.cpp
index b26cd0e77..861ff33a2 100644
--- a/src/doc/sprite.cpp
+++ b/src/doc/sprite.cpp
@@ -468,6 +468,8 @@ RgbMap* Sprite::rgbMap(const frame_t frame,
 
 void Sprite::addFrame(frame_t newFrame)
 {
+  const_iterator it;
+
   setTotalFrames(m_frames+1);
 
   frame_t to = std::max(1, newFrame);
```
You will get this error:
```
src/doc/sprite.cpp:471:18: error: no matching constructor for initialization of 'const_iterator'
  const_iterator it;
                 ^
src/doc/../doc/frames_iterators.h:30:5: note: candidate constructor not viable: requires single argument 'it', but no arguments were provided
    const_iterator(const frames::Ranges::const_iterator& it)
    ^
src/doc/../doc/frames_iterators.h:21:9: note: candidate constructor (the implicit copy constructor) not viable: requires 1 argument, but 0 were provided
  class const_iterator {
        ^
src/doc/../doc/frames_iterators.h:21:9: note: candidate constructor (the implicit move constructor) not viable: requires 1 argument, but 0 were provided
```
Which means that the `const_iterator` type is available in the whole `doc` namespace.

If you make these changes:
```diff
diff --git a/src/doc/frames_iterators.h b/src/doc/frames_iterators.h
index 0a2d03b1d..830aa5f16 100644
--- a/src/doc/frames_iterators.h
+++ b/src/doc/frames_iterators.h
@@ -15,7 +15,7 @@
 
 namespace doc {
 
-namespace frames {
+struct frames {
   typedef std::vector<FrameRange> Ranges;
 
   class const_iterator {
@@ -139,7 +139,7 @@ namespace frames {
     const T& m_frames;
   };
 
-} // namespace frames
+}; // struct frames
 } // namespace doc
 
 #endif  // DOC_FRAMES_ITERATORS_H_INCLUDED
diff --git a/src/doc/frames_sequence.h b/src/doc/frames_sequence.h
index e32bd314c..a51ea217c 100644
--- a/src/doc/frames_sequence.h
+++ b/src/doc/frames_sequence.h
@@ -10,14 +10,12 @@
 
 #include "doc/selected_frames.h"
 
-using namespace doc::frames;
-
 namespace doc {
 
   // This class is based in several code of the SelectedFrames class.
   // TODO: At some point we should remove the duplicated code between these
   // classes.
-  class FramesSequence {
+  class FramesSequence : public frames {
   public:
     FramesSequence() {}
     FramesSequence(const SelectedFrames& selectedFrames);
diff --git a/src/doc/selected_frames.h b/src/doc/selected_frames.h
index 566e5df0d..785ee555a 100644
--- a/src/doc/selected_frames.h
+++ b/src/doc/selected_frames.h
@@ -11,14 +11,12 @@
 
 #include "doc/frames_iterators.h"
 
-using namespace doc::frames;
-
 namespace doc {
 
   // The FramesSequence class is based in several code of this class.
   // TODO: At some point we should remove the duplicated code between these
   // classes.
-  class SelectedFrames {
+  class SelectedFrames : public frames {
   public:
     const_iterator begin() const { return const_iterator(m_ranges.begin()); }
     const_iterator end() const { return const_iterator(m_ranges.end()); }
```
And recompile you will get this error:
````
src/doc/sprite.cpp:471:3: error: unknown type name 'const_iterator'
  const_iterator it;
  ^
```
Which means that (e.g.) the `const_iterator` is not a type published outside `FramesSequence/SelectedFrames` bounds anymore.

martincapello: I see what you mean now, thank you for clarifying. I don't really like the semantics of adding a struct to make a couple of classes inherit from it just to make the namespaces work as we want, it just feels wrong to me. There is an alternative that I was avoiding because of the use of the "frames" namespace repeated times which I don't like either, but makes me feel better in terms of semantics:

```diff
diff --git a/src/doc/frames_sequence.h b/src/doc/frames_sequence.h
index e32bd314c..488df1048 100644
--- a/src/doc/frames_sequence.h
+++ b/src/doc/frames_sequence.h
@@ -10,8 +10,6 @@
 
 #include "doc/selected_frames.h"
 
-using namespace doc::frames;
-
 namespace doc {
 
   // This class is based in several code of the SelectedFrames class.
@@ -22,10 +20,10 @@ namespace doc {
     FramesSequence() {}
     FramesSequence(const SelectedFrames& selectedFrames);
 
-    const_iterator begin() const { return const_iterator(m_ranges.begin()); }
-    const_iterator end() const { return const_iterator(m_ranges.end()); }
-    const_reverse_iterator rbegin() const { return const_reverse_iterator(m_ranges.rbegin()); }
-    const_reverse_iterator rend() const { return const_reverse_iterator(m_ranges.rend()); }
+    frames::const_iterator begin() const { return frames::const_iterator(m_ranges.begin()); }
+    frames::const_iterator end() const { return frames::const_iterator(m_ranges.end()); }
+    frames::const_reverse_iterator rbegin() const { return frames::const_reverse_iterator(m_ranges.rbegin
()); }
+    frames::const_reverse_iterator rend() const { return frames::const_reverse_iterator(m_ranges.rend());
 }
 
     std::size_t size() const;
     std::size_t ranges() const { return m_ranges.size(); }
@@ -44,7 +42,7 @@ namespace doc {
     frame_t lowestFrame() const;
 
     void displace(frame_t frameDelta);
-    Reversed<FramesSequence> reversed() const { return Reversed(*this); }
+    frames::Reversed<FramesSequence> reversed() const { return frames::Reversed(*this); }
 
     FramesSequence makeReverse() const;
     FramesSequence makePingPong() const;
@@ -61,7 +59,7 @@ namespace doc {
     bool read(std::istream& is);
 
   private:
-    Ranges m_ranges;
+    frames::Ranges m_ranges;
   };
 
 } // namespace doc
diff --git a/src/doc/selected_frames.h b/src/doc/selected_frames.h
index 566e5df0d..3eec9e975 100644
--- a/src/doc/selected_frames.h
+++ b/src/doc/selected_frames.h
@@ -11,8 +11,6 @@
 
 #include "doc/frames_iterators.h"
 
-using namespace doc::frames;
-
 namespace doc {
 
   // The FramesSequence class is based in several code of this class.
@@ -20,10 +18,10 @@ namespace doc {
   // classes.
   class SelectedFrames {
   public:
-    const_iterator begin() const { return const_iterator(m_ranges.begin()); }
-    const_iterator end() const { return const_iterator(m_ranges.end()); }
-    const_reverse_iterator rbegin() const { return const_reverse_iterator(m_ranges.rbegin()); }
-    const_reverse_iterator rend() const { return const_reverse_iterator(m_ranges.rend()); }
+    frames::const_iterator begin() const { return frames::const_iterator(m_ranges.begin()); }
+    frames::const_iterator end() const { return frames::const_iterator(m_ranges.end()); }
+    frames::const_reverse_iterator rbegin() const { return frames::const_reverse_iterator(m_ranges.rbegin
()); }
+    frames::const_reverse_iterator rend() const { return frames::const_reverse_iterator(m_ranges.rend());
 }
 
     std::size_t size() const;
     std::size_t ranges() const { return m_ranges.size(); }
@@ -40,7 +38,7 @@ namespace doc {
     frame_t lastFrame() const { return (!m_ranges.empty() ? m_ranges.back().toFrame: -1); }
 
     void displace(frame_t frameDelta);
-    Reversed<SelectedFrames> reversed() const { return Reversed(*this); }
+    frames::Reversed<SelectedFrames> reversed() const { return frames::Reversed(*this); }
 
     SelectedFrames makeReverse() const;
     SelectedFrames makePingPong() const;
@@ -57,7 +55,7 @@ namespace doc {
     bool read(std::istream& is);
 
   private:
-    Ranges m_ranges;
+    frames::Ranges m_ranges;
   };
 
 } // namespace doc
```
Anyway, if you still think the struct changes worth it, I can introduce them and then relief my semantics pains with a bit of "Agaromba" 😄 
Just let me know what is your preference. Thanks.

dacap: Yeah, the struct is not the most elegant solution. It's a shame that we cannot use "using namespace" inside a class.

Your solution is good enough as we don't refer the iterators types outside these classes (or we use auto or range-based fors), other possibility would be adding (my initial idea but got some problems with the Reversed type and left it there):

```c++
  class SelectedFrames {
  public:
    using iterator = frames::iterator;
    // etc.
```
For each type, but we can leave this with `frames::`. I'll test this again and merge it 👍

martincapello:
Oh, I didn't pushed any changes about the "frames" namespace we were talking. I was working on another thing. I'm going to introduce the changes in a new PR.

PR #4238, opened 2023-12-22
Fx Alt+T -> Alt+V instantly crashes Aseprite without warning (fix #4221)

### LINKED ISSUE #4221: Alt+T -> Alt+V instantly crashes Aseprite without warning
Reported here:

https://steamcommunity.com/app/431730/discussions/2/4034725980842866910/

Gasparoken:
First approach.

fix #4221

### DIFF
--- a/src/ui/menu.cpp
+++ b/src/ui/menu.cpp
@@ -323,6 +323,9 @@ void MenuItem::setSubmenu(Menu* menu)
 
 void MenuItem::openSubmenu()
 {
+  MenuBaseData* base = get_base(this);
+  if (base && base->is_processing)
+    return;
   if (auto menu = static_cast<Menu*>(parent()))
     menu->highlightItem(this, true, true, true);
 }


src/ui/menu.cpp
```diff
@@ -630,6 +645,15 @@ bool MenuBox::onProcessMessage(Message* msg)
...
       if (menu) {
+        MenuItem* child_with_submenu_opened = nullptr;
+        findChildWithSubmenuOpened(menu, child_with_submenu_opened);
+        if (child_with_submenu_opened &&
+            child_with_submenu_opened->inBar() &&
+            this->type() == kMenuBarWidget &&
+            msg->modifiers() == kKeyAltModifier) {
+          cancelMenuLoop();
+        }
+
```
martincapello, line 656: I've been testing the fix and it works great. The only thing I'm not sure about is this piece of code, which I couldn't figure out how to make this if condition true. I also tested the fix without this code and it seems to work as well. So I would love to know what I am missing here.

martincapello (commented):
The fix works as expected, but I would need some clarification about an if condition.

src/ui/menu.cpp
Gasparoken: You're right! Thank you!

martincapello (approved):

PR #4239, opened 2023-12-22
Update INSTALL.md for windows compilation such that a misleading arch…

dev-nt90:
…itecture option is avoided when calling the developer command prompt

As it exists today, the provided `call "...VsDevCmd.bat" -arch=x64` command will output the developer command prompt at startup then close itself after a few seconds. What's going on, is the command prompt is silently failing, because it doesn't recognize the "x64" value for the -arch argument. 

This change makes the example `call` have a generic value for target architecutre, and links to the documentation where the valid values can be reviewed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -129,9 +129,9 @@ On SUSE:
 
 Open a command prompt window (`cmd.exe`) and call:
 
-    call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=x64
+    call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=<target architecture>
 
-The command above is required while using the 64-bit version of skia. When compiling with the 32-bit version, it is possible to open a [developer command prompt](https://docs.microsoft.com/en-us/dotnet/framework/tools/developer-command-prompt-for-vs) instead.
+The command above is required while using the 64-bit version of skia. When compiling with the 32-bit version, it is possible to open a [developer command prompt](https://docs.microsoft.com/en-us/dotnet/framework/tools/developer-command-prompt-for-vs) instead. See this section to specify the correct [target architecture](https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022#target-architecture-and-host-architecture).
 
 And then
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -129,9 +129,9 @@ On SUSE:
 
 Open a command prompt window (`cmd.exe`) and call:
 
-    call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=<target architecture>
+    call "C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\Tools\VsDevCmd.bat" -arch=amd64
 
-The command above is required while using the 64-bit version of skia. When compiling with the 32-bit version, it is possible to open a [developer command prompt](https://docs.microsoft.com/en-us/dotnet/framework/tools/developer-command-prompt-for-vs) instead. See this section to specify the correct [target architecture](https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022#target-architecture-and-host-architecture).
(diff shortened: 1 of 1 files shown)


dacap:
Hi @dev-nt90, I think most user will just copy and paste the command line example without even touching the `<target architecture>`. I'm not sure if we should incorporate this change to the guide.

What I cannot believe is why Microsoft removed the x64 CLI access to MSVC from the start menu and it defaulted to x86 (the least used arch).

dev-nt90:
@dacap That's fair. If we're concerned with the most common path, then what I think makes most sense is setting the arch value to "amd64", as that's what most people will be using, while leaving a reference to the target arch documentation for the more exotic cases.

dev-nt90:
@dacap the aforementioned changes are committed at e2438e1

dacap:
From my side `-arch=x64` just works. If you want to undo that change I can merge (squashing this PR) the other note.

Anyway I've found a way to run the x64 devtools from the Start menu: It's called `x64 Native Tools Command Prompt for VS 2022`, I'll add this in the guide.

dacap:
From `c:/Program Files/Microsoft Visual Studio/2022/Community/Common7/Tools/vsdevcmd/core/parse_cmd.bat` the `amd64` is converted to `x64` internally:

```
...
    @REM Set host and target architecture for tools that depend on this being
    @REM available. Note that we have special handling of "amd64" to convert to
    @REM "x64" due legacy usage of the former.
    if "%__VSCMD_ARG_TGT_ARCH%" NEQ "" (
        if "%__VSCMD_ARG_TGT_ARCH%"=="amd64" (
            set "VSCMD_ARG_TGT_ARCH=x64"
        ) else (
            set "VSCMD_ARG_TGT_ARCH=%__VSCMD_ARG_TGT_ARCH%"
        )
    ) else (
        set "VSCMD_ARG_TGT_ARCH=x86"
    )

    if "%__VSCMD_ARG_HOST_ARCH%" NEQ "" (
        if "%__VSCMD_ARG_HOST_ARCH%"=="amd64" (
            set "VSCMD_ARG_HOST_ARCH=x64"
        ) else (
            set "VSCMD_ARG_HOST_ARCH=%__VSCMD_ARG_HOST_ARCH%"
        )
    ) else (
        @REM By default, the host architecture will match the target
        @REM architecture, which was exported above.
        set "VSCMD_ARG_HOST_ARCH=%VSCMD_ARG_TGT_ARCH%"
    )
... 
```

dacap:
Closing this PR at the moment. I'm not sure if the architecture comment worth it, but in case it does, you can just force push an updated commit/solved conflict.

dev-nt90:
> From `c:/Program Files/Microsoft Visual Studio/2022/Community/Common7/Tools/vsdevcmd/core/parse_cmd.bat` the `amd64` is converted to `x64` internally:
> 
> ```
> ...
>     @REM Set host and target architecture for tools that depend on this being
>     @REM available. Note that we have special handling of "amd64" to convert to
>     @REM "x64" due legacy usage of the former.
>     if "%__VSCMD_ARG_TGT_ARCH%" NEQ "" (
>         if "%__VSCMD_ARG_TGT_ARCH%"=="amd64" (
>             set "VSCMD_ARG_TGT_ARCH=x64"
>         ) else (
>             set "VSCMD_ARG_TGT_ARCH=%__VSCMD_ARG_TGT_ARCH%"
>         )
>     ) else (
>         set "VSCMD_ARG_TGT_ARCH=x86"
>     )
> 
>     if "%__VSCMD_ARG_HOST_ARCH%" NEQ "" (
>         if "%__VSCMD_ARG_HOST_ARCH%"=="amd64" (
>             set "VSCMD_ARG_HOST_ARCH=x64"
>         ) else (
>             set "VSCMD_ARG_HOST_ARCH=%__VSCMD_ARG_HOST_ARCH%"
>         )
>     ) else (
>         @REM By default, the host architecture will match the target
>         @REM architecture, which was exported above.
>         set "VSCMD_ARG_HOST_ARCH=%VSCMD_ARG_TGT_ARCH%"
>     )
> ... 
> ```

Well first of all, thanks for taking a look.

This is interesting. I wonder why MSFT would document amd64 as an option but then tag it as legacy in the devcmd script. It's even listed as the option in that same file!

```
@echo Syntax: vsdevcmd.bat [options]
@echo [options] :
@echo     -arch=architecture : Architecture for compiled binaries/libraries
@echo            ** x86 [default]
@echo            ** amd64
@echo            ** arm
@echo            ** arm64
@echo     -host_arch=architecture : Architecture of compiler binaries
@echo            ** x86 [default]
@echo            ** amd64
```


Since opening this issue, I have unsurprisingly updated my local environment a bit. Maybe there was a window of time/updates where the x64 value didn't work? It seems unlikely, but I couldn't get this particular section of the guide to work without supplying the amd64 param. Next time I get around to compiling the project I'll give it a shot.

PR #4241, opened 2023-12-24
Add snap line to more angles (fix #1641)

### LINKED ISSUE #1641: Ability to snap line tool to more angles
The line tool in my opinion needs to be able to snap to the closest perfect line it can match.

What I mean with a perfect line is a line that has the same amount of pixels per "step" on every step. Right now it can snap to the same angles the rotation can snap to, which gives easy access to diagonal lines (1 pixel steps) and 25 degree angles which gives you 2 pixels per step. It would be nice to be able to hold alt or something to make it snap to 3 pixel steps and greater so you can always get easy "perfect" lines.

tsone:
Here's a small Christmas gift. More angles to snap lines. Issue #1641.

Added the drawing routine `algo_line_snap()` because existing `algo_line_perfect()` didn't seem to work. In fact, for best results, the line pixel run length (`m`), of the snapping done in `controllers.h`, should be passed somehow to `algo_line_snap()`. I notice `data` parameter in the function, but I don't know if you'd like me to add something in `data` just for this feature?

Longest pixel run of the snapped lines are currently hard-coded to 8 px limit. If you're interested, I could add a feature that dynamically changes this limit based on the line length. For example, 12 px runs could work for a line of 24 px or longer (or, max run = line length / 2). I experimented this but didn't yet find a good formula for this max run. Let me know if you're interested in this. Note, the tool feel of this dynamic snapping may be a bit random to users, so if you like me to pursue this I need feedback about the feel.

Todo:

- [ ] Line brush drawing routine in `algo`.
- [ ] (Optional) Dynamic line pixel run length.
- [x] Line tool "Draw from center" (Ctrl) mode.

### DIFF
--- a/src/app/tools/controllers.h
+++ b/src/app/tools/controllers.h
@@ -153,45 +153,16 @@ public:
     stroke[0] = m_first;
     stroke[1] = pt;
 
-    bool isoAngle = false;
+    int snapM = 0;
 
     if ((int(loop->getModifiers()) & int(ToolLoopModifiers::kSquareAspect))) {
-      int dx = stroke[1].x - m_first.x;
-      int dy = stroke[1].y - m_first.y;
-      int minsize = std::min(ABS(dx), ABS(dy));
-      int maxsize = std::max(ABS(dx), ABS(dy));
+      const int dx = stroke[1].x - m_first.x;
+      const int dy = stroke[1].y - m_first.y;
+      const int minsize = std::min(ABS(dx), ABS(dy));
 
       // Lines
       if (loop->getIntertwine()->snapByAngle()) {
-        double angle = 180.0 * std::atan(static_cast<double>(-dy) /
-                                         static_cast<double>(dx)) / PI;
-        angle = ABS(angle);
-
-        // Snap horizontally
-        if (angle < 18.0) {
-          stroke[1].y = m_first.y;
-        }
-        // Snap at 26.565
-        else if (angle < 36.0) {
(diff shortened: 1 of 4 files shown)


tsone:
At current state I'm waiting for your feedback on how to proceed.

dacap:
Hi @tsone! Thanks for this contribution as Christmas gift ☺️ 

I've just tested it and it works great, about the `MAX_M` parameter, I think it could be an option (not sure where to specify it, if in the context bar or in the preferences). E.g. I'd like a way to specify `MAX_M=2` to get the current behavior (I guess some users will prefer the current snap to iso angles instead of many angles).

tsone:
Thanks for testing!

I didn't think of isometric drawing, but now you mention it I understand the importance. Is 2:1 the only ratio for isometric drawing?

How about adding key action modifier to Alt for this new snapping? Ctrl would remain the same for isometric snapping in Freehand tool, and Shift in Line tool. Alt key seems to be free in both Freehand and Line tool, am I right?

tsone:
If having Alt key doesn't sound good, from your previous options I prefer to add it in the settings. However, I think isometric pixel art as a subcategory of pixel art, so such isometric snapping would be disabled by default. What is your opinion? As sidenote, I think in the future someone could add some global isometric art setting, turning Aseprite into an isometric pixel art editor specifically.

dacap:
Closing this to work in the rebase https://github.com/aseprite/aseprite/pull/5540

PR #4262, opened 2024-01-08
Allow selecting how the user wants to see the opacity/alpha values

### LINKED ISSUE #1544: Intuitive opacity / alpha values in Cel properties and color pickers
Adding alpha effects with the current Opacity bar when double-clicking over a Cel is not very efficient because you have to make an approximation to the desired level of opacity from 0 to 255.

I think that adding a box to change the percentage of the opacity with the keyboard can be very handful and people would be able to obtain a more precise effect on the range between 0% and 100% (more common values for drawing programs).

martincapello:
This PR introduces two new preferences for displaying and entering opacity and alpha values.
With this changes the user can select the range of values used for alpha and opacity values which are:
1. 0%-100% 
2. 0-255

By default Opacity values are expressed as 0%-100%, and Alpha values with 0-255.

Fix #1544

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Aseprite -->
-<!-- Copyright (C) 2018-2023  Igara Studio S.A. -->
+<!-- Copyright (C) 2018-2024  Igara Studio S.A. -->
 <!-- Copyright (C) 2014-2018  David Capello -->
 <preferences>
 
@@ -132,6 +132,10 @@
       <value id="BILINEAR" value="1" />
       <value id="BILINEAR_MIPMAP" value="2" />
       <value id="TRILINEAR_MIPMAP" value="3" />
+    </enum>
+    <enum id="AlphaRange">
+      <value id="EIGHT_BIT" value="0" />
+      <value id="PERCENTAGE" value="1" />
     </enum>
   </types>
 
@@ -430,6 +434,10 @@
       <option id="missing_profile" type="ColorProfileBehavior" default="ColorProfileBehavior::ASSIGN" />
       <option id="window_profile" type="WindowColorProfile" default="WindowColorProfile::MONITOR" />
       <option id="window_profile_name" type="std::string" />
+    </section>
+    <section id="range">
+      <option id="alpha" type="AlphaRange" default="AlphaRange::EIGHT_BIT" />
+      <option id="opacity" type="AlphaRange" default="AlphaRange::PERCENTAGE" />
     </section>
     <section id="canvas_size">
       <option id="trim_outside" type="bool" default="false" />
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -1,5 +1,5 @@
 # Aseprite
-# Copyright (C) 2018-2023  Igara Studio S.A.
+# Copyright (C) 2018-2024  Igara Studio S.A.
 # Copyright (C) 2016-2018  David Capello
 #
 # This work is licensed under the Creative Commons Attribution 4.0
@@ -1469,6 +1469,11 @@ use_embedded_cs = Use embedded profile
 convert_cs = Convert to working RGB space
 assign_cs = Assign working RGB space
 ask_cs = Ask
+alpha_and_opacity = Alpha && Opacity
+alpha_range = Alpha Range:
+opacity_range = Opacity Range:
+8bit_value = 0-255
+percentage = 0%-100%
 available_themes = Available Themes
 extension_themes = Extension Themes
 select_theme = &Select
--- a/src/app/ui/alpha_entry.cpp
+++ b/src/app/ui/alpha_entry.cpp
@@ -1,0 +1,67 @@
+// Aseprite UI Library
+// Copyright (C) 2024  Igara Studio S.A.
+//
+// This file is released under the terms of the MIT license.
+// Read LICENSE.txt for more information.
+
+#ifdef HAVE_CONFIG_H
+#include "config.h"
+#endif
+
+#include "app/pref/preferences.h"
+#include "app/ui/alpha_entry.h"
+#include "base/scoped_value.h"
+#include "gfx/rect.h"
+#include "gfx/region.h"
+#include "os/font.h"
+#include "ui/fit_bounds.h"
+#include "ui/manager.h"
+#include "ui/message.h"
+#include "ui/popup_window.h"
+#include "ui/scale.h"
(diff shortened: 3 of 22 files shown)


data/strings/en.ini
```diff
@@ -1469,6 +1469,11 @@ use_embedded_cs = Use embedded profile
 convert_cs = Convert to working RGB space
 assign_cs = Assign working RGB space
 ask_cs = Ask
+alpha_cs = Alpha && Opacity
+alpha_range_cs = Alpha Range:
+opacity_range_cs = Opacity Range:
+8bit_value_cs = 0-255
+percentage_cs = 0%-100%
```
dacap, line 1476: The `_cs` suffix means "color space", which I think it's not the case for these strings.

data/pref.xml
```diff
@@ -133,6 +133,10 @@
       <value id="BILINEAR_MIPMAP" value="2" />
       <value id="TRILINEAR_MIPMAP" value="3" />
     </enum>
+    <enum id="AlphaRange">
+      <value id="PERCENTAGE" value="0" />
+      <value id="EIGHT_BIT" value="1" />
```
dacap, line 138: This is a minor thing, but I'd switch the values: PERCENTAGE=1 and EIGHT_BIT (or raw) =0 (these values don't care too much, but sometimes in scripts we have to use them as they are, and we can see `0` as the "most basic representation" (raw bytes) and then any following number as special representations).

```suggestion
      <value id="EIGHT_BIT" value="0" />
      <value id="PERCENTAGE" value="1" />
```

dacap (changes requested):
So far it looks great 👍 I wasn't able to test this yet, but I just made some minor comments.

src/app/ui/alpha_entry.cpp
```diff
@@ -0,0 +1,256 @@
...
+#include "ui/theme.h"
+
+#include <algorithm>
+#include <cmath>
+
+namespace app {
+
+using namespace gfx;
+
+AlphaEntry::AlphaEntry(AlphaSlider::Type type)
```
dacap, line 34: I think this class has too much in common with IntEntry, isn't a way to derive IntEntry and avoid duplicating code?

src/app/ui/alpha_slider.cpp
```diff
@@ -0,0 +1,128 @@
...
+      if (max) *max = 100;
+      if (value) *value = std::round(((double)100)*((double)getValue())/((double)getMaxValue()));
+
+      return;
+    case AlphaRange::EIGHT_BIT:
+      return Slider::getSliderThemeInfo(min, max, value);
+  }
+}
+
+bool AlphaSlider::onProcessMessage(ui::Message* msg)
```
dacap, line 35: Some messages handling are similar to the original `Slider::onProcessMessage`, wouldn't be possible to expand Slider (e.g. with some virtual methods that can be override in AlphaSlider) to customize the behavior and avoid duplicating code?

src/app/ui/alpha_slider.h
```diff
@@ -0,0 +1,44 @@
...
+protected:
+  bool onProcessMessage(ui::Message* msg) override;
+
+private:
+  Type m_type;
+  // ui::SliderDelegate impl
+  std::string onGetTextFromValue(int value) override;
+  int onGetValueFromText(const std::string& text) override;
+};
+
```
dacap, line 41: Minor formatting changes can be made (+ moving AlphaSlider ctor to the .cpp file):
```suggestion
class AlphaSlider : public ui::Slider,
                    ui::SliderDelegate
{
public:
  enum Type { ALPHA, OPACITY };

  AlphaSlider(int value, Type type);
      
  void getSliderThemeInfo(int* min, int* max, int* value) const override;
  void updateValue(int value);
  
  app::gen::AlphaRange getAlphaRange() const {
    return (m_type == ALPHA ? Preferences::instance().range.alpha()
                            : Preferences::instance().range.opacity());
  }

protected:
  bool onProcessMessage(ui::Message* msg) override;

private:
  // ui::SliderDelegate impl
  std::string onGetTextFromValue(int value) override;
  int onGetValueFromText(const std::string& text) override;

  Type m_type;
};

```
Generally we want to put data members as the last items in the class definition (doesn't modify the structure of the class, but just for readability we want to put all member functions before data).

src/app/ui/alpha_slider.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include <algorithm>
+
+using namespace ui;
+using namespace app::gen;
+
+namespace app {
+
+void AlphaSlider::getSliderThemeInfo(int* min, int* max, int* value) const
+{
+  switch(getAlphaRange()) {
```
dacap, line 23:
```suggestion
  switch (getAlphaRange()) {
```

dacap (changes requested):
Some changes required.

Probably we could connect preference changes from the same `AlphaSlider` ctor (e.g. so they are controlled automatically e.g. on ColorPopup/ColorSliders):

alpha-slider

martincapello:
> Probably we could connect preference changes from the same AlphaSlider ctor (e.g. so they are controlled automatically e.g. on ColorPopup/ColorSliders):

Actually it seems that it is not that simple. The AlphaSlider doesn't contain the AlphaEntry, is the other way around. So I cannot update an entry from an AlphaSlider. In addition to this, the ColorSliders use a ColorEntry instance, which doesn't extend AlphaEntry at all. The ColorSliders just use the AlphaSlider, then to update the ColorEntry when the preference is changed, we should:
1. Connect to the preference change from the AlphaSlider ctor, then fire another signal from AlphaSlider to indicate this.
2. Connect to the AlphaSlider signal from the ColorSliders to be able to update the ColorEntry.

This actually doesn't seem practical...it actually involves more coding than connecting the ColorSliders directly to the preference change.

Maybe I'm missing something and there is an alternative that I don't see, please let me know if you have something else in mind.

dacap:
> Actually it seems that it is not that simple.

Yeah, it's more complex from what I thought, we can leave this as it is because the change of the preference is not something quite common (probably done just one time and then never more).

I'll test the PR a little more and then merge it as it looks great so far 👍

dacap:
It looks like dragging with the right mouse button on an AlphaSlider configured with % is not working correctly:

https://github.com/aseprite/aseprite/assets/39654/6ece43d0-6358-4e84-9d6f-4f279874064d

martincapello:
> It looks like dragging with the right mouse button on an AlphaSlider configured with % is not working correctly:

I just pushed some changes to slider.cpp to fix this.

dacap:
Nice work! I've just merged it 👍

PR #4264, opened 2024-01-10
Fix wrong edges on Brush Preview on Tilemap Mode (fix #4176)

### LINKED ISSUE #4176: Wrong edges on Brush Preview on Tilemap Mode
When 'Edges' are active on 'Brush Preview' + 'Tilemap Mode', the edges of the regular brush is displayed instead the bounds of the tile.

* Aseprite version: 1.3

Gasparoken:
fix #4176

### DIFF
--- a/src/app/ui/editor/brush_preview.cpp
+++ b/src/app/ui/editor/brush_preview.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2023  Igara Studio S.A.
+// Copyright (C) 2019-2024  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -17,7 +17,6 @@
 #include "app/color.h"
 #include "app/color_utils.h"
 #include "app/doc.h"
-#include "app/site.h"
 #include "app/snap_to_grid.h"
 #include "app/tools/controller.h"
 #include "app/tools/ink.h"
@@ -25,7 +24,6 @@
 #include "app/tools/point_shape.h"
 #include "app/tools/tool.h"
 #include "app/tools/tool_loop.h"
-#include "app/ui/color_bar.h"
 #include "app/ui/context_bar.h"
 #include "app/ui/editor/editor.h"
 #include "app/ui/editor/tool_loop_impl.h"
@@ -165,11 +163,14 @@ void BrushPreview::show(const gfx::Point& screenPos)
 
   tools::Tool* tool = m_editor->getCurrentEditorTool();
   const bool isFloodfill = tool->getPointShape(0)->isFloodFill();
-  // TODO add support for "tile-brushes"
+
+  // Get current tilemap mode
+  const TilemapMode tilemapMode = site.tilemapMode();
+
   gfx::Rect origBrushBounds =
     ((isFloodfill && brush->type() != BrushType::kImageBrushType) ||
-     site.tilemapMode() == TilemapMode::Tiles ? gfx::Rect(0, 0, 1, 1)
-                                              : brush->bounds());
+     tilemapMode == TilemapMode::Tiles ? gfx::Rect(0, 0, 1, 1)
+                                       : brush->bounds());
   gfx::Rect brushBounds = origBrushBounds;
 
   // Cursor in the screen (view)
@@ -187,9 +188,6 @@ void BrushPreview::show(const gfx::Point& screenPos)
 
   // Get the current tool
   tools::Ink* ink = m_editor->getCurrentEditorInk();
-
-  // Get current tilemap mode
-  TilemapMode tilemapMode = ColorBar::instance()->tilemapMode();
 
   // Setup the cursor type depending on several factors (current tool,
   // foreground color, layer transparency, brush size, etc.).
@@ -267,6 +265,12 @@ void BrushPreview::show(const gfx::Point& screenPos)
       break;
   }
 
+  // For tiles as we show the edges of the tile, we add the crosshair
+  // in the mouse position as a helper.
+  if (m_type & BRUSH_BOUNDARIES &&
+      tilemapMode == TilemapMode::Tiles)
+    m_type |= CROSSHAIR;
+
   if (m_type & SELECTION_CROSSHAIR)
     showPreview = false;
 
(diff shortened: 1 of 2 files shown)


martincapello (approved):
Have been testing this for a bit and seems to work fine!

Gasparoken:
The only thing I'm not sure about is when dynamics are enabled:
There are several options for this:
1. Always show the brush edges that correspond to the maximum size (like the current solution).
2. Don't show brush edges when Dynamics are enabled.
3. Show the brush edges as fast as the brush display changes due to Dynamics (I think this would have no real practical use and may introduce render lags because of this).
4. The brush edges corresponding to the maximum size are displayed while the brush is not drawing, then the brush edges disappear during the drawing state.
5. Other?

Personally, I'd like option 1 or 2.
What do you think @martincapello?

src/app/ui/editor/brush_preview.cpp
```diff
@@ -262,6 +260,8 @@ void BrushPreview::show(const gfx::Point& screenPos)
           if (showPreview)
             showPreviewWithEdges = true;
           m_type = BRUSH_BOUNDARIES;
+          if (tilemapMode == TilemapMode::Tiles)
+            m_type |= CROSSHAIR;
```
dacap, line 264: This should be valid for `app::gen::BrushPreview::EDGES` case too to avoid this bug:

edges-only

My recommended change, remove these lines:

```suggestion
```

And add after the `switch`:

```c++
  // For tiles as we show the edges of the tile, we add the crosshair
  // in the mouse position as a helper.
  if (m_type & BRUSH_BOUNDARIES &&
      tilemapMode == TilemapMode::Tiles) {
    m_type |= CROSSHAIR;
  }
```

src/app/ui/editor/brush_preview.cpp
```diff
@@ -165,11 +163,14 @@ void BrushPreview::show(const gfx::Point& screenPos)
 
   tools::Tool* tool = m_editor->getCurrentEditorTool();
   const bool isFloodfill = tool->getPointShape(0)->isFloodFill();
-  // TODO add support for "tile-brushes"
+
+  // Get current tilemap mode
+  TilemapMode tilemapMode = site.tilemapMode();
```
dacap, line 168: A little nice change is to put this as `const`:
```suggestion
  const TilemapMode tilemapMode = site.tilemapMode();
```
This indicates that the `tilemapMode` is used just like a constant value/an alias of `site.tilemapMode()`. (There are a lot of missing `const` in the code, so there is no problem if this is added or not).

src/app/ui/editor/brush_preview.cpp
```diff
@@ -512,13 +515,37 @@ void BrushPreview::invalidateRegion(const gfx::Region& region)
...
+                                                 const gfx::Point& spritePos)
+{
+  m_brushBoundaries.offset(-m_brushBoundaries.begin()[0].bounds().x,
+                           -m_brushBoundaries.begin()[0].bounds().y);
+  gfx::Point canvasPos = grid.tileToCanvas(grid.canvasToTile(spritePos));
+  m_brushBoundaries.offset(canvasPos.x - spritePos.x,
+                           canvasPos.y - spritePos.y);
+}
+
+void BrushPreview::generateBoundaries(Site &site, gfx::Point spritePos)
```
dacap, line 528: To avoid a copy of the `Site` we can use a `const Site&`:
```suggestion
void BrushPreview::generateBoundaries(const Site& site,
                                      const gfx::Point& spritePos)
```

src/app/ui/editor/brush_preview.h
```diff
@@ -10,6 +10,7 @@
 #pragma once
 
 #include "app/extra_cel.h"
+#include "app/site.h"
```
dacap, line 13: Not required as we are not using the `Site` value/size:
```suggestion
```
You can add a forward declaration below after `  class Editor;` line:
```c++
namespace app {
  class Editor;
  class Site;  // <--

  class BrushPreview ...
```

src/app/ui/editor/brush_preview.h
```diff
@@ -90,7 +91,12 @@ namespace app {
     doc::BrushRef getCurrentBrush();
     static doc::color_t getBrushColor(doc::Sprite* sprite, doc::Layer* layer);
 
-    void generateBoundaries();
+    // Offset re-calculation of brush boundaries only for TilemapMode::Tiles.
+    // Used within 'generateBoundaries' function.
+    void calculateTileBoundariesOrigin(const doc::Grid& grid,
+                                       const gfx::Point& spritePos);
+
+    void generateBoundaries(Site &site, gfx::Point spritePos);
```
dacap, line 99:
```suggestion
    void generateBoundaries(const Site& site,
                            const gfx::Point& spritePos);
```

src/app/ui/editor/brush_preview.h
```diff
@@ -141,6 +147,9 @@ namespace app {
     gfx::Rect m_lastBounds;
     doc::frame_t m_lastFrame;
 
+    doc::Layer* m_lastLayer;
```
dacap, line 150:
```suggestion
    doc::Layer* m_lastLayer = nullptr;
```

dacap (changes requested):
It works great 👍 there are just some minor changes required.

src/app/ui/editor/brush_preview.cpp
Gasparoken: Thanks! You added `const` to lock the values of the `Site` object inside the `generateBoundaries` function.
And you corrected `Site &site` as `Site& site` (I used to think `Site &site` is the same as `Site& site` @dacap could you confirm this?).

dacap: Sorry my first comment about "to avoid a copy" was wrong for `Site` (it was for `gfx::Point`) as originally it was `Site& site`, not `Site site`. (When we use value parameters like `Site site` is when a copy is made, I read it wrong when wrote that comment.)

About `Site &site` vs `Site& site` there is no difference, it's just the formatting we are using and we have to keep it consistent.

Gasparoken: Oh! `gfx::Point`, I omitted to modify it. I pushed a new change to include `const gfx::Point&`.
Thanks for the clarification!

PR #4266, opened 2024-01-14
Fix Improve UX to recover files (fix #4130)

### LINKED ISSUE #4130: Improve UX to recover files
Sometimes we receive mails about how to recover a session file with the session folder compressed, and they follow this guide: https://www.aseprite.org/docs/data-recovery/

I'm not sure why the data recovery fails, if it's because the UI is confusing or if there is another kind of problem, because after decompressing the folder and giving a try to the same steps, we (generally) can recover the files from the same UI (without further assistance, something they could have done in the first place).

Even more, in some extreme case we could offer the way to get the compressed session folder from the same UI.

Gasparoken:
To implement this fix is needed https://github.com/aseprite/laf/pull/75 first

I think this solution can be improved on Windows since each session 'pid' is checked by iteratively searching all running processes in the operating system.

The best thing to do would be to create an array of 'pid' numbers belonging to running processes whose process name is 'aseprite.exe', then check to see if the 'pids' of each session match any element of the 'pid' array.

fix #4130

### DIFF
--- a/src/app/crash/session.cpp
+++ b/src/app/crash/session.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2023  Igara Studio S.A.
+// Copyright (C) 2019-2024  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -127,10 +127,8 @@ const Session::Backups& Session::backups()
 bool Session::isRunning()
 {
   loadPid();
-  if (m_pid)
-    return base::is_process_running(m_pid);
-  else
-    return false;
+  return base::get_process_name(m_pid) ==
+    base::get_process_name(base::get_current_process_id());
 }
 
 bool Session::isCrashedSession()


src/app/crash/session.cpp
```diff
@@ -127,8 +127,12 @@ const Session::Backups& Session::backups()
 bool Session::isRunning()
 {
   loadPid();
+  std::string processName = "aseprite";
+  #if LAF_WINDOWS
+    processName = processName + ".exe";
+  #endif
   if (m_pid)
-    return base::is_process_running(m_pid);
+    return base::is_process_running(m_pid, processName.c_str());
```
martincapello, line 135: I saw the changes in laf for supporting this and they look good. I'm just wondering if we should get the current process name instead of using a hardcoded string?

Gasparoken: Yes! I agree with you.

Gasparoken:
Tested on macOS and Windows. Ready for review.

Issue #2680, opened 2021-03-25
Selection move of the hidden layer


magickbirb:
Selected area of the hidden layer is moving if I'm moving all layers in one particular frame. The only way to prevent it is to lock the hidden layers, but it'd be much easier if hiding it was enough
I asked on Discord and they said it's a bug, though I thought it was meant this way
Anyway I hope it'll be fixed soon, or at least an option will be added to chose if I want hidden layers to be moved

Aseprite v1.2.25
Windows 10 x64

dacap:
Thanks for reporting this @magickbirb, it will be fixed in the next release v1.2.26 ASAP

PR #4282, opened 2024-01-22
Drop selection if we hide the layer (fix #4179)

### LINKED ISSUE #4179: Drop selection if we hide the layer
If we move the selection (or just click it) and then hide the layer, we should drop the selection, because the MovingPixelsState is still running and if we change layer we'll cut the original selection (which is confusing because we didn't cut the selection, just hide the layer).

Reported here: https://community.aseprite.org/t/copy-paste-deletes-areas-in-hidden-layer/20621

Gasparoken:
fix #4179

### DIFF
--- a/src/app/ui/editor/moving_pixels_state.cpp
+++ b/src/app/ui/editor/moving_pixels_state.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2023  Igara Studio S.A.
+// Copyright (C) 2019-2024  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -754,6 +754,14 @@ void MovingPixelsState::onBeforeRangeChanged(Timeline* timeline)
   if (!isActiveDocument())
     return;
 
+  if (m_pixelsMovement)
+    dropPixels();
+}
+
+void MovingPixelsState::onBeforeLayerHiding(Timeline* timeline)
+{
+  if (!isActiveDocument())
+    return;
   if (m_pixelsMovement)
     dropPixels();
 }
--- a/src/app/ui/editor/moving_pixels_state.h
+++ b/src/app/ui/editor/moving_pixels_state.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2022  Igara Studio S.A.
+// Copyright (C) 2019-2024  Igara Studio S.A.
 // Copyright (C) 2001-2017  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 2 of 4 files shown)


martincapello:
I've been testing this and it fixes what the issue describes. But I think there is still a case where it is still a confusing behavior.
Let's suppose we have a sprite with 2 regular layers, with just 1 frame.
This is what it fixes (as far as I understood): 
1. Select something in "layer 1".
2. Click or move the selection.
3. Hide "layer 1". (with this fix drops the selection)
4. Select "layer 2".
5. Unhide "layer 1". Here, before this fix it removed the selected pixels in step 1, now it doesn't.

I found this other case:
1. Select something in "layer 1".
2. Hide "layer 1". (nothing to drop, we didn't click or move it yet)
3. Click or move the selection.
4. Select "layer 2".
5. Unhide "layer 1". Here you will see the selected pixels in step 1 removed.

Gasparoken:
Yeah! You're right. I'll fix it soon.

Gasparoken:
Well, I'm looking at this in more depth and conclude that we need to agree on a standard behavior about "hidden layers":
- Currently, it is allowed to "transform" or "shift pixels" on the current hidden layer
- On the other hand, the use of flip/filter/tools is not allowed.
- Multiple layer transformation isn't allowed.

IMO: hidden layer images should not be modified, which is logical. However, I think it can be useful for the artist to have non-visible layers also respond to modifications. Perhaps the option could be added as a global property in preferences to switch "Mods allowed in hidden layers".

In any case, I'll propose a solution for the present problem without modifying too much.

Gasparoken:
Ready for a new review. It's a different and simpler approach @martincapello.

martincapello:
I've tested the first case I mentioned in https://github.com/aseprite/aseprite/pull/4282#issuecomment-1927261252 but now, after step 5, it shows missing pixels on the bottom border and right border of the selection.
On the other hand, the second case works as the first, but has the same issue with the missing pixels.

Gasparoken:
Ready for review.

dacap:
I think I'll merge this as it is but add an extra change to avoid modifying hidden layers, that shouldn't never happen.

Probably this change https://github.com/aseprite/aseprite/compare/afbd4c8b251864215f14b0da97708d13119e62ea..6b7f08d2ba659df24fd115c6e29470918aba60e6 would not be needed after that, so maybe I'll be reverting that, but it needs some test.

dacap:
I think I'll revert this PR. Sorry.

dacap:
I'll try to fix this issue in other way because the merged PR added a regression that was fixed before #2680 in f3ab779bfded7710934cff6b9cc6b596756eccbb

PR #4294, opened 2024-02-03
Add sprite user data when exporting documents

skypjack:

---

As discussed [here](https://discord.com/channels/324979738533822464/348146938530496513/1203072349398634547) on discord, sprite user data are not exported when creating the json file.
This PR adds them to the `"meta"` section, as in:
```json
 "meta": {
  "app": "https://www.aseprite.org/",
  "version": "1.x-dev",
  "image": "test.png",
  "format": "RGBA8888",
  "size": { "w": 64, "h": 64 },
  "scale": "1",
  "documents": [
   { "name": "test.aseprite", "data": "my data" }
  ],
  "frameTags": [ ... ],
  "layers": [ ... ],
  "slices": [ ... ]
 }
```
I've no idea if this is a _correct_ result for the spirit of this software but I had to start somewhere, right? 🙂 
Let me know if this doesn't meet your tastes and what I can do to have this data in the final json.
I find this data to be pointless otherwise since one cannot export and thus use them properly.

### DIFF
--- a/src/app/doc_exporter.cpp
+++ b/src/app/doc_exporter.cpp
@@ -1434,6 +1434,31 @@ void DocExporter::createDataFile(const Samples& samples,
      << "\"h\": " << texture->height() << " },\n"
      << "  \"scale\": \"1\"";
 
+  {
+    bool firstDoc = true;
+  
+    std::set<doc::ObjectId> includedSprites;
+
+    for (auto& item : m_documents) {
+      if (firstDoc)
+        firstDoc = false;
+      else
+        os << ",";
+  
+      Doc* doc = item.doc;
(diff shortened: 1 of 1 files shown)


dacap:
Hi @skypjack! thanks for your contribution, I have to think a little more about the output format. For a one sprite -> sprite sheet, it looks like a "data" field instead of the "documents" might be preferable.

For multiple sprites -> texture atlas, at the moment we have the same issues with tags/layers/slices info from those multiple sprites/docs. Not sure what would be the best solution in these cases.

skypjack:
Let me know if you want me to change anything in the PR. 👍

PR #4325, opened 2024-02-21
Fix crash when trying to access a property of a Style which is nullptr (fix #4015)

### LINKED ISSUE #4015: Widget style can be nullptr in ui::Theme::calcMinSize()
Sentry Issue: [ASEPRITE-1CT](https://igara-studio.sentry.io/issues/4415719627/?referrer=github_integration)

```
EXCEPTION_ACCESS_VIOLATION_READ / 0x6c: Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0x6c
  File "theme.cpp", line 689, in ui::Theme::calcMinSize
    sz.w = style->minSize().w;
  File "widget.cpp", line 206, in ui::Widget::setStyle
    m_minSize = m_theme->calcMinSize(this, style);
  File "grid.cpp", line 128, in ui::Grid::setStyle
    Widget::setStyle(style);
  File "button_set.cpp", line 205, in app::ButtonSet::{ctor}::__l2::<lambda>::operator()
    setStyle(SkinTheme::instance()->styles.buttonset());
  File "type_traits", line 1576, in std::invoke
    return static_cast<_Callable&&>(_Obj)();
...
(17 additional frame(s) were not displayed)
```

Gasparoken:
The original description of the issue which this solution applies to is:
`Widget style can be nullptr in ui::Theme::calcMinSize() #4015`

Pending alerting the artist.

fix #4015


src/ui/theme.h
```diff
@@ -56,6 +56,8 @@ namespace ui {
     static int getStyleFlagsForWidget(const Widget* widget);
   };
 
+  static ui::Style m_defaultStyle(nullptr);
```
dacap, line 59: Move this global variable inside `Theme` class, making this a static member variable of `Theme`.

As it's now there will be multiple `m_defaultStyle` variables (one for each `#include "ui/theme.h"` we do).

src/ui/widget.cpp
```diff
@@ -199,14 +199,23 @@ void Widget::setTheme(Theme* theme)
...
   assert_ui_thread();
-
-  m_style = style;
-  m_border = m_theme->calcBorder(this, style);
-  m_bgColor = m_theme->calcBgColor(this, style);
-  m_minSize = m_theme->calcMinSize(this, style);
-  m_maxSize = m_theme->calcMaxSize(this, style);
-  if (style->font())
-    m_font = AddRef(style->font());
+  if (style) {
```
dacap, line 202: Same as in the grid case, I think this should never happen, we could do:
```
  ASSERT(style);
  if (!style)
    style = Theme::getDefaultStyle();
  ...
```
The default style should already have some default style constructed in some way so `m_theme->calc...` calls return the default values.

dacap (changes requested):

Gasparoken:
Ready for review.

src/ui/theme.h
```diff
@@ -165,6 +170,7 @@ namespace ui {
                            gfx::Size& sizeHint,
                            gfx::Border& borderHint,
                            gfx::Rect& textHint, int& textAlign);
+    static ui::Style* m_defaultStyle;
```
dacap, line 173: Use a unique pointer so it's deleted automatically:
```suggestion
    static std::unique_ptr<ui::Style> m_defaultStyle;
```

dacap (changes requested):

src/ui/theme.h
Gasparoken: Done.

dacap:
Can I review/take this one? I have a possible change to make without using pointers.

Gasparoken:
> Can I review/take this one? I have a possible change to make without using pointers.

Yes, OK.

dacap:
Merged in f22603caea379f0247e5596f4f0edd111594f00a

PR #4335, opened 2024-02-23
Fix Ignore any third-party theme that uses "default" as ID (fix #4226)

### LINKED ISSUE #4226: Ignore any third-party theme that uses "default" as ID
From ticket: https://igarastudio.zendesk.com/agent/tickets/5545

We must limit/ignore any theme extension that wants to replace the default theme and is not the official one installed in the data directory (this might be harder to detect in portable versions of Aseprite).

Gasparoken:
Pending alert artist when a theme with default ID is omitted.

fix #4226


Gasparoken:
Hello @martincapello. This solution hides any "aseprite-theme" themes from the `Extensions` and `Theme` lists.

src/app/commands/cmd_options.cpp
```diff
@@ -1766,6 +1770,20 @@ class OptionsWindow : public app::gen::Options {
...
 
+  // Function to determine if the input extension is the default theme
+  static bool isExtensionADuplicatedDefaultTheme(const Extension* e) {
+    if (!e->isDefaultTheme())
+      return false;
+    auto userThemePaths =
+      themeFolders({"extensions", skin::SkinTheme::kThemesFolderName});
+    for (auto& p : userThemePaths) {
+      // Has the user path (p) the same path of the extension (e->path())?
+      if (std::strncmp(e->path().c_str(), p.c_str(), p.size()) == 0)
```
martincapello, line 1781: Since e->path() and p are std::strings, this could be replaced by:
```c++
      if (e->path() == p)
```

src/app/commands/cmd_options.cpp
```diff
@@ -1741,9 +1744,10 @@ class OptionsWindow : public app::gen::Options {
     return base::normalize_path(rf.defaultFilename());
   }
 
-  static base::paths themeFolders() {
+  static base::paths themeFolders(const std::vector<std::string> filenames) {
```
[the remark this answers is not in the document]
dacap: There is `base::paths` for this kind of type (list of paths).

src/app/commands/cmd_options.cpp
Gasparoken: Unfortunately that expression does not work, nor does the following one:
```cpp
if (e->path().compare(p) == 0)
```
I'll return with
```cpp
if (std::strncmp(e->path().c_str(), p.c_str(), p.size()) == 0)
```

martincapello: Sorry, you are right, I didn't see that you where comparing up to p.size() characters and not the full strings.

In this case, the right one would be:
```c++
if ( e->path().compare(0, p.size(), p) == 0)
```
But I don't think it makes any difference...so I'm fine if you leave the std::strncmp() call as is.

martincapello (approved):

src/app/commands/cmd_options.cpp
```diff
@@ -1756,6 +1760,17 @@ class OptionsWindow : public app::gen::Options {
     return paths;
   }
 
+  static base::paths createUserDirPaths(const base::paths dirNames) {
```
dacap, line 1763: I'd change the name of this function to something like
```suggestion
  static base::paths getUserDirPaths(const base::paths dirNames) {
```
because when I saw it for first time I thought it was creating (mkdir) the directories.

dacap:
I've just tested with a simple [dacap-theme.aseprite-extension]([dacap-theme.zip](https://github.com/aseprite/aseprite/files/14981357/dacap-theme.zip)) (with the same content of the default aseprite-theme) and it allows me to add the extension:

bug2

After reopening the Preferences dialog the installed extensions is correctly not displayed anymore. But the extension keeps installed in the user directory (`c:/Users/David/AppData/Roaming/Aseprite/extensions/aseprite-theme`).

I think we should show an error message and not install/uncompress the extension in the user directory.

Gasparoken:
Ready for review.

src/app/extensions.h
```diff
@@ -21,6 +21,10 @@ namespace ui {
   class Widget;
 }
 
+namespace {
+  const char* kAsepriteDefaultThemeExtensionName = "aseprite-theme";
+}
```
dacap, line 26: Please don't use anonymous namespaces in header files. Move this member inside `Extension` class if needed.

dacap (changes requested):
I've found that this PR doesn't fix the issue because the default theme has the ID "default" (not the extension name "aseprite-theme"). 

The default theme styles/parts are loaded from the "default" theme:
https://github.com/aseprite/aseprite/blob/611713e5fcefd26b289a7e339730994dc8e23299/src/app/ui/skin/skin_theme.cpp#L319
where `pref.theme.selected.defaultValue()` is [default](https://github.com/aseprite/aseprite/blob/611713e5fcefd26b289a7e339730994dc8e23299/data/pref.xml#L227)

I think the confusion came from `Extension::isDefaultTheme()` member that is used to check if the extension is the default Aseprite theme **extension** (not the default ID which is "default", as the first commit of this PR specifies and should check).

Here a test extension (install it, restart, Aseprite doesn't start anymore): 
[invalid-theme.zip](https://github.com/aseprite/aseprite/files/15468789/invalid-theme.zip)

dacap:
Check the bot comments

dacap:
This PR has 5 commits (?)

dacap:
@Gasparoken see if you can fix the PR with just one commit rebasing it to the last commit of the main branch.

dacap:
@Gasparoken sorry I think I got confused, but probably I can squash all the commits from my side in one commit and do the rebase, don't worry, I'll try to fix it.

dacap:
@Gasparoken just reset your branch to the commit I pushed, your branch should match (same commit) that the one I pushed.

Gasparoken:
Ready for review.

data/strings/en.ini
```diff
@@ -62,6 +62,7 @@ deleting_tilemaps_will_delete_tilesets = Warning\n<<Deleting the following tilem
 cannot_file_overwrite_on_export = Overwrite Warning\n<<You cannot Export with the same name (overwrite the original file).\n<<Use "File > Save As" menu option in that case.\n||&OK
 cannot_open_file = Problem<<Cannot open file:<<{0}||&OK
 cannot_open_folder = Problem<<Cannot open folder:<<{0}||&OK
+cannot_open_theme = Error<<The selected theme ID is "default". Please change<<the theme ID inside "package.json" to make it work.||&OK
```
dacap, line 65: Is this message needed? Because it looks like in the place where `cannot_open_theme` is used it should be `cannot_install_default_extension` instead (e.g. when we use "Add Extension" we could show `cannot_install_default_extension`).

data/strings/en.ini
```diff
@@ -90,6 +91,7 @@ job_working = {0}<<Working...||&Cancel
 nothing_to_report = Crash Report<<Nothing to report||&OK
 install_extension = Warning\n<<Do you really want to install the given extension?\n<<"{0}"\n||&Install||&Cancel
 uninstall_extension_warning = Warning\n<<Do you really want to uninstall "{0}" extension?\n||&Yes||&No
+cannot_install_default_extension = Error\n<<This theme cannot be installed because its name matches the default aseprite theme\n||&OK
```
dacap, line 94: I'd change the message to something like:
```suggestion
cannot_install_default_extension = Error<<This extension cannot be installed because tries to replace the default theme.<<<<Contact the developer so they fix the theme ID string from \"default\" to something else.||&OK
```

src/app/commands/cmd_options.cpp
```diff
@@ -926,6 +926,18 @@ class OptionsWindow : public app::gen::Options {
       }
     }
 
+    // Get the extension information from the compressed
+    // package.json file.
+    const ExtensionInfo info =
+      App::instance()->extensions().getCompressedExtensionInfo(filename);
+    // Check if the filename corresponds to aseprite-default theme
+    if (base::string_to_lower(info.name) ==
+        Extensions::kAsepriteDefaultThemeName) {
```
dacap, line 935: Probably ExtensionInfo needs a `bool` property like `defaultTheme` instead of the `themeId` (below more details).

src/app/commands/cmd_options.cpp
```diff
@@ -926,6 +926,18 @@ class OptionsWindow : public app::gen::Options {
...
+      ui::Alert::show(
+        fmt::format(Strings::alerts_cannot_install_default_extension()));
```
dacap, line 937: No need for `fmt::format()`:
```suggestion
        Strings::alerts_cannot_install_default_extension());
```

src/app/extensions.cpp
```diff
@@ -51,12 +51,13 @@
 
 namespace app {
 
+const char* Extensions::kAsepriteDefaultThemeName = "aseprite-theme";
+
```
dacap, line 55: No need of this change if the `ExtensionInfo` includes a bool to know if it overrides the default theme.

src/app/extensions.cpp
```diff
@@ -283,6 +284,8 @@ void Extension::addTheme(const std::string& id,
                          const std::string& path,
                          const std::string& variant)
 {
+  if (id == "default" && !isDefaultTheme())
+    return;
```
dacap, line 288: Add some comment and extra constant (should we add an `ASSERT()` too?):
```suggestion
  // Avoid adding a theme with "default" ID if it's not
  // the default aseprite-theme extension.
  if (id == kAsepriteDefaultThemeName && !isDefaultTheme())
    return;
```
So we have `kAsepriteDefaultThemeExtensionName = "aseprite-theme"` and a new `kAsepriteDefaultThemeName = "default"`.

src/app/extensions.cpp
```diff
@@ -519,7 +522,7 @@ bool Extension::isCurrentTheme() const
 
 bool Extension::isDefaultTheme() const
 {
-  return (name() == kAsepriteDefaultThemeExtensionName);
+  return name() == Extensions::kAsepriteDefaultThemeName;
```
dacap, line 525: Rollback this change.

src/app/extensions.cpp
```diff
@@ -1003,6 +1006,11 @@ ExtensionInfo Extensions::getCompressedExtensionInfo(const std::string& zipFn)
     std::string err;
     auto json = json11::Json::parse(out.str(), err);
     if (err.empty()) {
+      if (json["contributes"].is_object()) {
+        auto themes = json["contributes"]["themes"];
+        if (themes.is_array() && themes[0].is_object())
+          info.themeId = themes[0]["id"].string_value();
```
dacap, line 1012: Check all themes, one extension can have multiple items in the "themes" array (and the second one could be called `"default"`):
```suggestion
        if (themes.is_array()) {
          for (...) {
            if (themes[i]["id"].string_value() == kAsepriteDefaultThemeName) {
              info.defaultTheme = true;
              break;
            }
          }
        }
```

src/app/extensions.h
```diff
@@ -54,6 +55,8 @@ namespace app {
       Max
     };
 
+    bool isDefaultTheme() const;
+
```
dacap, line 59: I think this shouldn't be needed with the new field in ExtensionInfo.

src/app/extensions.h
```diff
@@ -30,6 +30,7 @@ namespace app {
   class Extensions;
 
   struct ExtensionInfo {
+    std::string themeId;
     std::string name;
     std::string version;
     std::string dstPath;
```
dacap, line 36: Use a bool just to indicate that this extension info contains one theme that tries to overwrite the default one:
```suggestion
    std::string name;
    std::string version;
    std::string dstPath;
    bool defaultTheme = false;
```

dacap (changes requested):
Some changes to check all the information from `packages.json` + minor fixes.

src/app/extensions.h
Gasparoken: I think it is still necessary, since this function is used to identify `Extension` instances (not just `ExtensionInfo` instances).

Gasparoken:
I will squash the commits after the changes are approved.

dacap:
@Gasparoken: Using the changes in this PR, give a try to the following extension and tell me if you notice something strange when you try to install it: 
[override-aseprite-theme.zip](https://github.com/user-attachments/files/16396247/override-aseprite-theme.zip)

Gasparoken:
> @Gasparoken: Using the changes in this PR, give a try to the following extension and tell me if you notice something strange when you try to install it:

Yes, I noticed that override-aseprite-theme.zip tries to override the aseprite theme...

dacap:
Thanks @Gasparoken for the fixes, I've just merge this PR squashing it.

PR #4339, opened 2024-02-24
Enable websocket SSL/TLS support

samedii:
Just creating this to show that it should be relatively easy to support. Will close it. I expect there may be more considerations of course.

Compiled with this setup: https://github.com/nilsve/docker-aseprite-linux

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -767,6 +767,8 @@ if(ENABLE_SCRIPTING)
 
   if(ENABLE_WEBSOCKET)
     target_link_libraries(app-lib ixwebsocket)
+
+    target_compile_definitions(app-lib PUBLIC USE_TLS=1)
   endif()
 endif()
 


dacap:
Adding this flag makes IXWebSocket depends on OpenSSL or mbedtls (macOS is the only one that use its native securetranport framework), and at the moment we're not distributing these libraries. Anyway we want to make curl work on Linux (so we can enable Sentry on Linux too) and add possible HTTPS requests from Lua scripts. So adding mbedtls is an option here. Something to investigate a little more (as the Linux port needs the distribution of certificates/etc.).

samedii:
Yes but maybe it's nicer for end users if you link it statically instead of depending on shared. I don't think linux needs certificates, I think that's only for the server side implementation.

Nice to hear that you're open to something like this :+1: 

_I did a quick test doing static linking for everything but got stuck on libGL if I remember correctly._

_Note that only allowing insecure websockets is problematic sometimes because some providers seem to force upgrade to https._

Issue #4338, opened 2024-02-24
View the points of multiple colours in the colour picker.


Charlie-83:
When I select multiple colours from my colour palette, I would like for multiple points to appear in the colour picker. This would allow me to see where the selected colours lie in respect to each other. I generally like to have my colours tend towards blue as they get darker, this feature would let me easily see that this is the case. Currently, I move my mouse quickly back and forth across my colours to visualise this.

Charlie-83:
I would happily implement this myself (to be merged into master or not) if someone could give me some pointers

PR #4340, opened 2024-02-25
Multiple spectrum indicators 4338

Charlie-83:
#4338 

image

This is super rough right now but opening PR to gauge interest in this feature.

TODO:

- [ ] Multiple indicators for bottom bars
- [ ] Multiple indicators for other colour selectors
- [ ] Probably want to store the shade as `m_shade` on the colour selector

### DIFF
--- a/src/app/ui/color_spectrum.cpp
+++ b/src/app/ui/color_spectrum.cpp
@@ -14,6 +14,7 @@
 #include "app/color_utils.h"
 #include "app/ui/skin/skin_theme.h"
 #include "app/ui/status_bar.h"
+#include "app/ui/color_bar.h"
 #include "app/util/shader_helpers.h"
 #include "os/surface.h"
 #include "ui/graphics.h"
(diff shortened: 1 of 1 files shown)


dacap:
Thanks for this PR @Charlie-83, I've been testing this, and it looks a little confusing seeing all the indicators at the same time (with the same color).

Not sure if it's useful (probably it could be a switchable option, off by default), Probably seeing the indicators with a some alpha blending (50%) might be useful to differentiate between the current indicator and the other colors indicators, or avoid showing indicators for all colors and just show a dotted line connecting all points (and show the indicator only in the current color).

Charlie-83:
Thanks for having a look @dacap I really like the idea of showing one indicator and then a dotted line for the rest since the main point of the feature is to be able to visualize color curves.

This is currently just a rough hack. I'll find some time to work on it some more.

Charlie-83:
image
Here is what I have it looking like now

dacap:
I was doing some testing and I think we are getting closer. I'm not 100% convinced to enable this by default, but can be useful for some users.

Some issues I've found:

1) Probably it would make sense to draw a line between the near hue distance:

Screenshot from 2024-03-12 16-47-51

2) Sometimes the line is not visible

Screenshot from 2024-03-12 16-47-29

I didn't review the code yet.

Charlie-83:
I'll fix those issues. I agree that this is probably best as an option in the settings default off so I'll work on adding a setting for this too

Issue #2144, opened 2019-09-04
Enhancements to transformation of multiple cels


dacap:
The implementation of #1172 is going to be released in v1.2.14, but some missing important features:

- [ ] Preview of multi-layer transformation (at the moment only the active layer is previewed)
- [ ] Some commands like Ctrl+Shift+J could work with multiple frames now (e.g. creating a new layers cutting the selection of several frames)
- [ ] Paste on multiple-frames
- [ ] More?

m-ender:
Re "More?", it seems that the Shift command (i.e. cyclic translation of the selection) doesn't work on multiple cells yet. Would be really useful for longer sections where something moves a fixed distance at a time. (E.g. if I have a texture that I want to move one pixel a frame for 32 frames, I could just create 16 frames, the duplicate them, and then shift all of the copied frames by another 16 pixels... and then of course break that further down, so I only need five steps in this process rather than 32.)

lampysprites:
Regarding transforming the selection in multiple cels (which i think is related to all this?), I also suggest to maybe use a different cursor depending on whether a single cel, or multiple are going to be modified, or another kind of indication. Sometimes it's a very useful feature, and it's really nice to have it enabled. But when it modifies something offscreen, it's hard to track what changed and when. A fairly regular occurence when animating several layers btw.

PR #4342, opened 2024-02-26
Drop selection when we hide a layer that is being transformed (fix #4179, fix #3254)

### LINKED ISSUE #3254: Issue with copy/pasting on hidden layers
I've noticed over a hundred or so hours that ctrl+c / ctrl+v behaves in a weird and sometimes disturbing way when a hidden layer is selected. Whenever I have an area of the picture selected with the marquee tool, and I've picked a hidden layer in the layer list, pressing ctrl+c or ctrl+v (if I've noticed that correctly, still not entirely sure on the specifics of what happens) will erase whatever is in that area on the hidden layer without putting anything in the clipboard. It's usually not very problematic, but combined with a little bit of fatigue it sometimes leads to a huge load of mistakes :( I would love to see that fixed. Thanks!

### LINKED ISSUE #4179: Drop selection if we hide the layer

dacap:
This fixes several problems in MovingPixelsState where hidden layers were transformed anyway when we switched the visibility of a layer in this state.

Other fix was tried before in #3254 but we needed the onBefore/After layer visibility change notifications to make this work properly (i.e. drop pixels when the visiblity of a layer is changed).

The only drawback at this moment is that changing the visibility of the non-active layer when we are transforming multiple cels/timeline range can be confused because we don't have #2144/#2865 implemented yet.

This bug was originally reported here: https://community.aseprite.org/t/20621

It would be nice if @Gasparoken and @martincapello can test this.

### DIFF
--- a/src/app/ui/doc_view.cpp
+++ b/src/app/ui/doc_view.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2023  Igara Studio S.A.
+// Copyright (C) 2018-2024  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -478,7 +478,16 @@ void DocView::onTotalFramesChanged(DocEvent& ev)
 
 void DocView::onLayerRestacked(DocEvent& ev)
 {
-  m_editor->invalidate();
+  if (hasContentInActiveFrame(ev.layer()))
+    m_editor->invalidate();
+}
+
+void DocView::onAfterLayerVisibilityChange(DocEvent& ev)
+{
+  // If there is no cel for this layer in the current frame, there is
+  // no need to redraw the editor
+  if (hasContentInActiveFrame(ev.layer()))
+    m_editor->invalidate();
 }
 
 void DocView::onTilesetChanged(DocEvent& ev)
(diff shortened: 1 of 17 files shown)


src/app/ui/doc_view.cpp
```diff
@@ -653,4 +662,9 @@ void DocView::onCancel(Context* ctx)
   }
 }
 
+bool DocView::hasContentInActiveFrame(const doc::Layer* layer) const
+{
+  return (layer && layer->cel(m_editor->frame()));
+}
```
Gasparoken, line 668: Here if the `layer` is a `Group`, this function returns `false`, but this layer group can contain some layer with a cel. Then`DocView::onAfterLayerVisibilityChange()` doesn't invalidate properly the editor.
Recommended change:
````cpp
bool DocView::hasContentInActiveFrame(doc::Layer* layer) const
{
  if (layer && !layer->isGroup())
    return (layer && layer->cel(m_editor->frame()));

  // Recursive browsing of layers if 'layer' is a Group of Layers
  auto layers = dynamic_cast<LayerGroup*>(layer)->layers();
  for (auto& l : layers) {
    if (hasContentInActiveFrame(l))
      return true;
  }
  return false;

}```

Gasparoken (changes requested):
I found a couple of things:
1. Create a layer, Create a Group that contains the layer, try to hide the group --> the editor won't update (I recommended a solution for this).
2. Unrelated with this fix: An ASSERT was triggered in `expand_cel_canvas.cpp` line 278 (`ASSERT(isSelectionPreview());`) when:
- Create a layer
- Create a Group that contains the layer
- Select some area in the canvas and move it.
- Select tthe layer n the timeline
- Assert triggers
Omitting the ASSERT works fine.

PR #4343, opened 2024-02-26
Avoid moving the playback cue when decrementing a ping-pong tag's repeat field that has only one frame

### LINKED ISSUE #4336: Consistent crash on "exporting all frames" on a one frame animation with a tag that specifies "ping-pong" animation direction
Source: https://steamcommunity.com/app/431730/discussions/2/4345480979787790760/

martincapello:
Fix #4336


dacap:
@martincapello could we add a test for this in playback_tests?

Issue #4361, opened 2024-03-06
clang-tidy should recommend west const in some way


dacap:
clang-tidy recommends east const instead of west const which is the format we use.

Follow: https://github.com/llvm/llvm-project/issues/58221

PR #4377, opened 2024-03-19
Fix nested tag with "Ping-Pong" repeat mode causes to skip the first frame of the parent tag (fix #4271)

### LINKED ISSUE #4271: Nested tag with "Ping-Pong" repeat mode causes weird behaviors in Parent tag with any repeat mode
[Tag Repeat Bug.zip](https://github.com/aseprite/aseprite/files/13973733/Tag.Repeat.Bug.zip)

In the gif below, I have two animations built out of nested tags. In the first animation, both tags have "Forward" repeat. In the second animation, the parent tag has "Forward" repeat, while the nested tag has "Ping-Pong" repeat.

The first animation repeats correctly, but the second animation skips the first frame.

A Nested tag with "Ping-Pong" repeat will also cause a Parent tag with "Ping-Pong" or "Ping-Pong Reverse" repeat to break.
(Curiously, a Parent tag with "Reverse" repeat will play without issue.)

Aseprite v1.3.2-x64, Windows 11.

Gasparoken:
Issue #4271 has two issues (solved):
1. The nested tag with repeat mode "Ping-Pong" causes it to skip the first frame of the parent tag when it has to repeat the playback sequence.
2. Nested tag with "Ping-Pong" repeat mode causes to cancel the "Ping-Pong" sequence on its parent tag when it has to repeat the playback sequence.
And I found others (solved too):
3. Tag starting on frame 1 + "Ping-Pong Reverse" stops repeting the sequence on frame 1 when repeat = 1.
4. Tag starting on frame 1 + "Ping-Pong Reverse"  skips at least a sequence. For example: when repeat = 2 and aniDir = "Ping-Pong Reverse" the sequence is like aniDir = "Forward".

### DIFF
--- a/src/doc/playback.cpp
+++ b/src/doc/playback.cpp
@@ -1,5 +1,5 @@
 // Aseprite Document Library
-// Copyright (C) 2021-2022  Igara Studio S.A.
+// Copyright (C) 2021-2024  Igara Studio S.A.
 //
 // This file is released under the terms of the MIT license.
 // Read LICENSE.txt for more information.
@@ -167,7 +167,7 @@ void Playback::handleEnterFrame(const frame_t frameDelta, const bool firstTime)
     case PlayAll:
     case PlayInLoop: {
       const Tag* tag = this->tag();
-      const frame_t frame = m_frame;
+      frame_t frame = m_frame;
       const int forward = getParentForward();
 
       for (const Tag* t : m_tags) {
@@ -185,8 +185,15 @@ void Playback::handleEnterFrame(const frame_t frameDelta, const bool firstTime)
           }
           else {
             addTag(t, false, forward);
-            if (!firstTime)
+            if (!firstTime) {
               goToFirstTagFrame(t);
+              // Consideration for tests:
+              // Playback.OnePingPongInsidePingPongReverse
+              // Playback.OneReverseInsidePingPongReverse
+              // Playback.OnePingPongReverseInsideReverse
+              if (frame != m_frame)
+                handleEnterFrame(m_frame, firstTime);
+            }
           }
         }
       }
@@ -461,6 +468,7 @@ bool Playback::decrementRepeat(const frame_t frameDelta)
         newFrame = firstTagFrame(m_playing.back()->tag);
       }
       else {
+        // Note that 'tag' means 'the last tag removed from m_playing'
         newFrame = (frameDelta * forward < 0 ? tag->fromFrame()-1: tag->toFrame()+1);
       }
 
@@ -473,10 +481,96 @@ bool Playback::decrementRepeat(const frame_t frameDelta)
           stop();
           return false;
         }
-        if (newFrame < 0)
-          newFrame = m_sprite->lastFrame();
-        else if (newFrame > m_sprite->lastFrame())
-          newFrame = 0;
+        if (newFrame < 0) {
+          if (m_playing.empty() ||
+              m_playing.back()->tag->fromFrame() != 0 ||
+              m_playing.back()->tag->aniDir() == AniDir::REVERSE) {
+            m_frame = m_sprite->lastFrame();
+            return false;
+          }
+          // Special cases arise with PING_PONG_REVERSE aniDir and when
+          // the begining of the tag range matches with the first frame of
+          // the sprite.
+          // Consideration for tests inside:
+          // Playback.OnePingPongInsideOther
+          //     A            A
+          // >-------<    >-------<
+          //   B            B
+          // <--->        >---<
+          // 0 1 2 3 4    0 1 2 3 4
+          PlayTag* parentPlaying = m_playing.back().get();
+          // When parentPlaying is PING_PONG_REVERSE
+          // the next frame will be defined according:
+          // 1. The playloop has more repetitions to decrement
+          //    --> go to the next frame of the 'tag'
+          // 2. The playloop has no more repetitions to decrement
+          //    --> Start all the playloop again.
+          if (parentPlaying->repeat > 1) {
+            if (parentPlaying->tag->aniDir() == AniDir::PING_PONG_REVERSE)
+              parentPlaying->invertForward();
+            --parentPlaying->repeat;
+            newFrame = tag->toFrame() + 1;
+          }
+          else
+            continue;
+        }
+        else if (newFrame > m_sprite->lastFrame()) {
+          // If all the tags were played and
+          // the 'tag' range  ==  timeline range and
+          // the 'tag' is PING_PONG_REVERSE -->
+          // The playloop has to start on the last frame of
+          // the timeline, or the first frame of the most nested
+          // tag (on reverse direction).
+          // Consideration for tests:
+          // Playback.WithTagRepetitions
+          // Playback.OnePingPongInsideOther
+          // Playback.OnePingPongInsideOther14, 15, 18 and 19
+          //    A      <-- last tag removed from 'm_playing', i.e. 'tag'
+          // >-----<
+          //    B      <-- most nested tag
+          // ***-***
+          // 0 1 2 3
+          if (m_playing.empty() &&
+              tag->aniDir() == AniDir::PING_PONG_REVERSE &&
+              tag->fromFrame() == 0 &&
+              tag->toFrame() == m_sprite->lastFrame()) {
+            m_frame = m_sprite->lastFrame();
+            handleEnterFrame(frameDelta, false);
+            if (m_playing.size() > 1) {
+              m_playing.back()->invertForward();
+              goToFirstTagFrame(m_playing.back()->tag);
+            }
+            return false;
+          }
+
+          // 'tag' is contained by other tag and the last frame of each tag
+          // matches in the last frame of the sprite
+          if (!m_playing.empty() &&
+              tag->toFrame() == m_playing.back()->tag->toFrame()) {
+            PlayTag* parentPlaying = m_playing.back().get();
+            // The parentPlaying has no more repetitions to decrement
+            //    --> continue to remove the 'parentTag'
+            if (parentPlaying->repeat <= 1)
+              continue;
+            // Consideration for test:
+            // Playback.OnePingPongInsideOther
+            if (parentPlaying->tag->aniDir() == AniDir::PING_PONG ||
+                parentPlaying->tag->aniDir() == AniDir::PING_PONG_REVERSE) {
+              parentPlaying->invertForward();
+              newFrame = tag->fromFrame() - 1;
+            }
+            // Consideration for test:
+            // Playback.OnePingPongInsideForward2
+            else if (parentPlaying->tag->aniDir() == AniDir::FORWARD) {
+              --parentPlaying->repeat;
+              newFrame = parentPlaying->tag->fromFrame();
+            }
+            else
+              newFrame = 0;
+          }
+          else
+            newFrame = 0;
+        }
       }
 
       m_frame = newFrame;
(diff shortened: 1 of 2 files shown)


Gasparoken:
@martincapello this fix solves only the "1". I will create other commits for the other 3 sub-issues.

src/doc/playback.cpp
```diff
@@ -436,13 +437,21 @@ bool Playback::decrementRepeat(const frame_t frameDelta)
                  "repeat=", m_playing.back()->repeat,
                  "forward=", m_playing.back()->forward);
       // Tag has only 1 frame, then don't move the playback cue.
-      return tag->frames() > 1;
+      return tag->frames() > 1 &&
```
martincapello, line 440: I've been testing this changes, and they work as expected. But I think this can be simplified by just returning false in this line. Because the only place where the returned value of this function is used is in the following fragment of code in handleExitFrame():

```c++
        else if ((tag->aniDir() == AniDir::PING_PONG ||
                  tag->aniDir() == AniDir::PING_PONG_REVERSE)
                 && m_frame == lastTagFrame(tag)) {
          PLAY_TRACE("    Changing direction frame=", m_frame,
                     " forward=", forward, "->", -forward);

          // Changing the direction of the ping-pong animation
          m_playing.back()->invertForward();  
          return decrementRepeat(frameDelta);    <-------------- here
        }
```
And it actually makes sense, because in `decrementRepeat()`, after decrementing the repeat count, it goes to the first tag frame so we must avoid the moving to the next frame because it will jump the first frame otherwise.

Gasparoken: Sorry. I tried that solution but it didn't work in all cases. Simple test cases go wrong.
My first approach discarded: https://github.com/aseprite/aseprite/commit/2f72f3e7ec3334f2bd5624dadd49528a81d8d74c

martincapello:
Are items 3 and 4 actually issues? Because if you have a ping-pong reverse tag the playback goes backwards, so if you start at frame 1, you are starting at the end frame, then it immediately consumes a repeat.

src/doc/playback.cpp
martincapello: No, I meant to return false from decrementRepeat:
```c++
bool Playback::decrementRepeat(const frame_t frameDelta)
{
  bool move = true;
  while (true) {
    Tag* tag = this->tag();
    PLAY_TRACE("    Decrement tag", tag->name(),
               "repeat", m_playing.back()->repeat, "-1");

    if (m_playing.back()->repeat > 1) {
      --m_playing.back()->repeat;
      goToFirstTagFrame(tag);

      PLAY_TRACE("    Repeat tag", tag->name(), " frame=", m_frame,
                 "repeat=", m_playing.back()->repeat,
                 "forward=", m_playing.back()->forward);
      // Tag has only 1 frame, then don't move the playback cue.
      return false;
    }
   . . .
```

This way you are still using the cases where it returns true.

Gasparoken:
> Are items 3 and 4 actually issues? Because if you have a ping-pong reverse tag the playback goes backwards, so if you start at frame 1, you are starting at the end frame, then it immediately consumes a repeat.

This is the issue "3":
Mar-20-2024 18-03-32

And this is the issue "4":
Mar-20-2024 18-07-39

dacap:
3. and 4. look like bugs. The correct ping-pong behavior can be viewed with a couple of extra frames outside the tag (here a video of ping-pong with repeat=2, ping-pong reverse with repeat=2, and then with repeat=1):

https://github.com/aseprite/aseprite/assets/39654/f06c5592-3c20-47b3-a4df-078a481d31bd

Just in case, for every bug that it's found it'd be nice to add an automated test to reproduce the bug/fix.

martincapello:
Thank you both for the videos! Now I see.

Gasparoken:
@dacap you review the tests added in `playback_tests.cpp`. I will try to improve the readability of the code in`playback.cpp`.

src/doc/playback_tests.cpp
```diff
@@ -500,6 +525,349 @@ TEST(Playback, OnePingPongInsideOther3)
...
+  // <--->
+  // 0 1 2 3 4
+
+  Tag* tagA = make_tag("A", 0, 4, AniDir::PING_PONG, 2);
+  Tag* tagB = make_tag("B", 0, 2, AniDir::PING_PONG, 2);
+  auto sprite = make_sprite(5, { tagA, tagB });
+
+  Playback play(sprite.get(), 0, Playback::Mode::PlayInLoop);
+  expect_frames(play, {0,1,2,1,0 , 3,4,3 , 2,1,0,1,2 ,
+                       0,1,2,1,0 , 3,4,3 , 2,1,0,1,2 });
```
dacap, line 621: I stop reviewing cases here, I think they are good so far. As we talked, we have two options: 1) just keep the most relevant cases (but don't create a lot of more cases, or at least if more cases are add some extra variations can be tested, like adding more repeats), or 2) group all these OnePingPongInsideOther in one test case with a `for()` to test more combinations (just if the expected sequence can be easily generated).

dacap (changes requested):

src/doc/playback_tests.cpp
```diff
@@ -152,6 +152,31 @@ TEST(Playback, WithTagRepetitions)
   play = Playback(sprite.get(), 0, Playback::Mode::PlayAll);
   expect_frames(play, {0,1,2,1,2,3,0,0,0});
   EXPECT_TRUE(play.isStopped());
+
+  Tag* b = make_tag("B", 0, 3, AniDir::PING_PONG, 2);
+  sprite = make_sprite(4, { b });
+  play = Playback(sprite.get(), 0, Playback::Mode::PlayInLoop);
+  expect_frames(play, {0,1,2,3,2,1,0,
+                       0,1,2,3,2,1,0,
+                       0,1,2,3,2,1,0});
```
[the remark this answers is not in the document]
Gasparoken: Maybe it should be nice to have a checkbox to merge or not merge the last frame of a Ping Pong sequence. But for now it seems fine to me the way it is.

Gasparoken:
Ready for review. I'm not too happy with for loops, as they are probably harder to understand than individual cases.

src/doc/playback.cpp
```diff
@@ -167,7 +167,7 @@ void Playback::handleEnterFrame(const frame_t frameDelta, const bool firstTime)
     case PlayAll:
     case PlayInLoop: {
       const Tag* tag = this->tag();
-      const frame_t frame = m_frame;
+      frame_t frame = m_frame;
```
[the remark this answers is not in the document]
dacap: @Gasparoken this can be const:
```suggestion
      const frame_t frame = m_frame;
```
Just as a side note, clang-tidy recommend the east-const notation, but we use the west-const notation, more info in #4361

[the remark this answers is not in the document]
dacap: I see that this was already const originally, so there is no need for change in this line.

src/doc/playback_tests.cpp
```diff
@@ -464,39 +505,361 @@ TEST(Playback, PingPongWithInnerReverse)
...
 
+// OnePingPongInsideOther series
+std::vector<int> goRight(std::vector<int> range) {
+  std::vector<int> out;
+  if (range[0] > range[1])
+    return out;
+  for (int i=range[0]; i<range[1]+1 ; ++i)
+    out.push_back(i);
+  return out;
+}
```
dacap, line 516: No need to pass arguments as a std::vector:
```suggestion
static std::vector<int> goRight(const int a, const int b)
{
  std::vector<int> out;
  if (a > b)
    return out;
  for (int i=a; i<=b ; ++i)
    out.push_back(i);
  return out;
}
```

src/doc/playback_tests.cpp
```diff
@@ -464,39 +505,361 @@ TEST(Playback, PingPongWithInnerReverse)
...
+
+std::vector<int> goLeft(std::vector<int> range) {
+  std::vector<int> out;
+  if (range[0] > range[1])
+    return out;
+  for (int i=range[1]; i>range[0]-1 ; --i)
+    out.push_back(i);
+  return out;
+}
```
dacap, line 525: Same here:
```suggestion
static std::vector<int> goLeft(const int a, const int b)
{
  std::vector<int> out;
  if (a > b)
    return out;
  for (int i=b; i>=a; --i)
    out.push_back(i);
  return out;
}
```

src/doc/playback_tests.cpp
```diff
@@ -464,39 +505,361 @@ TEST(Playback, PingPongWithInnerReverse)
...
+
+void concat(std::vector<int>& a, std::vector<int>& b)
+{
+  if (b.empty())
+    return;
+  for (int i=0; i<b.size(); ++i)
+    a.push_back(b[i]);
+}
```
dacap, line 533: Added `&`, `const`, and no need to check for empty (as size will be 0 anyway):
```suggestion
static void concat(std::vector<int>& a, const std::vector<int>& b)
{
  for (size_t i=0; i<b.size(); ++i)
    a.push_back(b[i]);
}
```

src/doc/playback_tests.cpp
```diff
@@ -464,39 +505,361 @@ TEST(Playback, PingPongWithInnerReverse)
...
+        std::vector<int> temp;
+        //     A            A            A
+        // <------->    <------->    <------->
+        //   B              B              B
+        // *---*          *---*          *---*
+        // 0 1 2 3 4    0 1 2 3 4    0 1 2 3 4
+        if (A_aniDir == doc::AniDir::PING_PONG) {
+
+          // Start
+          temp = goRight({0, B_Range[0]-1});
```
dacap, line 697: Remove `{` and `}` from every `goRight()` and `goLeft()` call:
```suggestion
          temp = goRight(0, B_Range[0]-1);
```

dacap (changes requested):
So far it looks and works great :+1: Some minor changes to review.

src/doc/playback.cpp
```diff
@@ -473,10 +481,96 @@ bool Playback::decrementRepeat(const frame_t frameDelta)
...
-          newFrame = m_sprite->lastFrame();
-        else if (newFrame > m_sprite->lastFrame())
-          newFrame = 0;
+        if (newFrame < 0) {
+          if (m_playing.empty() ||
+              m_playing.back()->tag->fromFrame() != 0 ||
+              m_playing.back()->tag->aniDir() == AniDir::REVERSE) {
+            m_frame = m_sprite->lastFrame();
+            return false;
+          }
```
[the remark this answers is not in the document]
Gasparoken: You are right, I was testing this:
With newFrame < 0:
- `m_playing.empty()` never happen, because the only way to have newFrame < 0 is **if we have a tag** in REVERSE or PING PONG REVERSE which frame 0 is contained in that tag.
- `m_playing.back()->tag->fromFrame() != 0` never happens, for the same reason.
- `m_playing.back()->tag->aniDir() == AniDir::REVERSE` never happens, because the the newFrame is calculated in `newFrame = (frameDelta * forward < 0 ? tag->fromFrame()-1: tag->toFrame()+1);` which in reverse cases it **gives newFrame > 1 always**.

Therefore, no condition is needed.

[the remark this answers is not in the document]
dacap: Probably I'd put an assert but leave the if condition too (just to avoid a crash in production if there is a future bug).

Gasparoken:
I think I can squash all the code if you want.

dacap:
> I think I can squash all the code if you want.

No problem, I will merge it as it is :+1:

dacap:
There are several errors, I'll fix them and merge.

dacap:
Merged in 99e6e7bd8256cfdbb0d0c3b0ab28098ddb8a038a :+1:

Issue #4356, opened 2024-03-01
Linux: Exporting and confirming without selecting output folder with 3-dots button will export to Home instead of last folder


hsandt:
I export in chain many sprites and count on the last folder location to be respected, even if I create a new file (which is very common when copy-pasting a portion of an image into a new one in order to extract a sub-image and export it as a different image).

The 1st time, I carefully selected the output folder.

The 2nd time, I just checked that the same output folder was selected by clicking on the 3 dots menu: it was already select, excellent, so I closed the popup with Cancel and confirmed Export.

On further exports, I just blindly confirmed Export.

It turns out that only the 1st image was exported to the selected output folder.

Indeed, by default, the output folder is $HOME. The native output folder popup remembers the last output folder and displays it but it _doesn't mean that it was about to export there_. You need to confirm the folder once more even if you don't change the selection at all just to say "yes, I want to export there". Closing the output folder popup menu without confirming (by using the cross button or pressing Cancel) does *not* confirm the path, which would remain $HOME.

This means that every time you must click on the 3 dots button then press OK on the native folder selection popup just to confirm what the app seemed to have remembered, but didn't (maybe because it's native, it somewhat has an OS-level way to remember last folder selection but Aseprite doesn't know about it? but most people expect 3 dots menu to *edit the current output folder*)

This would be less of an issue if the output directory / full export path was shown in the Export File popup, see feature request: https://community.aseprite.org/t/show-path-in-export-as/20993

Even if it was though, I'd still need an option to actually remember the last export folder, else I'd still have to do the convoluted process I noted above.

And if the native popup was fixed to edit the current path but without this option, so $HOME at the beginning, then it'd be worse as I'd have to navigate back to the same output folder again.

In other words, ideally we'd have those features:
- display the output directory/full path in Export File popup as requested in https://community.aseprite.org/t/show-path-in-export-as/20993
- remember the last export location and reuse it for new files (for existing files in already opened tabs, it may be relevant to keep the last export location per file/tab as some workflows require different images to be exported to different folders - but for new files, exporting to $HOME is not very useful, so using the last export location as default makes sense) - there could be an option to decide in which contexts exactly we should reuse the last export location
- the native folder selection popup should always start at the current location shown in full path (if it exists - if someone removed a folder, then move up until you find a parent or grandparent that exists)

### Aseprite and System version

* Aseprite version: v1.2.3 (Humble Bundle deb)
* System: Linux Ubuntu 22.04

[...]

PR #4389, opened 2024-03-28
Show format within filename when exporting file

### LINKED ISSUE #4059: {tag} as folder name in "Export As" creating nested folders after first Export
Describe your bug report or feature request here
If you go to "Export As" and use {tag}/0001.png for example as your file name, it works as expected, creating a folder w/ the name of whatever the tag is.

However, if you export again, the original {tag}/ you wrote is no longer there in the autofilled export name, but it seems like aseprite still treats it as if it is there. So if you then write {tag}/0001.png again, it will export to {tag}/{tag}/0001.png.

All subsequent exports will be exported in this manner, and isn't able to be changed. Even closing and re-opening the project does not fix it.

* Aseprite version: v1.3-rc6-x64 Steam/beta/
* System: Windows

gui-marc:
### Bug Description
The `export_file_window` was displaying only the filename instead of the entire path, leading to confusion for users, especially when including dynamic placeholders such as {tag} in the file path.

### Cause of the Bug
Within the `ExportFileWindow` class, the method `setOutputFilename` split the provided path into `m_outputPath` and `m_outputFilename`, causing the window to display only the filename. Consequently, dynamic placeholders like {tag} were omitted from the displayed path, leading to ambiguity for users.

### Impact
This behavior caused confusion and made it difficult for users to accurately discern the file's destination, particularly when using dynamic placeholders. It also difficulted clearing dynamic placeholders from the export path, as it is not shown in the window.

### Solution
To address this issue, the following changes were implemented:

- Exposed `path_formats` defined in `filename_formatter`, modifying its type from `vector` to `array` for improved performance.
- During the assignment of `m_outputPath` and `m_outputFilename`, dynamic placeholders such as {tag} are now considered part of `m_outputFilename` and excluded from `m_outputPath`.

#### Example
With the new behavior, let's consider the export path: `/home/user/{tag}/test/{frame}/file.png`.

- `m_outputPath` will be `/home/user/`
- `m_outputFilename` will be `{tag}/test/{frame}/file.png`

Now, the displayed path will include the entire dynamic placeholder structure, providing users with clear visibility of the file's destination and enabling the editing of these parameters.

### Future Considerations
Maintainers should be aware of potential implications when modifying path-related functionalities in the future. Additionally, it's essential to maintain consistency in handling dynamic placeholders across the application to avoid similar issues.

Resolves #4059

I acknowledge that my contributions are licensed under the Individual Contributor License Agreement V4.0 ("CLA"), as stated in [CLA.md](https://github.com/igarastudio/cla/blob/main/cla.md).

I have signed the CLA following the steps provided [here](https://github.com/igarastudio/cla#signing).

### DIFF
--- a/src/app/ui/export_file_window.cpp
+++ b/src/app/ui/export_file_window.cpp
@@ -192,8 +192,14 @@ void ExportFileWindow::setAniDir(const doc::AniDir aniDir)
 
 void ExportFileWindow::setOutputFilename(const std::string& pathAndFilename)
 {
-  m_outputPath = base::get_file_path(pathAndFilename);
-  m_outputFilename = base::get_file_name(pathAndFilename);
+  if (base::get_file_path(m_doc->filename()).empty()) {
+    m_outputPath = base::get_file_path(pathAndFilename);
+    m_outputFilename = base::get_file_name(pathAndFilename);
+  }
+  else {
+    m_outputPath = base::get_file_path(m_doc->filename());
+    m_outputFilename = base::get_relative_path(pathAndFilename, base::get_file_path(m_doc->filename()));
+  }
 
   updateOutputFilenameEntry();
 }


gui-marc:
Hello @dacap, can you review this PR? I think it's a fast PR to review and it would be very useful for me.

Sorry to be so insistent, but this bug was causing me some trouble in a project I'm developing, and it's a little hard for me to compile aseprite in that workspace (I use the steam version there).

dacap:
Hi @gui-marc, I've just tested with an empty new sprite (that is not associated with a file on disk) and I got:

image

Anyway checking the code (and the problem to solve) I think we can approach this in a different way. In some way this is related to #4483, so we should find a way to display the output path, or the common path between the original sprite and the exported sprite (and add `{path}`+path elements to get to the output directory).

What I'm thinking is in something like this:

image

Where little capsules are the {path}, {tag}, {frame} elements and we should be able to add these easily (writing "{something}" or with some kind of menu/right-click).

Anyway as a first simple version that could be solved with this PR, we could just calculate the common path between the input sprite and the configured output path and remove it + adding all the path elements that goes to the output dir (that might include `../` elements if we are exporting in a parent dir, etc.). This doesn't require any of the changes in the file formatter, just calculating the correct output path/file comparing the selected output full path with the full input path.

gui-marc:
> Hi @gui-marc, I've just tested with an empty new sprite (that is not associated with a file on disk) and I got:
> 
> image

Strange, this seams to be a platform related problem. I couldn't reproduce it on Linux.

But thanks for the suggestion! I'll implement the way you suggested and test the behavior on different environments and then I'll reach out to you again when it's finished.

gui-marc:
>Anyway as a first simple version that could be solved with this PR, we could just calculate the common path between the input sprite and the configured output path and remove it + adding all the path elements that goes to the output dir (that might include ../ elements if we are exporting in a parent dir, etc.). This doesn't require any of the changes in the file formatter, just calculating the correct output path/file comparing the selected output full path with the full input path.

@dacap I don't understand what you mean by `input sprite path` and `configured output path`. Is it the path saved in the Document filename and the path set in the export file window? Also, I don't understand how to differenciate the path elements that goes to the output dir from the path dir itself without doing what I just did by finding the first `path_format` in the path string. My original PR just show the same output you suggested in the screenshot but only with text, not using these capsules.

image

dacap:
@gui-marc sorry for the confusion. For the moment you can ignore the screenshot with the capsules.

What I said that a possible PR that could fix now is the possibility to show the relative output path from the input sprite path (if possible). Here some examples:

Example A:
1. The current sprite we are working now is saved in `C:\User\David\A.aseprite`
2. We select _File > Export > Export As_ and choose `C:\User\David\Output\A.png`
3. The file name field in the Export dialog could show `Output\A.png`

Example B:
1. The current sprite is saved in `C:\User\David\Assets\B.aseprite`
2. We select _File > Export > Export As_ and choose `C:\User\David\Output\B.png`
3. The file name field in the Export dialog could show `..\Output\B.png`

Example C:
1. The current sprite is saved in `C:\User\David\Assets\B.aseprite`
2. We select _File > Export > Export As_ and read `C:\User\David\Assets\{tag}\{frame}\B.png` from the preferences
3. The file name field in the Export dialog could show `{tag}\{frame}\B.png` (the common path part is removed)

This means that the output file name could show a path relative to the path of the current file. Not sure if this makes sense or will be a lot easier to just show the full path. But it matches the current behavior when no path is specified (the output is done in the same directory of the current file).

We then could save in preferences (`[save_copy] filename`) this relative path (instead of full output path).

gui-marc:
> This means that the output file name could show a path relative to the path of the current file. Not sure if this makes sense or will be a lot easier to just show the full path. But it matches the current behavior when no path is specified (the output is done in the same directory of the current file).

@dacap from the implementation view, I've been looking through the code and didn't find any implemented method to, given two paths, return the relative path from one to another, also, if we further want to implement the feature with the capsules #4483, the logic I've already implemented could be fully reused. Regarding the `file_formatter`, I've just exposed the available path formats list, since it is not defined anywhere in the app, therefore I didn't bring any changes in the class behavior.

dacap:
> I've been looking through the code and didn't find any implemented method to, given two paths, return the relative path from one to another.

That's right, that's a function that this PR should implement in case we want to add this behavior.

> Also, if we further want to implement the feature with the capsules https://github.com/aseprite/aseprite/issues/4483, the logic I've already implemented could be fully reused. Regarding the file_formatter, I've just exposed the available path formats list, since it is not defined anywhere in the app, therefore I didn't bring any changes in the class behavior.

I'll review your PR but it will not be accepted as it is because it needs more testing (and some code changes). E.g. Just create a new sprite and try to <kbd>Ctrl+Shift+Alt+S</kbd> you will get the assert I show you or this behavior on Linux:

image

:point_up: that's just creating a new sprite and trying to exporting it.

Other example: If we specify `output/{tag}/test.png` then the dialog will show `{tag}/test.png` without the `output` the next time we use it (because in this patch we try to think that only template delimiters like `{tag}` can form the extra part of the output).

I'm not saying that the current version works better, but I'd prefer to merge a PR that fully fix this whole output path/file issue.

src/app/filename_formatter.h
```diff
@@ -87,6 +88,9 @@ namespace app {
   bool get_frame_info_from_filename_format(
     const std::string& format, int* startFrom, int* width);
 
+  // Types of formats that can be used in the filename formatter. E.g. "{frame}"
+  extern std::array<const char*, 12> path_formats;
```
dacap, line 92: We don't (try to not) define global variables. This could be something like `get_templates_list()` in case is required (I think this is not required for a proper fix).

dacap (changes requested):

gui-marc:
> ☝️ that's just creating a new sprite and trying to exporting it.

Ah ok, I've already fixed this issue in a local branch, but I didn't understand that it was related to the print you showed.

> We don't (try to not) define global variables. This could be something like get_templates_list() in case is required (I think this is not required for a proper fix).

Yes, it is not required if we just want to show the relative path.
>Other example: If we specify output/{tag}/test.png then the dialog will show {tag}/test.png without the output the next time we use it (because in this patch we try to think that only template delimiters like {tag} can form the extra part of the output).
> 
> I'm not saying that the current version works better, but I'd prefer to merge a PR that fully fix this whole output path/file issue.

Ok, i didn't get that this was also a problem related to this issue, but now I understand the problem.

@dacap thanks for the patience explaining this to me. I guess the better way is to implement your proposal here with the relative path and then I could send another PR in #4483 with the method I've already implemented here with the fixes and code changes.

I'll work on that. Thanks for the support!

dacap:
Hi @gui-marc, although using `std::filesystem::relative()` and we are using some C++17 functions, we are adopting new APIs at a slow rate, because we want to still compile for macOS 10.9 (and other old Linux distributions), I'll check this but probably these functions are not available yet on that macOS version (probably they were available since macOS 10.15).

As a side note this PR could be simplified with just one commit (removing all initial unnecessary changes now) leaving the ones in `ExportFileWindow`.

gui-marc:
This PR is not closed, I had to remove my commits due to a bug (i had to clone the project again on my pc so I could use the newer versions of the subrepositories). Then I removed all my commits of the branch and github closed the commit automatically...

gui-marc:
> Hi @gui-marc, although using `std::filesystem::relative()` and we are using some C++17 functions, we are adopting new APIs at a slow rate, because we want to still compile for macOS 10.9 (and other old Linux distributions), I'll check this but probably these functions are not available yet on that macOS version (probably they were available since macOS 10.15).

Yes @dacap  you are right, I've just checked and macOS really doesn't support `std::filesystem::relative()` yet. I'm rewriting the method using only strings.

dacap:
> I'm rewriting the method using only strings.

That will be great @gui-marc, one possibility would be to add a new function in the laf base/fs.h for this (e.g. base::get_relative_path). The good part about this is that you can add some tests in base/fs_tests.cpp to check if the function works as expected before using it on Aseprite.

gui-marc:
@dacap , I've made a last commit now only using strings and by my tests it seams to be working as spectated on Linux. Unfortunately, I cannot test on Windows nor Mac.

dacap:
I've just tested using:

image

And then exporting again will show:

image

About the `split()` function, you can use `base::split_string()`.

dacap:
And please remove all trailing whitespaces before committing:

image

src/app/ui/export_file_window.cpp
```diff
@@ -192,8 +192,13 @@ void ExportFileWindow::setAniDir(const doc::AniDir aniDir)
...
 {
-  m_outputPath = base::get_file_path(pathAndFilename);
-  m_outputFilename = base::get_file_name(pathAndFilename);
+  if (base::get_file_path(m_doc->filename()).empty()) {
+    m_outputPath = base::get_file_path(pathAndFilename);
+    m_outputFilename = base::get_file_name(pathAndFilename);
+  } else {
+    m_outputPath = base::get_file_path(m_doc->filename());
+    m_outputFilename = base::get_relative_path(pathAndFilename, base::get_file_path(m_doc->filename()));
+  }
```
gui-marc, line 201: @dacap , when we save a file (not export it), we cannot pass a dynamic placeholder like `{tag}`, so the relative path works as spected.

But, when we only export the file that is not saved, we can't get the path of the `.aseprite` file in `m_doc->filename()` and, therefore, it will just split the `pathAndFilename` in the old way.

For instance, the path `out/{tag}/test.png` would only show `test.png` in the input, causing the same bug as before.

The problem is that `m_docPref.saveCopy` has the full  `out/{tag}/test.png` path and we cannot get the relative path to an inexistent file.

In this case, what should we do? I thought about showing the absolute path but I'm not sure if this is the best approach.

dacap: @gui-marc I've just tested with your laf changes and it looks good so far :+1: At least it fixes the original issue for most cases (when the sprite is saved/has a filename associated).

We can merge this and I can think about a future fix for the other cases (there is a general UI problem to show the full path, or to show the real output folder when pressing the "..." dots, related to #4356).

dacap: I forgot, you'll need to create a PR for the laf project with your get-relative-path branch.

gui-marc: @dacap, ok, I was waiting for your reply here before creating the PR in the laf repo. I'm creating it now

gui-marc: @dacap I've just oppened the PR: https://github.com/aseprite/laf/pull/96

dacap:
@gui-marc due to GitHub limitations, I can:

1) cherry pick https://github.com/dacap/aseprite/commit/549f523fae51514e2503c1e7abbf9e5a1accc367 (with the laf submodule update + minor commit description changes/code formatting changes) and close this PR, or
2) you can update your branch commit with the changes from 549f523fae51514e2503c1e7abbf9e5a1accc367 and I can merge the PR (no need to close it)

Generally I'd just cherry pick and close the PR in these cases, but probably you need the "merge" status for your PR.

gui-marc:
@dacap I rebased it with https://github.com/dacap/aseprite/commit/549f523fae51514e2503c1e7abbf9e5a1accc367 but it added the previous commits in this PR (is this right?).

dacap:
> @dacap I rebased it with [dacap@549f523](https://github.com/dacap/aseprite/commit/549f523fae51514e2503c1e7abbf9e5a1accc367) but it added the previous commits in this PR (is this right?).

No, your branch should contain just one extra commit from the main branch.

You can start (reset hard) your gui-marc/issue-4059 branch to the last aseprite/main commit and cherry pick 549f523fae51514e2503c1e7abbf9e5a1accc367, then push

dacap:
Thanks for your PR and patience @gui-marc to continue working on this fix :+1:

gui-marc:
Thanks @dacap for your support and also your patience!

PR #4392, opened 2024-03-31
Add PICO-8 Extended color palette

musjj:
PICO-8 has a set of undocumented color palette as seen here: https://pico-8.fandom.com/wiki/Palette#128..143:_Undocumented_extra_colors. This adds an alternative PICO-8 palette that includes those colors.

I renamed `peach` from the original palette to `light_peach` to fit the new palette better (as suggested in the wiki above), but I left `indigo` as is (the wiki suggests `lavender`). Not sure if it matters at all since the names do not seem to show up in the UI.

### DIFF
--- a/data/extensions/pico8-palette/package.json
+++ b/data/extensions/pico8-palette/package.json
@@ -10,7 +10,8 @@
   ],
   "contributes": {
     "palettes": [
-      { "id": "PICO-8", "path": "./pico-8.gpl" }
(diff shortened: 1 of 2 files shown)


Chubercik (approved):

dacap:
Hi @musjj, would you mind sign our CLA? https://github.com/igarastudio/cla?tab=readme-ov-file#signing

musjj:
Do I need to share my real, full name? Or is a pseudonym ok?

PR #4416, opened 2024-04-11
Fix improve the octree quantization method using better Euclidean color distance (fix #2787)

### LINKED ISSUE #2787: Improve the octree quantization method using better Euclidean color distance
Implement a different method to calculate the Euclidean color distance in a more convenient color space.

Reported in community: https://community.aseprite.org/t/octree-color-indexed-conversion-testing/9682

Gasparoken:
First approach.

I'm still looking for a way to run the color mode conversion preview when changing the "color fit criteria" selector (combo box), just like it runs when changing the RGB mapping method.

fix #2787


src/doc/sprite.h
```diff
@@ -258,6 +260,9 @@ namespace doc {
     mutable RgbMapAlgorithm m_rgbMapAlgorithm;
     mutable std::unique_ptr<RgbMap> m_rgbMap;
 
+    // Last 'color fit criteria' used on indexed conversions
+    mutable FitCriteria m_fitCriteria = FitCriteria::DEFAULT;
```
dacap, line 264: Do we really need this in `doc::Sprite`?

Gasparoken: You're right! We don't need it, and we don't need `mutable RgbMapAlgorithm m_rgbMapAlgorithm;` either. I'll add this parameters into class Rgbmap.

dacap: Actually moving it to the RgbMap is a similar case. If you think it makes more sense to move them inside RgbMap go for it (not sure if this FitCriteria is only used inside the RgbMap and not in the findBest of the palette).

One important thing I'm missing and probably I thought incorrectly was that this FitCriteria was something used in the *Sprite > Color Mode > More Options > Advance* only. But it looks like you are planning as part of the whole program configuration. Something particular for the Octree (if I'm guided by the #2787 title), or something that can be used in both rgbmaps (rgb5a3/octree). What is the plan with this PR?

Gasparoken: In a first approximation, I thought of it only for octree quantization. But I tested it with RGB5A3 and it can be processed without problems. So I extended it for both rgbmaps.
I added a global setting to set the default value in Advanced conversion `RGB->Indexed`, then added this setting to work with `Sprite > ColorMode > Indexed` too.

We can stop here and wait for the artist's reception to see if it is feasible to extend this option further (include parameter in lua for example).

All of these other 'color fit criteria' in the `Palette::findBestFit` function can be useful for artists who want to fit any RGB image to a custom palette and achieve a more appropriate color selection.

Gasparoken: @dacap I will limit the feature to be used only in "Sprite > Color Mode > Advanced"

Gasparoken:
This new parameter called 'color fit criteria' is used to compare colors under different criteria in the `Palette::findBestFit` function.
A possible alternative to emplace the 'color fit criteria'  conversion process is within the Rgbmap class.
I'll try to refactor this to include the conversion function inside Rgbmap only.

dacap:
I'm trying to finally merge this PR, I've just squashed all commits [here](https://github.com/dacap/aseprite/tree/add-color-fit-criteria-squashed) to understand the change as a whole locally (as some commits removed code and then other commits add it back, I was able to see better where the code was moved in the squashed commits, probably we'd merge this squashed).

Anyway, the first error I saw (this was detected by a compiler warning):

https://github.com/aseprite/aseprite/pull/4416/commits/8a17be1d1106e75c59e6f9e81cf002167a7a7abe#diff-01e5af79bba18b3ca9eaa00307326da1d8ec4b46330c3a5c0996443969dd2d3dR157

https://github.com/aseprite/aseprite/commit/8a17be1d1106e75c59e6f9e81cf002167a7a7abe#diff-0bb1cd5267ecb4d20ff5b0987863cbcdfa257ff2803120a430d5d4f8127127caR58

I'm not sure if you were able to test this correctly, as I'm seeing that `Sprite::rgbMap()` uses this method but the method does nothing in both cases (OctreeMap and RgbMapRGB5A3). Check that `void fitCriteria(const FitCriteria fitCriteria)` was using `==` operator when `=` should be used.

I've fixed this bug in my squashed branch, and moved all common methods to `RgbMapBase` as the fields are now there and we can avoid to duplicate this code in derived classes. Basically the hierarchy now is:

```
   RgbMap > RgbMapBase > OctreeMap
                       > RgbMapRGB5A3
```

I've removed duplicated code in `OctreeMap::mapColor(r, g, b, a)` which is defined in `RgbMap::mapColor(r, g, b, a)`.

dacap:
If you agree with the changes, I can continue reviewing the commit (probably you can force push the squashed commit). I'd also recommend you to take care of some aseprite-bot comments (there is a limit of 25 comments per push, so if you fix some of the comments more important ones might be generated).

Gasparoken:
Just a few comments about the purpose of this feature and its implementation:

The rgbmap is the function that maps the colors of an RGB image to the colors available in the palette. That is, the rgbmap can replace the color of each RGBA pixel with an index (palette index) according to a color comparison criterion (fit criteria).
In few words, changing the fit criteria affects the result of a image conversion.

The 'fit criteria' will help us to recolor an RGB image with a limited color palette taking into account different color perception criteria (color spaces: RGB, linearized RGB, CIE XYZ, CIE LAB).

A side effect of having an rgbmap with a certain 'fit criteria' is that any new color that may appear in the sprite, for example via ‘Paste’ from the clipboard, should be passed through the same mapping function.
It follows that the rgbmap + fit criteria should be associated with the sprite from the creation of the sprite and/or the conversion of it to indexed. This property should be able to be modified through the sprite properties. Ideally, the rgbmap algorithm and fit criteria would be stored as persistent data in the sprite.

However, the issue of including this data in the sprite could be a feature to be implemented in the future.
For now:
1. The rgbmap algorithm and fit criteria will be defined as  _Preferences > Experimental > Color Quantization_. in the following cases:
    - New indexed sprite
    - Open an indexed sprite
    - Simple conversion RGB to indexed
2. In _Sprite > Color Mode > More Options_ you can choose the desired rgbmap algorithm and fit criteria.

dacap:
Merged in 3cc1c63274dac90360b8194f66101f86f1a670b6 and e1bd7990a3467cae8d49049685df493872e829b8 to resolve some conflicts with the latest `main`. Added 3d683ac6aa33dcb972df87f95ab70f172b3eeacb with a minor layout change. It works great :+1: (tested with [this image](https://community.aseprite.org/uploads/default/original/2X/3/347e8eb9f223710f8cda1a7e5429350f91041ff7.png) from the post and the [Apollo](https://lospec.com/palette-list/apollo) palette).

PR #4420, opened 2024-04-17
Fix don't paint transparent color via Eraser tool in Background layers (fix #4370)

### LINKED ISSUE #4370: Downloaded PNG in RGB mode fails to support transparency: erase uses secondary color and export PNG replaces transparent color with black
I often have this issue on images downloaded directly from the Internet, typically on The Spriters Resource. This happens when opening the PNG directly in Aseprite (not creating a new file then copying image rectangles). Note that opening a PNG sets the Color mode to **RGB** by default, so this is not an issue related to using Indexed Color mode (and even then, behavior differs from real Indexed Color mode).

Since spritesheets are often distributed with a transparency index color, I must manually remove the background color (by either using the magic wand + delete, paint bucket with transparent color or, when some background pixels are inside the sprites, the eraser + right-click to replace background color to transparent color (added myself to the palette).

However, I noticed the following:
- the eraser, bucket fill with transparent color and delete actions **replace the color with the secondary color**, as mentioned in https://github.com/aseprite/aseprite/issues/196 (very old but same symptom) and https://community.aseprite.org/t/bug-using-erase-causes-secondary-color-to-be-applied/4076
- the eraser + right-click to replace is the only way I found to actually replace background color with transparent color
- after using eraser + right-click to get actual transparency, the editor will show the "transparent grid" properly, but on export to PNG, it will generate **black instead of transparent color**

It seems that somewhat the PNG metadata cannot be changed even after saving the file as an .aseprite file.

Switching to Indexed Color mode and back to RGB doesn't help. Switching to Indexed Color mode and defining background color as the index color doesn't work either. All 3 color modes have the issue.

**Repro steps**

1. Download a spritesheet from the net, such as https://www.spriters-resource.com/master_system/sonicthehedgehog/sheet/5859/
2. Open in directly in Aseprite (Open with... or drag and drop)
3. Verify that it is RGB mode
4. Try the eraser or bucket fill with transparent color => fills with secondary color
5. Select background color as primary color. Add transparent color to palette, select as secondary color. Use the eraser + right-click with big brush size to replace background with transparency => it works
6. Export PNG => background is now black
7. Try different Color modes and repeat

**Workaround**

Create a brand new file in Aseprite in RGB mode, and copy image content from the PNG into that file, instead of opening the PNG itself in Aseprite.

Here is an example with a standard export on the left, and creating a new RGB file and copying content then exporting on the right:

image

* Aseprite version: v1.3.2 (official)
* System: Linux Ubuntu 22.04

### Other related threads

https://community.aseprite.org/t/my-eraser-draws-black-color-instead-of-erasing/2556/4

Gasparoken:
Original issue #4370 description:
"Downloaded PNG in RGB mode fails to support transparency: erase uses secondary color and export PNG replaces transparent color with black"

fix #4370

### DIFF
--- a/src/app/app.cpp
+++ b/src/app/app.cpp
@@ -894,7 +894,37 @@ int app_get_color_to_clear_layer(Layer* layer)
       color = ColorBar::instance()->getBgColor();
     else
 #endif
-      color = app::Color::fromRgb(0, 0, 0); // TODO get background color color from doc::Settings
+    {
+      auto c = Preferences::instance().colorBar.bgColor();
+      if (layer->sprite()->pixelFormat() == IMAGE_INDEXED) {
+        if (c.getType() == Color::IndexType) {
+          color =
+            (c.getIndex() < layer->sprite()->palette(0)->size() ?
+              c :
+              app::Color::fromIndex(0));
+        }
+        else {
+          Palette* pal = layer->sprite()->palette(0);
+          color_t bg;
+          if (c.getType() == Color::RgbType) {
+            bg = rgba(c.getRed(), c.getGreen(), c.getBlue(), 255);
+            bg = pal->findBestfit(rgba_getr(bg),
+                                  rgba_getg(bg),
+                                  rgba_getb(bg), 255, -1);
+          }
+          else {
+            ASSERT(c.getType() == Color::GrayType);
+            bg = graya(c.getGray(), 255);
+            bg = pal->findBestfit(graya_getv(bg),
+                                  graya_getv(bg),
+                                  graya_getv(bg), 255, -1);
+          }
+          color = app::Color::fromIndex(bg);
+        }
+      }
+      else
+        color = app::Color::fromRgb(0, 0, 0);
+    }
   }
   else // All transparent layers are cleared with the mask color
     color = app::Color::fromMask();
--- a/src/app/tools/ink_processing.h
+++ b/src/app/tools/ink_processing.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2023  Igara Studio S.A.
+// Copyright (C) 2019-2024  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -570,6 +570,12 @@ public:
   ReplaceInkProcessing(ToolLoop* loop) {
     m_color1 = loop->getPrimaryColor();
     m_color2 = loop->getSecondaryColor();
+    if (loop->getLayer()->isBackground()) {
+      switch (loop->sprite()->pixelFormat()) {
+        case IMAGE_RGB: m_color2 |= rgba_a_mask; break;
+        case IMAGE_GRAYSCALE: m_color2 |= graya_a_mask; break;
+      }
+    }
     m_opacity = loop->getOpacity();
   }
 
--- a/src/app/script/app_object.cpp
+++ b/src/app/script/app_object.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2023  Igara Studio S.A.
+// Copyright (C) 2018-2024  Igara Studio S.A.
 // Copyright (C) 2015-2018  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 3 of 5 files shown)


martincapello (commented):
The fix works as intended. I'm just not sure if it works as expected. Because the paint bucket behaves different when trying to paint a background layer with a color that has an alpha < 255. For instance, with this fix, a full transparent black will be painted as black by the eraser but as nothing (no color is painted) by the paint bucket.

Gasparoken:
@martincapello and I verified the correct behavior of Paint Bucket Tool (there was only one confusion about the ink used during testing).

martincapello (approved):

src/app/tools/inks.h
```diff
@@ -366,7 +366,8 @@ class EraserInk : public BaseInk {
       }
       case ReplaceFgWithBg:
         loop->setPrimaryColor(loop->getFgColor());
-        loop->setSecondaryColor(loop->getBgColor());
+        loop->setSecondaryColor(loop->getBgColor() |
+          (loop->getLayer()->isBackground() ? ColorAMask : 0));
```
dacap, line 370: This works only for RGB images, but here `getBgColor()` and `setSecondaryColor()` could be handling an index or a grayscale color (the bug is reproducible in a grayscale image even with this patch).

Gasparoken:
I'll add tests for this.

Gasparoken:
I'm testing this fix. It's not able to review yet, anyway could you give me an opinion about the approach @dacap? (mainly, `ink_processing.h`, - I already changed `palette(0)` -> `palette(loop-> getFrame()` -)

martincapello (approved):
LGTM!

src/app/tools/ink_processing.h
```diff
@@ -143,6 +143,18 @@ class CopyInkProcessing : public SimpleInkProcessing<CopyInkProcessing<ImageTrai
...
+        case IMAGE_INDEXED: {
+          Palette* pal = loop->sprite()->palette(loop->getFrame());
+          color_t c = pal->getEntry(m_color);
+          m_color = (rgba_geta(c) < 255 ?
+                      pal->findBestfit(rgba_getr(c),
+                                       rgba_getg(c),
+                                       rgba_getb(c),
+                                       255,
+                                       -1) :
+                      m_color);
```
dacap, line 155: The issue here (super improbable case but possible) is that we lost the original index of the secondary color, e.g. try to load this file in this PR:
 
[try-to-paint-with-index-2.zip](https://github.com/user-attachments/files/16640741/try-to-paint-with-index-2.zip)

and try to paint with index 2, you will get index 1, and this is the most simple case in indexed color mode (as the "copy" in CopyInkProcessing says): we have to just copy the same index color.

I think indexed mode with transparent colors is a special case were we shouldn't take care in these cases, i.e. the background layer can contain specific indexes that might be transparent/semi-transparent (probably something to fix in the rendering side in a future, or just leave it as it is).

src/app/app.cpp
```diff
@@ -894,7 +894,27 @@ int app_get_color_to_clear_layer(Layer* layer)
...
+                                     rgba_getb(c),
+                                     255,
+                                     -1) :
+                    c);
+          color = app::Color::fromIndex(c);
+        }
+      }
+      else
+        color = app::Color::fromRgb(0, 0, 0);
+    }
```
dacap, line 917: Probably we can just return `Preferences::instance().colorBar.bgColor()` converted to the sprite color mode (?). Not sure.

Gasparoken: Mmh, sounds good to me. I changed this to simply "convert bgColor preference to sprite color mode" as you recommended. Also, assign an index = 0 when `Preferences::instance().colorBar.bgColor()` is beyond the palette size.

Gasparoken:
@dacap you can check it out when you can.

dacap:
Merged in 4a91d150af97e70e588e82d698dc8c5e0db8e6fe with the following changes: afb8a3d94a328a2b525b0edf838abff1eabd32c1 + 240d481645ccc0b4c8cda8347e6fc95d35291ab1 + 9429d915ae3cee9c0d21a9ea54bac39e544be932

PR #4421, opened 2024-04-17
Fix crashing segfault when calling: app.command.CloseFile() or app.command.GotoPreviousTab() , in --batch mode (fix #4352)

### LINKED ISSUE #4352: Crashing segfault when calling: app.command.CloseFile() or app.command.GotoPreviousTab() , in --batch mode
https://github.com/aseprite/aseprite/assets/34615798/ce986c7d-aaba-45fb-a850-30ab0bf3b827

    - v1.3 master branch
    - commit: `73fe8099d96479364f92ff6317a5687498d43d36`
* System: Arch Linux

lua script:
```lua
-- these aren't the problem but are included to simulate my big script that has this issue in case.
--app.open(app.activeSprite.filename)
--app.open(app.activeSprite.filename)

-- running either of these in batch mode segfaults!
--app.command.GotoPreviousTab()
app.command.CloseFile()
--app.command.GotoNextTab()
--app.command.CloseAllFiles()
```

command:
```
aseprite image.png --script "test_crash.lua" --batch
```

NOTE: It crashes even without the `app.open()` calls but I included it in there because I have a big script that opens 3-4 tabs so I thought it was a memory limit issue.

Gasparoken:
fix #4352

### DIFF
--- a/src/app/commands/cmd_close_file.cpp
+++ b/src/app/commands/cmd_close_file.cpp
@@ -1,4 +1,5 @@
 // Aseprite
+// Copyright (C)      2024  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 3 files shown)


dacap:
I'll merge this as it is because it fixes #4352. As a future task we'll have to implement these commands for batch processing but it's unrelated to this fix :+1: and it needs a kind of refactor (probably several commands should be handled by the `app::Context` and the `app::UIContext` override the behavior to deal with the `Workspace`, but for batch using a simpler default implementation).

Issue #3779, opened 2023-03-30
Cmake CMAKE_OSX_SYSROOT flag is incorrect in INSTALL.MD


sinistersnare:
Describe your bug report or feature request here

The [docs for installing on MacOS](https://github.com/aseprite/aseprite/blob/main/INSTALL.md#macos-details) use the `CMAKE_OSX_SYSROOT` flag, which seems to be out of date. On my computer, I found via `xcrun --show-sdk-path`:

```sh
-DCMAKE_OSX_SYSROOT=/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk
```

Maybe its just me, but I think the docs should mention to run that command to find the appropriate path.

### Aseprite and System version

* Aseprite version: commit 733f7ec (main for me)
* System: MacOS Monterey, M1.

Happy to make the change, just wanted to make a tracking issue so I can get to it later.

martincapello:
Thank you for creating an issue for this. We will evaluate it. Feel free to create a PR, just remember that you will need to sign our CLA first to let us merge your PR:
https://github.com/aseprite/aseprite/blob/main/CONTRIBUTING.md#forking--pull-requests

PR #4436, opened 2024-04-23
fix deprecated doc error in macOS section of INSTALL.md

melanihodge:
Fix the deprecated documentation for the INSTALL.md file according to this open issue ([Cmake CMAKE_OSX_SYSROOT flag is incorrect in INSTALL.MD #3779](https://github.com/aseprite/aseprite/issues/3779)).

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -163,29 +163,48 @@ More information in [issue #2449](https://github.com/aseprite/aseprite/issues/24
 
 ## macOS details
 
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @melanihodge but I don't see a significant improvement in the guide with this change. We could just add the `xcrun --show-sdk-path` in the note about the `CMAKE_OSX_SYSROOT` location.

Issue #4123, opened 2023-11-06
[Feature Request] Provide support for requiring c libraries in scripts


zarstensen:
Currently, loading C or C++ shared libraries in scripts is not fully supported in Aseprite. It is possible to load these types of libraries on Windows if you fiddle with the package.cpath and package.searchers values, or load them directly using package.loadlib, however it would be ideal to fully support loading shared libraries out of the box. Additionally, it is impossible to load shared libraries on Linux or MacOS, as lua is not linked against 'dl' on these platforms.

Supporting loading shared libraries would enable development of more advanced scripts, as it would allow them to make use of features lua does not support (ex. multi threading), or allow usage of popular lua packages, which require shared libraries to function (ex. LuaSocket).

I would be able to open a PR of a [possible implementation](https://github.com/zarstensen/aseprite/tree/script-c-library-loading), if this feature is something you would like to see in Aseprite.

dacap:
Hi @zarstensen, actually the main issue with loading shared libraries are that 1) you have to highly trust the other party, and 2) we'd need to offer some kind of SDK to compile the libraries with the exact same Lua implementation.

At the moment this is low priority as we'd prefer to make the scripting Lua API bigger instead.

MinekPo1:
sorry for posting on a dead thread, but it appears that the main hurdle in loading c libraries currently is the way aseprite includes lua headers requires that the library be also compiled with c++. it appears (see for example [lua-users.org – "Re: lua.h and c++" by David Burgess on 2004-11-28](http://lua-users.org/lists/lua-l/2004-09/msg00566.html)) that the lua developers intend on includes being wrapped in a `extern "C"` block. [a comment in aseprite source code](https://github.com/aseprite/aseprite/blob/main/src/app/script/luacpp.h#L12-L14) suggests not putting all includes in an `extern "C"` block is intentional, not sure if it was considered that this would be the result. `extern "C"` also does not in any way hinder c++ error handling from what I understand.

The source releases of lua include a header called `lua.hpp`, with the content as shown in the mailing list linked above, though for some reason it does not appear to be present in the github mirror. If putting all includes in an `extern "C"` block is unwanted, it appears one can instead add `#define LUA_API extern "C"`, to only apply the modifier to functions meant as part of the lua ABI.

[...]

PR #4445, opened 2024-05-01
[lua] FFI function binding

lampysprites:
Hello! I saw #4123, and had something done about that locally by compiling cffi statically with lua. Which is close to what luajit has - alternative to "luaopen_xxxxx" type of shared libraries, and unlike it, it binds normal dll's without initializing them with lua state. It's still a workaround though, which might end up bothersome to support.

Changes:
- added `ffi` global to lua, which is the [cffi-lua module](https://github.com/q66/cffi-lua/blob/master/docs/api.md)
- added libbfi and cffi-lua submodules, ~~both of which i patched a bit - cmake file in libffi, and some overloads in cffi-lua that don't pass msvc.~~ (upd: no longer needed)
- added `ENABLE_FFI` compilation flag

Benefits are pretty obvious:
- compared to loadlibrary, ABI and compiler compatibility becomes irrelevant.
- compared to interpreter, way way faster execution - namely, for anything that requires reading/writing image data. 
- allows to do a whole lot of things with only limitation that the application state is not shared.

Downsides:
- no security whatsoever, as pointed out by @ dacap on #4123:
- loadlib stil won't work, thus neither will already existing lua modules.
- it's harder than scripting - requires some basic understanding of C and memory management. And even though it's copy-paste, it's still bothersome to create bindings for libraries that use transparent pointers.
- crashes are very easy to enjoy in just a few lines of code.
- dependency on libdl on linux - could that affect distribution?

Example/tests:
Source and compiled version: [ffi_test.zip](https://github.com/aseprite/aseprite/files/15179469/ffi_test.zip) - includes minimal example; image modification; using external library (miniaudio)

overall a binding can look like this:

```lua
local ffi_test = ffi.load("ffi_test.dll")
ffi.cdef("void grayscale(unsigned char * bytes, unsigned long length);")
local function grayscale(image)
  local bytes = image.bytes
  local sz = image.rowStride * image.height
  ffi_test.grayscale(ffi.cast("unsigned char *", bytes), ffi.cast("unsigned long", sz))
  image.bytes = bytes
end

```

-----------------


lampysprites:
sorry abt the mess - i used wrong patch of cffi-lua; but macos issue is something else and it'll take a bit of looking

lampysprites:
Issues with compilation fixed, hopefully. 
Also figured how to keep build scripts separately (`third_party/ffi-cmake`) - everything now works with upstream repos without my oonga-boonga patches.
The last remaining addition i can think of is adding a flag for linking to shared libffi, like it is done for some deps. But also it is how it is not done for other deps, so tell me.

dacap:
Hi @lampysprites, are you still interested in this PR and FFI solution? We'd like to review it soon (sorry for the late response :cry:)

lampysprites:
Hiii @dacap! It's same as before, an experiment. Nothing crirical depends
on it, and with inevitable stability and safety problems I think the cons
outweigh the pros, by a lot. A loooot.

Although if you see a point in keeping it, I am happy to help updating it
to the current state.

But right now I'm not using this hack myself. It was for a
calculation-heavy processing back then but figured something else to avoid
rebuilding the editor with every update.

I do admit though that lua performance can be infuriating. Tried to add
median filter (off-screen) not long ago and it took seconds. I don't think
that's anywhere close to common use, more like an example of a thing that
led me to trying FFI.

ckaiser:
@lampysprites One of the things we're keen to explore with @dacap is a better way to let people implement custom image formats, the API is done but as you say Lua performance can be iffy and being able to leverage existing code/libraries would be ideal, so having a way to let extension authors call native code more easily is something I'll be looking into, if not exactly this PR then probably something similar

lampysprites:
Custom formats are a great feature to have indeed.
In that case I'll check if it still works tomorrow-ish. I think you have to
experience using it first.

lampysprites:
Hi! I updated this pr to main branch to the point calling a dll works, but also broke the build script a bit because for cffi-lua makefile is outside the repo. There's also some kind of a bug in grayscale in that dll evidently.

PR #4446, opened 2024-05-02
Fix scripting app.command.Cut() and app.command.Paste() does not work in --batch mode (fix #4354)

### LINKED ISSUE #4354: Scripting app.command.Cut() and app.command.Paste() does not work in --batch mode
I'm trying to do a magic wand pixel selection on 1 layer and cut and paste to a new layer. 

This is the minimum reproducible version because my real script is more complicated than this.

https://github.com/aseprite/aseprite/assets/34615798/744ebd07-3c03-45df-8aa5-64fe026e92af

```lua
local sprite = app.activeSprite
local from_layer = sprite.layers[1]
local frame_num = 1
local from_cel = from_layer:cel(frame_num)

app.useTool {
    tool = "magic_wand",
    layer = from_layer,
    cel = from_cel,
    button = MouseButton.LEFT,
    points = {from_cel.position},
    selection = SelectionMode.REPLACE
}
app.command.Cut()
local to_layer = sprite:newLayer()
app.activeLayer = to_layer
app.activeFrame = to_layer.sprite.frames[1]
app.activeLayer = to_layer
app.command.Paste()

local without_ext = app.fs.filePathAndTitle(app.activeSprite.filename)
app.command.SaveFile {
    ["filename"] = without_ext .. ".ase",
    ["filename-format"] = ".ase",
}
```

### How to reproduce:

Extract folder and open command prompt in directory and run:
[one_pixel_paste.zip](https://github.com/aseprite/aseprite/files/14457020/one_pixel_paste.zip)
```
aseprite.exe ./one_pixel.png ./test_cut.lua
```

    - v1.3 master branch
    - commit: `73fe8099d96479364f92ff6317a5687498d43d36`
* System: Arch Linux

Gasparoken:
It fixes the original problem, but I need advice on this approach.

fix #4354


Gasparoken:
Ready for advice.

src/app/commands/cmd_cut.cpp
```diff
@@ -10,10 +11,16 @@
...
+#include "app/context_access.h"
+#include "app/tx.h"
 #include "app/ui/input_chain.h"
+#include "app/util/clipboard.h"
+#include "app/util/range_utils.h"
 
 namespace app {
 
+class Clipboard;
+
```
dacap, line 23: This is not needed because we are already including `"app/util/clipboard.h"`.

src/app/commands/cmd_cut.cpp
```diff
@@ -30,12 +37,18 @@ CutCommand::CutCommand()
 
 bool CutCommand::onEnabled(Context* ctx)
 {
-  return App::instance()->inputChain().canCut(ctx);
+  return App::instance()->isGui() ?
```
dacap, line 40: Use `ctx->isUIAvailable()` when possible:
```suggestion
  return ctx->isUIAvailable() ?
```

src/app/commands/cmd_cut.cpp
```diff
@@ -30,12 +37,18 @@ CutCommand::CutCommand()
...
+    App::instance()->inputChain().canCut(ctx) : true;
```
dacap, line 41: Is this true? Or should we check that `ctx->clipboard()`? Should `Context` offer a default input chain?

src/app/commands/cmd_paste.cpp
```diff
@@ -30,12 +33,17 @@ PasteCommand::PasteCommand()
 
 bool PasteCommand::onEnabled(Context* ctx)
 {
-  return App::instance()->inputChain().canPaste(ctx);
+  return ctx->isUIAvailable() ?
+    App::instance()->inputChain().canPaste(ctx) :
+    ctx->clipboard()->isImageAvailable();
```
dacap, line 38: Same case as in cut, probably we can implement a default input chain for batch processing.

src/app/util/clipboard.cpp
```diff
@@ -290,7 +290,7 @@ bool Clipboard::copyFromDocument(const Site& site, bool merged)
     (mask ? new Mask(*mask): nullptr),
     (pal ? new Palette(*pal): nullptr),
     nullptr,
-    true,                       // set native clipboard
+    App::instance()->isGui() ? true : false, // set native clipboard
```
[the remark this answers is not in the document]
dacap: I think we can just leave the `true` as it was (probably some batch scripts will want to copy some content to the native clipboard).

src/app/util/clipboard.cpp
```diff
@@ -487,7 +504,9 @@ void Clipboard::paste(Context* ctx,
 
     case ClipboardFormat::Image: {
       // Get the image from the native clipboard.
-      if (!getImage(nullptr))
+      if (App::instance()->isGui() && !getImage(nullptr))
+        return;
```
dacap, line 508: I think this change is not needed (?). Again a script could paste the native clipboard content if possible.

src/app/util/clipboard.cpp
```diff
@@ -805,6 +824,13 @@ ImageRef Clipboard::getImage(Palette* palette)
   return m_data->image;
 }
 
+bool Clipboard::isImageAvailable() const
+{
+  return m_data->image &&
+    m_data->image->width() > 0 &&
+    m_data->image->height() > 0;
+}
+
```
dacap, line 833: Probably this should go to some CliClipboardManager (which could implement an InputChain and be the default one) to handle all the cut/paste/copy logic for scripts (when UI and editors are not available). Not sure but I think it might be a possible design.

src/app/util/clipboard.cpp
```diff
@@ -379,6 +394,8 @@ void Clipboard::cut(ContextWriter& writer)
       tx.commit();
     }
     writer.document()->generateMaskBoundaries();
+    if (!App::instance()->isGui())
+      return;
 #ifdef ENABLE_UI
     update_screen_for_document(writer.document());
 #endif
```
dacap, line 401:
```suggestion
#ifdef ENABLE_UI
    if (writer.context()->isUIAvailable())
      update_screen_for_document(writer.document());
#endif
```

src/app/commands/cmd_paste.cpp
```diff
@@ -30,12 +33,17 @@ PasteCommand::PasteCommand()
...
 }
 
 void PasteCommand::onExecute(Context* ctx)
 {
-  App::instance()->inputChain().paste(ctx);
+  if (ctx->isUIAvailable())
+    App::instance()->inputChain().paste(ctx);
+  else if (ctx->clipboard())
+    ctx->clipboard()->paste(ctx, false);
```
dacap, line 46: Probably we required a `origin` argument for the paste (at least initially) to paste the image in any position:

```c++
struct PasteParams : public NewParams {
  Param<gfx::Point> origin { this, gfx::Point(), "origin" };
};

class PasteCommand : public CommandWithNewParams<PasteParams> {
public:
  PasteCommand();
...
```

src/app/util/clipboard.cpp
```diff
@@ -364,6 +364,21 @@ void Clipboard::cut(ContextWriter& writer)
...
+            doc::Image* image(new_image_from_mask(site, true));
+            setData(image,
+                    (mask ? new Mask(*mask): nullptr),
+                    (pal ? new Palette(*pal): nullptr),
+                    nullptr,
+                    false,
+                    site.layer() && !site.layer()->isBackground());
+          }
+        }
+      }
```
dacap, line 381: Why is this required? Didn't `copyFromDocument()` make the copy?

dacap (changes requested):
Some comments to complete the implementation.

Gasparoken:
@dacap When you are available could you give me your opinion on this new approach?

src/app/cli/default_cli_delegate.cpp
```diff
@@ -143,6 +145,8 @@ void DefaultCliDelegate::exportFiles(Context* ctx, DocExporter& exporter)
 int DefaultCliDelegate::execScript(const std::string& filename,
                                    const Params& params)
 {
+  ScriptInputChain scriptInputChain;
+  App::instance()->inputChain().prioritize(&scriptInputChain, nullptr);
```
dacap, line 149: I'm not 100% sure, probably we should do this only when `-batch` was specified (?)

Remember that we can specify something like `aseprite.exe file.aseprite -script script.lua`

You can test if the script is executed with the UI/Editor/DocView available in that case (and using the `ScriptInputChain` might broke that case where the `DocView` can handle the `InputChain`).

dacap (changes requested):
Some changes, I think this is a proper implementation. Delegating the cut/copy/paste commands to a special InputChainElement. Made some comments to review.

src/app/commands/cmd_undo.cpp
```diff
@@ -74,51 +74,53 @@ void UndoCommand::onExecute(Context* context)
...
-  // Get the stream to deserialize the document range after executing
-  // the undo/redo action. We cannot yet deserialize the document
-  // range because there could be inexistent layers.
   std::istream* docRangeStream;
-  if (m_type == Undo)
-    docRangeStream = undo->nextUndoDocRange();
-  else
-    docRangeStream = undo->nextRedoDocRange();
+  if (context->isUIAvailable()) {
+    if (gotoModified) {
```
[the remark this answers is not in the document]
Gasparoken: You are right. Not necessary.

src/app/util/clipboard.cpp
```diff
@@ -190,7 +190,7 @@ Clipboard::~Clipboard()
 
 void Clipboard::setClipboardText(const std::string& text)
 {
-  if (use_native_clipboard()) {
+  if (use_native_clipboard() && App::instance()->isGui()) {
```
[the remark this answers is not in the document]
Gasparoken: I did this because in some places after 'if(use_native_clipboard())' I found calls to the 'set/getNativeBitmap' or 'setNativePalette' functions which in turn call the 'native_window_handle()' which crashed without the UI.
So to unify criteria, I used (use_native_clipboard() && App::instance()->isGui()) in every place where use_native_clipboard was used.
I'll see what I can do with this.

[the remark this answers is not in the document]
dacap: I think that we can fix `native_window_handle()`:
```
  void* native_window_handle() {
    return os::instance()->defaultWindow()->nativeHandle();
  }
```
returning `nullptr` if the `defaultWindow()` is nullptr.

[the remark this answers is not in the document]
Gasparoken: Thanks David, it's the best solution.

PR #4453, opened 2024-05-06
Fix duplicate shortcut key configuration items in Keyboard shortcuts (fix #4387)

### LINKED ISSUE #4387: Duplicate shortcut key configuration items in Keyboard shortcuts Window
From  https://github.com/aseprite/languages/issues/29#issuecomment-1848991843

aseprite version: 1.3.5
The problem still exists.

image
image

~~There seems to be some issues with the UI here. The column `action` width is not enough and cannot be adjusted, causing text to be obscured.~~

Gasparoken:
fix #4387


martincapello (commented):
I haven't tested this, but seems that it should work because it removes duplicates. I'm just concerned about the underlaying cause of the duplicates. Maybe we could try to avoid duplicates in the first place?

Gasparoken:
> I haven't tested this, but seems that it should work because it removes duplicates. I'm just concerned about the underlaying cause of the duplicates. Maybe we could try to avoid duplicates in the first place?

Most duplicates appear because some commands have "params" and their "friendlyString" does not concatenate those specific parameters.
For example TiledMode has:
```
TiledMode::NONE
TiledMode::X_AXIS
TiledMode::Y_AXIS
TiledMode::BOTH
```
But in the command list box is showed:
```
Tiled Mode
Tiled Mode
Tiled Mode
Tiled Mode
Tiled Mode
```
Other duplicates I still don't know why they are duplicated (one of them is TiledMode::NONE which appears twice)

So, I'll concatenate the param string to the base command.

The following commands has duplicates:
```
Change Color Mode: Indexed
Contract Selection
Export Sprite Sheet
Flip Canvas Horizontally
Frame Properties
Go to Next Frame
Go to Previous Frame
Launch
Layer Properties
Load Palette
Open Browser
Open Sprite
Playback Speed 1x
Run Script: …
Save Palette
Select Used Colors
Set Loop Selection
Set Palette Entry Size
Take & Open Screenshot (sRGB Color Profile
Tiled Mode
Tileset Mode: Auto
Zoom 600%
Zoom 800%
Zoom 1600%
```

Gasparoken:
@martincapello What do you think about this approach?
Pending duplicates:
```
Contract Selection
Frame Properties
Go to Next Frame
Go to Previous Frame
Layer Properties
Load Palette
Playback Speed 1x
Select Used Colors
Set Loop Selection
Set Palette Entry Size
Take  Open Screenshot (sRGB Color Profile
Tileset Mode: Auto
```

Gasparoken:
New approach to review. This will compile if https://github.com/aseprite/laf/pull/88 is implemented first.

Pending:
There are 4 items that depend on 'keyContext', i.e. the difference between key Items are only the context:
'Set Loop Section'
'Go to Next Frame'
'Go to Previous Frame'
'Layer Properties' (in this case by default, there are 2 key board shortcuts for the same command 'F2' and 'Shiift+P')

 To go ahead I need to know if this approach is OK.

dacap:
I'll review this layer.

Gasparoken:
> I'll review this later.

Ok, I will just push-force the last movement of the removeEscapeCharFromText() function from 'laf/base/string' to 'ui/intern'

dacap:
I'll take this from here, rebase it to the latest main refactors, and test it.

I'll refactor the `isSkipListing` to something like `isShortcutListed` (with a new `onShortcutListed`). This is to follow the enabled/checked logic which test for true, instead of false (as this "skip listing").

dacap:
@Gasparoken call me when you are free to know how to approach this.

PR #4461, opened 2024-05-11
Update INSTALL.md

AhmedThe13thUser:
correcting the Windows details part.


dacap:
Hi @AhmedThe13thUser, thanks for your contribution. Actually I'm not sure what anti-virus are you using but I'm using Avira and doesn't require to be turned off (generally I'd not recommend that).

PR #4465, opened 2024-05-13
[lua] suggestion: move canvas behind foreground dialog

lampysprites:
It would be useful to allow moving and zooming on canvas when showing dialog in foreground, similar to built-in filter dialogs, e.g. like so:
```lua
d = Dialog()
d:show{ move = true }
```
Without it things work same as before. It's really nice to have for previewing color adjustment or other intermediary changes to the image.

There's one problem though: it's still possible to use ctrl key to move layers which feels like a problem. Filters avoid this because the sprite is locked ("Sprite is used by backup/data recovery task") but I'm not sure it's okay to do so here because the script also needs to use it? I wonder if there is a simple solution to this.

-----------

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -25,6 +25,10 @@
 #include "app/ui/expr_entry.h"
 #include "app/ui/filename_field.h"
 #include "app/ui/main_window.h"
+#include "app/ui/editor/editor.h"
+#include "app/ui/toolbar.h"
+#include "app/ui/context_bar.h"
+#include "app/tools/tool_box.h"
 #include "base/paths.h"
 #include "base/remove_from_container.h"
 #include "ui/box.h"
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @lampysprites, I'll try to merge this tomorrow, probably with few changes. E.g. Instead of `move=true`, the parameter could be `moveTool=true` or `moveCanvas=true`, I have to think about it (the `move` alone looks like something related to the `Dialog`, e.g. `move=false` could lock the dialog movement).

dacap:
Thanks again lampy, I'll merge this as it is and make some adjustments in the following commits (syntax/formatting, renaming `move` param, and disabling the <kbd>Ctrl</kbd> modifier).

dacap:
The parameter will be named `hand` (for the Hand tool):
```lua
local d = Dialog()
d:show{ hand=true }
```

PR #4500, opened 2024-05-24
DPaint2 style "spread" modifier on shape tools

patrickmeehan:

### DIFF
--- a/data/gui.xml
+++ b/data/gui.xml
@@ -671,6 +671,7 @@
       <!-- Modifiers for two-or-more-points tools -->
       <key action="MoveOrigin" shortcut="Space" />
       <key action="RotateShape" shortcut="Alt" />
+      <key action="Spread" />
 
       <!-- Without default shortcuts -->
       <key action="LeftMouseButton" />
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -943,6 +943,7 @@ move_origin = Move Origin
 square_aspect = Square Aspect
 draw_from_center = Draw From Center
 rotate_shape = Rotate Shape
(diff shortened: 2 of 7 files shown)


dacap:
One thing I forgot to mention, I like the idea to being possible to "stamp/stroke" the shape we are making without ending the loop, but I'm not sure if the "spread" term makes too much sense for this right now (it looks like a DPaint term era).

I was thinking that probably a nice idea would be that the Stroke command (<kbd>S</kbd> key) could take this role, so calling the Stroke command in the ToolLoop could stamp the shape, the same for the Fill command (<kbd>F</kbd> key) to fill the shape.

Anyway I've just tested this PR and the spread works in a different way which might be useful for some users, so probably we could keep this as it is (without a default keyboard modifier initially). The only problem I've found with this is that releasing the Spread modifier does't repaint the whole editor to invalidate/repaint the old areas. E.g.:

https://github.com/aseprite/aseprite/assets/39654/e23cb49c-83c9-47b4-a173-b3f5a83cc8a9

patrickmeehan:
Thanks, David! I'll hunt down that bug and update the PR as soon as I have a moment to get back to it. Also, I didn't see the CLA in my inbox yet. Will it be coming from Igara's domain? I will re-apply for it if I don't see it in the next few days.

dacap:
> Will it be coming from Igara's domain? 

Hi @patrickmeehan, there should be a mail to sign the CLA titled "Igara Studio needs your signature for the Individual CLA" from "David Capello [mail@signnow.com](mailto:mail@signnow.com)" to finish the CLA signing process.

PR #4506, opened 2024-05-27
Improved list-layers command

requizm:
Old output was not respecting group names. 

Old:
```
layer1
layer1.1
layer2.1
```

New:
```
group1/
  layer1.1
  layer1.2
  group1.1/
    layer1.1.1
group2/
  layer2.1
```


src/doc/layer.cpp
```diff
@@ -540,6 +540,20 @@ void LayerGroup::allTilemaps(LayerList& list) const
   }
 }
 
+std::string LayerGroup::visibleLayerHierarchyAsString(std::string indent) const
```
requizm, line 543: This might be not correct place for this function, I have doubts but I couldn't find a good place though

src/app/cli/default_cli_delegate.cpp
```diff
@@ -63,8 +63,7 @@ void DefaultCliDelegate::afterOpenFile(const CliOpenFile& cof)
     return;
 
   if (cof.listLayers) {
-    for (doc::Layer* layer : cof.document->sprite()->allVisibleLayers())
-      std::cout << layer->name() << "\n";
+    std::cout << cof.document->sprite()->visibleLayerHierarchyAsString() << "\n";
```
requizm, line 66: Why are we not creating a new command? I think unnecessary. If the user wants to see _only_ layers, it can check with last `/` character

dacap:
Hi @requizm, thanks for your PR. Could you please check your mail box? There should be a mail to sign the CLA titled "Igara Studio needs your signature for the Individual CLA" from "David Capello <mail@signnow.com>" to finish the CLA signing process.

requizm:
@dacap done

src/doc/layer.cpp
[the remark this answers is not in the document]
requizm: `misc-no-recursion`? I saw lots of recursive functions in the code base 🐸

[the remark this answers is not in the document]
dacap: This check is a warning just to warn that non-recursive functions are preferred. I think this "visibleLayerHierarchyAsString()" is super specific for the CLI. Anyway probably I should impl a generic `for_each_layer()` for these cases.

I can take that job after merging the PR (if we do merge it).

src/app/cli/default_cli_delegate.cpp
dacap: I did a search in GitHub and there are places where changing the --list-layers behavior can break existing scripts (e.g. https://github.com/ZenCoderGames/ZRoguelikeGame/blob/6b21228a98c33bf8a5ef8251e564f85c226ebc66/addons/AsepriteWizard/aseprite/aseprite.gd#L118)

I think that for any change to this behavior we should add a new parameter.

dacap:
@requizm could you please fix your commit message to something like this:

```
[cli] Add -list-layer-hierarchy option

Example output:
group1/
  layer1.1
  layer1.2
  group1.1/
    layer1.1.1
group2/
  layer2.1
```

The most important thing is a blank line between the summary and the description of the commit.

requizm:
@dacap is this going to be merged? if yes, what is the release schedule?

dacap:
@requizm I'll review and test this now so we can merge it. Once it's merged it should be released in [v1.3.8](https://github.com/aseprite/aseprite/milestone/56)

dacap:
Just merged, thanks @requizm 👍

Issue #3619, opened 2022-12-10
Texture atlas export/import from the UI


dacap:
We should have in the UI:

- [ ] A way to specify several sprites in the "Export Sprite Sheet" dialog and create the texture atlas + JSON file
- [ ] A way to import that same JSON to create the original sprites

This should work regardless if we specified -split-tags, or -split-layers, etc. In this way gamedevs will be ready to take care of all cases in a proper/standarized way (at the moment there are cases that are not correctly handled like #3611).

Issue #3611, opened 2022-12-03
CLI: frame tags incorrect when exporting a spritesheet from multiple files


alexgorcrom:
While running the bellow command (via gitbash for the path niceties) over two different files, each with a single layer and multiple tags:
```
 ~/Downloads/Aseprite-v1.2.40/Aseprite-v1.2.40/Aseprite.exe -b king.aseprite silver_knight.aseprite --sheet ../spritesheets/characters.png --scale 2 --sheet-pack --list-tags --filename-format {tag}_{tagframe} --format json-array --data ../spritesheets/characters.json
```

The produced json has all the frames in the array, however the produces frame tags section in the `meta` object is problematic:
```
{
  ...
  "meta": {
    ...
    "frameTags": [
      { "name": "king_idle", "from": 0, "to": 7, "direction": "forward" },
      { "name": "king_run", "from": 8, "to": 15, "direction": "forward" },
      { "name": "king_death", "from": 16, "to": 21, "direction": "forward" },
      { "name": "king_attack1", "from": 22, "to": 25, "direction": "forward" },
      { "name": "king_attack2", "from": 26, "to": 29, "direction": "forward" },
      { "name": "king_attack3", "from": 30, "to": 33, "direction": "forward" },
      { "name": "king_hit", "from": 34, "to": 37, "direction": "forward" },
      { "name": "silver_knight_idle", "from": 0, "to": 10, "direction": "forward" },
      { "name": "silver_knight_run", "from": 11, "to": 18, "direction": "forward" },
      { "name": "silver_knight_attack1", "from": 19, "to": 25, "direction": "forward" },
      { "name": "silver_knight_attack2", "from": 26, "to": 32, "direction": "forward" },
      { "name": "silver_knight_death", "from": 33, "to": 43, "direction": "forward" },
      { "name": "silver_knight_hit", "from": 44, "to": 47, "direction": "forward" }
    ]
  }
}
```

From the above it is both the `king_idle` and `silver_knight_idle` have overlapping `from->to` frame numbers. This makes it pretty much impossible to extract which frames in the array are meant to be for which tag, when attempting to do something as follows (in peudo code):

```
json frames = json_get_array( root, "frames );
json frame_tags = json_get_array( json_get_obj( root, "meta" ), "frameTags")

for( tag in frame_tags ) {
  int from, to;
  if( tag.name == "<tag>" ) {
    from = tag.from;
    to = tag.to;
  } 
  for( i = from; i <= to; i++ ) {
[...]

PR #4512, opened 2024-05-31
Emit sprite UserData into data file

Nycto:

---

This change updates the JSON data file exports so it include the sprite level user data as part of the `meta` field. Without this change, the user data doesn't exist in the exported JSON file at all.


dacap:
Hi @Nycto, `app::DocExporter` cannot be a `doc::WithUserData` because a `doc::WithUserData` is a `doc::Object`. In case `app::DocExporter` must store a `doc::UserData`, it can just have a field to store the sprite user data.

Anyway there are several issues related to texture atlases where a redesign of the JSON format is required, e.g. when multiple sprites are exported to the same sprite sheet https://github.com/aseprite/aseprite/issues/3611#issuecomment-1345280312 and #3619

Issue #2173, opened 2019-10-11
Move Selected Slices Together with Selected Layers


lawrence-laz:
Currently we can select multiple slices and move them, then select relevant part of the image and move that, like so:
![aseprite-slice-request-example](https://user-images.githubusercontent.com/8823448/66669927-35cb2780-ec61-11e9-831e-0de9541b4e97.gif)

I would like to be able to do this in a single action. This would particularly be useful for organizing big sprite sheets. 

Maybe we could have some modifier key to move an image together with selected slices?

dacap:
Hi @lawrence-laz, yeah it would great to have that implemented, there were plans to add this functionality in the last slice improvements, but we need to implement the animation support first #721

PR #4533, opened 2024-06-12
Add "Transform" checkbox to slices and introduce moving slice content

### LINKED ISSUE #3419: Add slice tool toggle to transform content along with slice
Some times when I'm using slices I want to organize and move around the slices. It would be nice if the slice tool had a toggle to transform the currently selected layer's contents when the slice is being moved or scaled.

* Aseprite version: v1.3-beta16-x64, Steam beta
* System: Linux, 5.18.9, Fedora 36

martincapello:
Adds the checkbox that allows the user to enable the slice content transformation along the slice modification.

Fix #3419 and #2173


martincapello:
I've decided to change the approach to this feature because the PixelsMovement class has a lot of logic that it is not needed for this and also doesn't work well when multiple slices are transformed at the same time (for instance, the size of the slices doesn't match the areas being resized). Also, it seems that it wouldn't fit the 9-slices cases very well. So I'm taking some of the logic that I need from PixelMovement class and customizing it a bit. The https://github.com/aseprite/aseprite/pull/4533/commits/1efc563c3f6eef511bf613616c1d28eba1ae1243 commit has just the basics of the new approach, it is not complete yet, I'm still working on it.

martincapello:
At this point this are the missing things:
1) Tilemap layers support. 
2) 9-slices support.

With this approach the flickering is gone. Commits until this point should not be modified, so feel free to take a look at any time that works for you @dacap. I will add more commits addressing the missing bits.

martincapello:
Still missing the things mentioned, but at least some improvements were made, like reusing the existent ExpandCelCanvas instead of creating a new command. Which algo helps in getting the tilemap layer support done.

martincapello:
This is almost ready for review. I just have to test it more. The tilemap layer support is not tested at all with 9-slices yet.
Also, there is a crash when undoing transformations on a layer with an empty cel...

martincapello:
Okay, I think this is ready for review. There are still the following known issues:
1. Transforming slices when a mix of image layers and tilemap layers are selected AND tilemap mode is active makes ASSERTs fail in debug mode, and crashes the program in release mode. When the same is done but using pixels mode it seems to work fine.
2. There is no way to "pause" the slice transformation once started. When the user releases the mouse button the changes are applied. Hence each transformation is destructive and this is most noticeable in 9-slices transformations, because if someone wants to modify the center each drag & drop of it will destroy a bit more the original pixels, because on each drag the pixels are grabbed from the state they were in the last drop. Anyway, IMO for a first version of the feature I think it might be left as is.

Gasparoken:
I started to test this, and feels good! I'll start the code review.

About:
> 2\. There is no way to "pause" the slice transformation once started. When the user releases the mouse button the changes are applied. Hence each transformation is destructive and this is most noticeable in 9-slices transformations, because if someone wants to modify the center each drag & drop of it will destroy a bit more the original pixels, because on each drag the pixels are grabbed from the state they were left in the last drop. Anyway, IMO for a first version of the feature I think it might be left as is.

I think it might be a good idea to not scale the image inside the slice, just allow movement of the slice with its content (i.e. lock the offset between the image position and the origin of the 'slice' boundaries). I believe this because a very popular use of 'slices' such as remapping a 'slice' and its image content could lead to unintentional distortion of the image (due to scaling).
What do you think @dacap?

Gasparoken:
About:
> 1. Transforming slices when a mix of image layers and tilemap layers are selected AND tilemap mode is active makes ASSERTs fail in debug mode, and crashes the program in release mode. When the same is done but using pixels mode it seems to work fine.

I found that the selected layers is a mix of tilemap and image layers...
1. If the last selected layer is a Tilemap layer + TilemapMode::Tiles -> crash on Release (the problem you mentioned).
2. If the last selected layer is a Tilemap layer + TilemapMode::Pixels is OK!
3. If the last selected layer is an image layer, the image in the tilemap may disappear depending on where you place the "slice".

dacap:
> I think it might be a good idea to not scale the image inside the slice, just allow movement of the slice with its content (i.e. lock the offset between the image position and the origin of the 'slice' boundaries). I believe this because a very popular use of 'slices' such as remapping a 'slice' and its image content could lead to unintentional distortion of the image (due to scaling).

I didn't test this PR yet, are you saying that we should lock the resize of slices in this case? Or for some specific cases (9-slices) we shouldn't resize the center and do something different (e.g. like repeating a pattern)?

> I also think the active 'slice' should be a floating image and only stamp it when the 'slice' loses focus.

I think this would be the best implementation, probably too hard to implement but it should work similar to the selection movement: when the transformation loop / dropPixels() is done, we finish the transaction.

Martin said that as a first implementation it might work as this is, probably it's a good idea to merge this ASAP and then iterate with a better impl.

Gasparoken:
> I didn't test this PR yet, are you saying that we should lock the resize of slices in this case?

I was thinking that 'slice' movement is the most common use of this feature. On the other hand, scaling an image and its 'slice' at the same time would almost never be used, mainly because pixelart scaling can destroy detailed images.
So my conclusion is: restrict the 'slice' transformation to motion only and thus avoid unintentional modifications.
It's just an idea.

> Or for some specific cases (9-slices) we shouldn't resize the center and do something different (e.g. like repeating a pattern)?

Repeating the pattern inside the center of 9-slices is an interesting idea. It would be requested in a new issue.

> Martin said that as a first implementation it might work as this is, probably it's a good idea to merge this ASAP and then iterate with a better impl.

Yes, I agree.

src/app/cmd/clear_slices.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+      Doc* doc = static_cast<Doc*>(sc.cel()->document());
+      color_t bgcolor = doc->bgColor(sc.cel()->layer());
+
+      modify_tilemap_cel_region(
+        &m_seq, sc.cel(), nullptr,
+        gfx::Region(sc.mask.bounds()),
+        m_tilesetMode,
+        [sc, bgcolor](const doc::ImageRef& origTile,
+                        const gfx::Rect& tileBoundsInCanvas) -> doc::ImageRef {
```
Gasparoken, line 115: Minor indentation adjustment

src/app/ui/editor/moving_slice_state.cpp
```diff
@@ -9,19 +9,24 @@
...
 #endif
 
-#include "app/ui/editor/moving_slice_state.h"
-
 #include "app/cmd/set_slice_key.h"
+#include "app/cmd/clear_slices.h"
 #include "app/context_access.h"
 #include "app/tx.h"
 #include "app/ui/editor/editor.h"
+#include "app/ui/editor/moving_slice_state.h"
```
Gasparoken, line 17: The '#include' line from the header file corresponding to the current cpp file always appears at the top. This should be reverted.

src/app/ui/editor/moving_slice_state.cpp
```diff
@@ -48,34 +56,271 @@ MovingSliceState::MovingSliceState(Editor* editor,
     }
   }
 
+  editor->getSite(&m_site);
```
Gasparoken, line 59: When the selected layers are a mix of tilemap and image layers + if the last layer chosen in the timeline is a Image Layer --> the ColorBar does not have the TilesetMode buttons and the code assumes 'Manual' at this point. This causes a loss of drawing if the slice is moved outside the original area where the tiles were drawn in the tilemap layer.
```suggestion
  editor->getSite(&m_site);
  if (!m_site.layer()->isTilemap()) {
    m_site.tilemapMode(TilemapMode::Pixels);
    m_site.tilesetMode(TilesetMode::Auto);
  }
```

src/app/cmd/clear_slices.cpp
```diff
@@ -0,0 +1,152 @@
...
+void ClearSlices::restore()
+{
+  for (auto& sc : m_slicesContents) {
+    if (!sc.copy)
+      continue;
+
+    copy_image(sc.cel()->image(),
+              sc.copy.get(),
+              sc.cropPos.x,
+              sc.cropPos.y);
```
Gasparoken, line 147: Minor indentation adjustment required

src/doc/mask.cpp
```diff
@@ -129,6 +128,67 @@ void Mask::copyFrom(const Mask* sourceMask)
...
+        case IMAGE_GRAYSCALE: {
+          LockImageBits<GrayscaleTraits> grayBits(image);
+          auto grayIt = grayBits.begin();
+#if _DEBUG
+          auto grayEnd = grayBits.end();
+#endif
+          for (; maskIt != maskEnd; ++maskIt, ++grayIt) {
+            ASSERT(grayIt != grayEnd);
+            color_t c = *grayIt;
+            *maskIt = (graya_geta(c) >= 128); // TODO configurable threshold
```
Gasparoken, line 166: Same case as before:
```suggestion
            *maskIt = (graya_geta(c) > 0);
```

Gasparoken:
When a moved 'slice' is moved off the canvas, the content outside the canvas boundaries is lost. Other side effect of:

> 2. There is no way to "pause" the slice transformation once started. When the user releases the mouse button the changes are applied...

Anyway, this is a first approach.

Gasparoken:
@dacap  you can check it when you can.

martincapello:
> I was thinking that 'slice' movement is the most common use of this feature. On the other hand, scaling an image and its 'slice' at the same time would almost never be used, mainly because pixelart scaling can destroy detailed images. So my conclusion is: restrict the 'slice' transformation to motion only and thus avoid unintentional modifications. It's just an idea.
> 

Yes, I also think the most useful feature is to be able to move the slice along with its content....however, if any scaling was accidentally made, the user can always just undo it. Also, I would leave the scaling as is, and then wait for feedback from users. If there are several users that found out that slice scaling is actually more annoying than useful, then we can remove it later.

dacap:
I didn't review this yet, but just testing the resize of 9-slice, I noticed that probably a quick fix could be done:

resize

:point_up: I think we can offer the final result as a resize of the 9-slice by default, i.e. stretching the center areas (top/left/right/bottom and center) instead of resizing the top-right, right, bottom-right, bottom, and left-bottom corners/edges).

martincapello:
> ☝️ I think we can offer the final result as a resize of the 9-slice by default, i.e. stretching the center areas (top/left/right/bottom and center) instead of resizing the top-right, right, bottom-right, bottom, and left-bottom corners/edges).

I like this! Will do it.

src/app/ui/editor/moving_slice_state.h
```diff
@@ -25,26 +31,213 @@ namespace app {
...
+      gfx::Border border() const {
+        gfx::Border border;
+        if (isNineSlice()) {
+          border = gfx::Border(oldKey.center().x,
+                               oldKey.center().y,
+                               oldKey.bounds().w-oldKey.center().x2(),
+                               oldKey.bounds().h-oldKey.center().y2());
+        }
+        return border;
+      }
```
martincapello, line 210: I think I could refactor this into a SliceKey gfx::Border border() method. Let me know if it makes sense and I'll change this.

dacap:
I've just pushed the [slices-transform-switch](https://github.com/aseprite/aseprite/tree/slices-transform-switch) branch rebased to the latet beta to fix some conflicts (you can reset your branch there and push again to update the PR).

dacap:
It looks great, there are minor things that I didn't mention just to merge is right now. I didn't test it too much but it's going to `beta` anyway, so it'll be better if we get some feedback from users ASAP.

Issue #3628, opened 2022-12-14
Encountering bug while compiling aseprite on Ubuntu


GAGANsinghmsitece:
# Describe your bug report or feature request here
I'm trying to compile aseprite on my Lubuntu system following the instructions. When I running these set of commands in the build directory of aseprite,

```
export CC=clang-10
export CXX=clang++-10
cmake \
  -DCMAKE_BUILD_TYPE=RelWithDebInfo \
  -DCMAKE_CXX_FLAGS:STRING=-stdlib=libc++ \
  -DCMAKE_EXE_LINKER_FLAGS:STRING=-stdlib=libc++ \
  -DLAF_BACKEND=skia \
  -DSKIA_DIR=$HOME/deps/skia \
  -DSKIA_LIBRARY_DIR=$HOME/deps/skia/out/Release-x64 \
  -DSKIA_LIBRARY=$HOME/deps/skia/out/Release-x64/libskia.a \
  -G Ninja \
  ..

```

I'm getting this message in console

```

CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
WEBP_LIBRARIES
    linked by target "app-lib" in directory /home/gagan/Projects/games/temp/aseprite/src/app

-- Configuring incomplete, errors occurred!
See also "/home/gagan/Projects/games/temp/aseprite/build/CMakeFiles/CMakeOutput.log".
See also "/home/gagan/Projects/games/temp/aseprite/build/CMakeFiles/CMakeError.log".

```
### Aseprite and System version

* Aseprite version: v1.2.40
* System: Lubuntu, 20.04 LTS

StarTech13:
Same problem with me on Ubuntu 22.10. Aseprite version v1.2.40

pahaze:
Do you have `libwebp-dev` installed? Just encountered this myself and didn't have it installed, but even after installing `libwebp-devel` (I'm on openSUSE), I still get it and have to manually set the libraries.

StarTech13:
Thanks! It worked. `libwebp-dev` should be added to the linux dependencies in the install.md.

pahaze:
Will make a PR including the dependency for other distros later. Also happy New Years!

dacap:
Hi there, sorry but I think main branch doesn’t need the dependency anymore because we are using the one included with skia.

pahaze:
[...]

PR #4549, opened 2024-06-26
Update INSTALL.md

kevinforrestconnors:
The instructions were not clear to me.  

The Dependencies section lists 3 "main" dependencies:

- CMake is already available on most systems, and people can figure out very quickly if they don't have it
- Ninja is installed by the linux `apt-get` command
- Skia is not, and this was not clear to me

Furthermore, the Linux dependencies section has a lot of other dependencies (not sure if they are just required for Ninja or not, but they aren't mentioned in the Dependencies section).

I don't know if this is the *best* way to resolve the confusion, but issues like this https://github.com/aseprite/aseprite/issues/3628#issuecomment-1426960575 are evidence that I am not the only one who had confusion

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -71,6 +71,8 @@ might work).
 
 ## Linux dependencies
 
+> Note that `apt-get` will not install the [required Skia library](https://github.com/aseprite/aseprite/blob/main/INSTALL.md#dependencies) on the `asesprite-m102` branch, you will need to install this dependency separately.
+
 You will need the following dependencies on Ubuntu/Debian:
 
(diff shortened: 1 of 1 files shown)


dacap:
> I have signed the CLA following the steps given in https://github.com/igarastudio/cla#signing

Hi @kevinforrestconnors, we don't have a record of the signed CLA.

About this change, the Skia dependency is named in the `Dependencies` section (is the first thing you have to read) because it does't depend on the platform (all platforms depend on it).

I'm closing this PR because in other case we should clarify this point on each platform (and it's already in the general `Dependencies` section).

PR #4550, opened 2024-06-26
[ui] Fixed slice order in File > Export > Export As > Area

### LINKED ISSUE #4521: Sort slices by name in File > Export > Export As...
This request was received via email and it looks quite easy to fix: we should sort the slices by name in "Export As" dialog as in the context bar (right now the "Export" dialog shows slices by creation order), e.g.:

image

image

requizm:
Fix #4521


dacap:
Thanks @requizm, I'll merge this as it is, probably with a minor helper function so we can use the same code here and in the context bar:

https://github.com/aseprite/aseprite/blob/79efbd76a1b55ef98dceb929f122b198058dea24/src/app/ui/context_bar.cpp#L1750

Issue #4511, opened 2024-05-30
JSON export of Cel Properties does not expose Opacity 


Florent-Steiner:
When I export a file that has cell properties, the opacity field is not exposed.

Example:
![image](https://github.com/aseprite/aseprite/assets/555678/4a70b15e-fb6e-4297-912e-9fe121c9b968)

gives:
![image](https://github.com/aseprite/aseprite/assets/555678/7bcbb92c-4456-4aa0-9a82-130936c931b5)

### Aseprite and System version

* Aseprite version: 1.3.7-x64, Steam
* Windows 11

Florent-Steiner:
Hi, thanks for your reply.
The issue is not in "Layer Properties" but in "Cel Properties"

PR #4552, opened 2024-06-27
Added opacity field when exporting a cell

requizm:
Fix: #4511

### DIFF
--- a/src/app/doc_exporter.cpp
+++ b/src/app/doc_exporter.cpp
@@ -1569,6 +1569,7 @@ void DocExporter::createDataFile(const Samples& samples,
               os << ", ";
 
             os << "{ \"frame\": " << cel->frame();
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @requizm, I'm not sure if we should export the opacity if it's 255. What do you think?

dacap:
I'll merge this and remove the opacity when it's 255.

requizm:
> Thanks @requizm, I'm not sure if we should export the opacity if it's 255. What do you think?

Why not? What is the disadvantage?

dacap:
> Why not? What is the disadvantage?

I've added an extra change after the merge explaining the rationale behind this: https://github.com/aseprite/aseprite/commit/4e5dfa9d2c520b9e98dfc5b0fcd759df6a523b84

Issue #4487, opened 2024-05-22
image:resize is broken, width and/or height sometimes off by 1


RefinedSoftwareLLC:
image:resize is broken. Resizing from some source sizes to some target sizes fail to leave the image in the target size:

image:resize(22, 22) -- first run
assert image.w == 22 & image.h == 22 -- pass
image:resize(218, 218) -- then run
assert image.w == 218 & image.h == 218 -- fails, size is 217, 217
image:resize(218, 218) -- run again
assert image.w == 218 & image.h == 218 -- pass

A side effect of this bug is that resize(w, h) after resize(w, h) should be a no-op but is not.
Width and Height can independently be off by 1.
image:resize(1, 1) then image:resize(218, 218) just once passes.

### Aseprite and System version

* About screen does not have button to copy versions nor is a commit-hash displayed
* Aseprite version: Aseprite_1.3.6-x64.AppImage and Aseprite_1.3.7-x64.AppImage
* System: Linux
Operating System: Bazzite 40
KDE Plasma Version: 6.0.4
KDE Frameworks Version: 6.2.0
Qt Version: 6.7.0
Kernel Version: 6.8.9-301.fsync.fc40.x86_64 (64-bit)
Graphics Platform: X11

dacap:
Thanks for reporting this bug @RefinedSoftwareLLC, was able to reproduce it with your script.

```lua
local image = Image(app.image)
image:resize(22, 22) -- first run
print(image.width .. "x" .. image.height)
image:resize(218, 218) -- then run
print(image.width .. "x" .. image.height)
image:resize(218, 218) -- run again
print(image.width .. "x" .. image.height)
```

PR #4553, opened 2024-06-27
Fixed size calculation on resize_image

requizm:
Fix: #4487


dacap:
Thanks a lot @requizm for this fix! :+1:

PR #4560, opened 2024-07-13
prevent windows from only opening in the default workarea (multimon)

ajerick:
In multi-monitor setups, native windows, e.g. the preview window, will, when re-opened, be moved/clamped to the monitor of the main aseprite window even if the re-opened window was previously and explicitly moved to a different monitor.

This solution replaces the proposed get_workspace_region() implementation.
Pro:
- Doesn't introduce invalid workspace in non-rectangular multi-monitor setups. 
Con:
- Creates weird clamping behavior if a native window tries to open between the border of two or more monitors.

I signed the form, however I'll have it stated here again for peace of mind:


ajerick:
X11 is a pain to work with. With this other [laf PR](https://github.com/aseprite/laf/pull/101), working with multiple monitors is much better on X11 and windows.  I can give a crack at preventing clamping of windows between monitor borders later, but for now the behavior is consistent across X11 and windows. I believe the clamping bug is worth dealing with for the time being if it means windows can behave normally on other monitors.

ajerick:
Removed broken submodule reference in commit.

dacap:
Thanks @ajerick for this fix, I'm just testing and going to merge this with another rebased PR.

dacap:
Thanks @ajerick, finally merged in #5426

PR #4582, opened 2024-07-31
Add new frame menu to status bar button

ckaiser:
Adds a menu with new frame options when right-clicking the "+" button in the status bar:

https://github.com/user-attachments/assets/a4d3da28-ea07-4a08-a4f6-92f40fccf63c

It also adds a new plus sign icon, `icon_add`, since for themes that are using regular fonts the plus sign was off-center and smaller than usual, I've also adjusted the status bar's border since that was also affecting button sizing and their centering.

Example of a misaligned button using the [Wild Night](https://wildleoknight.itch.io/aseprite-wild-night-theme) theme:
image

In order to make the feature work at all I added the `RightClick` signal to `BaseButton`, this IMO improves consistency across the app since before there were some buttons that did nothing when right clicked, and others where right click meant a full click, so if you were testing for example if a button had extra functionality (like the animation play button) by right clicking it, this could result in accidental operations, so this makes buttons behave in the way that users usually expect them to. Ideally these buttons that have extra right-click functionality should have some sort of visual indicator, other than the current tooltip-only solution.

My main justification for the inclusion of this feature is that it increases discoverability of additional new frame options, since that might not be fully obvious from just the button, and it helps as a stepping stone into using the keyboard shortcuts.


dacap:
Hi @ckaiser! Thanks for your PR! I'll take a look to this next week, meanwhile, at a glance, it looks like the dark theme change broke the `pal_resize` icon:

break-pal-resize-icon

And the `+` sign has two columns:

image

I like the right-click idea for this "+" button and probably a future feedback for "right-clickable" buttons. I'm not sure how to make it in a way that is not visually annoying, probably the feedback could disappear once we right-click a button (?).

martincapello:
I have just took a quick look to this PR and it looks great! Just noted a couple of things that I will add as comments in the code.

data/extensions/aseprite-theme/dark/theme.xml
```diff
@@ -694,8 +695,9 @@
             <background color="check_hot_face" state="selected" />
             <icon part="pal_options" />
         </style>
-        <style id="new_frame_button" extends="mini_button" />
-        <style id="color_button" extends="mini_button" font="mini" padding-bottom="1" />
```
martincapello, line 698: Why is color_button removed from dark theme.xml?

src/ui/button.h
```diff
@@ -34,13 +34,15 @@ namespace ui {
     WidgetType behaviorType() const;
     // Signals
     obs::signal<void()> Click;
+    obs::signal<void()> RightClick;
```
martincapello, line 37: I think I would rename this to "ContextClick" instead. I took a look at some gui frameworks (unity and android) and liked this name because it is more generic, or less implementation oriented.

dacap: We talked about this in the interview with @ckaiser, he saw the RightClick on ButtonSet so it's based on the existing naming.

Anyway we could use ContextClick in a future when a tablet-like UI is introduced. Because there are other elements that use "right click" like

https://github.com/aseprite/aseprite/blob/6f65acb54f0d739b996d107989762db85c386ed0/src/ui/base.h#L31

But I guess several renames will be needed.

data/extensions/aseprite-theme/dark/theme.xml
ckaiser: Whoops, copy-paste accident. Fixed!

src/ui/button.h
ckaiser: Yeah the addition was mostly for internal and user consistency, if everything is already mouse-focused there's going to have to be either some kind of abstraction layer on top that maps whatever other input exists (touchscreen/pen/etc) to what is essentially going to be a mouse, or there's gonna have to be a pretty big refactor of input stuff, the former is probably the easier route to start with since there's some UI conventions that'd take some more planning like holding buttons to trigger context menus or swiping gestures, but it's still kinda nice to be able to do those with a mouse too if available.

dacap:
Thanks @ckaiser! 👍 I've just tested and it looks great, I'll merge this squashed right now.

PR #4587, opened 2024-08-07
Needed libwebP as a dependency on Arch Linux

gabespound:
Went to compile on arch linux and found that I needed another dependency.  Thought i'd make a small contribution for the free software by updating the install.md


dacap:
Thanks @gabespound, I'll merge this if it fixed the webp support for you.

PR #4590, opened 2024-08-09
[docs] Improve phrasing for Tileset image data length

drpaneas:
Clarified the description for the data length field of the compressed Tileset image when flag 2 is set. The way it is written written now is misleading and I originally thought that the length field is also compressed.


dacap:
Thanks @drpaneas!

Issue #2852, opened 2021-07-26
Improve selection tools feedback


dacap:
I think there were an issue related to this, but I wasn't able to find it.

All selection tools (mainly Rectangular Marquee) can be improved: instead of showing a negative black & white fill, we could simplify the feedback showing only the new area that will be selected with a similar [marching ants](https://en.wikipedia.org/wiki/Marching_ants) pattern.

dacap:
Implementing this should fix #4318

PR #4595, opened 2024-08-13
Reset mask when clicking outside of bounds with the magic wand tool

### LINKED ISSUE #4490: Magic Wand tool doesn’t deselect a current selection when clicking outside of the canvas or when clicking on a selection again
Aseprite installer, v1.3.6-x64 in Windows and installer, v1.3.6-arm64 on MacOS.

Originally reported here: https://community.aseprite.org/t/magic-wand-tool-doesnt-deselect-a-current-selection-when-clicking-outside-of-the-canvas-or-when-clicking-on-a-selection-again/22171

I found that when using the Magic Wand tool, if you click outside of the canvas in attempt to deselect a current selection, it doesn’t deselect but actually re-select the closest pixel to the mouse cursor on the X axis. Unlike when using the the Marquee tool which deselects correctly. I found that double clicking outside of the canvas with the Magic Wand sometimes deselect a current selection, but it is not very stable as it sometimes works and sometimes doesn’t (perhaps it depends on how fast you double-click). So you’d need to use the ESC key to deselect, which isn’t very convenient when using a stylus.
Magic_Wand_Deselection_Issue
The correct behavior should be that you could either click outside of the canvas to deselect a current selection, or to left-click again on the current selection to deselect it, just like in Photoshop.

ckaiser:
Fixes #4490, now clicking outside of bounds resets the selection.

https://github.com/user-attachments/assets/dc1724c9-cf10-4968-9ac1-409f3ea48db0

### DIFF
--- a/src/app/tools/tool_loop_manager.cpp
+++ b/src/app/tools/tool_loop_manager.cpp
@@ -120,14 +120,6 @@ void ToolLoopManager::pressButton(const Pointer& pointer)
     return;
   }
 
-  if (m_toolLoop->getController()->isOnePoint() &&
-     m_toolLoop->getInk()->isSelection() &&
-     !m_toolLoop->getSrcImage()->bounds().contains(pointer.point())) {
-    // Reset the mask if we're clicking out of bounds when we're using a one-point selection tool (magic wand)
-    m_toolLoop->setMask(nullptr);
-    return;
-  }
-
   m_stabilizerCenter = pointer.point();
 
   Stroke::Pt spritePoint = getSpriteStrokePt(pointer);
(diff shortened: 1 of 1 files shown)


dacap:
Hi @ckaiser! I was testing the PR and it differs from the other selection-like tools where the selection is deselected when we release the mouse button (instead of when we press the button like it's happening here).

I've tried to run this on debug mode (on Linux and Windows) and there were some asserts failing, mainly because no points are added in the stroke after `ToolLoopManager::pressButton()` (i.e. `m_toolLoop->getController()->pressButton()` is not called and the `ToolLoop` is not canceled).

ckaiser:
@dacap Moved it to after the press, the only issue now is that the selection behavior paints the background/whatever would've been selected, I'll look into fixing that logic but it might be something to look at for #2852

dacap:
> the only issue now is that the selection behavior paints the background/whatever would've been selected

That's right, no a big deal at the moment, at least it works better in this way (no assert fails / the tool loop logic is prepared to work in this way).

> I'll look into fixing that logic

I can merge this PR squashed and you can take a look to that later with other PR. I think the issue here is the same as the paint bucket: if we click outside the canvas we take the closest point to the sprite bounds. (Be aware that the gradient tool should continue working / filling even when we start pressing the mouse button outside the sprite bounds.)

> but it might be something to look at for #2852

Exactly! We can merge this fix as a workaround now, but a proper solution would be to refactor the selection tools to use other kind of logic (this is huge project/refactor, something to do in several tasks/steps, we can talk about this later).

PR #4604, opened 2024-08-19
Avoid drawing empty strings, other text fixes

ckaiser:
Avoids sending empty strings to the graphics API, adds some asserts to prevent it from happening in future.

Uses measureText() from the font to measure text length instead of drawing.

Improves underline bounds calculations

### DIFF
--- a/src/ui/theme.cpp
+++ b/src/ui/theme.cpp
@@ -186,7 +186,6 @@ void Theme::setDecorativeWidgetBounds(Widget* widget)
       widget->setBounds(buttonBounds);
       break;
     }
-
   }
 }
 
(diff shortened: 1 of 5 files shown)


dacap:
I'll make some comments in the commits but only to left record about it. Don't worry about these comments, I'm not expecting fixes because I'll cherry-pick some hunks of this PR anyway.

src/ui/theme.cpp
```diff
@@ -592,16 +596,18 @@ void Theme::measureLayer(const Widget* widget,
...
-            style->font() != widget->font()) {
-          text::Font* font = style->font();
-          textSize = gfx::Size(Graphics::measureUITextLength(widget->text(), font),
-                               font->height());
-        }
-        else {
-          // We can use Widget::textSize() because we're going to use
-          // the widget font and, probably, the cached TextBlob.
-          textSize = widget->textSize();
+        if (!widget->text().empty()) {
```
dacap, line 599: It looks like we cannot filter out this case because sometimes we "size hint" widgets with empty text, e.g. with this change:

Screenshot from 2024-08-22 20-15-45

Screenshot from 2024-08-22 20-15-51

While just commenting this condition we get the expected result:

Screenshot from 2024-08-22 20-16-37

Screenshot from 2024-08-22 20-16-31

ckaiser: Should still avoid drawing it and use the font height or similar value to get the size right, I can do the change if you want

dacap: Don't worry, I'll take from here testing in all platforms and making comments just to leave a record about it.

dacap:
Cherry-picked in aed9127b5107202b0a150f1da8a499b5c71206ee and 7d80ff120c74b7ce1de30f189cdd876a8cccbeae.

I've discarded the changes about the underline position/thickness calculation (at least temporarily, we can create other PR in the future about this), mainly because I prefer that the underline position/thickness is completely solved from the laf side (so we can use `text::FontMetrics` without depending on the `text::FontType`).

PR #4605, opened 2024-08-19
Add Image.context property to allow getting a GraphicsContext from an Image

### LINKED ISSUE #4330: Get a GraphicsContext to modify an Image/Sprite
I though there were an issue created for this one, anyway it would be nice to be able to get a GraphicsContext to modify an Image. The tricky part here is to be able to undo all the modifications when the GraphicsContext is not used anymore.

Requested in the past and recently:
* ~~https://community.aseprite.org/t/how-to-draw-a-sprite-with-graphics-context/20572~~
* https://community.aseprite.org/t/why-is-this-returning-a-nil/21525

The function `wrap_docimage_in_sksurface()` can be used for this purpose (not sure if this would work on all color modes).

EDIT: The crossed out link is about drawing in a GraphicsContext received in the onpaint event, so it is actually a different issue, also it is not actually an issue, since the user was trying to draw an sprite using GraphicsContext::drawImage, which receives an Image as a parameter, not an Sprite. Someone from the community gave the user a right answer, so there is no need to create an new github issue.

martincapello:
Fix #4330

### DIFF
--- a/src/app/color_utils.cpp
+++ b/src/app/color_utils.cpp
@@ -37,22 +37,12 @@ gfx::Color color_utils::blackandwhite_neg(gfx::Color color)
     return gfx::rgba(0, 0, 0);
 }
 
-app::Color color_utils::color_from_ui(const gfx::Color color, PixelFormat format)
+app::Color color_utils::color_from_ui(const gfx::Color color)
 {
-  switch (format) {
-    case IMAGE_GRAYSCALE:
-      return app::Color::fromGray(gfx::getr(color), gfx::geta(color));
-    case IMAGE_INDEXED: {
-      const int i = gfx::geta(color);
-      return app::Color::fromIndex(i);
-    }
-    case IMAGE_RGB:
-    default:
-      return app::Color::fromRgb(gfx::getr(color),
-                                 gfx::getg(color),
-                                 gfx::getb(color),
-                                 gfx::geta(color));
-  }
+  return app::Color::fromRgb(gfx::getr(color),
+                             gfx::getg(color),
+                             gfx::getb(color),
+                             gfx::geta(color));
 }
 
 gfx::Color color_utils::color_for_ui(const app::Color& color)
--- a/src/app/color_utils.cpp
+++ b/src/app/color_utils.cpp
@@ -37,12 +37,22 @@ gfx::Color color_utils::blackandwhite_neg(gfx::Color color)
     return gfx::rgba(0, 0, 0);
 }
 
-app::Color color_utils::color_from_ui(const gfx::Color color)
+app::Color color_utils::color_from_ui(const gfx::Color color, PixelFormat format)
 {
-  return app::Color::fromRgb(gfx::getr(color),
-                             gfx::getg(color),
-                             gfx::getb(color),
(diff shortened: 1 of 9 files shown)


martincapello:
@dacap Build is failing for cli setup because I changed something to include the GraphicsContext class when UI is not enabled. Since now the user can create a GraphicsContext based on an image. The thing is that when building as cli, the skia backend is not used and this is what causes problems.
So I'm wondering what to do. Should we avoid compiling Image.context property when UI is not enabled? This would let me revert the change to include GraphicsContext when UI is not enabled.

dacap:
> Should we avoid compiling Image.context property when UI is not enabled?

That's right, `Image.context` should return nil if Skia is not available.

dacap:
I'll make a few comments about this one but will review it later.

src/app/color_utils.cpp
```diff
@@ -37,12 +37,22 @@ gfx::Color color_utils::blackandwhite_neg(gfx::Color color)
     return gfx::rgba(0, 0, 0);
 }
 
-app::Color color_utils::color_from_ui(const gfx::Color color)
+app::Color color_utils::color_from_ui(const gfx::Color color, PixelFormat format)
```
dacap, line 40: `gfx::Color` is always a 32-bit RGBA value. Here is like you are using a `gfx::Color` like a `doc::color_t`, probably you are looking for a new function, the inverse of `color_for_image()`.

dacap (commented):
I'll continue later. Just in case, seeing the commits I'll merge this squashed later.

src/app/color_utils.cpp
dacap: I'll review this later because the API is a little messy, mainly because the new kind of pixel format expected in `os::Paint`/SkPaint. From the docs of SkPaint:

    Sets alpha and RGB used when stroking and filling. The color is a 32-bit value,
    unpremultiplied, packing 8-bit components for alpha, red, blue, and green.

But it depends on the pixel format of the surface, so probably we need a new kind of color/type for these cases.

martincapello: > But it depends on the pixel format of the surface, so probably we need a new kind of color/type for these cases.

Yes, seems that one can set any unpremultiplied ARGB (8bit per component) color value to SkPaint and then the surface will just take what it needs depending on the SkColorType it has.

dacap: Probably we can leave it as it is (because Skia itself expect the values in each channel of the RGBA value depending on the format), but I'd leave `color_from_ui()` as it was, and move the new implementation (the one with the `switch (format)`) inside `GraphicsContext_get_color()` itself, because it's the only place we need this (I guess).

martincapello: > but I'd leave `color_from_ui()` as it was, and move the new implementation (the one with the `switch (format)`) inside `GraphicsContext_get_color()` itself, because it's the only place we need this (I guess).

Sure, no problem, will make this change and see how it works. Thanks!

dacap:
Merged in 7e504ad5ea46d5e9aba6f29f4c92b8d8bbcb6318

PR #4611, opened 2024-08-23
Release mouse before command execution on dragging value state

### LINKED ISSUE #4484: Ctrl key alone is not useful to drag brush size value
From: https://community.aseprite.org/t/ctrl-z-and-lost-brushstrokes/22112

Configuring <kbd>Ctrl</kbd> key to drag the brush size:
image

Then each time we use <kbd>Ctrl+Z</kbd>, the first stroke is ignored.

Liebranca:
Fix #4484

Same as the old commit, I merely rebased to drop unrelated changes.

I haven't been able to sign the document yet, due to what appears to be a browser error, so just in case: I agree that my contributions are licensed under the Individual Contributor License Agreement V4.0 ("CLA") as stated in https://github.com/igarastudio/cla/blob/main/cla.md


dacap:
Just tested and it works great :+1:

Issue #1279, opened 2016-10-05
Aseprite only CLI


JosefKuchar:
Hello,
Can I compile aseprite without GUI, only CLI tools ? 
Thx

dacap:
Hi @JosefKuchar, that's not possible at this moment but I'd like to make that possible in a near future. Related to #139

jacksarick:
What needs to change to make this happen? I'm happy to help

dacap:
@jacksarick I think it's a pretty big change for someone that want to contribute for the very first time. I would prefer smaller bug fixes first and then trying to contribute to something big.

Anyway, for some reference, the basic idea is that [this statement](https://github.com/aseprite/aseprite/blob/master/src/app/app.cpp#L231) must be completely ignored from the compilation process (and all the dependencies should be ignored too/not linked, i.e. UI library isn't needed, etc.)

dacap:
New `cmake -DENABLE_UI=OFF` flag added. Another discussion in [t/1351](https://community.aseprite.org/t/integrating-aseprite-with-travis-ci/1351)

PR #4615, opened 2024-08-26
Use right click action with shading ink (fix #4520)

### LINKED ISSUE #4520: Right-clicking with shading ink should do the configured action for right-click when no color range is selected
Some special right-click options for the Editor:

image

might result in confusing situations when the shading mode is selected. I think a possible simple fix is that when the shading ink mode is enabled, if no range of colors is selected

image

right-click could just act with the configured option (as left click already paints normally like in simple ink)

Liebranca:
Save the number of selected entries in `PaletteView` so that `ActiveToolManager` can access this value in `isToolAffectedByRightClickMode`.

If the number of colors that were last manually selected is less than two, allow the configured right click action during shading mode.

This value ignores reversal; it will only change uppon a new selection, be it a single color or another range.

### DIFF
--- a/src/app/tools/active_tool.cpp
+++ b/src/app/tools/active_tool.cpp
@@ -237,6 +237,15 @@ void ActiveToolManager::setSelectedTool(Tool* tool)
 bool ActiveToolManager::isToolAffectedByRightClickMode(Tool* tool)
 {
   bool shadingMode = (Preferences::instance().tool(tool).ink() == InkType::SHADING);
+
+  if (shadingMode) {
+
+    ColorBar* colorbar = ColorBar::instance();
+    if (colorbar->getPaletteView()->getLastSelectionSize() < 2)
+      shadingMode = false;
+
+  }
+
   return
     ((tool->getInk(0)->isPaint() && !shadingMode) ||
      (tool->getInk(0)->isEffect())) &&
--- a/src/app/tools/active_tool.cpp
+++ b/src/app/tools/active_tool.cpp
@@ -9,6 +9,7 @@
 #include "config.h"
 #endif
 
+#include "app/app.h"
 #include "app/tools/active_tool.h"
 
 #include "app/color.h"
@@ -18,6 +19,7 @@
 #include "app/tools/pointer.h"
 #include "app/tools/tool_box.h"
 #include "app/ui/color_bar.h"
+#include "app/ui/context_bar.h"
 
 namespace app {
 namespace tools {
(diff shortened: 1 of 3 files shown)


src/app/tools/active_tool.cpp
```diff
@@ -237,6 +237,15 @@ void ActiveToolManager::setSelectedTool(Tool* tool)
 bool ActiveToolManager::isToolAffectedByRightClickMode(Tool* tool)
 {
   bool shadingMode = (Preferences::instance().tool(tool).ink() == InkType::SHADING);
+
+  if (shadingMode) {
+
+    ColorBar* colorbar = ColorBar::instance();
+    if (colorbar->getPaletteView()->getLastSelectionSize() < 2)
+      shadingMode = false;
```
dacap, line 245: Be aware that the user can change the selected shades from the `ContextBar` itself:

image

The information of the active shades can be accessed through the `ContextBar::getShade()` member function.

dacap (changes requested):
Hi @Liebranca, check the review comments and the actions: https://github.com/aseprite/aseprite/actions/runs/10568125733 when compiling with `ENABLE_UI=OFF` is failing (there is no `ColorBar` when compiling without UI).

Liebranca:
Hi @dacap, added a bit of logic to my scripts to automatically compile with and without UI, so now I should catch this early.

Discarded the initial unorthodox workaround for the more straightforward approach through `ContextBar`, and placed the new check behind a preprocessor switch (else left as it was).

However, it seems that excluding some of these functions themselves from compilation altogether via such a switch may be more appropriate, since they're only needed by UI elements in the first place. What do you think?

dacap:
> it seems that excluding some of these functions themselves from compilation altogether via such a switch may be more appropriate, since they're only needed by UI elements in the first place. What do you think?

Yeah, there are some work already done in `src/app/CMakeLists.txt` to only include UI-dependent files when `ENABLE_UI` is true, e.g.:

https://github.com/aseprite/aseprite/blob/e87f6df72b03bfc5c17bc3e53dd88145e1b0f368/src/app/CMakeLists.txt#L252

But it's like a "refactor in progress". We have the inverse case too, i.e. commands that are included only when `ENABLE_UI` is on, but should be always available (like `InvertMask`, etc.):

https://github.com/aseprite/aseprite/blob/e87f6df72b03bfc5c17bc3e53dd88145e1b0f368/src/app/commands/commands_list.h#L64

The case of the `ActiveToolManager` is a little complex because it's used in the scripting API too. Probably for simple things, like just storing the active tool, without even calling functions related to UI events (newToolSelectedInToolBar, regularTipProximity, etc.). I don't think it's worth refactoring at the moment.

Even more, this `ENABLE_UI` was created to fix #1279 but never really used (and now we are compiling the full version in GitHub CI), so I'm not sure if we should keep the flag or just remove it and simplify the maintenance. I'll create a new issue about this.

src/app/tools/active_tool.cpp
```diff
@@ -9,6 +9,7 @@
 #include "config.h"
 #endif
 
+#include "app/app.h"
```
dacap, line 12: Oh a little thing I missed at a glance is that we generally include the header file related to the `.cpp` as the first header file (after the `config.h` thing) and then all other files, e.g. to check that `"app/tools/active_tool.h"` includes everything it needs.

PR #4616, opened 2024-08-27
[lua] Add get/set the foreground/background tile (fix #4403)

### LINKED ISSUE #4403: Scripting API: access to selected tile in APP object
Possibility to access the current selected tile object from the active tileset through object APP via scripting, for example:
app.fgTile or app.bgTile
And then access to tile properties.

* Aseprite version: v1.3.2-x64 installer
* System: Windows 10

Gasparoken:
This approach relies on the fact that the existing fgTile and bgTile are just indexes. I think other refactor is possible using some other type of Tile object, but much more work is needed to achive it.

fix #4403 

Pending: add documentation on aseprite/api

### DIFF
--- a/src/app/script/app_object.cpp
+++ b/src/app/script/app_object.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2023  Igara Studio S.A.
+// Copyright (C) 2018-2024  Igara Studio S.A.
 // Copyright (C) 2015-2018  David Capello
 //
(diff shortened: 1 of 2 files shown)


dacap:
Some of this work was already done in 9429d915ae3cee9c0d21a9ea54bac39e544be932, I'll review this later and see what we can cherry-pick from it.

dacap:
Merged in 0f7cac0f0cc93a9a1e06afe5fef3fd641c92eb3b

Issue #4622, opened 2024-08-29
Cannot uninstall extensions


dacap:
It looks like v1.3.8 introduced a regression (detected in #4620) where we cannot uninstall extensions because they are all identified as built-in.

PR #4625, opened 2024-08-29
Fix extensions wrongly being detected as "built-in"

ckaiser:
The extra trailing path slash was making the comparison fail, must've been one of the changes to the base functions. Fixed it by just normalizing and removing any trailing slashes. Fixes [#4622](https://github.com/aseprite/aseprite/issues/4622)

### DIFF
--- a/src/app/extensions.cpp
+++ b/src/app/extensions.cpp
@@ -827,7 +827,8 @@ Extensions::Extensions()
       if (!base::is_directory(dir))
         continue;
 
-      const bool isBuiltinExtension = m_userExtensionsPath != base::get_file_path(dir);
+      const bool isBuiltinExtension =
+        (m_userExtensionsPath != base::get_file_path(dir));
 
       auto fullFn = base::join_path(dir, kPackageJson);
       fullFn = base::normalize_path(fullFn);
--- a/src/app/extensions.cpp
+++ b/src/app/extensions.cpp
@@ -800,7 +800,7 @@ Extensions::Extensions()
   // Create and get the user extensions directory
   {
     ResourceFinder rf2;
-    rf2.includeUserDir("extensions/.");
+    rf2.includeUserDir("extensions");
     m_userExtensionsPath = rf2.getFirstOrCreateDefault();
     m_userExtensionsPath = base::normalize_path(m_userExtensionsPath);
     if (!m_userExtensionsPath.empty() &&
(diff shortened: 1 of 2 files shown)


ckaiser:
I figured out why it was behaving this way, the `normalize_path` function was changed is now behaving correctly.
Because the initial given path was `"extensions/." which is not a valid Windows path, it's now appending the // because of the new logic, which matches the one in std::filesystem, so I think the actual solution is to change how we're using the function and just call it with `"extensions"`, which is the same way it is called in the rest of the code.

I found another place where this can happen, when detecting scripts, so I changed it there as well.

I'm keeping the other aesthetic changes to the code even if they're not strictly necessary for the fix.

Gasparoken:
Thank you @ckaiser. This fix works on the extension folder, the aesthetic changes are OK in my opinion.
On the other hand, when I tried going to the _File > Scripts_ menu, it showed the content of the folder
`/Users/gasparcapello/Library/Application Support/Aseprite` instead of 
`/Users/gasparcapello/Library/Application Support/Aseprite/scripts` (I'm on macOS).
This is because `scripts` without the trailing slash results in a search for the path where the `scripts` _file_ is located, rather than searching the path of the `scripts` _directory_.
In conclusion, please modify
`rf.includeUserDir("scripts");` to
`rf.includeUserDir("scripts/");`
(I haven't seen any effect by putting the final dot, at least in this case)

ckaiser:
The script submenu loader was trimming the path by calling `get_file_path` on it after, which was probably necessary with the old behavior, without it now it should work fine.

Gasparoken:
Great! I'll assign it to @dacap for final review/merge.

src/app/extensions.cpp
```diff
@@ -817,33 +817,33 @@ Extensions::Extensions()
...
-        const bool isBuiltinExtension =
-          (m_userExtensionsPath != base::get_file_path(dir));
+    for (auto& fn : base::list_files(extensionsDir)) {
+      const auto dir = base::join_path(extensionsDir, fn);
+      if (!base::is_directory(dir))
+        continue;
 
-        auto fullFn = base::join_path(dir, kPackageJson);
-        fullFn = base::normalize_path(fullFn);
+      const bool isBuiltinExtension = m_userExtensionsPath != base::get_file_path(dir);
```
dacap, line 830: Keep the parenthesis and the line break as it was: https://github.com/aseprite/aseprite/blob/c8f018f2f1a7f050ca79aa6b766bb4d4fb86168f/docs/CODING_STYLE.md?plain=1#L32-L34

```c++
      const bool isBuiltinExtension =
        (m_userExtensionsPath != base::get_file_path(dir));
```

dacap (changes requested):
Looks good, I've tested it on Linux and it works :+1: I've made a minor comment about the coding style.

PR #4635, opened 2024-09-03
Remove ENABLE_UI flag

### LINKED ISSUE #4619: Should we remove ENABLE_UI?
The `ENABLE_UI` flag makes the development a little more complex because generally we always compile with this option enabled. It was originated for #1279 but we have never released a CLI-only version of the program. Even more, probably we can solve the CLI-only problem with other easier approach instead of `#ifdef/#endif`'ing the whole code.

I think we have some options:

1. Remove `ENABLE_UI` and suppose we always have the GUI functions available, with a `LAF_BACKEND=none` the creation of the main window will fail anyway (and this could be the CLI-only version).
2. Sometimes the `ENABLE_UI` helps to think about the CLI-only situation where the GUI is not available (i.e. we're running with the `-batch` parameter) and we try to use some widget at runtime but it fails because it's uninitialized. Anyway sometimes the compilation with `ENABLE_UI=off` doesn't catch all possible errors.

dacap:
Fix #4619, we finally remove the `ENABLE_UI` flag to make the development a little easier. A CLI-only version (#1279) can be created with `LAF_BACKEND=none` anyway, and we can check the GUI existence with `App::isGui()` or `Context::isUIAvailable()`.


dacap:
Rejected by my own bot.

PR #4643, opened 2024-09-07
Use FlattenLayers for MergeDownLayerCommand

### LINKED ISSUE #4468: Merge layers taking into account z-index
When a user merge layers the resulting layer should be created taking into account the z-index of the merged layers.

Originally reported here:
https://community.aseprite.org/t/merge-with-z-index/22106

Liebranca:
Rework of the `cmd::FlattenLayers` implementation to accomodate the 'Merge Down' command as a special case. Fix #4468.

This includes a multitude of changes and additions to said implementation, listed below. Because not all of these modifications are directly related to the linked issue, feedback on what could be improved or skipped altogether would be greatly appreciated.

Summary of changes:
- Resetting of cel opacity for the destination layer, as well as the layer's visibility, opacity and blend mode, in cases where flattening is done onto an existing layer rather than a new one.
- Unconditional use of `shrink_bounds` to ensure layer edges are displayed correctly.
- Swapped the usage of `CopyRect` command for a `ReplaceImage`. This was done merely because I could not get the former to work properly when flattening onto an existing, non-background layer.
- Changed the boolean `newBlend` parameter of `cmd::FlattenLayers` (and the `m_newBlendMethod` member variable) for an integer `options` and `m_options` respectively, to enable and disable behaviours (defined in the `cmd::FlattenLayers::Options` enum) through a single bitmask value. These options are:
  - `NewBlendMethod`: identical to `newBlend`.
  - `Inplace`: utilize the last layer in the provided list as destination for flattening rather than spawning a new layer to flatten onto. Similar to utilizing the background layer.
  - `MergeDown`: invoke `notifyLayerMergedDown` before issuing the removal of source layers.
  - `ExtendCanvas`: give a larger size to the temporary image being rendered to, when cels in a frame are partially or completely outside of the canvas. Without this option, pixels outside of those bounds are discarded.
- All three of the new options detailed above are enabled when `cmd::FlattenLayers` is invoked through `MergeDownLayerCommand`. When invoked through `FlattenLayersCommand`, the new options are ignored altogether, thus retaining the behaviour prior to this rework, save for the resetting of visibility, opacity and blend mode, and the adjusting of layer edges.

I've left note of a few smaller details via comments in the code itself, mostly to explain what exactly a series of lines are doing. Please, let me know if any of them are too redundant.

### DIFF
--- a/src/app/cmd/flatten_layers.cpp
+++ b/src/app/cmd/flatten_layers.cpp
@@ -18,7 +18,6 @@
 #include "app/cmd/remove_layer.h"
 #include "app/cmd/remove_cel.h"
 #include "app/cmd/replace_image.h"
-#include "app/cmd/set_layer_flags.h"
 #include "app/cmd/set_layer_name.h"
 #include "app/cmd/set_layer_opacity.h"
 #include "app/cmd/set_layer_blend_mode.h"
@@ -40,7 +39,7 @@ namespace cmd {
 
 FlattenLayers::FlattenLayers(doc::Sprite* sprite,
                              const doc::SelectedLayers& layers0,
-                             const int options)
+                             const Options options)
   : WithSprite(sprite)
 {
   m_options = options;
@@ -54,7 +53,6 @@ FlattenLayers::FlattenLayers(doc::Sprite* sprite,
 
 void FlattenLayers::onExecute()
 {
-
   Sprite* sprite = this->sprite();
   auto doc = static_cast<Doc*>(sprite->document());
 
@@ -73,48 +71,23 @@ void FlattenLayers::onExecute()
   if (list.empty())
     return;                     // Do nothing
 
-  // Extend the drawable area beyond the sprite bounds
+  // Set the drawable area to a union of all cel bounds
   // when this option is enabled
   ImageSpec spec = sprite->spec();
-  int area_x = 0;
-  int area_y = 0;
-  if ((m_options & Options::ExtendCanvas) != 0) {
-    // Get initial dimentions of draw area
-    int mX = spec.width();
-    int mY = spec.height();
-
-    // Check those bounds against every frame of every layer that
-    // will be merged
+  gfx::Rect area;
+  if (m_options.dynamicCanvas) {
     for (frame_t frame(0); frame<sprite->totalFrames(); ++frame) {
       for (Layer* layer : layers) {
         Cel* cel = layer->cel(frame);
-        if (!cel)
-          continue;
-
-        // Is this cel outside of the current dimentions?
-        const gfx::Rect bounds = cel->bounds();
-        const int x = (bounds.x < 0 ? bounds.w+spec.width()+abs(bounds.x):
-                                      bounds.w+bounds.x);
-        const int y = (bounds.y < 0 ? bounds.h+spec.height()+abs(bounds.y):
-                                      bounds.h+bounds.y);
-
-        if (x > mX)
-          mX = x;
-
-        if (y > mY)
-          mY = y;
+        if (cel)
+          area |= cel->bounds();
       }
     }
-
-    // Update size of draw area
-    if (mX != spec.width()) {
-      spec.setWidth(mX*2);
-      area_x = mX/2;
-    }
-    if (mY != spec.height()) {
-      spec.setHeight(mY*2);
-      area_y = mY/2;
-    }
+    spec.setSize(area.size());
+  }
+  // Otherwise use the sprite's bounds
+  else {
+    area.setSize(spec.size());
   }
 
   // Create a temporary image.
@@ -131,7 +104,7 @@ void FlattenLayers::onExecute()
   }
   // Get bottom layer when merging layers in-place, but only if
   // we are not flattening into the background layer
-  else if ((m_options & Options::Inplace) != 0) {
+  else if (m_options.inplace) {
     flatLayer = static_cast<LayerImage*>(list.front());
     bgcolor = sprite->transparentColor();
   }
@@ -145,7 +118,7 @@ void FlattenLayers::onExecute()
   }
 
   render::Render render;
-  render.setNewBlend((m_options & Options::NewBlendMethod) != 0);
+  render.setNewBlend(m_options.newBlendMethod);
   render.setBgOptions(render::BgOptions::MakeNone());
 
   {
@@ -154,17 +127,14 @@ void FlattenLayers::onExecute()
     RestoreVisibleLayers restore;
     restore.showSelectedLayers(sprite, layers);
 
-    // Map canvas to extended draw area
-    const gfx::ClipF area(
-      0, 0,
-      -area_x, -area_y,
-      spec.width(), spec.height());
+    // Map draw area to image coords
+    const gfx::ClipF area_to_image(0, 0, area);
 
     // Copy all frames to the background.
     for (frame_t frame(0); frame<sprite->totalFrames(); ++frame) {
       // Clear the image and render this frame.
       clear_image(image.get(), bgcolor);
-      render.renderSprite(image.get(), sprite, frame, area);
+      render.renderSprite(image.get(), sprite, frame, area_to_image);
 
       // Get exact bounds for rendered frame
       gfx::Rect bounds = image->bounds();
@@ -200,7 +170,7 @@ void FlattenLayers::onExecute()
         if (!newFlatLayer) {
           executeAndAdd(new cmd::SetCelOpacity(cel, 255));
           executeAndAdd(new cmd::SetCelPosition(cel,
-            area.src.x+bounds.x, area.src.y+bounds.y));
+            area.x+bounds.x, area.y+bounds.y));
         }
 
         // Modify destination cel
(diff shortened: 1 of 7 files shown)


src/app/cmd/flatten_layers.cpp
```diff
@@ -66,8 +73,52 @@ void FlattenLayers::onExecute()
...
+        const int y = (bounds.y < 0 ? bounds.h+spec.height()+abs(bounds.y):
+                                      bounds.h+bounds.y);
+
+        if (x > mX)
+          mX = x;
+
+        if (y > mY)
+          mY = y;
+      }
+    }
```
dacap, line 107: Probably something we can use to simplify this is an union of all `cel->bounds()`, something like:

```c++
    gfx::Rect extends;
    for (frame_t frame(0); frame<sprite->totalFrames(); ++frame) {
      for (Layer* layer : layers) {
        Cel* cel = layer->cel(frame);
        if (cel)
          extends |= cel->bounds();
      }
    }
```
And use `extends.origin()` to replace `mX` and `mY` (we could even avoid creating a whole canvas image if it's not required for the merge down).

src/app/cmd/flatten_layers.cpp
```diff
@@ -97,44 +154,94 @@ void FlattenLayers::onExecute()
...
+      list.front()->parent(), flatLayer, list.front()));
+
+  }
+  // Reset layer properties when flattening in-place
+  else {
+    executeAndAdd(new cmd::SetLayerOpacity(flatLayer, 255));
+    executeAndAdd(new cmd::SetLayerBlendMode(
+      flatLayer, doc::BlendMode::NORMAL));
+    executeAndAdd(new cmd::SetLayerFlags(
+      flatLayer, doc::LayerFlags::Visible));
```
dacap, line 243: Probably it'd be better to leave the flags as they were, with this PR the resulting destination layer is being locked.

src/app/cmd/flatten_layers.cpp
```diff
@@ -97,44 +154,94 @@ void FlattenLayers::onExecute()
...
-                            gfx::Clip(0, 0, image->bounds())));
+        // Reset cel properties when flattening in-place
+        if (!newFlatLayer) {
+          executeAndAdd(new cmd::SetCelOpacity(cel, 255));
+          executeAndAdd(new cmd::SetCelPosition(cel,
+            area.src.x+bounds.x, area.src.y+bounds.y));
+        }
+
+        // Modify destination cel
+        executeAndAdd(new cmd::ReplaceImage(sprite, cel_image, new_image));
```
dacap, line 207: Just as a side comment: Something for the future (not for this PR, to keep it as simple as possible), here we could do something like in [FilterManagerImpl::apply()](https://github.com/aseprite/aseprite/blob/81bf86f9e700ca2e75434256a986edf1f4388297/src/app/commands/filters/filter_manager_impl.cpp#L223) using PatchCel/CopyRegion for certain cases (to reduce the undo information). But this is not a priority.

dacap (changes requested):
Hi @Liebranca! It looks great so far, some minor changes/comments here. It'd be nice to add a test in `tests/scripts/merge_down_bugs.lua` for the specific bug with z-index if you can.

src/app/commands/cmd_merge_down_layer.cpp
```diff
@@ -81,85 +76,17 @@ void MergeDownLayerCommand::onExecute(Context* context)
...
-  document->notifyLayerMergedDown(src_layer, dst_layer);
-  document->getApi(tx).removeLayer(src_layer); // src_layer is deleted inside removeLayer()
+  DocRange range;
+  range.selectLayer(writer.layer());
+  range.selectLayer(dst_layer);
+  const bool newBlend = Preferences::instance().experimental.newBlend();
+
+  tx(new cmd::FlattenLayers(sprite,
+    range.selectedLayers(),
+    int(newBlend) |
```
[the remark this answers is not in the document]
dacap: I think both options are valid (we've used both in Aseprite).
1. In case you go with flags, you can use the new [LAF_ENUM_FLAGS()](https://github.com/aseprite/laf/commit/6abf784cd9a131c85cab016ddbc5fa22199c9211) macro to overload the `operator|`.
2. Otherwise, using bitfields, we'll need an `Options()` constructor (e.g. like [DocDiff](https://github.com/aseprite/aseprite/blob/ec42689b820cc025e685e4f2bc3f230a76bb1e89/src/app/doc_diff.h#L29))

Both options are fine :+1:

src/app/cmd/flatten_layers.cpp
Liebranca: I like this, it's much simpler logic, and can indeed be used to make the temporary image only as big as necessary rather than always enlarging it. I'll try to think of a more appropriate name for the `ExtendCanvas` option, so that it better reflects this more dynamic behavior.

src/app/cmd/flatten_layers.cpp
Liebranca: Right, I forgot to mention, it's possible to merge down when the destination layer isn't visible, which results in the source layer seemingly disappearing after merging as visibility remains disabled. The first time it happened I thought for a second that I had somehow deleted them, so I added this bit.

Another way to prevent such confusion would've been disabling from `MergeDownLayerCommand::onEnabled()` when the destination isn't visible, but honestly this is a very small detail, so it can be left as it was.

dacap: You are right :thinking: Even worse, if the layer at the bottom is locked, the merge down command works anyway. We could leave this functionality as it's working now (leaving the flags as they were), and I'll create another issue about this.

src/app/cmd/flatten_layers.cpp
```diff
@@ -97,44 +127,89 @@ void FlattenLayers::onExecute()
...
-        cel_image = cel->imageRef();
+        const ImageRef cel_image = cel->imageRef();
         ASSERT(cel_image);
 
-        executeAndAdd(
-          new cmd::CopyRect(cel_image.get(), image.get(),
-                            gfx::Clip(0, 0, image->bounds())));
+        // Reset cel properties when flattening in-place
+        if (!newFlatLayer) {
+          executeAndAdd(new cmd::SetCelOpacity(cel, 255));
```
dacap, line 171: We could create `SetCelOpacity()` only when it's required (this is something that wasn't present in the old impl of merge down but it'd be nice to have):
```suggestion
          if (cel->opacity() != 255)
            executeAndAdd(new cmd::SetCelOpacity(cel, 255));
```
One thing I noticed is the z-index of the resulting cel, I'm not sure if we should do something like:
```c++
          if (cel->zIndex() != newZIndex)
            executeAndAdd(new cmd::SetCelZIndex(cel, newZIndex));
```
The value of `newZIndex` could be just `0`, but there are some special cases that we could take care (I'll attach an example in the general comment).

src/app/cmd/flatten_layers.cpp
```diff
@@ -97,44 +127,89 @@ void FlattenLayers::onExecute()
...
+  if (newFlatLayer) {
+    executeAndAdd(new cmd::AddLayer(
+      list.front()->parent(), flatLayer, list.front()));
+
+  }
+  // Reset layer properties when flattening in-place
+  else {
+    executeAndAdd(new cmd::SetLayerOpacity(flatLayer, 255));
+    executeAndAdd(new cmd::SetLayerBlendMode(
+      flatLayer, doc::BlendMode::NORMAL));
```
dacap, line 211: Same here as with the cel opacity:
```suggestion
    if (flatLayer->opacity() != 255)
      executeAndAdd(new cmd::SetLayerOpacity(flatLayer, 255));

    if (flatLayer->blendMode() != doc::BlendMode::NORMAL) {
      executeAndAdd(new cmd::SetLayerBlendMode(
                      flatLayer, doc::BlendMode::NORMAL));
    }
```

dacap (changes requested):
It looks great so far! There are some special cases with z-index that might be fixed but I'm not sure if it's worth to fix these cases now (we could just set the z-index of all resulting cels to 0):

Peek 2024-09-17 18-42

Example file: [merge-down-test.zip](https://github.com/user-attachments/files/17035262/merge-down-test.zip)

Liebranca:
Hi @dacap! A note about adjusting the z-index, both changing and not changing the value can affect the final image in different scenarios. Setting it to zero on the destination could give a similar result if there's a cel below it with a z-index greater than zero, for instance:

```
// Before merging
(layer 2)  [Z:0] <-- source
(layer 1)  [Z:1] <-- destination, rendered on top
(layer 0)  [Z:1] <-- layer below destination

// After merging
(merged)   [Z:0] <-- destination
(layer 0)  [Z:1] <-- layer below, rendered on top
```

This means some edge cases are unsolvable too (cannot preserve the exact same final image without outright flattening everything) when more than two layers are involved. I'm not sure what could be done.

Anyway, setting the z-index to zero, as if a new cel was created, would be consistent with opacity and blend mode being reset back to the default, so that may be the more predictable behaviour.

dacap:
That's right. I think setting the z-index to 0 is the simplest way to approach this.

There is that (solvable) edge case where the resulting layer after a merge down should have an adjusted z-index (to be on top of a specific layer), but I don't think it's worth it. It's already working a lot better than the current `main`.

dacap (approved):

dacap:
It works great @Liebranca! I'll merge it as it is, there is a minor memory overhead now that we use `ReplaceImage` instead of `CopyRect`, but as [I mentioned](https://github.com/aseprite/aseprite/pull/4643#discussion_r1752237044) it's something we can address in the future.

Issue #4027, opened 2023-08-31
Moving cel action does not respect 'Snap To Grid'


Gasparoken:
The moving cel action via `MovingCelState::onCommitMouseMove`, does not respect `Snap To Grid` in any case.

Aseprite version: 1.3rc6

Liebranca:
Quick test -- borrowing (or rather pasting in) some bits from `PixelsMovement::moveImage` seems to work fine in the basic case:

```cpp
bool snapToGrid = (Preferences::instance().selection.snapToGrid() &&
                   editor->docPref().grid.snap());

celBounds.x += intOffset.x;
celBounds.y += intOffset.y;

if (snapToGrid) {
  gfx::Rect gridBounds = editor->getSite().gridBounds();
  gfx::PointF gridOffset(
    snap_to_grid(
      gridBounds,
      gfx::Point(celBounds.origin()),
      PreferSnapTo::ClosestGridVertex));

  celBounds.setOrigin(gridOffset);
}

cel->setBounds(gfx::Rect(celBounds));
```

Repeating this operation almost verbatim for `MovingCelState::onMouseUp` was needed to make the cel stay in place after letting go of the mouse, so perhaps this can be it's own function, if only to avoid duplication.

I'm not entirely sure if what I'm doing here is correct, however, but it looks like a place to start.

dacap:
I'm testing #4646 and I'm still not sure if the cel movement should snap to grid with the cel origin, or relative to the cel current position:

![snap-cel](https://github.com/user-attachments/assets/f19aa4bd-ef70-47a5-bbce-61f8874cf2ad)

@Gasparoken do you remember what user reported this problem? (just to check their use case)

dacap:
I'm not sure if it'd make sense to have some kind of pivot option to snap cels to that pivot position:

![image](https://github.com/user-attachments/assets/783e315b-785b-4b4d-99a7-0ad26dae42f6)

There are future plans to include a pivot for cels (#1357) but it's not planned yet.

Liebranca:
Just as an idea, we can make it "magnetic" too. For example:
```C++
void MovingCelState::calcPivot()
{
  // Make a union of all cel bounds
  gfx::RectF bounds;
  for (size_t i=0; i<m_celList.size(); ++i)
    bounds |= m_celStarts[i];

  // Consider all nine points
  float midw = bounds.w * 0.5;
  float midh = bounds.h * 0.5;
  gfx::PointF pivot;
[...]

PR #4646, opened 2024-09-09
 Snap cels to grid on moving cel action (fix aseprite#4027)

Liebranca:
Used the eponymous function from `app/snap_to_grid` to align the offsets and final position of cels to the grid during `MovingCelState`, when 'Snap To Grid' is enabled. Just for completeness, I've included scaling in this fix, though right now it should only affect reference layers far as I can tell.

A small detail: I've made it so the origins of cel bounds aren't individually snapped to the grid in cases where multiple layers are selected, simply because it felt weird when I tested it. In such situations, the cels will simply move together in grid-sized intervals. This can be easily reversed by changing the `snapToGrid && !m_multiLayer` conditionals to just `snapToGrid`.

### DIFF
--- a/src/app/ui/editor/moving_cel_state.cpp
+++ b/src/app/ui/editor/moving_cel_state.cpp
@@ -17,6 +17,7 @@
 #include "app/context_access.h"
 #include "app/doc_api.h"
 #include "app/doc_range.h"
+#include "app/snap_to_grid.h"
 #include "app/tx.h"
 #include "app/ui/editor/editor.h"
 #include "app/ui/editor/editor_customization_delegate.h"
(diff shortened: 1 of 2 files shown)


dacap:
Hi @Liebranca, I've just [tested this](https://github.com/aseprite/aseprite/issues/4027#issuecomment-2359395335) and I'm still not sure how the snap to grid should behave for cels (probably with a pivot option? not sure). As a side note I noticed that the mouse cursor is unaligned to the snap position (i.e. the snap position is too far from the cursor position, I guess).

Liebranca:
Hi @dacap! I added a placeholder that a pivot option could take advantage of later on. A reference point and the result of `calcFullBounds` are used to temporarily offset the grid, and that controls where the cels will snap to. If the reference point is set to `(0,0)`, then the cel or selected range will move relative to the initial position.

So an option, if added in the future, could set that reference point to another value to change this behaviour. For instance, setting the reference point to `calcFullBounds().point2()` would snap to the lower right corner.

dacap:
@Liebranca I've just tested it and it works great :raised_hands: I'll merge it right now.

Issue #4642, opened 2024-09-06
Configure pre-commit hooks: lint commit / clang-tidy / clang-format


dacap:
It'd be great to create/configure some scripts to lint commits / clang-tidy-diff / clang-format-diff as pre-commit hooks for Git locally, and write a section in the [contributing guide](https://github.com/aseprite/aseprite/blob/main/CONTRIBUTING.md) about how to do it (on all platforms).

Our [lint commit action](https://github.com/aseprite/aseprite/blob/main/.github/workflows/lint_commit.yml) is based on [serenityOS one](https://github.com/SerenityOS/serenity/blob/master/.github/workflows/lintcommits.yml). They already have a pre-commit hook [script](https://github.com/SerenityOS/serenity/blob/master/Meta/lint-commit.sh) that is configured as [described here](https://github.com/SerenityOS/serenity/blob/master/CONTRIBUTING.md#commit-hooks) (be aware that our lint commits actions are a little different).

Other projects to see:
* https://github.com/barisione/clang-format-hooks

ckaiser:
I'm a fan of the way [Godot does it](https://github.com/godotengine/godot/blob/master/.pre-commit-config.yaml), using [pre-commit](https://pre-commit.com/).

It's one-command and it resolves dependencies automatically and in a self-contained way.

dacap:
Pre-commit project looks great (I’m not a big fan of Python as a dependency but we can give it a try).  We can locate all required scripts into a misc/ folder if needed. Would you like to give a try @ckaiser?

ckaiser:
I like Python for this kind of stuff because it's pretty ubiquitous as a dependency and more cross-platform than bash scripts which I can definitely run on Windows with something like Git Bash but it's kind of an annoyance versus more first-party support which can have dedicated platform detection for it, and it's better to write IMO.

I'm down to go for it!

PR #4654, opened 2024-09-12
Add pre-commit configuration and documentation.

ckaiser:
Implements [#4642](https://github.com/aseprite/aseprite/issues/4642), which adds `clang-format`, `clang-tidy`, and `codespell` as pre-commit hooks.

`clang-tidy` is not run automatically, that one is manual but it's nice to have to make it easier to run with `pre-commit run --hook-stage manual clang-tidy`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,38 @@
+exclude: 'third_party/.*|laf/.*'
+
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.6.0
+    hooks:
+      - id: end-of-file-fixer
+      - id: trailing-whitespace
+
+  - repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.5.5
+    hooks:
+      - id: remove-tabs
+        args: [--whitespaces-count, '8']
+
+  - repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: v18.1.8
+    hooks:
+      - id: clang-format
+        files: \.(c|h|cpp|hpp|cc|hh|cxx|hxx)$
+        types_or: [text]
+
+  - repo: https://github.com/pocc/pre-commit-hooks
+    rev: v1.3.5
+    hooks:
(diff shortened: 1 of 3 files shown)


dacap:
Thanks @ckaiser, I'll try to test this soon, just checking the `codespell` plugin, does it works with camelCase? because we use it a lot. I found this: https://github.com/codespell-project/codespell/issues/196

About `clang-format`, I'm not sure if we should format the whole file or just the diff. I think that it'd be more practical to format all the files for the author, but for merging purpose this might be a nightmare.

It looks like there were some attempts to include `git-clang-format` in `mirrors-clang-format` but it was never added in the repo: https://github.com/pre-commit/mirrors-clang-format/pull/16

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
+exclude: |
+  (?x)^(
+    .*third_party/.*|
+    .*laf/.*|
+  )
```
dacap, line 5: I've tested on Linux and it looks like all .cpp files are being excluded, changed this to
```suggestion
exclude: 'third_party/.*|laf/.*'
```
and it started to work.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+repos:
```
dacap, line 7: Some extra hooks to add:
```suggestion
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.6.0
    hooks:
      - id: end-of-file-fixer
      - id: trailing-whitespace

  - repo: https://github.com/Lucas-C/pre-commit-hooks
    rev: v1.5.5
    hooks:
      - id: remove-tabs
        args: [--whitespaces-count, '8']

```

dacap (changes requested):
Some changes to make it work on Linux (didn't test on Windows and macOS yet).

dacap:
I've cherry picked this into 102624cad3c433e8c09fe1cae9f8ccfea344a9db with a few changes later (disabled codespell + updated packages).

PR #4660, opened 2024-09-14
Use new list_files, parse dates to check if sessions are old

ckaiser:
Use the new filtering in `base::list_files` and sessions parse the filename dates to avoid overscanning when the directories are copy/pasted or their edit date changes for some reason.

### DIFF
--- a/src/app/crash/log.h
+++ b/src/app/crash/log.h
@@ -10,7 +10,10 @@
 
 #include "base/debug.h"
 
-// TODO should we log some events in release?
-#define RECO_TRACE(...) // TRACE(__VA_ARGS__)
+#ifdef _DEBUG
+  #define RECO_TRACE(...) TRACE(__VA_ARGS__)
+#else
+  #define RECO_TRACE(...)
+#endif
 
 #endif
--- a/src/app/crash/log.h
+++ b/src/app/crash/log.h
@@ -10,10 +10,7 @@
 
 #include "base/debug.h"
 
-#ifdef _DEBUG
-  #define RECO_TRACE(...) TRACE(__VA_ARGS__)
-#else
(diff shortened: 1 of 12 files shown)


src/app/crash/log.h
```diff
@@ -10,7 +10,10 @@
 
 #include "base/debug.h"
 
-// TODO should we log some events in release?
```
ckaiser, line 13: @dacap I changed this to work only in dev builds because it was pretty helpful to debug session issues (which we'll probably be doing for the performance problems it currently has), but I can undo it if it's not desired.

dacap: I'd prefer to keep the `RECO_TRACE()` as it was, so we enable it only in case that we need to debug this (instead tracing all this info on Debug builds).

src/app/crash/session.cpp
```diff
@@ -147,18 +149,40 @@ bool Session::isOldSession()
...
+  if (parts.size() == 3 && parts[0].size() == 8) {
+    try {
+      sessionTime = base::Time(filenamePartToInt(parts[0].substr(0, 4)),
+                               filenamePartToInt(parts[0].substr(4, 2)),
+                               filenamePartToInt(parts[0].substr(6, 2)),
+                               0,
+                               0,
+                               0);
+    }
+    catch (std::exception& ex) {
```
dacap, line 167: Super minor thing (I think there are a lot of `catch()` statements without `const` that should be `const` out there, so nothing to worry about it):
```suggestion
    catch (const std::exception& ex) {
```

dacap (changes requested):
Looks great so far! Just some minor comments.

After merging this I'll give a try to move the listing of files to a background thread as we talked.

PR #4666, opened 2024-09-17
Add symmetry to image brushes

Gasparoken:
During diagonal symmetry development I implemented the flip of image brushes itself.
This is the version for 4 quadrant symmetry (current).
@dacap Can you give me an overview of the approach?

Note 1: I need to review the evidence, I will be more comfortable with the approval of the approach anyway.
Note 2: Tiled Mode need some fix in some cases.

### DIFF
--- a/src/app/tools/ink.h
+++ b/src/app/tools/ink.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2022  Igara Studio S.A.
+// Copyright (C) 2018-2024  Igara Studio S.A.
 // Copyright (C) 2001-2017  David Capello
 //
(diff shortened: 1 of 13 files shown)


dacap:
> @dacap Can you give me an overview of the approach?

I wasn't able to check this in detail yet. We could implement the diagonal symmetry ASAP with this approach (creating requested flipped versions inside the brush).

On the other hand I see that there are a lot of bugs with the current symmetry impl (e.g. enabling Dynamics creates non-symmetrical pixels). The solution would be to move the symmetry part into other part of the ToolLoop pipeline (instead of copying strokes, copying pixels), but that will be for another issue (I'll create a new one soon, I don't know if it already exists).

Issue #4610, opened 2024-08-22
Add feedback to "Recover Files" option when there are a lot of files/sessions to recover


dacap:
This requires some testing, but I think there is a "little" hang of the UI when we press the Recover Files button (probably a `base::list_files()` blocking the whole UI or something like that).

Reports:
- https://steamcommunity.com/app/431730/discussions/0/4427688158474122991/
- https://steamcommunity.com/app/431730/discussions/2/4758703722109585564/
- https://steamcommunity.com/app/431730/discussions/2/4757578099471857947/

dacap:
Seeing several reports recently where the "Recover Files" option is taking time/freezes the program, not sure if there is a regression of some kind (?)

ckaiser:
I've been doing some investigating and it's a bit of a maze but hopefully I'll have something concrete soon, I've already found some optimizations for the raw crawling of files but I'm still unclear on how and why it's doing some of the work since it's a bit inconsistent

dacap:
At the moment we've fixed this issue with #4660 and #4672 improving the performance to list/display files/items. If further performance issues are reported we can reopen this.

PR #4672, opened 2024-09-19
Don't read the magic number of all objects to show a backup description (#4610)

dacap:
For #4610

Now we lazily initialize the description of each backup on each session. This means that only when we have to display the item on the screen (onPaint) we'll ask for the description/doc info (width/height/color mode, etc.). We've also removed the check of all magic numbers of every single object in the backup when we only need the doc description.

@ckaiser assigning this to you so you can give some feedback and give it a try.

### DIFF
--- a/src/app/crash/read_document.cpp
+++ b/src/app/crash/read_document.cpp
@@ -91,7 +91,12 @@ public:
       if (!id || !ver)
         continue;               // Error converting strings to ID/ver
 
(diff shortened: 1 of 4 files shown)


ckaiser:
Looks great! Tested it with big sessions and it turns a multi-minute hang into an instant operation 🎉

dacap:
Thanks @ckaiser! I'll merge it :+1:

PR #4680, opened 2024-09-24
Fix Sentry wrapper compilation

ckaiser:
Accidentally broke this with #4660 and CI doesn't compile with sentry enabled (and neither did I until today!) so it didn't catch it


dacap:
Great @ckaiser! One thing that might help you to compile Aseprite is using `-D "CRASHPAD_ZLIB_SYSTEM:BOOL=OFF"` (in case you have a linker error about `zlibstatic` when the crash handler is linked).

Issue #2082, opened 2019-06-05
Show multiplied grid cell number


hiddenasbestos:
Currently the status bar shows the x and y grid cell the mouse pointer is in, which is great.
I'd like to request an enhancement to this which would also show the multiplied grid cell number, as in `x + y * w`  where w is the number of horizontal cells in the image.

The benefit here is for retro tile maps, where a single number is used to specify the tile rather than an x/y co-ordinate. Doing this multiplication for me will save a ton of time when referencing sprites from a sheet.

dacap:
Already implemented with #4706, ebd17b0ced4ce14f68c4828e766e16c174aa217d and 5464c4a3c2a614b65b547a17b70b003f741de369

dacap:
![statusbar2](https://github.com/user-attachments/assets/dfc94864-c926-4a58-961b-1b4a43fb18c2)

PR #4706, opened 2024-10-16
Show multiplied grid cell index

Liebranca:
Implements #2082. The cell index is displayed in the same way the tile index is shown for a tilemap layer, as both values serve more or less the same purpose.

To try and avoid confusion, I've made it so the cell index only appears if the active layer is not a tilemap, else the tile index logic is used. I'd like some feedback on this as I'm not entirely sure how obvious this would be from the user's perspective.

Perhaps it'd be more clear if both values are displayed? I thought of using `()` parentheses for the cell index rather than `[]` brackets, so as to differentiate them. Maybe this could be done only when both indices are displayed.

### DIFF
--- a/src/app/ui/editor/standby_state.cpp
+++ b/src/app/ui/editor/standby_state.cpp
@@ -601,17 +601,6 @@ bool StandbyState::onUpdateStatusBar(Editor* editor)
         gfx::Point pt = grid.canvasToTile(gfx::Point(spritePos));
         buf += fmt::format(" :grid: {} {}", pt.x, pt.y);
 
-        // Show the grid cell index
-        int columns = int(std::floor(
-          sprite->bounds().w/grid.tileSize().w));
-        int rows = int(std::floor(
-          sprite->bounds().h/grid.tileSize().h));
-        int column = pt.x%columns;
-        int row = pt.y%rows;
-        if (row < 0) row = row + rows;
-        if (column < 0) column = column + columns;
-        buf += fmt::format(" :arrow_down: {}", column+row*columns);
-
         // Show the tile index of this specific tile
         if (site.layer() &&
             site.layer()->isTilemap() &&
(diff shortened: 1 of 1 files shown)


src/app/ui/editor/standby_state.cpp
```diff
@@ -614,6 +614,21 @@ bool StandbyState::onUpdateStatusBar(Editor* editor)
...
+        else if ((!site.layer() ||
+            !site.layer()->isTilemap()) &&
+            (spritePos.x >= 0 && spritePos.x <= sprite->width() &&
+            spritePos.y >= 0 && spritePos.y <= sprite->height())) {
+          int columns = int(std::floor(
+            sprite->bounds().w/grid.tileSize().w));
+          int rows = int(std::floor(
+            sprite->bounds().h/grid.tileSize().h));
+          int column = pt.x%columns;
+          int row = pt.y%rows;
```
dacap, line 627: I'd prefer to check for columns != 0 and rows != 0 before doing this.

src/app/ui/editor/standby_state.cpp
```diff
@@ -614,6 +614,21 @@ bool StandbyState::onUpdateStatusBar(Editor* editor)
             buf += fmt::format(" [{}{}]", ti, str);
           }
         }
+        // Show the grid cell index
+        else if ((!site.layer() ||
+            !site.layer()->isTilemap()) &&
+            (spritePos.x >= 0 && spritePos.x <= sprite->width() &&
+            spritePos.y >= 0 && spritePos.y <= sprite->height())) {
```
dacap, line 621: We can use
```suggestion
            sprite->bounds().contains(spritePos)) {
```

dacap (changes requested):
Sorry for the late testing about this PR @Liebranca. As you said I think it might be confusing using the same notation `[]` for two different purposes: in tilemaps it shows the tile index, and with this PR in normal layers it would shows the grid tile index in a linear array.

I also think that this field might probably be most useful in tilemap layers, so it should appear in all kind of layers. Not sure about the notation, but maybe showing something like:

```c++
          buf += fmt::format(" :search: {}", column+row*columns);
```

And I'm not sure what we should do with partially visible grid tiles:

https://github.com/user-attachments/assets/27421cc6-d4b7-42fe-933b-1a07eb9a328d

Liebranca:
Hi @dacap! About partially visible cells, I purposely made it so the index wraps back to the beggining or end of the row/column. That way a cell being "cut" by grid offset gives the same index as the one on the opposite side, as if the missing portion of the tile was continued there.

However, this may not make that much sense after all. A possible alternative I'm testing right now is considering such partially visible cells as individual tiles, rather than portions of a single tile spread across two to four corners of the screen.

PR #4710, opened 2024-10-18
Fix scaling of brush type popup window's icons

### LINKED ISSUE #4386: `brush_type` popup window icons use incorrect scaling ratio
version: main branch after 1.3.5

## Settings 1: 

image
image

## Settings 2:

image
image

It seems that the scaling ratio of these three icons is related to the font size, which is a bit strange. They should load according to the value of `uiscaling`.

By the way, for the default theme, should the following two be equivalent (In terms of display effect)?

```xml
<theme name="" screenscaling="2" uiscaling="1">
```

```xml
<theme name="" screenscaling="1" uiscaling="2">
```

martincapello:
Fix #4386

### DIFF
--- a/src/app/ui/brush_popup.cpp
+++ b/src/app/ui/brush_popup.cpp
@@ -80,6 +80,11 @@ public:
     : m_brushes(App::instance()->brushes())
     , m_brush(brush)
     , m_slot(slot) {
+    initTheme();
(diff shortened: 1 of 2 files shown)


Gasparoken:
Works perfectly. LGTM!

PR #4718, opened 2024-10-22
Fix MouseEnter and MouseLeave

### LINKED ISSUE #4240: Pixel-hunting with Screen Scale 100%, UI Scale 200%
Test

martincapello:
Callback messages were not being handled and since MouseEnter and MouseLeave events were wrapped into CallbackMessages they were lost and so never got called. 
Fix #4240.

### DIFF
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -498,6 +498,7 @@ void Manager::generateMessagesFromOSEvents()
           set_mouse_cursor(kArrowCursor);
           mouse_display = display;
         });
+        msg->setRecipient(this);
         enqueueMessage(msg);
         lastMouseMoveEvent = osEvent;
         break;
@@ -513,7 +514,7 @@ void Manager::generateMessagesFromOSEvents()
             mouse_display = nullptr;
           }
         });
-
+        msg->setRecipient(this);
         enqueueMessage(msg);
 
         // To avoid calling kSetCursorMessage when the mouse leaves
@@ -1664,6 +1665,11 @@ bool Manager::onProcessMessage(Message* msg)
       }
       else
         return false;
+    }
+    case kCallbackMessage: {
+      CallbackMessage* callback = static_cast<CallbackMessage*>(msg);
+      callback->call();
+      return true;
     }
 
   }
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -1666,11 +1666,6 @@ bool Manager::onProcessMessage(Message* msg)
       else
         return false;
     }
-    case kCallbackMessage: {
-      CallbackMessage* callback = static_cast<CallbackMessage*>(msg);
-      callback->call();
-      return true;
-    }
 
   }
 
--- a/src/ui/widget.cpp
+++ b/src/ui/widget.cpp
@@ -1604,6 +1604,11 @@ bool Widget::onProcessMessage(Message* msg)
         return true;
       }
       break;
+    case kCallbackMessage: {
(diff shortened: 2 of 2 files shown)


dacap:
:+1: It looks like `kMouseEnter/LeaveMessage` are generated only by the first `os::Event::MouseMove` (or any other mouse event using `Manager::setMouse()`), and `os::Event::MouseEnter/Leave` are totally ignored. I'll test this a little more and merge it ASAP.

src/ui/manager.cpp
```diff
@@ -1665,6 +1666,11 @@ bool Manager::onProcessMessage(Message* msg)
       else
         return false;
     }
+    case kCallbackMessage: {
+      CallbackMessage* callback = static_cast<CallbackMessage*>(msg);
+      callback->call();
+      return true;
+    }
```
dacap, line 1673: Probably this should be in `ui::Widget::onProcessMessage()`.

martincapello: Now that I'm thinking about this a bit more...why do we need the CallbackMessage? Couldn't we just wrap the `os::Event::MouseEnter/Leave` in a `os::Event::Callback` instead? So we avoid to have two different callback handling mechanisms?

I will try this to see what happens, please let me know if you see a downside that I'm missing.

martincapello: I see...
I'm just wondering why are we pushing a CallbackMessage instead of translating the `os::Event::MouseEnter/Leave` to the UI's `kMouseEnterMessage/LeaveMessage` corresponding message and handle those in the `Manager::onProcessMessage` method instead. 
Do you think it worth trying to refactor this to remove the CallbackMessage? Because it might end up being impossible if I found something that is too complex to refactor.

dacap: Check the commit description bd91a6430fdc04c3bc4d35b5d0c11de80c763198, the issue is `setMouse()`: it enqueues the corresponding events `kMouseLeaveMessage/kMouseEnterMessage`, but modifies the `HAS_MOUSE` flags of the widget hierarchy and the `mouse_widget` variable. So to keep in sync the message processing with these flags/variables the message is enqueued (so these variables are set in the same order as messages are being processed by each widget).

Anyway, as the `CallbackMessage` wasn't working, probably the fix was just a side effect of processing MouseMove events correctly. I don't exactly remember what specific issue I fixed with this commit, probably something with Windows and multiple input devices like stylus+mouse+trackpad.

If you want to dedicate to refactor and transform all this logic you can go and give a try. Just don't introduce the `hasMouseOver()` function (as with `hasMouse()` should be enough/work as expected since that patch). And don't delete the `CallbackMessage`, keep it working (the `Widget`/any receipt should process this message).

martincapello: Yes, I've checked out the commit description already and saw that the setMouse function is enqueuing the `kMouseLeave/Enter` messages. But, I've also realized that only from the `os::Event::MouseLeave` handler, inside the callback function, setMouse(nullptr) is being called, which means that the `kMouseLeaveMessage` is generated there. In the other hand, no `kMouseEnterMessage` is generated from the `os::Event::MouseEnter` handler.

I'll try to refactor this, and won't introduce hasMouseOver() function back. Also I will keep the CallbackMessage and move the handling logic into the Widget class.

dacap: In that case send the PR to beta. (Or create two PRs, one for main with this fix, and one for beta with the refactor).

martincapello: Makes sense, I will create two PRs

Issue #4708, opened 2024-10-17
Group buttons outside the window bounds


dacap:
There are several UI issues when the window is too small and we could fix. The most important ones are the context bar:

![image](https://github.com/user-attachments/assets/d44bcedf-fb33-4f89-aa86-4339174fd90b)

And the tool box:

![image](https://github.com/user-attachments/assets/869ab01f-3b83-42d4-9b74-93cd754a1d84)

But it can happen in other cases.

One solution would be to group buttons in on last "..." button:

![image](https://github.com/user-attachments/assets/d4844c35-af5e-48d5-8036-11068d72dc2a) ![image](https://github.com/user-attachments/assets/bbe24fd9-7eac-4703-84dd-8b594dd5e2a5)

ckaiser:
I've got a first draft of a solution for the toolbar, as seen here:

https://github.com/user-attachments/assets/6514a08c-577f-48ba-8558-5897c3202345


The only thing that's a bit janky with this solution is that when you get to very silly heights it results in very silly popup lengths:
![image](https://github.com/user-attachments/assets/8d15f3bf-c37f-42f4-a13a-de44d8883bba)

but it's _technically_ still usable, so it's a definite improvement over what's currently happening.

If you like the solution @dacap I'll clean it up and open a PR.

dacap:
@ckaiser it looks great! you can open the PR as it is, and in a future we could see how to rearrange the buttons better.

With your video I've noticed another (unrelated) issue: https://github.com/aseprite/aseprite/issues/4719, but that is for other PR.

PR #4720, opened 2024-10-24
Handle toolbar in small sizes with an overflow button

ckaiser:
Fixes the toolbar side of #4708

### DIFF
--- a/src/app/ui/toolbar.cpp
+++ b/src/app/ui/toolbar.cpp
@@ -6,7 +6,7 @@
 // the End-User License Agreement for Aseprite.
 
 #ifdef HAVE_CONFIG_H
-#include "config.h"
+  #include "config.h"
 #endif
 
 #include "app/ui/toolbar.h"
@@ -18,6 +18,7 @@
 #include "app/modules/gfx.h"
 #include "app/tools/active_tool.h"
 #include "app/tools/tool_box.h"
+#include "app/tools/tool_group.h"
 #include "app/ui/keyboard_shortcuts.h"
 #include "app/ui/main_window.h"
 #include "app/ui/preview_editor.h"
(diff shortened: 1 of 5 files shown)


src/app/ui/toolbar.cpp
```diff
@@ -616,6 +675,95 @@ void ToolBar::drawToolIcon(Graphics* g, int group_index, SkinPartPtr skin, os::S
...
+
+    ToolGroup* tool_group = *it;
+    for (ToolIterator bit = toolbox->begin(); bit != toolbox->end(); ++bit) {
+      Tool* tool = *bit;
+      if (tool->getGroup() == tool_group) {
+        auto item = new ButtonSet::Item();
+        // TODO: Tooltip with tool->getText(), implement in ButtonSet?
+        item->setIcon(theme->getToolPart(tool->getId().c_str()));
+        item->setId(tool->getId().c_str());
+        item->setMaxSize(toolIconSize);  // TODO: Need a better way to enforce the button/icon size, this one doesn't look great
```
ckaiser, line 723: @dacap Is there a way to constrain the icon size here without changing how ButtonSet renders things?
image

It doesn't look super great with some of the icons, but the default size without specifying anything is massive:
image

dacap:
One thing I didn't mention about it, is that we should try to avoid merging clang-format changes at the moment. The #4654 was not yet merged because I have some branches modifying several files (including the toolbar to support a left-side toolbar :cold_sweat:).

ckaiser:
Ah, yeah - I wanted to run it mostly for clang-tidy but it formatted everything and I didn't have it committed beforehand 😭. Most of the changes in the header are indents so I'll just revert them

dacap:
The grouped popup at the bottom is smaller than the regular group used in other tool groups:

image

It should be of the same size that a regular tool group.

dacap:
Still the buttons are misaligned (compare the rectangle/ellipses as a regular tool group and as items of the last hidden groups):

image

dacap:
I've just merged 9b90159d1bdd019e24ef41f6fc76650843d99c2c with some extra patches in a600c0012727d61827b6131ba3bb822313ccfb18 (some details in its commit description):

https://github.com/user-attachments/assets/b78550f7-650a-4934-bbe8-eec0d8539321

PR #4723, opened 2024-10-24
Fix Magic Wand doesn't work correctly in Tiled Mode (fix #4659)

### LINKED ISSUE #4659: Magic Wand doesn't work correctly in Tiled Mode
In Tiled Mode, Magic Wand works only when we click in the top-left ninth:

Peek 2024-09-12 19-20

From: https://community.aseprite.org/t/i-cant-select-with-the-magic-wand-tool/23399/6?u=dacap

Gasparoken:
fix #4659

### DIFF
--- a/src/app/ui/editor/tool_loop_impl.cpp
+++ b/src/app/ui/editor/tool_loop_impl.cpp
@@ -360,6 +360,15 @@ public:
             == app::gen::PixelConnectivity::EIGHT_CONNECTED);
   }
 
+  bool isPointInsideCanvas(const gfx::Point& point) override {
+    const int a = ((getTiledMode() == TiledMode::X_AXIS ||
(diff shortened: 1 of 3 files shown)


src/app/ui/editor/tool_loop_impl.cpp
```diff
@@ -360,6 +360,15 @@ class ToolLoopBase : public tools::ToolLoop {
             == app::gen::PixelConnectivity::EIGHT_CONNECTED);
   }
 
+  bool isPointInsideTheCanvas(const gfx::Point& point) override {
```
martincapello, line 363: Just a minor thing, I think I would rename this as isPointInsideCanvas().

martincapello (approved):
Works great! I just did a minor comment, feel free to omit it if you like the name as is.

dacap:
Works great!

PR #4741, opened 2024-10-28
Refactor MouseEnter and MouseLeave events

martincapello:
Instead of wrapping the MouseEnter and MouseLeave OS events handling inside a UI CallbackMessage, we now enqueue the corresponding kMouseEnterMessage and kMouseLeaveMessage.
This is actually almost the same we were doing before using the CallbackMessage, but here I'm adding a couple of lines to set the mouse widget (and thus enqueue the UI message) when a `os::Event::MouseEnter` is received.
Also, the HAS_MOUSE flag setting was moved to the Widget class, making it simpler because it takes advantage of the fact that the MouseEnterMessage/MouseLeaveMessage are propagated to the parent already, so each widget can set the state of its own flag. This also has the benefit that we avoid to disable the HAS_MOUSE flag before the MouseLeaveMessage is handled by the widgets, so they have a chance to know if they had the mouse before the MouseLeaveMessage.

### DIFF
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -488,34 +488,29 @@ void Manager::generateMessagesFromOSEvents()
       }
 
       case os::Event::MouseEnter: {
-        auto msg = new CallbackMessage([osEvent, display]{
-          if (get_multiple_displays()) {
-            if (osEvent.window()) {
-              ASSERT(display != nullptr);
-              _internal_set_mouse_display(display);
-            }
+        if (get_multiple_displays()) {
+          if (osEvent.window()) {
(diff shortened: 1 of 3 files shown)


dacap:
LGTM :+1:

dacap:
I've just found a bug on macOS with this PR (probably on other platforms), click the Text combo box size and then try to click other size. Without this PR it works:

https://github.com/user-attachments/assets/1876a776-9b0d-4276-adbc-c58daf827418

And now it doesn't work:

https://github.com/user-attachments/assets/9808256f-f96c-44c2-a66a-9952a4410c94

I'm going to create a new issue about this.

PR #4750, opened 2024-10-31
Add "Apply" button to dialogs used to apply filters

### LINKED ISSUE #4678: Add 'Apply' button to edit dialogs like Replace Color
From: https://community.aseprite.org/t/replace-color-adding-apply-button/23485/2
This can improve editing speed when multiple changes are required.

* Aseprite version: 1.3.8.1

martincapello:
With this change a user can apply the filter without having to re-open the window. Fix #4678


Gasparoken:
It works perfectly. Perhaps it can be merged as is. LGTM!

However, I must anticipate possible complaints from some artists that could be addressed in new issues:
1. After an Apply is pressed, we cannot undo. To correct the filter we need to press Cancel, then Undo, open the filter dialog again. 
2. When Apply button is pressed, filter is applied, then preview does not show result of first filter step, instead it shows result of filtered result of filtered image since sliders/inputs have values ​​different than 0. Maybe, one option is to reset sliders/inputs after running Apply (but this can be a problem in some cases, we need to deliberate which is better in this case).

Gasparoken (approved):

martincapello:
Thanks for the review!

> 1. After an Apply is pressed, we cannot undo. To correct the filter we need to press Cancel, then Undo, open the filter dialog again.

Yes, maybe we could consider letting the user undoing while applying.

> 2. When Apply button is pressed, filter is applied, then preview does not show result of first filter step, instead it shows result of filtered result of filtered image since sliders/inputs have values ​​different than 0. Maybe, one option is to reset sliders/inputs after running Apply (but this can be a problem in some cases, we need to deliberate which is better in this case).

Let's see what @dacap thinks about these issues.

dacap:
Probably we could put the buttons in the:
```
[   OK   ]
[ Apply  ]
[ Cancel ]
```
order as in the Preferences dialog:

image

PR #4762, opened 2024-11-04
Fix duplicate Sprite or Layer don't duplicate z-indexes (fix #4665)

### LINKED ISSUE #4665: Duplicate Sprite & Duplicate Layer don't duplicate z-indexes
Just duplicating a sprite (Sprite > Duplicate) or a layer (Layer > Duplicate) will not duplicate the z-index of each cel of that sprite layers/specific layer.

EDIT: @Gasparoken wasn't able to reproduce this bug so it looks like other kind of problem, maybe related to a specific file/combination with other thing. Duplicating this sprite: [duplicate-sprite-bug.zip](https://github.com/user-attachments/files/17317517/duplicate-sprite-bug.zip) fails.

Gasparoken:
fix #4665

### DIFF
--- a/src/app/doc.cpp
+++ b/src/app/doc.cpp
@@ -528,6 +528,7 @@ void Doc::copyLayerContent(const Layer* sourceLayer0, Doc* destDoc, Layer* destL
       if (it != linked.end()) {
         newCel.reset(Cel::MakeLink(sourceCel->frame(),
                                    it->second));
+        newCel->setZIndex(sourceCel->zIndex());
       }
       else {
         newCel.reset(create_cel_copy(nullptr, // TODO add undo information?


dacap:
@martincapello I'm taking this one for review.

src/app/cmd/copy_cel.cpp
```diff
@@ -116,12 +116,18 @@ void CopyCel::onExecute()
...
+      dstCelZOrder = dstCel->zIndex();
       executeAndAdd(new cmd::RemoveCel(dstCel));
+    }
 
     if (srcCel) {
-      if (createLink)
+      if (createLink) {
         dstCel = Cel::MakeLink(m_dstFrame, srcCel);
+        dstCel->setZIndex(dstCelZOrder);
+      }
```
dacap, line 130: This `dstCelZOrder` is not required if we keep the change in `Cel::MakeLink()` as it was, because:
1. `dstCelZOrder` is used only when `srcCel != nullptr`
2. so `dstCelZOrder = srcCel->zIndex()`
3. and `dstCelZOrder` is used only in `dstCel->setZIndex(dstCelZOrder);` after `dstCel = Cel::MakeLink(m_dstFrame, srcCel);`

So if we set the zindex in `MakeLink()` (as we was doing), we don't require this change at all.

src/app/doc.cpp
```diff
@@ -528,6 +528,7 @@ void Doc::copyLayerContent(const Layer* sourceLayer0, Doc* destDoc, Layer* destL
       if (it != linked.end()) {
         newCel.reset(Cel::MakeLink(sourceCel->frame(),
                                    it->second));
+        newCel->setZIndex(sourceLayer0->cel(sourceCel->frame())->zIndex());
```
dacap, line 531: This can be:
```
        newCel->setZIndex(sourceCel->zIndex());
```
as `sourceCel` is already equal to `sourceLayer0->cel(sourceCel->frame())`.

src/doc/cel.cpp
```diff
@@ -54,9 +54,7 @@ Cel* Cel::MakeCopy(const frame_t newFrame,
 Cel* Cel::MakeLink(const frame_t newFrame,
                    const Cel* other)
 {
-  Cel* cel = new Cel(newFrame, other->dataRef());
-  cel->setZIndex(other->zIndex());
-  return cel;
+  return new Cel(newFrame, other->dataRef());
```
dacap, line 57: Revert this, we can copy the z-index of the given cel and adjust later if needed.

dacap (changes requested):
I think the only required change to fix this bug is:
```
newCel->setZIndex(sourceCel->zIndex());
```
after making the `Cel::MakeLink()` in `Doc::copyLayerContent()`.

PR #4767, opened 2024-11-06
Add parameters to the DuplicateSprite command (fix #4755)

### LINKED ISSUE #4755: Add parameters for Duplicate Sprite command
The `DuplicateSprite` command code seems to be quite old and doesn't support any parameters, it would be nice to have these, e.g.:
``` lua
app.command.DuplicateSprite {
    ui = false,
    filename = "name or path"
}
```

ckaiser:
Implements #4755, adds `ui`, `flatten` and `filename` params:

```lua
app.command.DuplicateSprite{
	ui=false,
	flatten=true,
	filename="duplicate.ase"
}
```

### DIFF
--- a/src/app/commands/cmd_duplicate_sprite.cpp
+++ b/src/app/commands/cmd_duplicate_sprite.cpp
@@ -10,6 +10,8 @@
 
 #include "app/app.h"
 #include "app/commands/command.h"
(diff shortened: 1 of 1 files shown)


dacap:
It works great! :raised_hands:

Issue #1303, opened 2016-10-24
Associate user data to an empty cel


dacap:
We should be able to associate user data to an empty cel. Or avoid deleting the cel automatically when it contains associated user data (related to #559).

--

For the developer: a cel with a `nullptr` Image might be problematic in the actual code, as a first step we might keep a cel with an empty Image (e.g. an Image with one transparent pixel).

PR #4769, opened 2024-11-06
Fix 'cel.image = nil' is not allowed (fix #4514)

### LINKED ISSUE #4514: cel.image = Image(cel.image, rectangle) can error
`cel.image = Image(cel.image, rectangle)` can error with `[string "internal"]:1: bad argument #2 to 'f' (ImageObj expected, got nil)` without any line number nor stack trace.

`Image(cel.image, rectangle)` can return nil if rectangle is zero width and zero height yet `cel.image` can't be set to this same nil.

Expected behavior:

`Image(cel.image, rectangle)` should not return nil but instead Image(1, 1, cel.image.colorMode) otherwise the colorMode is lost.

`cel.image = nil` should either do:

1: Equivalent of `cel.image = Image()` which will set image to a 1px by 1px transparent image.
 
2: Have a sane error message with a line number and stack trace AND update https://www.aseprite.org/api/image#image to say that it can return nil and you must check for it yourself.

Aseprite v1.3.7-x64 appimage

Gasparoken:
fix #4514

### DIFF
--- a/src/app/script/cel_class.cpp
+++ b/src/app/script/cel_class.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2023  Igara Studio S.A.
+// Copyright (C) 2018-2024  Igara Studio S.A.
 // Copyright (C) 2018  David Capello
 //
 // This program is distributed under the terms of
@@ -9,6 +9,7 @@
 #include "config.h"
 #endif
 
+#include "app/cmd/remove_cel.h"
 #include "app/cmd/replace_image.h"
 #include "app/cmd/set_cel_opacity.h"
 #include "app/cmd/set_cel_position.h"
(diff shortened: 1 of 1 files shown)


martincapello (commented):
The code looks good, but I've been testing this a bit. 
Supposing that the active sprite has the first frame with a cel with some image, then the following script fails when trying to set back the cel.image to a valid image:
```lua
  local sprite = app.sprite
  local cel = sprite.cels[1]
  cel.image = nil;
  cel.image = Image(32, 32)  -- <--- Error here
```

This is the error:
image

Which actually makes sense to any person that knows the internals, because we are deleting the cel, not just the image when setting cel.image to nil.

But if I do this:
```lua
  local sprite = app.sprite
  local cel = sprite.cels[1]
  cel.image = Image(32, 32)
```
There is no error and the image is set to an empty 32x32 rectangle.

So, I believe that from a script developer's point of view this behavior is quite confusing. 

I think that if there were a way to actually keep the cel, it would be more intuitive.

dacap:
> So, I believe that from a script developer's point of view this behavior is quite confusing.
> 
> I think that if there were a way to actually keep the cel, it would be more intuitive.

@martincapello At the moment it's the best we can do, please refer to #1303 for further discussion about how to solve this in a future (there are a lot of issues related to cels being lost, and not just with scripts, but just painting/erasing pixels).

PR #4770, opened 2024-11-06
Add "ui" parameter to MaskByColorCommand

### LINKED ISSUE #2774: Lua Scripting Feature: Add UI=false to MaskByColor Command.
Describe your bug report or feature request here

It would be really great to get the UI=true/false added to the MaskByColor Command via lua scripting. While this command is very useful it's made it quite difficult to automate because of the popup and having to press ok each time.

That being said the ui=true/false could be added to all dialog pop up type commands but I'd love to see this one added if anything :D

1.3-beta4-x64 / Windows 10

Thank you guys for all your hard work on this btw! I know scripting is not the top priority but its such an amazing feature and allows us to do so so much more! :D

martincapello:
Now the user can execute the MaskByColorCommand without displaying the UI by using the "ui=false" parameter. 
Fix #2774.

### DIFF
--- a/src/app/commands/cmd_mask_by_color.cpp
+++ b/src/app/commands/cmd_mask_by_color.cpp
@@ -65,15 +65,6 @@ public:
     // ADD/SUBTRACT/INTERSECT Selection Mode
     , m_isOrigMaskVisible(reader.document()->isMaskVisible()) {
 
-    if (!params.color.isSet())
-      params.color(ColorBar::instance()->getFgColor());
-
-    if (!params.tolerance.isSet())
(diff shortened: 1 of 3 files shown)


Gasparoken (commented):
It's not related with your commit, anyway please change `App::instance()->isGui()` by `context->isUIAvailable()`  (David's old recommendation).

Gasparoken:
LGTM!
This is a thought, not a real problem: I just noticed that with the UI available and using the script:
```lua
app.command.MaskByColor { ui=false }
```
the rest of the default parameters are set to `0` instead of taking the current values ​​from the editor, for example: the default `color` is the mask color rather than the foreground color of the palette.

PR #4772, opened 2024-11-07
Add columns and rows fields to Import Sprite Sheet

### LINKED ISSUE #1287: Add option for known number of elements in Import Sprite Sheet
When importing a sprite sheet, it would be nice being able to specify how many columns and rows the sprite sheet is, instead of the size of each frame.

It is especially useful when the spritesheet has a nime like "animationName_rowsXcolums.png" and moving the sliders or calculating the frame size takes way longer than just saying the number of elements in the file
- Aseprite (version number, installer/portable/Steam/beta/dev): 1.1.9
- System (Win/OSX/Linux, version, distribution): all

martincapello:
Adds two new fields to the "Import Sprite Sheet" dialog: columns and rows. Fix #1287 

~~EDIT: Just realized that I should add new parameters to `app.command.ImportSpriteSheet` to support specifying columns and rows.~~


dacap:
About b53901dab924a56a63edeb13d5d5aeb44c3c6445, I think other kind of fix is needed, if you use the arrow keys it will generate the `onClick()` event too, e.g. on the "Import Sprite Sheet" dialog click "Include partial tiles at bottom/left edges" and then press the <kbd>Up arrow</kbd> (any key that makes the focus moves will do it). Probably the same logic as the kKeyDownMessage is needed for check boxes:

`if ((hasFocus() && (scancode == kKeySpace)) || mnemonicPressed)`

The button case is filtered by the `if (isSelected()) {` I guess.

martincapello:
I've force pushed the commits with the suggested changes. I've just removed the hasFocus() call from the if condition because it was already checked in the parent if.

dacap:
Something minor but that it'd be nice to have is to disable (or hide? put to 1? disable and put to 1?) the Columns or Rows field if we select Vertical Strip or Horizontal Strip respectively:

image

src/app/commands/cmd_import_sprite_sheet.cpp
```diff
@@ -443,6 +554,21 @@ void ImportSpriteSheetCommand::onExecute(Context* context)
     document = context->activeDocument();
     if (!document)
       return;
+
+    Sprite* sprite = document->sprite();
+
+    auto newFrameBounds = calcFrameSize(sprite->size(), params.columns(), params.rows(), params.frameBounds().origin(), params.padding());
```
dacap, line 560: Remember to break long lines.

PR #4775, opened 2024-11-08
Fix dialog title issue with close button (fix #4735)

### LINKED ISSUE #4735: Width-defining Dialog titles issue with close button

When creating a dialog with default bounds, if the title is the width-defining element of that dialog's bounds, it will overlap with the close button, making the left part of it impossible to click on.

Adding a small margin when the title is the width-defining element of a dialog should fix the issue

Example script:

``` Lua

function generateDialogMain()
	DialogMain = Dialog{
		id="ThisIsAVeryLongNameForADialogDontYouThinkBecauseIdo",
		title="ThisIsAVeryLongNameForADialogDontYouThinkBecauseIdo"
	}

	DialogMain:separator{}
	DialogMain:separator{}
	DialogMain:separator{}

	DialogMain:button{
		id="ThisIsAVeryLongNameForADialogDontYouThinkBecauseIdo_button",
        text="OK"
	}

	DialogMain:show{wait=false}

	return DialogMain
end

generateDialogMain()
```

* Aseprite version: v1.3.9.2
* System: Windows

Gasparoken:
fix #4735

### DIFF
--- a/src/ui/window.cpp
+++ b/src/ui/window.cpp
@@ -55,6 +55,28 @@ public:
     setType(kWindowTitleLabelWidget);
     initTheme();
   }
+private:
+  void onSetDecorativeWidgetBounds() override {
+    Label::onSetDecorativeWidgetBounds();
+    // Adjust the bounds of the window title label if a mini-button
+    // overlaps it.
+    if (parent()->window()) {
+      Window* window = parent()->window();
+      gfx::Border margin(0, 0, 0, 0);
+      if (window->style())
+        margin = window->style()->margin();
+      int mostLeftMiniButtonPosition = window->bounds().x2() - margin.right();
+      for (auto child : window->children()) {
+        if (child->isDecorative() &&
+            (child != this) &&
+            child->bounds().x < mostLeftMiniButtonPosition)
+          mostLeftMiniButtonPosition = child->bounds().x;
+      }
+      gfx::Rect newBounds(bounds());
+      newBounds.w -= bounds().x2() - mostLeftMiniButtonPosition;
+      setBounds(newBounds);
+    }
+  }
 };
 
 


martincapello (commented):
It works great! 
However I have to say that I would like not having to rely in the order of the widgets (title and close button in this case)...not sure if this is possible, but one alternative I was thinking of is by using a HBox to make the title expansive and thus avoid the overlapping. I don't know if this has any drawback or if it is even possible. Or maybe it doesn't worth the extra effort because this fix works already.
I'm saying this because the pick function was okay IMO, it just failed to pick a widget when there was overlapping between another widget. In fact the version proposed here, using the reverse order, can fail too if the children are not in the correct order.
So, I feel that the fix should involve only the Window class and leave the pick function as it was.

Gasparoken:
> It works great! However I have to say that I would like not having to rely in the order of the widgets (title and close button in this case)...not sure if this is possible, but one alternative I was thinking of is by using a HBox to make the title expansive and thus avoid the overlapping. I don't know if this has any drawback or if it is even possible. Or maybe it doesn't worth the extra effort because this fix works already. I'm saying this because the pick function was okay IMO, it just failed to pick a widget when there was overlapping between another widget. In fact the version proposed here, using the reverse order, can fail too if the children are not in the correct order. So, I feel that the fix should involve only the Window class and leave the pick function as it was.

Thanks for your feedback. Anyway, I still think it's better to test the last widget first, since the last item in `m_children` is the widget "closest to the artist" and the first item in `m_children` is the "furthest" one. So, if two overlapping widgets can interact with the cursor, I think the natural behavior would be for the visible widget to take priority over the hidden one.
Perhaps this never caused any problems because dialogs are designed so that widgets don't overlap, except in the title bar.

What do you think @dacap?

martincapello:
@dacap Yesterday we were talking with Gaspar about what I'd commented...and despite I also think that changing the order to pick siblings is not THE fix...he made a point about keeping the pick order consistent with the rendering order...which means that if two sibling widgets are rendered one on top of the other (I know this shouldn't happen, but if it does...), the pick algorithm should prefer picking the last rendered, just because it feel more natural/logical.
So now I'm actually in favor of changing the order of picking. But as David also said, we should fix the issue by avoiding overlapping between widgets in the first place.

dacap:
I think there are 3 things we can do:

1) We could reverse the order of widgets on pick (just to handle buggy cases with overlapped widgets when we resize a window smaller of what it should be)
2) Take care of IGNORE_MOUSE flag to ignore widgets with this flag on the pick process (ui::Label has this flag)
3) Fix the width of the title label (so it's not rendered overlapped with the close button or any other decorative widget)

There are other issues with the Preview window title bar that I think will be fixed with those fixes:

image

src/ui/widget.cpp
```diff
@@ -499,7 +499,9 @@ Widget* Widget::pick(const gfx::Point& pt,
       (bounds().contains(pt))) {
     picked = this;
 
-    for (Widget* child : m_children) {
+    auto it = m_children.rbegin();
+    for (; it != m_children.rend(); ++it) {
+      const Widget* child = *it;
```
martincapello, line 504: Alright, this change breaks the Aseprite menus in single window UI mode...they do nothing when clicking the options. I'm investigating a bit why is this happening...

martincapello: > Alright, this change breaks the Aseprite menus in single window UI mode...they do nothing when clicking the options. I'm investigating a bit why is this happening...

This happens because the manager (app::CustomizedGuiManager) has their children not ordered as the rendering order. In this case, the menu goes first and the MainWindow goes last in the children list. Hence, when iterating in reverse order the MainWindow is picked and the mouse down message is sent to it instead of the menu.

martincapello: So...maybe is not a good idea to change the order...because it involves more changes to fix this...on the other hand, maybe it would be nice to fix how the manager is storing its children...but not sure how much work it would take and which other issues could arise.

src/ui/window.h
```diff
@@ -14,13 +14,23 @@
...
 #include "ui/hit_test_event.h"
+#include "ui/label.h"
 #include "ui/widget.h"
 
 namespace ui {
 
   class ButtonBase;
   class Label;
 
+  class WindowTitleLabel : public Label {
```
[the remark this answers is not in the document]
Gasparoken: Ah, you're right! This was an old change due to a discarded approach.

src/ui/widget.cpp
```diff
@@ -737,6 +737,10 @@ void Widget::saveLayout()
 void Widget::setDecorativeWidgetBounds()
 {
   onSetDecorativeWidgetBounds();
+  // Adjust the bounds of the window title label if a mini-button
+  // overlaps it.
+  if (parent() && parent()->window())
+    parent()->window()->adjustTitleLabelBounds();
```
martincapello, line 743: I think we can avoid creating a new window method (adjustTitleLabelBounds) if we override the onSetDecorativeWidgetBounds() in the window class and put the logic there after calling Widget::onSetDecorativeWidgetBounds().

martincapello (commented):
Just realized that we might not need a new public method in the Window class.

src/ui/widget.cpp
Gasparoken: You're right! I followed your advice, but I assigned 'onSetDecorativeWidgetBounds' to 'Window TitleLabel' instead of 'Window'.

src/ui/window.cpp
```diff
@@ -55,6 +55,29 @@ class WindowTitleLabel : public Label {
     setType(kWindowTitleLabelWidget);
     initTheme();
   }
+private:
+  void onSetDecorativeWidgetBounds() override {
+    if (theme())
+      theme()->setDecorativeWidgetBounds(this);
```
martincapello, line 61: We can replace these two lines by a call to Label::onSetDecorativeWidgetBounds(); which basically calls Widget::onSetDecorativeWidgetBounds() and does the same as these two lines.

martincapello (commented):
Looking good! Just left one final comment, just a detail. Feel free to assign this issue to David once addressed.

martincapello (approved):

dacap:
I've just rebased and apply clang-format to the commit here: https://github.com/dacap/aseprite/tree/fix-close-button (you can continue from here)

Just tested the #4735 script and it looks like the title is being clipped when painted for first time (this is with multi-window option off):

image

https://github.com/user-attachments/assets/a3649855-9f57-4ffa-b2bf-0ee12aabcd97

dacap:
Why was this PR closed? You can hard reset (`git reset --hard`) your `fix-close-button` to the rebased commit.

Gasparoken:
Oh sorry, I just deleted the branch on `Gasparoken/fix-close-button` (I think that action closed the PR automatically).
I still fixing this issue based in your fork at https://github.com/dacap/aseprite/tree/fix-close-button. Then I'll push-force my new solution.

dacap:
Generally you in these cases you can just reset hard your branch and push force, the PR will be automatically updated (there is no need to delete your branch). Not sure what happens if you re-create the branch.

PR #4780, opened 2024-11-12
Add 'beforesitechange' event (fix #4569)

### LINKED ISSUE #4569: New beforesitechange event
It'd be nice to have a beforesitechange event for scripts to detect if a new sprite is going to be selected in case we want to save a [document preference](https://www.aseprite.org/api/app_preferences#apppreferencesdocument) before the sprite is changed/closed.

EDIT: This feature is not fully implemented yet until we fix #4785, in the meantime we could disable the event from the API.

ckaiser:
Checked with most actions that can result in a site change but I'm not 100% sure I'm not missing one, need to see that it solves what script authors need the event for.

### DIFF
--- a/src/app/context_observer.h
+++ b/src/app/context_observer.h
@@ -16,7 +16,8 @@ namespace app {
   class ContextObserver {
   public:
     virtual ~ContextObserver() { }
-    virtual void onActiveSiteChange(const Site& site) { }
+    virtual void onActiveSiteChange(const Site& site) { };
+    virtual void onBeforeActiveSiteChange(const Site& fromSite) { };
   };
 
 } // namespace app
--- a/src/app/script/events_class.cpp
+++ b/src/app/script/events_class.cpp
@@ -157,6 +157,7 @@ class AppEvents : public Events
 public:
   enum : EventType {
     Unknown = -1,
+    BeforeSiteChange,
     SiteChange,
     FgColorChange,
     BgColorChange,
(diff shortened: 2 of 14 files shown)


src/app/script/events_class.cpp
```diff
@@ -213,8 +222,11 @@ class AppEvents : public Events
 
   void onRemoveLastListener(EventType eventType) override {
     switch (eventType) {
+      case BeforeSiteChange:
+        [[fallthrough]];
       case SiteChange:
         App::instance()->context()->remove_observer(this);
+        m_addedObserver = false;
```
ckaiser, line 229: @dacap Question on how this might work: I added the `m_addedObserver` boolean to avoid doubling app on observers for SiteChange/BeforeSiteChange, but for removing the listener I'm assuming `onRemoveLastListener` would delete the observer if i do something like:

> Listen to sitechanged -> Listen to beforesitechanged -> Remove sitechanged

I think with the code how it is right now this will remove the observer, making the remaining beforesitechange event unable to trigger, what's the best way to solve this? Just keeping track of the fact that I may be listening to both events and only remove the observer when I'm listening to 0?

dacap: Interesting, probably this `m_addedObserver` can be a counter, so when it starts/goes back to `m_addedObserver==0` we add/remove the observer.

src/app/context_observer.h
```diff
@@ -15,8 +15,9 @@ namespace app {
 
   class ContextObserver {
   public:
-    virtual ~ContextObserver() { }
-    virtual void onActiveSiteChange(const Site& site) { }
+    virtual ~ContextObserver() = default;
+    virtual void onActiveSiteChange(const Site& site) { };
+    virtual void onBeforeActiveSiteChange(const Site& fromSite) { };
```
dacap, line 20: As a side note, it's not necessary to change the destructor to `= default` and even more, the `{ }` is shorter :upside_down_face: Also `;` is not needed in these cases:
```suggestion
    virtual ~ContextObserver() { }
    virtual void onActiveSiteChange(const Site& site) { }
    virtual void onBeforeActiveSiteChange(const Site& fromSite) { }
```

dacap:
It looks great so far! There are some places that I think we might fail to generate this signal correctly (so I think they will be future bugs to fix). I'm testing and already found a bug, but I'll create another issue related to that. I'll merge it as it is :+1:

PR #4797, opened 2024-11-19
Fix locking a layer reverts changes (fix #4586)

### LINKED ISSUE #4586: Locking a layer reverts changes
I have found a bug: When I lock a layer and switch to another layer, the last change may be reverted. I have tested this for the following specific steps, but it is possible that it also occurs for other types of changes.

Here are the steps to reproduce:
* Create an empty image
* Add a second layer (Layer 2)
* Select Layer 1
* B for brush tool
* Draw something
* M for marquee tool
* Ctrl-A to select all
* Arrow up to move the Layer 1 contents one pixel up
* Lock Layer 1
* Select Layer 2

Now the contents of layer one should automatically revert to their position before they were shifted up.

2024-08-04_14h56_16

* Aseprite version: 1.3.7-x64, Steam
* System: Windows 10

Gasparoken:
fix #4586

Some comment about the approach:
This solution solves the problem and seems like a more intuitive behavior.

Anyway, I don't like the following case:
1. `lock` the layer while there is an active selection.
2. `unlock` the layer without any movement of the selection.
Result: The original selection mask will be a new mask containing any parts that were overlapped before step 1. See the following video to understand:
Nov-19-2024 20-05-15
Note: this was also happening with the layer visibility state change.
I need some feedback on the approach.

### DIFF
--- a/src/app/doc_observer.h
+++ b/src/app/doc_observer.h
@@ -10,118 +10,121 @@
 #pragma once
 
 namespace doc {
-  class Remap;
+class Remap;
 }
 
 namespace app {
-  class Doc;
-  class DocEvent;
+class Doc;
+class DocEvent;
 
-  class DocObserver {
-  public:
-    virtual ~DocObserver() { }
+class DocObserver {
+public:
+  virtual ~DocObserver() { }
 
-    virtual void onCloseDocument(Doc* doc) { }
-    virtual void onFileNameChanged(Doc* doc) { }
+  virtual void onCloseDocument(Doc* doc) { }
+  virtual void onFileNameChanged(Doc* doc) { }
 
-    // General update. If an observer receives this event, it's because
-    // anything in the document could be changed.
-    virtual void onGeneralUpdate(DocEvent& ev) { }
+  // General update. If an observer receives this event, it's because
+  // anything in the document could be changed.
+  virtual void onGeneralUpdate(DocEvent& ev) { }
 
-    virtual void onColorSpaceChanged(DocEvent& ev) { }
-    virtual void onPixelFormatChanged(DocEvent& ev) { }
-    virtual void onPaletteChanged(DocEvent& ev) { }
+  virtual void onColorSpaceChanged(DocEvent& ev) { }
+  virtual void onPixelFormatChanged(DocEvent& ev) { }
+  virtual void onPaletteChanged(DocEvent& ev) { }
 
-    virtual void onAddLayer(DocEvent& ev) { }
-    virtual void onAddFrame(DocEvent& ev) { }
-    virtual void onAddCel(DocEvent& ev) { }
-    virtual void onAddSlice(DocEvent& ev) { }
-    virtual void onAddTag(DocEvent& ev) { }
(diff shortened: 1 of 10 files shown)


src/app/doc_observer.h
```diff
@@ -97,8 +97,10 @@ namespace app {
     // The collapsed/expanded flag of a specific layer changed.
     virtual void onLayerCollapsedChanged(DocEvent& ev) { }
 
-    // The visibility flag of a specific layer is going to change/changed.
-    virtual void onBeforeLayerVisibilityChange(DocEvent& ev, bool newState) { }
+    // The 'visibility' or 'editable' flag of a specific layer is going
+    // to change/changed.
+    virtual void onBeforeLayerStatusChange(DocEvent& ev, bool newState) { }
```
martincapello, line 102: I don't like changing the name of this function just because when the lock switches it has to behave the same as when changing visibility. Also there is a onAfterLayerVisibilityChange that now wouldn't have the onBefore.
So, I see two alternatives:
1. Keep the new name, change the onAfterLayerVisibilityChange by onAfterLayerStatusChange for consistency. And add something (a new parameter maybe?) to let the implementer know if the status change is about the visibility or the editable flag.
2. Revert this change to preserve onBeforeLayerVisibilityChange as it was. Add a new onBeforeLayerEditableChange and use it.

By the way, I like more the second approach, but that's just my opinion.

martincapello (commented):
It works good! Just left a comment about part of the implementation.

Gasparoken:
I see that `onBeforeLayerVisibilityChange` and `onBeforeLayerEditableChange` have unused arguments (`editor` and `newState`). I left it as is, since that was the case in the existing `onBeforeLayerVisibilityChange` function. Personally I would remove them.

src/app/ui/editor/moving_pixels_state.cpp
```diff
@@ -597,6 +597,29 @@ void MovingPixelsState::onBeforeLayerVisibilityChange(Editor* editor,
   }
 }
 
+void MovingPixelsState::onBeforeLayerEditableChange(Editor* editor,
+                                                    doc::Layer* layer,
+                                                    bool newState)
+{
```
martincapello, line 603: Now I think that since `onBeforeLayerEditableChange` and `onBeforeLayerVisibilityChange` do the same, you could create a private method that would be called by both. With this I'm saying that in my previous review what I didn't like was changing the interface of DocObserver. But I do like to avoid code duplication when possible.

martincapello (commented):
Almost there! Left one last comment and that's it. About removing the unused parameters, it seems like a good idea but I'm not sure if they are there just to keep consistency with other methods of the same class, in which case I'm not against leaving them there...so I'll leave that decision to David.

Gasparoken:
My first commit with the pre-commit implementation.

Gasparoken:
I'm leaving it as is despite clang-tidy's recommendations. I'm afraid of making changes and having to go back with the modifications for style/readability reasons.

dacap:
Rebased and merge in 912af16453ae76e823675200fc48209437cedff0

PR #4814, opened 2024-11-21
Fix windows on visible and dialog on open

### LINKED ISSUE #4805: Clicking "Disable Snap to Grid" resizes the main window (using multiple windows)
Strange regression in v1.3.10 reported in: https://community.aseprite.org/t/23912

martincapello:
Fix #4805 (`Window::onVisible()` not taking into account if the display was owned)
and https://community.aseprite.org/t/dialog-show-bounds-rectangle-is-broken-in-1-3-10/23917 (`DialogWindow::onOpen()` not calling `Window::onOpen()`)

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -78,6 +78,7 @@ protected:
     if (m_handTool && Editor::activeEditor()) {
       enableHandTool(true);
     }
+    WindowWithHand::onOpen(ev);
   }
 
   void onBeforeClose(CloseEvent& ev) override {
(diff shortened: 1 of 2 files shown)


src/app/script/dialog_class.cpp
```diff
@@ -78,6 +78,7 @@ class DialogWindow : public WindowWithHand {
     if (m_handTool && Editor::activeEditor()) {
       enableHandTool(true);
     }
+    Window::onOpen(ev);
```
dacap, line 81: Should be `WindowWithHand::onOpen()`

martincapello: Damn, give me a min

Issue #720, opened 2015-07-09
Isometric object creation tools


AkashaRepo:
Asepirite has grids and pixel grids which are handy, but they are mostly limited to flat square things. It would be nice if there were tools for making isometic art easier, isometric angle grids, and perhaps a tool to generate perfect isometric shapes of arbitrary size like this one http://codepen.io/nosir/details/ganrh

axelbsa:
+1+

dacap:
From [Steam forums](http://steamcommunity.com/app/431730/discussions/0/343786745997996334/#c343787283744545837):

> Right now, even if you don't have full-on isometric-recognition mode implemented, could you at least please add (x-delta, y-delta) options for the Tiling mode? Something as simple as that would help with isometric tile editing immensely, and I don't think it should be too difficult to implement.

dacap:
On the [community forum too](https://community.aseprite.org/t/isometric-on-aseprite/214).

petracles:
+1 on this, and thanks for listening! Being able to adjust the ratios of the canvas' squares would be a great solution to this.

ellogwen:
thank you very much for this :)

dacap:
Other from [the forum](https://community.aseprite.org/t/isometric-feature/1173)

dacap:
Nobody is working on this one at the moment @jaybill, if you have some mockups about this they will be welcome 👍 I think the first idea that came to my mind is the possibility to create isometric grids (like the regular grid), where snap to grid should work in those new kind of isometric grids.

RyanJEC:
I would also like to see offsets for Tiled Mode so would be easier to make seamless isometric tiles or just make an Isometric Tiled Mode.

pgoral:
I would be also nice to create circle in isometric mode

ScepticDope:
@AkashaRepo maybe check out this plugin? https://darkwark.itch.io/isobox-for-aseprite It creates isometric box shapes and seems to work just fine in v1.2.11

skunheal:
No news on this yet?

dacap:
There are no news about this one yet. We're working in a tilemap editor for v1.3, we hope to be able to introduce isometric grids with tilemaps (maybe v1.3-beta2 or 3).

l4tte:
A tilemap editor? That's going to be great! +1 to the grid but def do the tile editor first, really excited to use it

aadamsx:
[...]

PR #4818, opened 2024-11-21
[WIP] Isometric grid

Liebranca:
Implements an isometric grid and 'Snap To' for it. Part of #720.

The projection angle, so to speak, depends entirely on grid size, and so multiple styles are supported from the get-go:

cubes

But although this is very flexible, it has it's share of issues: not every line angle will give an equally good-looking result for pixel art.

A summary of what's in this PR:

- Drawing of the isometric grid itself, plus a pixel-precise version when zoomed in.

- Added the `Grid::Type` enum to identify which type of grid a sprite is using, plus a drop-down element in grid settings to select between orthogonal (default) or isometric, as well as handling the persistence of this setting.

- Implementation of 'Snap To' for isometric, with a few adjustments for the line tool so that it can be used for drawing.

- When using the line tool and with 'Snap To' enabled, vertical lines can be selectively offset one pixel to the left or right by holding or releasing the 'Angle Snap' key. I added this merely so I could draw tileable cubes quicker as I did a lot of that during testing. I left it there just in case.

And a few things that are still pending or need further polish:

- Offering better control over the pixel-precise grid, for instance, giving an option to toggle off shared edges between cells, or allowing setting specific border width and height.

- Following from the offsetting of vertical lines, I would give the option to set a specific offset value for it, as this comes into play when selecting or moving isometric tiles, and it's not the same offsets for every style.

- I applied the aforementioned line adjustment to the Freehand controller, but it's not a 100% correct, as the end point of one line doesn't always match the start point of the next.

- Because odd sizes are generally problematic, I've set them to be (in a sense) silently ignored via sharing of edges to make them evenly divisible by two, but this doesn't always work as intended.

- Due to how lines intertwine at certain angles, the direction of a stroke matters when drawing (`line(A, B) != line(B, A)`), which may be a bit confusing. I've remedied it for a handful of cases when using the line tool with 'Snap to Grid' enabled, but this is not a complete fix.

- Similarly, length of a stroke is also a factor for `TwoPointsController`, also dependant on line angle.

- There are no adjustments for rectangles and ellipses, only lines.

- With some odd grid sizes, like 30x15 or 18x9, the snap point will be off by one pixel.


dacap:
@Liebranca thanks a lot for all this work! :pray: I'll try to test this step by step, starting tomorrow. It'd be great to integrate a first iteration of the feature soon in the beta.

src/app/tools/controllers.h
```diff
@@ -17,6 +17,94 @@ namespace app { namespace tools {
...
 
+// Adjustment for snap to isometric grid
+static void snap_isometric_line(ToolLoop* loop, Stroke& stroke, bool lineCtl)
+{
+  // Get last two points
+  Stroke::Pt& a = stroke[stroke.size() - 2];
+  Stroke::Pt& b = stroke[stroke.size() - 1];
+
+  // Get function invoked by line tool
+  bool lineTool = (string_id_to_brush_type(loop->getTool()->getId()) == kLineBrushType);
```
dacap, line 28: Do we need to check the line tool or the line brush type? because here we are passing the tool ID to a function that receives a brush type (and I think it works only by chance because `"line"` is a brush type and a tool ID).

Probably you are looking for something like:
```suggestion
  const bool lineTool = (loop->getTool()->getId() == WellKnownTools::Line);
```
But you will need to add:
```c++
const char* WellKnownTools::Line = "line";
```
in `tool_box.cpp` and `.h`.

src/app/ui/editor/editor.cpp
```diff
@@ -1145,21 +1148,156 @@ void Editor::drawGrid(Graphics* g,
   grid_color =
     gfx::rgba(gfx::getr(grid_color), gfx::getg(grid_color), gfx::getb(grid_color), alpha);
 
-  // Draw horizontal lines
-  int x1 = spriteBounds.x;
-  int y1 = gridF.y;
-  int x2 = spriteBounds.x + spriteBounds.w;
-  int y2 = spriteBounds.y + spriteBounds.h;
+  // Orthogonal grid
+  if (isPixelGrid || getSite().sprite()->gridType() == doc::Grid::Type::Orthogonal) {
```
dacap, line 1152: We have to avoid calling `getSite()` as much as possible (the site collects a lot of information that might be unnecessary):
```suggestion
  if (isPixelGrid || m_sprite->gridType() == doc::Grid::Type::Orthogonal) {
```

src/app/tools/controllers.h
```diff
@@ -17,6 +17,94 @@ namespace app { namespace tools {
...
+      // when the order of the two points is inverted, so we try to
+      // detect this edge case and flip the points.
+      //
+      // TODO: this fix only works for two-point lines. Support
+      // for freehand strokes would require changes to intertwiners,
+      // not just the freehand controller itself.
+      if (lineTool && vto.x < 0 && a.x % (grid.w - !line.evenWidth)) {
+        auto tmp = a;
+        a = b;
+        b = tmp;
```
dacap, line 88: You can use `std::swap()` (needs `<utility>` probably)
```suggestion
        std::swap(a, b);
```

src/doc/grid.cpp
```diff
@@ -177,4 +179,60 @@ std::vector<gfx::Point> Grid::tilesInCanvasRegion(const gfx::Region& rgn) const
...
+  evenWidth = sz.w % 2 == 0;
+  evenHeight = sz.h % 2 == 0;
+  evenHalfWidth = ((sz.w / 2) % 2) == 0;
+  evenHalfHeight = ((sz.h / 2) % 2) == 0;
+  squareRatio = ABS(sz.w - sz.h) <= 1;
+  oddSize = !evenWidth && evenHalfWidth && !evenHeight && evenHalfHeight;
+
+  // TODO: add 'share edges' checkbox to UI.
+  // For testing the option, set this 'false' to 'true'
+  shareEdges = !(false && !squareRatio && evenHeight);
```
dacap, line 193: I'm thinking if the shareEdges could be an argument of this function, or another type of grid `IsometricSharedEdges` or something like that (and passing the grid type to `IsometricGuide` ctor).

dacap (changes requested):
I was finally able to give a try and test a little this implementation. So far so good 🚀 There are some final details to merge the PR, and after that we can fix any future issue/bug getting some user feedback.

Liebranca:
Hi, @dacap! I've separated the changes into two commits as the modification to `snap_to_grid()` meant I had to touch a few additional files, hopefully splitting it up makes it easier to amend/revert things later on if we need to.

A quick note about `GridSettingsCommand`/`OptionsWindow`, I added two helper functions to convert the string to a `Grid::Type` and viceversa, I'm not really sure which file it'd be best to put them in. For now I've placed these functions in `app/cmd/set_grid_type.h` as both classes already include this header. I noticed a few other minor things I initially missed regarding setting/getting the default type so I've addressed them as well.

I also added a few methods to `doc::Grid` to try and take advantage of the object now being used more directly, and also a new type of grid to act as toggle for the `shareEdges` option. I've updated the strings to convey the difference between the two types of isometric grid, however the new type (I've called it "thick edges" for now) doesn't work if the grid's height is uneven, and in such cases the two types are effectively the same, we may want to communicate this somehow (could there be a tooltip perhaps?).

Finally, I noticed a couple minor problems with snapping, the more important being in `MovingCelState` as the `calcPivot()` method didn't account for the grid being isometric (I think I was missing patch #4646 when I first wrote this), so I added a small fix for these as well.

dacap:
I've tried to use the GitHub web UI to resolve conflicts but it merged the beta branch into this PR. I'll try to rebase the PR to the beta branch.

dacap:
Rebased in https://github.com/aseprite/aseprite/pull/5793

PR #4828, opened 2024-11-26
Avoid endless loop when dropping unsupported files

martincapello:
When unsupported files where dropped on the timeline the DropOnTimeline command got stuck trying to process the unsupported file forever.

I've encountered this issue when trying reproduce [ASEPRITE-2QB](https://igara-studio.sentry.io/share/issue/2ab93fe53a054489853abae0758b4d41/)


src/app/cmd/drop_on_timeline.cpp
```diff
@@ -116,6 +116,12 @@ bool DropOnTimeline::getNextDocFromPaths(Doc** srcDoc)
   if (!fop)
     return false;
 
+  if (fop->hasError()) {
+    console.printf(fop->error().c_str());
+    m_paths.erase(m_paths.begin());
+    return true;
+  }
+
```
dacap, line 124: Probably we can add a comment like
```suggestion
  if (fop->hasError()) {
    console.printf(fop->error().c_str());
    
    // Remove the first item from the m_paths so we don't process
    // in the next getNextDocFromPaths() call.
    m_paths.erase(m_paths.begin());
    return true;
  }

```
Also it might be nice to always remove that element. I know `for (const auto& filename : fopFilenames) { ... }` will do that job, but just in case.

martincapello: > Also it might be nice to always remove that element. I know for (const auto& filename : fopFilenames) { ... } will do that job, but just in case.

I liked this, so I did it. Thanks!

PR #4835, opened 2024-11-28
Fix inactive window passes events to window below (fix #4561)

### LINKED ISSUE #4561: Inactive windows in multiple-window UI mode pass events to below windows
UI with multiple windows is enabled!

When attempting to drag the preview window from the red highlighted area, which becomes immovable when another windows is open, the windows it's intersecting with is moved instead, acting as if the user attempted to move the other window:
image
Clicking on the yellow part would click the button below instead

* Aseprite version: 1.3.7, built locally, no changes
* System: Windows 10

Gasparoken:
This is a first approach, it may be wrong and for sure I need some advice.
Ideally I would like to make it so that at the OS level another window cannot be brought to the front/activated if there is already a modal window open.
Currently, this fix allows a non-modal window to be selected and brought to the front but does not allow events to propagate to the modal window.

fix #4561


Gasparoken:
I will try to find an alternative solution at the os-laf level.

dacap:
I'm trying to review all pending PRs. Testing this I'm not sure if a "isOnTop()" should be treated as a modal window (e.g. the combobox popup is the only one calling "setOnTop()" and it shouldn't a modal window).

I'll merge a modified version of this patch calling
```c++
spec.modal(window->isForeground());
```
Anyway the `modal` parameter is only processed on macOS, so I think Windows will still have the original issue #4561 (it's still present on Linux even with this patch).

dacap:
Merged in d9785e5c363a555fcda30ad0a256abc59f5f6550

PR #4837, opened 2024-11-29
Add a helper to View to standardize scrolling behavior

ckaiser:
While working on the multiline entry box I noticed that we have the same duplicated code in most view-operated widgets, so I figured I'd abstract it into a static method similar to `View::getView` so if we ever want to make changes to scrolling behavior or defaults, we can do it from one place.


dacap:
LGTM! :+1:

PR #4841, opened 2024-11-30
Try to use TextBlobs to paint ui::Style's text layers

dacap:
Improves general performance to draw most widgets.


ckaiser:
Same idea I had in my multiline entry branch, you love to see it 😄
image

dacap:
> Same idea I had in my multiline entry branch, you love to see it 😄

:smile: sorry for the missing method, I'll push the `drawTextBlob()` change only to beta (as I've found an error in this PR with widgets using the mini font) so we both have it available.

PR #4850, opened 2024-12-03
Add app.clipboard Lua API (fix #2073)

### LINKED ISSUE #2073: Lua:  Clipboard access in the commands API
1.- It would be great having access to the clipboard in the API like:

app.command.PasteClipboardText

It would allow to create third party interaction inside Aseprite.

2.- And adding custom keyboard shortcuts for scripts would be also a great feature

Aseprite v1.2.11 and above.

ckaiser:
Implements a native clipboard interface API for Lua, fixes #2073

```lua
-- Empties the clipboard
app.clipboard.clear()

-- Content check
app.clipboard.hasText
app.clipboard.hasImage

-- Quick access
app.clipboard.text
app.clipboard.image

-- Full table
app.clipboard.content = {
 image,
 selection,
 palette,
 tileset,
 text
}
```

### DIFF
--- a/src/app/script/app_clipboard_object.cpp
+++ b/src/app/script/app_clipboard_object.cpp
@@ -30,7 +30,7 @@ struct Clipboard {};
   if (!ask_access(L, nullptr, mode, ResourceType::Clipboard))                                      \
     return luaL_error(L,                                                                           \
                       "the script doesn't have %s access to the clipboard",                        \
-                      mode == FileAccessMode::Read ? "read" : "write");
+                      (mode) == FileAccessMode::Read ? "read" : "write");
 
 int Clipboard_clear(lua_State* L)
 {
@@ -190,23 +190,35 @@ int Clipboard_set_content(lua_State* L)
     return luaL_error(L, "clipboard content must be a table");
 
   int type = lua_getfield(L, 2, "image");
-  if (type != LUA_TNIL)
+  if (type != LUA_TNIL) {
     image = may_get_image_from_arg(L, -1);
+    if (!image)
+      return luaL_error(L, "invalid image provided");
+  }
   lua_pop(L, 1);
 
   type = lua_getfield(L, 2, "mask");
-  if (type != LUA_TNIL)
-    mask = may_get_docobj<Mask>(L, -1);
+  if (type != LUA_TNIL) {
+    mask = get_mask_from_arg(L, -1);
+    if (!mask)
+      return luaL_error(L, "invalid mask provided");
+  }
   lua_pop(L, 1);
 
   type = lua_getfield(L, 2, "palette");
-  if (type != LUA_TNIL)
+  if (type != LUA_TNIL) {
     palette = may_get_docobj<Palette>(L, -1);
+    if (!palette)
+      return luaL_error(L, "invalid palette provided");
+  }
   lua_pop(L, 1);
 
   type = lua_getfield(L, 2, "tileset");
-  if (type != LUA_TNIL)
+  if (type != LUA_TNIL) {
     tileset = may_get_docobj<Tileset>(L, -1);
+    if (!tileset)
+      return luaL_error(L, "invalid tileset provided");
+  }
   lua_pop(L, 1);
 
   type = lua_getfield(L, 2, "text");
@@ -214,13 +226,14 @@ int Clipboard_set_content(lua_State* L)
     const char* tableText = lua_tostring(L, -1);
     if (tableText != nullptr && strlen(tableText) > 0)
       text = std::string(tableText);
+    else
+      return luaL_error(L, "invalid text provided");
   }
   lua_pop(L, 1);
 
   if (!image && !mask && !palette && !tileset && !text.has_value()) {
     // If we aren't provided anything valid in the table, clear the clipboard.
-    clip::clear();
-    return 0;
+    return Clipboard_clear(L);
   }
 
   if (image != nullptr && text.has_value())
--- a/src/app/script/app_clipboard_object.cpp
+++ b/src/app/script/app_clipboard_object.cpp
@@ -156,7 +156,7 @@ int Clipboard_get_content(lua_State* L)
   lua_setfield(L, -2, "mask");
 
   if (bitmapResult && palette)
-    push_docobj<Palette>(L, palette);
+    push_palette(L, palette);
   else
     lua_pushnil(L);
   lua_setfield(L, -2, "palette");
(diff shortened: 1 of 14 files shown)


dacap:
I'll take a look right now...

> do we want to prompt or at least notify the user that a script has read their clipboard? Having a permission for it maybe?

Yes, we should require a prompt/security flag for this specific permission (when image or text is read/get from the clipboard).

src/app/script/app_clipboard_object.cpp
```diff
@@ -0,0 +1,138 @@
...
+int Clipboard_is_image(lua_State* L)
+{
+  lua_pushboolean(L, clip::has(clip::image_format()));
+  return 1;
+}
+
+int Clipboard_is_empty(lua_State* L)
+{
+  // Using clip::has(clip::empty_format()) had inconsistent results, might as well avoid false positives
+  // by just checking the two formats we support
```
dacap, line 47: Reading your comment I'm thinking that probably we should remove this `isEmpty` function/property. This was a point of discussion in a issue from the clip project in the past: https://github.com/dacap/clip/issues/28

src/app/script/app_clipboard_object.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+  doc::Image* image = nullptr;
+  doc::Mask* mask = nullptr;
+  doc::Palette* palette = nullptr;
+  doc::Tileset* tileset = nullptr;
+  const bool result = app::Clipboard::instance()->getNativeBitmap(
+    &image, &mask, &palette, &tileset
+  );
+
+  // TODO: If we get a tileset, should we convert it to an image?
```
dacap, line 64: Good question, and probably this will make things a little harder. `set/getNativeBitmap()` will use a custom format to copy/paste images inside Aseprite/between Aseprite instances, and that custom format saves the original image format (e.g. indexed / tilemap) and the original palette/tileset.

This `app.clipboard.image` getter doesn't return this information. One option would be to return multiple values, in the same order, e.g.
```lua
local image, mask, palette, tileset = app.clipboard.image
```
But it's quite confusing.

src/app/script/app_clipboard_object.cpp
```diff
@@ -0,0 +1,138 @@
...
+{
+  auto* image = may_get_image_from_arg(L, 2);
+  if (!image)
+    return luaL_error(L, "invalid image");
+
+  const bool result = app::Clipboard::instance()->setNativeBitmap(
+    image,
+    nullptr,
+    get_current_palette(),  // TODO: Not sure if there's any way to get the palette from the image
+    nullptr,
```
dacap, line 92: This is the best we got at the moment. If we specify an isolated image.

Anyway it'd be probably better to use the active context site (the app::Site contains the active palette and tileset).

src/app/script/app_clipboard_object.cpp
```diff
@@ -0,0 +1,138 @@
...
+    image->maskColor()  // TODO: Unsure if this is sufficient.
```
dacap, line 93: Should be enough.

src/app/script/app_clipboard_object.cpp
```diff
@@ -0,0 +1,138 @@
...
+  return 0;
+}
+
+int Clipboard_set_text(lua_State* L)
+{
+  const char* text = lua_tostring(L, 2);
+  if (text != NULL && strlen(text) > 0 && !clip::set_text(text))
+    return luaL_error(L, "failed to set the clipboard text to '%s'", text);
+
+  return 0;
```
dacap, line 108: I'd return an error only if the `set_text()` function fails, and set an empty string in the other cases:
```suggestion
  const char* text = lua_tostring(L, 2);
  if (!clip::set_text(text ? std::string(text): std::string()))
    return luaL_error(L, "failed to set the clipboard text to '%s'", text);

  return 0;
```

ckaiser:
Renamed "mask" to "selection" since that's what it's known as in scripting, API now looks like:

```lua
-- Empties the clipboard
app.clipboard.clear()

-- Content check
app.clipboard.hasText
app.clipboard.hasImage

-- Quick access
app.clipboard.text
app.clipboard.image

-- Full table
app.clipboard.content = {
 image,
 selection,
 palette,
 tileset,
 text
}
```

I was wondering: Do we need hasText & hasImage when now app.clipboard.text && app.clipboard.image return `nil` instead of failing when they get called? I think they're kinda redundant.

We could also add a helper to copy the current sprite to the clipboard with all the data we can have (custom format, etc) since the way to do it right now is slightly cumbersome, like:
```lua
app.clipboard.content = {
    image = app.image,
    palettte = sprite.palettes[1],
    mask = sprite.selection,
    selection = sprite.selection,
    tileset = app.layer.tileset
  }
```

could become:
```lua
app.clipboard.copyActive() 
``` 
or something like that, maybe? Just an idea.

src/app/script/security.cpp
```diff
@@ -93,7 +94,7 @@ std::string get_script_filename(lua_State* L)
   const char* source = lua_tostring(L, -1);
   std::string script;
   if (source && *source)
-    script = source+1;
+    script = source;
```
dacap, line 97: About your question:

> It was also saying "nternal" because the text was getting +1'd for some reason? Unclear why that was there and the git blame was not particularly illuminating since it's always been like this.

The `source` field contains a `"@"` at the beginning for files, e.g.

Create a `test.lua` in your `scripts` folder:
```lua
print(debug.getinfo(1, "S").source)
```

And it will print:
```
@/home/david/.config/aseprite/scripts/test.lua
```

We need this +1 to get the proper filename.

dacap (changes requested):
I wasn't able to fully test the API yet, but I've rebased/reformatted this PR into: https://github.com/dacap/aseprite/tree/feature/clipboard-api-rebased with some comments.

> I was wondering: Do we need hasText & hasImage when now app.clipboard.text && app.clipboard.image return nil instead of failing when they get called? I think they're kinda redundant.

I'd prefer to keep these functions just in case an user needs to check the content type in some place (e.g. a isEnabled kind of function), and get the content in other place (e.g. executeCommand kind of function), as hasText/hasImage are faster than getting the whole content.

> We could also add a helper to copy the current sprite to the clipboard with all the data we can have (custom format, etc) since the way to do it right now is slightly cumbersome...

Probably `app.command.Copy()` already does that if select the whole canvas? E.g.
```lua
app.command.MaskAll()
app.command.Copy()
```

ckaiser:
Hm, yeah, reading permission makes sense but write permission could be annoying - personally I like when software explicit tells me it's gonna mess with my clipboard (I have my phone set to tell me whenever my clipboard changes) so for me it seemed reasonable but for a regular user it might be unnecessary and unwelcome. It's up to what you think is the most consistent with how the rest of the API behaves wrt system access.

Issue #3718, opened 2023-03-02
Saved Undo History


ElderlyGlek:
I would like the undo history to be stored in .aseprite files, for example, this may be needed for creating timelapses. This idea turned out to be too short, I have nothing more to add.

dacap:
Hi @ElderlyGlek, actually this feature was requested several times in the past:

* https://github.com/aseprite/aseprite/issues/738
* https://github.com/aseprite/aseprite/issues/1324 (as a [separated .undo file](https://github.com/aseprite/aseprite/issues/1324#issuecomment-260881123) too)
* https://github.com/aseprite/aseprite/issues/183#issuecomment-236030642
* https://community.aseprite.org/t/save-undo-history-of-project/4642
* https://community.aseprite.org/t/undo-history-save-restore/14247

ElderlyGlek:
> Hi @ElderlyGlek, actually this feature was requested several times in the past:
> 
>     * [History of all things - Buffing up the Undo feature #738](https://github.com/aseprite/aseprite/issues/738)
> 
>     * [Save/restore the whole session (+undo information) automatically #1324](https://github.com/aseprite/aseprite/issues/1324) (as a [separated .undo file](https://github.com/aseprite/aseprite/issues/1324#issuecomment-260881123) too)
> 
>     * [Redesign .aseprite file format (it might be a .zip file) #183 (comment)](https://github.com/aseprite/aseprite/issues/183#issuecomment-236030642)
> 
>     * https://community.aseprite.org/t/save-undo-history-of-project/4642
> 
>     * https://community.aseprite.org/t/undo-history-save-restore/14247

And… do you have plans for this?

dacap:
Reopening this issue just to be the tracking issue of the "save undo history" in some way. There are plans to save the undo history in a separate file first, and then inside .aseprite files if possible.

PR #4859, opened 2024-12-03
Revamp and fix DropOnTimeline command

### LINKED ISSUE #4858: Crash when undoing the drop of a layer after being deleted
Sentry Issue: [ASEPRITE-2QT](https://igara-studio.sentry.io/issues/6094992079/?referrer=github_integration)

```
EXCEPTION_ACCESS_VIOLATION_READ / 0xb6: Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0xb6
  File "xutility", line 5987, in std::_Find_unchecked
    const auto _Result    = _STD _Find_vectorized(_First_ptr, _STD _To_address(_Last), _Val);
  File "xutility", line 6024, in std::find
    _STD _Seek_wrapped(_First, _STD _Find_unchecked(_STD _Get_unwrapped(_First), _STD _Get_unwrapped(_Last), _Val));
  File "layer.cpp", line 57, in doc::Layer::getPrevious
    std::find(m_parent->layers().begin(),
  File "drop_on_timeline.cpp", line 237, in app::cmd::DropOnTimeline::onUndo
    layerBefore = layer->getPrevious();
  File "cmd.cpp", line 52, in app::Cmd::undo
    onUndo();
...
(20 additional frame(s) were not displayed)
```

Been able to reproduce this as follows:
1. ✨ Create a new file
2. ✨ Drop a png or jpg file at the bottom of the "Layer 1".
3. ✨ Delete the layer that was just created because of the drop.
4. ✨ Undo the deletion.
5. 💥 Try to undo the drop. You can't, it crashes.

martincapello:
This PR essentially fixes #4858, but it does so by revamping the `DropOnTimeline` command after I realized that it is actually not practical to store the dropped layers in a vector of pointers to layers because of how the `ObjectId`s impact on the logic needed to keep them in sync. So I used the same approach as in other commands: use a `stringstream` object to store the data (dropped layers) that needs the redo operation.

As a side effect, I believe that there is a high chance that the following crashes are fixed as well:
[ASEPRITE-2QF](https://igara-studio.sentry.io/issues/6085369687/events/ff0e50ded22d4bcbc4330aa777f531b9/)
[ASEPRITE-2QB](https://igara-studio.sentry.io/issues/6084029246/events/c3edf8fd1b1449a010abf680ebd561c5/)

### DIFF
--- a/src/app/cmd/drop_on_timeline.cpp
+++ b/src/app/cmd/drop_on_timeline.cpp
@@ -21,10 +21,15 @@
 #include "app/util/layer_utils.h"
 #include "app/util/open_file_job.h"
 #include "app/tx.h"
+#include "base/serialization.h"
 #include "doc/layer_list.h"
+#include "doc/layer_io.h"
+#include "doc/subobjects_io.h"
 #include "render/dithering.h"
 
 #include <algorithm>
+
+using namespace base::serialization::little_endian;
 
 namespace app {
 namespace cmd {
@@ -45,6 +50,10 @@ DropOnTimeline::DropOnTimeline(app::Doc* doc,
   ASSERT(m_layerIndex >= 0);
   for(const auto& path : m_paths)
     m_size += path.size();
+
+  // Zero layers stored.
+  write32(m_stream, 0);
+  m_size += sizeof(uint32_t);
 }
 
 DropOnTimeline::DropOnTimeline(app::Doc* doc,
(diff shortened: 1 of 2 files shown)


dacap:
A couple of points before I review this:
1) Probably we could use some smart pointers for the Doc* (check #4850)
2) Although we are using `std::stringstream` to serialize `doc::Object`s, in a future I'd like to avoid doing so, and serialize objects only when it's necessary (e.g. to save undo info to disk/.aseprite format, https://github.com/aseprite/aseprite/issues/3718#issuecomment-1453512657 ), but I have that on mind from a long time but never created the issue (it's a refactor for the future), I'll create a new issue about that now

dacap:
> So, how do you think we should deal with this situation?

At the moment probably the best is to keep the layers serialized, and then investigate #4860 to avoid the serialization on all `Cmd`s (anyway we'll need some way to keep the serialization available to save it in the undo history inside .aseprite files in the future).

dacap:
Rebased, clang-formatted and merged in: 62d9318196e0610bb65568ca2720694378424f18 + da76c8a0fb073cb44afed6426a14f3ecc76db34f + 2f2f522108aee0fabaaff55dd72530ea0884c8b4.

PR #4873, opened 2024-12-10
Show export location

### LINKED ISSUE #4483: Show / make explicit export location
Some recurrent UI inconvenience when we export (File > Export > Export As / Export Sprite Sheet) is to show/select the location where the export will be done.

Some ideas to show the path or {elements}:

image

Liebranca:
Repurposed the already existing `FilenameField` class for use with `ExportFileWindow` and `ExportSpriteSheetWindow`. Fix #4483.

Changes to`FilenameField`:

- Moved most of the path-related logic in `ExportFileWindow` to `FilenameField`.
- Pressing the button now shows a menu with options for choosing a file, and whether to display it as absolute or relative.
- Added the class itself to the switches on `gen/ui_class.cpp` and `app/widget_loader.cpp` so that it can be used in xml files as `<filename />`. The default type is `ButtonAndEntry`, overridden by the xml attribute `button_only`.
- Added a member variable for use with the 'ask overwrite' popup, set to true by default, and then set to false when the file selector is invoked.

Notes:
- There's a small problem with 'filename' in some places refering to the *name* of the file without path (eg: `base::get_file_name()`), while in other places it refers to the full absolute path (`app::Doc::filename`). It's a very minor thing, but leaving it like this could be confusing.
- The new menu shows when `type == ButtonOnly`. I'm not sure whether to disable this or not.
- `FilenameField` could perhaps handle other bits of logic related to file extension.
- `ExportSpriteSheetWindow` used to invoke the 'ask overwrite' popup for both files regardless of whether one of them had already been confirmed for overwrite. Now it will only ask whether to overwrite a file if the corresponding boolean is set to true.

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -161,6 +161,7 @@
       <option id="keep_closed_sprite_on_memory" type="bool" default="true" />
       <option id="keep_closed_sprite_on_memory_for" type="double" default="15.0" />
       <option id="show_full_path" type="bool" default="true" />
+      <option id="edit_full_path" type="bool" default="false" />
       <option id="timeline_position" type="TimelinePosition" default="TimelinePosition::BOTTOM" />
       <option id="timeline_layer_panel_width" type="int" default="100" />
       <option id="show_menu_bar" type="bool" default="true" />
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -665,9 +665,6 @@ sensors_tweaks = Sensor Threshold
 
 [export_file]
 title = Export File
-choose_file = Choose File...
-relative_path = Relative Path to "{}"
-absolute_path = Absolute Path
 output_file = Output File:
 resize = Resize:
 area = Area:
@@ -1698,6 +1695,9 @@ cancel = Cancel
 
 [select_file]
 text = Select File
+choose = Choose File...
+relative_path = Relative Path to "{}"
+absolute_path = Absolute Path
 browse = ...
 
 [selection_mode]
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -665,6 +665,9 @@ sensors_tweaks = Sensor Threshold
 
 [export_file]
 title = Export File
+choose_file = Choose File...
+relative_path = Relative Path to "{}"
+absolute_path = Absolute Path
 output_file = Output File:
 resize = Resize:
 area = Area:
--- a/data/widgets/export_file.xml
+++ b/data/widgets/export_file.xml
@@ -5,9 +5,7 @@
   <window id="export_file" text="@.title" help="exporting">
     <grid columns="3">
       <label text="@.output_file" />
-      <entry id="output_filename" cell_align="horizontal" maxsize="1024" maxwidth="256" />
-      <button id="output_filename_browse" text="..." style="mini_button" />
-
+      <filename id="output_field" cell_align="horizontal" cell_hspan="2" />
       <label id="resize_label" text="@.resize" />
       <hbox cell_hspan="2" cell_align="horizontal">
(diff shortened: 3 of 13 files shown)


dacap:
I've changed the target branch to beta, and rebased these commits: https://github.com/dacap/aseprite/tree/show-output-folder (already with clang-format).

I'll try to test a little more. @Liebranca if you update your branch to the rebased version, this PR will be updated automatically (and probably the CI checks will pass)

Liebranca:
Hi @dacap! I've added a few things to `Dialog:file{}` to try and better match how `FilenameField` behaves on both of the 'Export As' windows, please let me know if you have any feedback.

dacap:
@Liebranca great! I'll be reviewing this PR today and tomorrow. I'm trying to rebase the whole PR because it has a conflict with ebc4624d59f0486f9502bfe7b17966473e609908, trying to check where is `base::get_absolute_path()` is necessary :thinking:

dacap:
I [rebased this branch](https://github.com/dacap/aseprite/tree/show-output-folder) again to the latest beta commit to fix the conflict. I was able to test this on Windows too (and [found a bug](https://github.com/aseprite/laf/commit/07f4e9577d67aed16bd04efd545a18aa26e2037e) in `get_absolute_path()`).

It looks great! I really like how it's working this now. There are a couple of changes I'd made (probably you can `rebase --interactive` to edit the commits):

1. About https://github.com/aseprite/aseprite/pull/4873/commits/8d2f0380d97b03e58a368dd8fecb41a793879cba, we can add the `basepath` argument but without these specials `{doc}` or `{cwd}` modifiers, as the user already can access these fields from [app.sprite.filename](https://www.aseprite.org/api/sprite#spritefilename) and [app.fs.currentPath](https://www.aseprite.org/api/app_fs#appfscurrentpath), and the "doc" term is not used in scripting (only "sprite"), but it's better if we keep this parameter simple, and in this way we avoid adding a new kind of filename formatting (which means more documentation).
2. Remove `FilenameFieldObserver` class, and do the `Preferences::instance().general.editFullPath.AfterChange.connect()` inside `FilenameField::FilenameField(...)` constructor. Remember to create a `obs::scoped_connection` member in `FilenameField` so this observer is disconnected automatically in its destructor. E.g. You can find some examples in the `Editor`: [m_fgColorChangeConn](https://github.com/aseprite/aseprite/blob/5e6927ed4e362c42ce8a3aecb62e6f64a3c8a5a5/src/app/ui/editor/editor.h#L442), and [connect() call](https://github.com/aseprite/aseprite/blob/5e6927ed4e362c42ce8a3aecb62e6f64a3c8a5a5/src/app/ui/editor/editor.cpp#L178-L179).
3. In `FilenameField::onBrowse()` you can select the `relative` or `absolute` item depending on the current `Preferences::instance().general.editFullPath()` value to display a checkbox in the current option, e.g.:
   ```diff
   diff --git a/src/app/ui/filename_field.cpp b/src/app/ui/filename_field.cpp
   index 5fd0e7823..8bc248371 100644
   --- a/src/app/ui/filename_field.cpp
   +++ b/src/app/ui/filename_field.cpp
   @@ -124,6 +124,10 @@ void FilenameField::onBrowse()
      ui::MenuItem choose(Strings::select_file_choose());
      ui::MenuItem relative(Strings::select_file_relative_path(m_pathBase));
      ui::MenuItem absolute(Strings::select_file_absolute_path());
   +
   +  relative.setSelected(!m_editFullPath);
   +  absolute.setSelected(m_editFullPath);
   +
      menu.addChild(&choose);
      menu.addChild(new ui::MenuSeparator());
      menu.addChild(&relative); 
   ```
   image
   In this way the user can expect a "do nothing" action if the current checked option is selected again from the menu.

PR #4894, opened 2024-12-23
Regression: Edit > Keyboard Shortcuts > Action Modifiers is empty (fix #4893)

### LINKED ISSUE #4893: Regression: Edit > Keyboard Shortcuts > Action Modifiers is empty
It looks a regression in v1.3.9, the list is empty (wasn't the case in v1.3.8):

image

Gasparoken:
It fixes the regression inserted in https://github.com/aseprite/aseprite/issues/4387.
Before this fix 'Action' keys were not taken into account in the "Key::isListed()" function.
It was changed the name of 'Key::isListed()' x 'Key::skipListing()' to avoid confussion.
It was also changed the name of 'Command::isListed()'  x  'Command::isOnList()'.

fix #4893

### DIFF
--- a/src/app/commands/cmd_keyboard_shortcuts.cpp
+++ b/src/app/commands/cmd_keyboard_shortcuts.cpp
@@ -582,10 +582,11 @@ private:
     fillWheelActionsList();
     fillDragActionsList();
 
+    // Fill the 'Commands' and 'Action Modifier' lists
     for (const KeyPtr& key : m_keys) {
       if (key->type() == KeyType::Tool || key->type() == KeyType::Quicktool ||
           key->type() == KeyType::WheelAction || key->type() == KeyType::DragAction ||
-          key->isListed()) {
+          !key->isListed()) {
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


dacap:
Is it necessary to do the rename from isListed to isOnList? Also I'd avoid the skipListing and try to follow the positive logic with isListed. If we can do the fix in the most smallest patch, the better.

Gasparoken:
I renamed the functions mainly because I found a confussion about the name: I was using 'isListed' as two opposing functions depending on whether I used it in 'Key' or 'Command'.
I interpreted 'Key::isListed' as "Is the key already listed?", meaning "it didn't need to be listed", i.e. it was a "skip list". 
In contrast, 'Command::isListed' meant "the command must be included in the list".
So this confussion pushed my to rename the functions.

Now the 'Key::isListed' is used as "the key must be included in the list".

I was going to rename it to 'hasToBeListed' or something similar, but this works more for my own tastes. So I'm going to stick to PR "as simple as possible". Feel free to rename the functions.

dacap:
Merged in ff0dfb83314f62cefaba3da09ae89afce946826e changing the name of `Key::isListed()` to `Key::isCommandListed()` so the function makes sense only for `KeyType::Command`. Simplified the condition to filter out items in the `fillAllLists()` for-loop.

PR #4898, opened 2024-12-26
Move font_paths files from aseprite to laf

martincapello:
Moves the font_path* files from Aseprite source into LAF's "os" namespace. To know more about why I did this, see the description of https://github.com/aseprite/laf/pull/123

Also, there is a comment in the CMakeLists.txt that says:
```
# These specific-platform files should be in an external library
# (e.g. "base" or "os").
```

WARNING: This PR needs https://github.com/aseprite/laf/pull/123 merged first to make it work properly.


martincapello:
Closed because this is going to be deprecated.

PR #4904, opened 2024-12-29
Add .version and .displayName to Lua Plugin table (fix #4730)

### LINKED ISSUE #4730: Add a way to access package.json information from Plugin properties/functions

It would be great to have access to all information from the package.json file - especially the plugin version for version management purposes - as Plugin properties or getter functions.

* Aseprite version: v1.3.9.2
* System: Windows

ckaiser:
Fixes #4730 "partially": provides access only to .version and .displayName. I'm not sure if there's much value in providing access to the rest of the properties, maybe giving a way to load it using `json.decode()` and just be returned as an arbitrary table? It'd be a bunch of duplicated information so I wouldn't do it unless there's demand for it, version number and display name are easy wins.

It might also be worth it to expose the version as a LAF `Version` object instead of just a raw string so script authors can use the comparison functions, but we'd have to clarify/standardize the "version" field so it uses something that that parser can understand, is that worth doing do you think @dacap? Could be something for a follow-up PR, even just offering a `Version.is_newer(string, string)` function might be simpler than a whole new object type.

### DIFF
--- a/src/app/extensions.cpp
+++ b/src/app/extensions.cpp
@@ -1058,7 +1058,7 @@ Extension* Extensions::installCompressedExtension(const std::string& zipFn,
   {
     json11::Json::object obj;
     obj["installedFiles"] = json11::Json(installedFiles);
-    json11::Json json(obj);
+    const json11::Json json(obj);
 
     const std::string fullFn = base::join_path(info.dstPath, kInfoJson);
     LOG("EXT: Saving list of installed files in <%s>\n", fullFn.c_str());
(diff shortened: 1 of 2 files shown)


dacap:
> It might also be worth it to expose the version as a LAF `Version` object instead [...]

Actually the Version object is already available. Check the [version_class.cpp](https://github.com/aseprite/aseprite/blob/main/src/app/script/version_class.cpp) file and [app.version](https://github.com/aseprite/aseprite/blob/ebc4624d59f0486f9502bfe7b17966473e609908/src/app/script/app_object.cpp#L687) property.

Probably the `plugin.version` can return a [Version object](https://aseprite.org/api/version).

ckaiser:
Oh, I can't believe I missed that 🫠, ok rebased and switched `.version` to return a Version object

PR #4925, opened 2025-01-09
Add fitCriteria parameter to ChangePixelFormat command

### LINKED ISSUE #4781: Add bestFitCriteria param to ChangePixelFormat command
Requested here: https://bsky.app/profile/joshuaskelly.bsky.social/post/3las4bof6zk2g

martincapello:
Fix #4781 

This PR also includes:
- A refactor of ChangePixelFormatCommand to make it inherit from CommandWithNewParams.
- A new "dithering-factor" parameter to being able to specify this for the "error-diffusion" dithering algorithm.
- A commit to avoid the BayerMatrix object slicing, not actually needed.

### DIFF
--- a/src/app/commands/cmd_change_pixel_format.cpp
+++ b/src/app/commands/cmd_change_pixel_format.cpp
@@ -30,6 +30,7 @@
 #include "app/ui/editor/editor_render.h"
 #include "app/ui/rgbmap_algorithm_selector.h"
 #include "app/ui/skin/skin_theme.h"
+#include "doc/color_mode.h"
 #include "doc/image.h"
 #include "doc/layer.h"
 #include "doc/sprite.h"
@@ -70,7 +71,7 @@ public:
   ConvertThread(const doc::ImageRef& dstImage,
                 const doc::Sprite* sprite,
                 const doc::frame_t frame,
-                const doc::PixelFormat pixelFormat,
+                const doc::ColorMode colorMode,
                 const render::Dithering& dithering,
                 const gen::ToGrayAlgorithm toGray,
                 const gfx::Point& pos,
@@ -83,11 +84,11 @@ public:
     , m_thread([this,
                 sprite,
                 frame,
-                pixelFormat,
+                colorMode,
                 dithering,
                 toGray,
                 newBlend]() { // Copy the matrix
-      run(sprite, frame, pixelFormat, dithering, toGray, newBlend);
+      run(sprite, frame, colorMode, dithering, toGray, newBlend);
     })
   {
   }
@@ -105,7 +106,7 @@ public:
 private:
   void run(const Sprite* sprite,
            const doc::frame_t frame,
-           const doc::PixelFormat pixelFormat,
+           const doc::ColorMode colorMode,
            const render::Dithering& dithering,
            const gen::ToGrayAlgorithm toGray,
            const bool newBlend)
@@ -121,7 +122,7 @@ private:
 
     render::convert_pixel_format(tmp.get(),
                                  m_image.get(),
-                                 pixelFormat,
+                                 (PixelFormat)colorMode,
                                  dithering,
                                  sprite->rgbMap(frame),
                                  sprite->palette(frame),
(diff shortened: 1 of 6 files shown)


src/app/commands/cmd_change_pixel_format.cpp
```diff
@@ -460,163 +461,82 @@ class ColorModeWindow : public app::gen::ColorMode {
 
 } // anonymous namespace
 
-class ChangePixelFormatCommand : public Command {
+struct ChangePixelFormatParams : public NewParams {
+  Param<bool> ui{ this, false, "ui" };
+  Param<PixelFormat> format{ this, IMAGE_RGB, "format" };
```
dacap, line 466: Change this to `ColorMode`. `PixelFormat` is an deprecated enum to represent the color mode, maybe in the future we could just remove it, or use it as a real "pixel format" (e.g. to use more formats, like R10G10B10A2).

We don't want to publish the `PixelFormat` enum in scripting API, we already have `ColorMode`. 

```c++
  Param<ColorMode> colorMode{ this, ColorMode::RGB, { "colorMode", "format" } };
```

We don't support command aliases yet, only params aliases (like this `"colorMode"/"format"`), but in a future we should change `ChangePixelFormat` to `ChangeColorMode` or something like that.

src/app/commands/cmd_change_pixel_format.cpp
```diff
@@ -673,32 +595,32 @@ void ChangePixelFormatCommand::onExecute(Context* context)
...
-                                 m_dithering,
-                                 m_rgbmap,
-                                 get_gray_func(m_toGray),
+                                 params().format(),
+                                 render::Dithering(params().dithering(), matrix, params().factor()),
+                                 params().rgbmap(),
+                                 get_gray_func(params().toGray()),
                                  &job,
-                                 m_fitCriteria)); // SpriteJob is a render::TaskDelegate
+                                 params().fitCriteria())); // SpriteJob is a render::TaskDelegate
```
dacap, line 605: Not introduced in this PR, but introduced here https://github.com/aseprite/aseprite/commit/3cc1c63274dac90360b8194f66101f86f1a670b6#diff-6a0da30ae1d087883f14800a3979cf9ffbab2e035cc9cc23b1fede5d83278046L682, the `SpriteJob` comment is for the `&job` line. We can fix this in this commit.

src/app/commands/cmd_change_pixel_format.cpp
```diff
@@ -460,163 +461,82 @@ class ColorModeWindow : public app::gen::ColorMode {
...
 
-class ChangePixelFormatCommand : public Command {
+struct ChangePixelFormatParams : public NewParams {
+  Param<bool> ui{ this, false, "ui" };
+  Param<PixelFormat> format{ this, IMAGE_RGB, "format" };
+  Param<render::DitheringAlgorithm> dithering{ this,
+                                               render::DitheringAlgorithm::None,
+                                               "dithering" };
+  Param<std::string> matrix{ this, "", "dithering-matrix" };
+  Param<double> factor{ this, 1.0, "dithering-factor" };
```
dacap, line 471: Add param aliases (these are easier to use in Lua):
```suggestion
  Param<std::string> matrix{ this, "", { "ditheringMatrix", "dithering-matrix" } };
  Param<double> factor{ this, 1.0, { "ditheringFactor", "dithering-factor" } };
```

src/app/commands/new_params.cpp
```diff
@@ -252,6 +258,60 @@ void Param<gen::SelectionMode>::fromString(const std::string& value)
...
+{
+  if (base::utf8_icmp(value, "rgb") == 0)
+    setValue(doc::PixelFormat::IMAGE_RGB);
+  else if (base::utf8_icmp(value, "grayscale") == 0 || base::utf8_icmp(value, "gray") == 0)
+    setValue(doc::PixelFormat::IMAGE_GRAYSCALE);
+  else if (base::utf8_icmp(value, "indexed") == 0)
+    setValue(doc::PixelFormat::IMAGE_INDEXED);
+  else
+    setValue(doc::PixelFormat::IMAGE_RGB);
+}
```
dacap, line 287: Remove this, use `ColorMode` instead.

src/app/commands/new_params.cpp
```diff
@@ -425,6 +485,42 @@ void Param<gen::SelectionMode>::fromLua(lua_State* L, int index)
...
+
+template<>
+void Param<doc::PixelFormat>::fromLua(lua_State* L, int index)
+{
+  if (lua_type(L, index) == LUA_TSTRING)
+    fromString(lua_tostring(L, index));
+  else
+    setValue((doc::PixelFormat)lua_tointeger(L, index));
+}
```
dacap, line 504: Remove this, use `ColorMode` instead.

dacap (changes requested):

dacap:
Merged in 5e40e5544 + 91b360319 + b1d739608

Patch 5e40e5544 includes the following 3 fixups squashed. About the object slicing issue, although only an issue when derived classes have additional members, it's better just to completely avoid it. I gonna make DitheringMatrix a final class.

PR #4926, opened 2025-01-10
Fixed zlib include dir when Aseprite is used as sub-project

bjorn:
For the Qt "[qaseprite](https://github.com/mapeditor/qaseprite)" image plugin, we're using Aseprite as git submodule so that we can link to its laf, dio and render libraries.

In this case, `CMAKE_BINARY_DIR` refers to the build directory of qaseprite. We need to use `CMAKE_CURRENT_BINARY_DIR` instead, which always refers to the right location for the generated `zconf.h`.

See also https://github.com/mapeditor/qaseprite/pull/6.

---

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -234,7 +234,7 @@ else()
   set(ZLIB_LIBRARIES ${ZLIB_LIBRARY})
   set(ZLIB_INCLUDE_DIRS
     ${ZLIB_DIR}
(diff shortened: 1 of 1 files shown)


dacap:
Hi @bjorn, thanks for these PRs. Please check your inbox, there should be a mail to sign from mail@signnow.com.

bjorn:
@dacap Alright, I've signed the agreement.

PR #4934, opened 2025-01-11
Handle screen DPI changes on Windows

lampysprites:
Corresponds to changes explained in aseprite/laf#127

---

### DIFF
--- a/src/app/modules/gui.cpp
+++ b/src/app/modules/gui.cpp
@@ -157,7 +157,7 @@ static bool create_main_window(bool gpuAccel, bool& maximized, std::string& last
     // Change the scale value only in the first run (this will be
     // saved when the program is closed).
     if (scale == 0)
-      Preferences::instance().general.screenScale(main_window->scale());
+      Preferences::instance().general.screenScale(main_window->baseScale());
 
     main_window->setGpuAcceleration(gpuAccel);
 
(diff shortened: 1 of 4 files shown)


martincapello (commented):
Nice work! However, this PR is not pointing to the latest laf changes. I think you made a force push in your branch in laf and then forgot to update the submodule here. I realized this after testing your laf changes in the branch pointed by this PR 😬, which has an issue with the scaling...I will test again with the latest changes to see how it works.

martincapello (commented):
I've been testing these changes and they work very well! However, I found an issue when changing the scale of a display (from Windows configuration) while a floating window (like the preview window) is being shown. The floating windows are not receiving the WM_DPICHANGED message. But then, if I move or scale the floating window then it accommodates. 
~I think that there are a big chances that it is a Windows bug that I'm not sure we will be able to workaround (maybe trying to update floating windows when a WM_DPICHANGED is received in the main window?)~

I've found the following stack overflow post about the described issue: 
https://stackoverflow.com/questions/77270061/window-wont-receive-wm-dpichanged-message-when-set-window-style-to-ws-ex-toolwi

EDIT: I've crossed out that statement because I'm not sure if it is a Windows bug anymore...I've tried with some Gimp windows with the WS_EX_TOOLWINDOW and they worked.

EDIT2: This is a laf issue actually, I've tested this with the hello_laf example. More details in https://github.com/aseprite/laf/pull/127

martincapello (approved):
Despite the laf issue mentioned (which is not our fault actually), I think this PR looks good.

EDIT: Sorry, forgot that the laf submodule must be updated before merging this.

dacap:
Thanks @lampysprites! I'll cherry-pick this PR squashed (and with prek hook).

I'll take note about @martincapello bug in another issue, probably we can run the DPI adjustment in floating windows manually when the main window receives a DPI change.

dacap:
Merged in https://github.com/aseprite/aseprite/commit/bbe70d852711609d8d9383aa122a28245503be19

PR #4935, opened 2025-01-12
Fix invalid tileset cache after color remap

tedle:
Changing a color palette and then remapping it does not invalidate a tileset's cache. Immediately saving to the .aseprite format after this will use the stale cache data instead of the current sprite data, resulting in the changes not being saved.

This is because `RemapColors::incrementVersions` only increments an image's version
https://github.com/aseprite/aseprite/blob/9779470d0eb02ec937b2203e4de1e1cb06414fa2/src/app/cmd/remap_colors.cpp#L48
And `ase_file_write_tileset_chunk` only checks a tileset's version
https://github.com/aseprite/aseprite/blob/9779470d0eb02ec937b2203e4de1e1cb06414fa2/src/app/file/ase_format.cpp#L1492

This clip shows a tilemap layer and a normal layer becoming desynced from each other after re-opening a file because the tilemap data was not saved:

https://github.com/user-attachments/assets/e81266cb-220e-4b71-9142-15f9d02d213c

### DIFF
--- a/src/app/cmd/remap_colors.cpp
+++ b/src/app/cmd/remap_colors.cpp
@@ -15,6 +15,7 @@
 #include "doc/image.h"
 #include "doc/remap.h"
 #include "doc/sprite.h"
+#include "doc/tilesets.h"
 
 namespace app { namespace cmd {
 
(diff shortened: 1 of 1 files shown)


dacap:
Thanks a lot @tedle for this patch, I was able to tested and it works great! :+1:

PR #4943, opened 2025-01-17
Add UUIDs to layers

### LINKED ISSUE #3683: Create GUIDs for layers to keep track of renamed layers on external importers
Sometimes an external tool will need to import layers and keep track of the same layer (even if the layer is renamed). Probably we could save a GUID on memory and on disk for each layer (this GUID wouldn't be editable).

martincapello:
Fix #3683

This PR adds a new checkbox to the "Sprite properties" window to enable/disable to usage of UUIDs in layers. Also adds a field in the "Layer properties" window that shows the current UUID, only if the aforementioned checkbox is checked.

### DIFF
--- a/docs/ase-file-specs.md
+++ b/docs/ase-file-specs.md
@@ -79,7 +79,7 @@ A 128-byte header (same as FLC/FLI header, but with other magic number):
                   1 = Layer opacity has valid value
                   2 = Layer blend mode/opacity is valid for groups
                       (composite groups separately first when rendering)
-                  4 = Layers are using UUIDs
+                  4 = Layers have an UUID
     WORD        Speed (milliseconds between frame, like in FLC files)
                 DEPRECATED: You should use the frame duration field
                 from each frame header
@@ -203,8 +203,8 @@ entire layers layout:
     STRING      Layer name
     + If layer type = 2
       DWORD     Tileset index
-    UUID        Layer's universally unique identifier (present only if the
-                [main header](#header) "Flags" field has the bit 3 enabled)
+    + If file header flags have bit 4:
+    UUID        Layer's universally unique identifier
 
 ### Cel Chunk (0x2005)
 
--- a/docs/ase-file-specs.md
+++ b/docs/ase-file-specs.md
@@ -79,6 +79,7 @@ A 128-byte header (same as FLC/FLI header, but with other magic number):
                   1 = Layer opacity has valid value
                   2 = Layer blend mode/opacity is valid for groups
                       (composite groups separately first when rendering)
+                  4 = Layers are using UUIDs
     WORD        Speed (milliseconds between frame, like in FLC files)
                 DEPRECATED: You should use the frame duration field
                 from each frame header
@@ -202,6 +203,8 @@ entire layers layout:
     STRING      Layer name
     + If layer type = 2
       DWORD     Tileset index
+    UUID        Layer's universally unique identifier (present only if the
+                [main header](#header) "Flags" field has the bit 3 enabled)
 
 ### Cel Chunk (0x2005)
 
--- a/src/app/commands/cmd_layer_properties.cpp
+++ b/src/app/commands/cmd_layer_properties.cpp
@@ -464,8 +464,7 @@ private:
       m_userDataView.setVisible(false, false);
     }
 
-    bool uuidVisible = m_document && m_document->sprite() &&
-                       m_document->sprite()->useUuidsForLayers();
+    bool uuidVisible = m_document && m_document->sprite() && m_document->sprite()->uuidsForLayers();
     uuidLabel()->setVisible(uuidVisible);
     uuid()->setVisible(uuidVisible);
 
--- a/src/app/commands/cmd_layer_properties.cpp
+++ b/src/app/commands/cmd_layer_properties.cpp
@@ -32,6 +32,7 @@
 #include "app/ui/timeline/timeline.h"
 #include "app/ui/user_data_view.h"
 #include "app/ui_context.h"
+#include "base/convert_to.h"
 #include "base/scoped_value.h"
 #include "doc/image.h"
 #include "doc/layer.h"
@@ -462,6 +463,14 @@ private:
       opacity()->setEnabled(false);
       m_userDataView.setVisible(false, false);
     }
+
+    bool uuidVisible = m_document && m_document->sprite() &&
+                       m_document->sprite()->useUuidsForLayers();
+    uuidLabel()->setVisible(uuidVisible);
+    uuid()->setVisible(uuidVisible);
+
+    if (uuidVisible)
+      uuid()->setText(m_layer ? base::convert_to<std::string>(m_layer->uuid()) : "");
 
     if (tileset()->isVisible() != tilemapVisibility) {
       tileset()->setVisible(tilemapVisibility);
--- a/src/app/commands/cmd_layer_properties.cpp
+++ b/src/app/commands/cmd_layer_properties.cpp
@@ -470,7 +470,7 @@ private:
     uuid()->setVisible(uuidVisible);
 
     if (uuidVisible)
-      uuid()->setText(m_layer ? base::convert_to<std::string>(m_layer->uuid()) : "");
+      uuid()->setText(m_layer ? base::convert_to<std::string>(m_layer->uuid()) : std::string());
 
     if (tileset()->isVisible() != tilemapVisibility) {
       tileset()->setVisible(tilemapVisibility);
--- a/src/doc/layer.cpp
+++ b/src/doc/layer.cpp
@@ -36,9 +36,6 @@ Layer::Layer(ObjectType type, Sprite* sprite)
          type == ObjectType::LayerTilemap);
 
   setName("Layer");
-  // Always generate a UUID for this layer, but take into account that it could
-  // be replaced. For instance, when loading a layer that already had a UUID.
-  m_uuid = base::Uuid::Generate();
 }
 
 Layer::~Layer()
--- a/src/doc/layer.cpp
+++ b/src/doc/layer.cpp
@@ -1,5 +1,5 @@
 // Aseprite Document Library
-// Copyright (C) 2019-2021  Igara Studio S.A.
+// Copyright (C) 2019-2025  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -36,6 +36,9 @@ Layer::Layer(ObjectType type, Sprite* sprite)
          type == ObjectType::LayerTilemap);
 
   setName("Layer");
+  // Always generate a UUID for this layer, but take into account that it could
+  // be replaced. For instance, when loading a layer that already had a UUID.
+  m_uuid = base::Uuid::Generate();
 }
 
 Layer::~Layer()
--- a/src/doc/layer.h
+++ b/src/doc/layer.h
@@ -147,9 +147,7 @@ private:
   Sprite* m_sprite;          // owner of the layer
   LayerGroup* m_parent;      // parent layer
   LayerFlags m_flags;        // stack order cannot be changed
-  mutable base::Uuid m_uuid; // The UUID is generated the first time the "HasUUID" flag
-                             // is activated when it is a null UUID. If this field had
-                             // a valid UUID already, it won't be replaced by a new one.
+  mutable base::Uuid m_uuid; // lazily generated layer's UUID
 
   BlendMode m_blendmode;
   int m_opacity;
--- a/src/doc/layer.h
+++ b/src/doc/layer.h
@@ -129,7 +129,12 @@ public:
   int opacity() const { return m_opacity; }
   void setOpacity(int opacity) { m_opacity = opacity; }
 
-  const base::Uuid& uuid() const { return m_uuid; }
+  const base::Uuid& uuid() const
+  {
+    if (m_uuid == base::Uuid())
+      m_uuid = base::Uuid::Generate();
+    return m_uuid;
+  }
   void setUuid(const base::Uuid& uuid) { m_uuid = uuid; }
 
   virtual Grid grid() const;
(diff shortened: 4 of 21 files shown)


martincapello:
I know we talked about adding some preference to enable/disable UUIDs and use a layer chunk flag to indicate that the layer has a UUID. But then it made more sense to me to have a flag in the header part (and hence at the sprite level instead of program preferences) to indicate that all the layers will have a UUID, because I believe that if the usage of UUID is enabled, then all layers must have an UUID, otherwise none of them should, it seems like an all or nothing flag to me.

Anyway, we could also add a preference to set a default for every new sprite.

I'm still working on the IO side of this.

Gasparoken:
It would be nice if UUIDs were implemented for all cases. In certain scripts when you need to save the layer structure in a custom file by the user and then, in a different session you want to take the same layer structure.

I had to deal with the lack of this for the first time was in my script ([PathAnimatorTool](https://gasparoken.itch.io/path-animator-script)). I had included a feature that simplified the re-animation of an already made animation (re-animation was useful when the path, or the images, or the 'frame rate' were modified, i.e. without modifying the layer structure or parameterization). To achieve this I had to collect all the layers involved in the animation so I had to search for them by name. I knew there was the possibility of error if the artist changed the name of a layer, however I don't think I had a simple way to avoid it.
Another example where I could have used layer UUIDs in the same script: when defining the special layer that contains the path I had to do it by the layer name. I could have used the USER DATA to stamp some ID, but it would have been an unstable solution since the user can easily modify it.

Gasparoken:
IMO, layer UUIDs should be implemented as default for any sprite, and only disable it for very specific cases where memory is very limited (I think these cases will be rare).

dacap:
Just in case we talked about this with Martín, the issue here is that just opening an existing .aseprite file (creating all UUIDs for all layers automatically) and saving the .aseprite will make the .aseprite bigger (and the content of the file will change). And probably this will be unnecessary for most users (only for users using the Unity importer).

In my opinion, layer UUID shouldn't be included in any sprite until the user (and probably an extension?) request them. By default we shouldn't include new unnecessary information in .aseprite files.

Gasparoken:
Not only importers may need it. Certain scripts could make use of it:

I had to deal with the lack of this in my script ([PathAnimatorTool](https://gasparoken.itch.io/path-animator-script)). I had included a feature that simplified the re-animation of an already made animation (re-animation was useful when the path, or the images, or the 'frame rate' were modified, i.e. without modifying the layer structure or parameterization). To achieve this I had to collect all the layers involved in the animation so I had to search for them by name. I knew there was the possibility of error if the artist changed the name of a layer, however I don't think I had a simple way to avoid it.
Another example where I could have used layer UUIDs in the same script: when defining the special layer that contains the path I had to do it by the layer name. I could have used the USER DATA to stamp some ID, but it would have been an unstable solution since the user can easily modify it.

martincapello:
Let's say we make the UUID a new, and mandatory layer property. 
Then, let's imagine that a user opens a sprite that doesn't have UUIDs.
What should Aseprite do?
1. Open the sprite and mark it as unsaved? Because now it has these new properties that will change the sprite's file if saved.
2. Open the sprite as usual and just save the UUIDs silently when saved?

Also, what should happen when a script that expect layer's UUIDs loads a sprite that doesn't have UUIDs (because it was created in an old Aseprite version)? Because we could generate the UUID on the fly, but then, if the sprite is not saved, the next time the UUIDs will be different.

ckaiser:
My 2c:

Saving it transparently is IMO the best option, even if it might cause the files to get bigger, it's not a significant increase in size and idk about you but I'm not kilobyte-peeping my files (but then again I'm not a pixel artist so maybe y'all are really into kilobytes?)

For scripts, we could prompt the user with something like the permissions popup when code tries to read an UUID for a file that has none, and give them the option to: generate & save, generate & proceed without saving, abort - that way if the script is doing something and they wanna undo it or not save the results, they can just do that, but if they don't care and the script is something that's non-destructive, they can just save with the UUIDs. The popup can explain that the file was created with an older version, etc.

Keep in mind my opinion is from someone who's just been a fly on the wall for this feature's development so, grain of salt.

martincapello:
> it's not a significant increase in size and idk about you but I'm not kilobyte-peeping my files (but then again I'm not a pixel artist so maybe y'all are really into kilobytes?)

It is not about size I think, it is more about modifying the users files without let them know. Even with a change as insignificant as 16 bytes per layer. We could maybe warn the user when opening files without UUIDs per layers...but it seems like bothering a majority of users that are not using UUIDs. And, if the majority of users are not using UUIDs, then does actually worth saving the UUIDs for all of them?

Anyway, I'm not sure what's best...this are just some thoughts

Gasparoken:
One possible impact: it could cause confusion/queries to artists who use a version control software (like git), and perform a save action in a aseprite file without modifying anything, then get the surprise of the file change.

martincapello:
This is how this PR behaves right now:
In "Sprite Properties" I put a new "Use Universally Unique Identifiers for layers" checkbox:
Screenshot 2025-01-21 at 5 41 00 PM

Then, when it is checked, the "Layer properties" dialog looks like this:
Screenshot 2025-01-21 at 5 41 16 PM

The idea is that when a sprite with this checkbox enabled is saved, it would be saved with a UUID per layer. Otherwise, it would be saved without any UUID per layer.

One inconvenience of this approach, is that the user has to activate the checkbox every time a new sprite is created...so if the user wants to work mostly with sprites that require the UUIDs per layer feature, maybe it would be useful to have a default value for the checkbox in the preferences dialog that would be used every time a new sprite is created.

ckaiser:
IMO adding a checkbox and user option increases the maintenance burden on the codebase (something it already has coming out the ears), I think just being clear in the documentation and news that "Aseprite 1.<whichever> adds this to the sprite file format to improve things" - I don't think I'd be too concerned to see new data in my re-saved files when I upgrade, I can expect version changes to have some format/metadata differences.

martincapello:
I'm marking this as ready for review because it aligns with the following David's statement:
>In my opinion, layer UUID shouldn't be included in any sprite until the user (and probably an extension?) request them. By default we shouldn't include new unnecessary information in .aseprite files.

However, I will be happy to make any changes needed in case we decide to go for transparent/mandatory UUIDs for layers.

martincapello:
This is missing:
- Lua properties to access the layers UUIDs and the sprite's flag that indicate if it is using UUIDs.
- Tests.

I will tackle these once we agreed about the current approach. Just to avoid working in something that we are not sure yet.

dacap:
I'll review this next week. Some points about the discussion:

1. Although we've introduced changes in the past that broke some third-party encoders (e.g. removing the old color2 chunk 10dda30a15a58d09e561a59594f348b4db3a4405), it was done to fix some bugs and reduce the file size
2. Making the file larger without any benefit to most users makes no sense, even if it makes the life easier for us as developers
3. I'm thinking about this: why would we like to uniquely identify a layer, and not every single possible object inside the sprite? would not be better to just include this data in the "user data" chunk (as a new flag/field) and make it available for every single object (layer/cels/slices/etc.)

About:

> [...] we could generate the UUID on the fly, but then, if the sprite is not saved, the next time the UUIDs will be different.

IMO there is no problem about this, the script should not persist the UUID in external files, everything related to the generated UUID should be kept inside the saved file (if it's saved) or discarded (if it's not saved). So the UUID is auto-generated if a script reads it.

martincapello:
> I'm thinking about this: why would we like to uniquely identify a layer, and not every single possible object inside the sprite? would not be better to just include this data in the "user data" chunk (as a new flag/field) and make it available for every single object (layer/cels/slices/etc.)

I'm not sure if it actually makes sense to identify every single object with an UUID...maybe if we were about to identify every single object, an integer would make more sense? I mean, I would imagine an .aseprite file as a "database", where each element has its own id. The advantage of using UUID is that we are freed of managing the ids for each element...but size is much bigger (4 bytes for an integer id against 16 bytes for the UUID).

Anyway, maybe just using the user data properties to store the UUID for layers would be the best alternative if could offer some way (via aseprite extension maybe?) to hook into the layer's creation to let the aseprite extension's dev to store an UUID in the layer's user data properties. Then, the external tool plugin dev could provide a small Aseprite's extension to do this. Thus we wouldn't have to add anything new to .aseprite file specs, nor modify Aseprite code to mold to a specific external tool requirement.

docs/ase-file-specs.md
```diff
@@ -79,6 +79,7 @@ A 128-byte header (same as FLC/FLI header, but with other magic number):
                   1 = Layer opacity has valid value
                   2 = Layer blend mode/opacity is valid for groups
                       (composite groups separately first when rendering)
+                  4 = Layers are using UUIDs
```
dacap, line 82:
```suggestion
                  4 = Layers have an UUID
```

docs/ase-file-specs.md
```diff
@@ -202,6 +203,8 @@ entire layers layout:
     STRING      Layer name
     + If layer type = 2
       DWORD     Tileset index
+    UUID        Layer's universally unique identifier (present only if the
+                [main header](#header) "Flags" field has the bit 3 enabled)
```
dacap, line 207: Links don't work in Markdown code blocks:
```suggestion
    + If file header flags have bit 4:
      UUID      Layer's universally unique identifier
```
Here bit 4 is not fourth (or fifth) bit, but value 4, in the User Data Chunk we are already using this notation.

src/app/commands/cmd_layer_properties.cpp
```diff
@@ -463,6 +464,14 @@ class LayerPropertiesWindow : public app::gen::LayerProperties,
       m_userDataView.setVisible(false, false);
     }
 
+    bool uuidVisible = m_document && m_document->sprite() &&
+                       m_document->sprite()->useUuidsForLayers();
+    uuidLabel()->setVisible(uuidVisible);
+    uuid()->setVisible(uuidVisible);
+
+    if (uuidVisible)
+      uuid()->setText(m_layer ? base::convert_to<std::string>(m_layer->uuid()) : "");
```
dacap, line 473:
```suggestion
      uuid()->setText(m_layer ? base::convert_to<std::string>(m_layer->uuid()) : std::string());
```

src/doc/layer.cpp
```diff
@@ -36,6 +36,9 @@ Layer::Layer(ObjectType type, Sprite* sprite)
          type == ObjectType::LayerTilemap);
 
   setName("Layer");
+  // Always generate a UUID for this layer, but take into account that it could
+  // be replaced. For instance, when loading a layer that already had a UUID.
+  m_uuid = base::Uuid::Generate();
```
dacap, line 41: Probably I'd prefer to generate the `m_uuid` in the `uuid()` getter when `m_uuid` is empty (`m_uuid == Uuid()`).
```c++
const base::Uuid& Layer::uuid() const
{
  if (m_uuid == base::Uuid())
    m_uuid = base::Uuid::Generate();
  return m_uuid;
}
```
Converting `m_uuid` in `mutable`.

src/doc/layer.h
```diff
@@ -138,6 +142,9 @@ class Layer : public WithUserData {
   Sprite* m_sprite;     // owner of the layer
   LayerGroup* m_parent; // parent layer
   LayerFlags m_flags;   // stack order cannot be changed
+  base::Uuid m_uuid;    // The UUID is generated the first time the "HasUUID" flag
+                        // is activated when it is a null UUID. If this field had
+                        // a valid UUID already, it won't be replaced by a new one.
```
dacap, line 147: I think this comment wasn't updated for the latest version of this PR.

src/doc/layer_io.cpp
```diff
@@ -1,10 +1,11 @@
 // Aseprite Document Library
-// Copyright (c) 2019 Igara Studio S.A.
+// Copyright (c) 2019-2025 Igara Studio S.A.
 // Copyright (c) 2001-2018 David Capello
 //
 // This file is released under the terms of the MIT license.
 // Read LICENSE.txt for more information.
 
+#include "doc/uuid_io.h"
```
dacap, line 8: Please move this `#include` below with the other ones.

`"config.h"` and `"doc/layer_io.h"` must be the first ones.

src/doc/layer.h
```diff
@@ -128,6 +129,9 @@ class Layer : public WithUserData {
   int opacity() const { return m_opacity; }
   void setOpacity(int opacity) { m_opacity = opacity; }
 
+  const base::Uuid& uuid() const { return m_uuid; }
+  void setUuid(const base::Uuid& uuid) { m_uuid = uuid; }
```
dacap, line 133: One little nice check that we can add on debug mode is that we can set the UUID of the layer just one time if the UUID is generated inside `uuid()`:
```suggestion
  void setUuid(const base::Uuid& uuid)
  {
    ASSERT(m_uuid == base::Uuid());
    m_uuid = uuid;
  }
```

src/doc/layer_io.cpp
```diff
@@ -112,6 +114,10 @@ Layer* read_layer(std::istream& is, SubObjectsFromSprite* subObjects, const Seri
 {
   ObjectId id = read32(is);
   std::string name = read_string(is);
+  base::Uuid uuid;
+  if (serial >= SerialFormat::Ver3)
+    uuid = read_uuid(is);
```
dacap, line 119: I'd try to save the UUID as the last new field of the layer stream (instead of shifting all old fields), i.e. saving the UUID after the user data.

Why this? Because probably and older decoder could read the layer data of a newer encoder, it means that a session of a future version can be recovered with previous version of Aseprite (losing the UUID, but at least reading all the other data).

src/doc/sprite.h
```diff
@@ -272,6 +275,9 @@ class Sprite : public WithUserData {
   // giving the possibility to handle tiles exclusively to a plugin.
   std::string m_tileManagementPlugin;
 
+  // This setting indicates if the layers of this sprite are using UUIDs.
+  bool m_useUuidsForLayers = false;
```
dacap, line 279: Probably this and getter/setters could be called just `uuidsForLayers` (removing the "use" verb).

src/doc/uuid_io.h
```diff
@@ -0,0 +1,23 @@
...
+// This file is released under the terms of the MIT license.
+// Read LICENSE.txt for more information.
+
+#ifndef DOC_UUID_IO_H_INCLUDED
+#define DOC_UUID_IO_H_INCLUDED
+#pragma once
+
+#include "base/uuid.h"
+
+#include <iostream>
```
dacap, line 13: I think `<iosfwd>` is enough here:
```suggestion
#include <iosfwd>
```

dacap:
One extra minor comment I forgot: please see if you can rebase this to the latest beta commit, there is a minor conflict (some minor thing about the order of some #includes), so we can later merge the PR.

Issue #4256, opened 2024-01-03
Countour Tool: pixel perfect doesnt work after a straight line


EthanHunt404:
this is an old issue regarding the countour tool, Its quite simple if you try doing a straight line with the shift key and continue after the pixel perfect feature stops working

Steps to recreate:
- create a new document, any size, any color mode;
- select the coutour tool(shortcut D);
- draw a pixel anywere in the canvas, thats point A;
- hold shift, aseprite will preview the line from point A to whatever other pixel your mouse in hovering;
- press to confirm whenever you want but dont stop, immediatly start dragging it to continue drawing with the countour tool, you can stop holding shift now, it doesnt change whats happening
its at this point the bug shows itself.

### Aseprite and System version

* Aseprite version: 1.3.2, Steam version
* System: Windows 10 amd graphics card.

EthanHunt404:
1.3.2 behavior
https://www.loom.com/share/2588d1d2563c41a893680bef38944743
1.2.40 behavior
https://www.loom.com/share/aff49720730845aab15ac66b78ce32c0

i believe this happens ever since the start of 1.3 betas on steam too.

Gasparoken:
Thanks for reporting this. I was able to reproduce it.

EthanHunt404:
i just wanna say that the fix mister @Gasparoken proposed is working wonders, i compiled aseprite myself with the changes he proposed and it absolutely worked, not only countour tool, but the Pen and Eraser also were fixed.

EthanHunt404:
why did you reopen this @dacap?

EthanHunt404:
hey i want to comment on this again, im not sure why this has been moved to another milestone, the fix was suitable and it didnt seen to cause any other unintended effects, im relly invested on this specifically, cause its one of the main reasons i bought aseprite way back when, and it heavily hinders work, so i wanted to hear a reason to not implement what mr. @Gasparoken proposed.

PR #4949, opened 2025-01-23
Fix "Straight Line from Last Point” bug in Tilemaps (fix #4623)

### LINKED ISSUE #4623: “Straight Line from Last Point” behave wrong in Tilemaps
Reported in the forum: https://community.aseprite.org/t/straight-line-from-last-point-behave-wrong-in-tilemaps/22912/1

Gasparoken:
fix #4623

### DIFF
--- a/src/app/tools/intertwiners.h
+++ b/src/app/tools/intertwiners.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2024  Igara Studio S.A.
+// Copyright (C) 2018-2025  Igara Studio S.A.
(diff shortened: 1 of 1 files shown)


dacap:
I'll be testing this, the fix does work, but I'd like to know if it doesn't bring back #4256 as 611713e5fcefd26b289a7e339730994dc8e23299 introduced the change now reverted here in this PR.

dacap:
Yes, it looks like this fix brings back https://github.com/aseprite/aseprite/issues/4256#issuecomment-1878073706, we could try to find some fix for both issues.

dacap:
It works great in all cases :+1:

PR #4951, opened 2025-01-24
Add Aseprite Format section in the preferences dialog to introduce the "save raw cels" option

### LINKED ISSUE #4901: Add option to generate raw cel chunks in 'ase_file_write_cel_chunk' function
It would be really nice if there was an option for generating raw cel chunks, especially since the code to generate them exists but is never called.

martincapello:
Fix #4901

And the caacd8f7fa0ec70dddea83209b5462a673214ee1 commit fixes the setting of the format options for the document when a single file is loaded. It worked for file sequences but not for single files.


src/ui/await.h
```diff
@@ -0,0 +1,54 @@
...
+  ev.setType(os::Event::Callback);
+  ev.setCallback([func, &promise]() {
+    try {
+      promise.set_value(func());
+    }
+    catch (...) {
+      promise.set_exception(std::current_exception());
+    }
+  });
+  os::queue_event(ev);
```
martincapello, line 47: Note to myself:
- consider the possibility that the gui is not present. I think that some cli tests are hanging because of this.
- handle the exception set in the promise.

data/widgets/options.xml
```diff
@@ -562,6 +563,13 @@
           </hbox>
         </vbox>
 
+        <!-- Aseprite Format -->
+        <vbox id="section_aseprite_format">
+          <separator text="@.section_aseprite_format" horizontal="true" />
+          <check id="save_raw_cels" text="@.save_raw_cels"
+                 pref="aseprite_format.save_raw_cels" />
```
dacap, line 570: I think these kind of options can be an enum (similar to [ColorProfileBehavior](https://github.com/aseprite/aseprite/blob/8d30a18ed5eb269981cd7a53ce9f36a3fc902b4f/data/pref.xml#L107)) where we have 3 options: 1) default (use the most compressed/latest version of cels/aseprite specs), 2) use the same format as in the file, or 3) use the raw cel format.

Probably we need an `AsepriteOptions` (derived from `FormatOptions`) for this kind of options.

src/app/file/ase_format.cpp
```diff
@@ -1016,8 +1017,14 @@ static void ase_file_write_cel_chunk(FILE* f,
       link = nullptr;
   }
 
+  bool saveRawCels = ui::await<bool>([]() -> bool {
+    auto& pref = Preferences::instance();
+    return pref.asepriteFormat.saveRawCels();
+  });
+
```
dacap, line 1024: Please don't access `Preferences` from file encoders/decoders (not even using threads). There is an already established mechanism for this:
1. If it's an option to be asked from the UI each time we save we can access preferences from `onAskUserForFormatOptions()` and that goes to the `FileOptions` instance.
2. Use `FileOpConfig` and `FileOpConfig::fillFromPreferences()`

Probably this requires a combination of both where the `AsepriteOptions` is modified (or created) when saving the file depending on the preferences (not asking the user when saving, or in `onAskUserForFormatOptions()` without showing UI, just accessing the `Preferences` instance and updating the `AsepriteOptions`).

src/ui/await.h
```diff
@@ -0,0 +1,54 @@
...
+// and returns the result.
+template<typename T>
+T await(std::function<T()>&& func)
+{
+  if (ui::is_ui_thread()) {
+    return func();
+  }
+
+  std::promise<T> promise;
+  std::future<T> future = promise.get_future();
```
dacap, line 34: We are not using std::promise/future. Anyway we could start using them if they are needed, but generally we have `base::task` which is cancellable.

dacap (changes requested):
Some comments so far.

src/ui/await.h
martincapello: Sorry, but I'm not sure what you mean saying that we have base::task. Because I don't see it as a replacement of promises and futures, I think they are different things. If I understood correctly, base::task manages the execution of a function in a new thread. Promises enables existing threads to communicate between them, so there is no thread creation, they enable a thread to synchronize with another, among other things.  In this case,  promises  seemed to me like a great option. However,  since there's a different mechanism to get the preferences,  will use it.

dacap: Sorry, I read this code too quickly and thought `ui::await()` was doing something similar to `std::async()` (creating a new thread), but I see that it is communicating with the already existing UI thread. We are not using `std::async/promise/future` in the whole code base, but probably we could use `promise/future` if needed (I guess macOS 10.9 supports it).

Some points I want to say about this:

1. Any kind of sync needed from a worker thread towards the UI (and more if it's for such a simple thing like accessing a flag/option from the UI) is a code smell, we must try to launch the worker thread with all the required options from the UI, and avoid accessing the UI thread from the worker as much as possible
2. When I said we have `base::task`, actually it's to replace `std::async()`, but it's when the UI is calling a background thread to do some work (that can be cancelled + with progress bar). We have `app::Job` (old) and `app::Task` (new, which use `base::task_token`) too, but this doesn't apply in your case as it's the inverse (a worker trying to get info from the UI, which generally is not something I'd expect).

data/widgets/options.xml
martincapello: One question about this. If the user has the "use the same format as in the file" option selected, what cel format should we use when saving a new .aseprite file?

dacap: > I guess the default

Yes.

> How should I get the current cel_type when saving an .aseprite file when the user has this preference set to "use the same format as in the file" option?

This must be stored in `AsepriteOptions` after reading the `.aseprite` file (the `FormatOptions` pointer is stored in `app::Doc::formatOptions()`). When the user saves the sprite, if the option is "use the same format as in the file" and the `AsepriteOptions` exists in the `Doc`, you use those options (which is the one that was read from the original `.aseprite` file).

martincapello: I'm facing a problem when loading the .aseprite file. I get access to the cel_type in the `AsepriteDecoder::readCelChunk` method, but at that point I don't have access to any app::Doc instance. Which, as far as I saw, [it is constructed after sprite decoding](https://github.com/aseprite/aseprite/blob/81bd37551b8a58275c5ce6b2aaadc88533bbb587/src/app/file/ase_format.cpp#L306).

So, I've been thinking that maybe I could set the `FileOp`'s `FormatOptionsPtr`, since the decoder has a delegate that has a pointer to the `FileOp`. However I think there is no method to get the `FileOp` from the delegate. Hence I'm wondering if I should try an approach like this, by adding something to get the `FileOp` pointer from the delegate, then use it from `AsepriteDecoder::readCelChunk` to set its `FileOp`'s `FormatOptionsPtr` there, and finally, after the `Doc` creation, set the `Doc`'s `FormatOptions` pointer got from the `FileOp`.

dacap: I think you could just add a member variable in `AsepriteDecoder` to store the cel type read from the file, and then get this value in `AseFormat::onLoad()`.

dacap:
It works great, I'll merge it as it is. The only thing I'd like to rethink is the "Aseprite Format" section, it looks like a too advance tweak to be so accessible.

Issue #3131, opened 2022-01-14
Textbox UI logic


dacap:
There are some features that require the same logic for multiline text editing/navigation:

- [x] #28
- [x] #3130
- [ ] #1967 

It includes a similar logic to `ui::Entry` but for multiple lines (a text editor for multiple lines, vertical navigation with Up/Down).

balldrix:
Can we have any updates on this request/issue please?
https://github.com/aseprite/aseprite/issues/3130

dacap:
Probably multilines userdata / textboxes will be pushed a little into the future (not in the first beta).

PR #4968, opened 2025-02-01
[WIP] TextEdit Widget (implements #3131)

ckaiser:
Supersedes #4743, implements #3131.

Uses some of the skeleton of features intended for the feedback dialog just as a way to test things quickly (Help -> Feedback opens an edit) - will be removed before merging.

This is a completely new widget designed from scratch, so in theory it could replace the other multiline and text entry widgets (would need to have a readonly and one-line toggles added, but it'd be pretty simple) - but for now I need more eyes on it because I've been churning at it for a million years.

Known issues:
 - Caret to mouse position is a bit offset with TTF fonts - maybe? There's some strange scrolling/positioning bugs so it might be related to that, needs more testing.
 - Double click to select word behavior does not match Entry.
 - Haven't tested RTL languages yet.
 - Code still has some TODOs (which might be out of date).


src/app/ui/textedit.cpp
```diff
@@ -0,0 +1,748 @@
...
+    size_t advance = 0;
+    bool found = false;
+
+    line.blob->visitRuns([&](const text::TextBlob::RunInfo& run) {
+      if (found) {
+        return;
+      }
+
+      for (int i = 0; i < run.glyphCount; ++i) {
+        const gfx::RectF& glyphBounds = run.getGlyphBounds(i).offset(gfx::PointF(0, lineStartY));
```
dacap, line 548: This is not a full review and I'm not sure if it's working with MSVC, but with clang-18 on Linux the clicking/selecting a char with line>1 and column>1 didn't work, and just changing this to:
```suggestion
        gfx::RectF glyphBounds = run.getGlyphBounds(i).offset(gfx::PointF(0, lineStartY));
```
Will fix the selection in multiple lines/clicking specific chars in any line. Never use a reference to a temporal object :pray:

dacap:
I'll start reviewing this today, about:

> it could replace the other multiline and text entry widgets (would need to have a readonly and one-line toggles added, but it'd be pretty simple)

I'd prefer to keep both widgets separated, as the single line `ui::Entry` has some extra features (suffix, etc.) and combine (probably in a base class) the common functionality of both widgets.

The first thing I see is that the `TextEdit` widget is in `app-lib` (`src/app/ui/textedit.h`) but it should be in `src/ui/textedit.h`. This will limit the usage of the `Strings` class, `SkinTheme` class, etc.

A problem with `Entry::showEditPopupMenu()` is that it uses raw strings, I'll try to fix this first in `beta` branch. Using some strings in `en.ini` and a delegate similar to `ClipboardDelegate` but to translate strings from the UI. So then `TextEdit` can be moved to the `ui-lib`.

I'll continue reviewing the code as it is.

data/strings/en.ini
```diff
@@ -1900,3 +1906,9 @@ toggle_left_diagonal = Toggle -45° Symmetry
 show_options = Symmetry Options
 reset_position = Reset Symmetry to Center
 reset_position_to_view_center = Reset Symmetry to View Center
+
+[textedit]
+copy = &Copy
+cut = Cu&t
+paste = &Paste
+select_all = Select &All
```
dacap, line 1914: This is a good idea, I'll try to add these strings to `en.ini` and start using them in `ui::Entry` in some way (so then you can move `TextEdit` to the `ui-lib`).

dacap:
Added a new ui::TranslationDelegate in c01dc3472ac802ca6028c8637b83674f4d398805

dacap:
As the change is too big and I've already some local fixes, but cannot make a PR for this (as it includes temporal code, like the feedback command/dialog), I can cherry pick/integrate the changes from e572c83129bad749cf22042c90147910223e1c9d to create the TextEdit widget and continue working from there.

dacap:
Other possibility is to adjust this PR in such a way to merge it, but removing all the Feedback stuff.

Another note: Please follow the CODING_STYLE guide, mainly the one for switch/cases: https://github.com/aseprite/aseprite/blob/3e3dd2a653a6741aeaed18cd8327ec7c396d4b7b/docs/CODING_STYLE.md?plain=1#L83-L87

ckaiser:
Moved to ui-lib, did a full rebase and I'm now using the TranslationDelegate - I've removed the Feedback stuff and found a nicer way to test the functionality that can actually just go into a PR, in the layer user data, seems to be working great and closes another issue while we're at it:
image

dacap:
I'll be reviewing this one today.

dacap (changes requested):
There is a lot to review yet, but probably it'd be better to fix these issues, create a good first commit, and I can send you PR to your branch to continue fixing issues from there, in other case I'll make a lot of comments here and probably it's better to just continue working together in one branch.

dacap:
There are other changes in `.xml` files to make the user data text field expansible:
```diff
diff --git a/data/widgets/layer_properties.xml b/data/widgets/layer_properties.xml
index 6c4ef4115..2793f566f 100644
--- a/data/widgets/layer_properties.xml
+++ b/data/widgets/layer_properties.xml
@@ -4,7 +4,7 @@
 <gui>
 <window id="layer_properties" text="@.title">
   <vbox>
-    <grid id="properties_grid" columns="3">
+    <grid id="properties_grid" columns="3" expansive="true">
       <label text="@.name" />
       <entry text="" id="name" magnet="true" maxsize="256" minwidth="64" cell_align="horizontal" />
       <button id="user_data" icon="icon_user_data" tooltip="@general.user_data" />
diff --git a/data/widgets/slice_properties.xml b/data/widgets/slice_properties.xml
index a683919c5..549a86f1e 100644
--- a/data/widgets/slice_properties.xml
+++ b/data/widgets/slice_properties.xml
@@ -5,7 +5,7 @@
 <gui>
 <window id="slice_properties" text="@.title" help="slices#slice-properties">
   <vbox>
-    <grid id="properties_grid" columns="3">
+    <grid id="properties_grid" columns="3" expansive="true">
       <label id="label1" text="@.name" />
       <entry id="name" maxsize="256" magnet="true" cell_align="horizontal" expansive="true" />
       <button id="user_data" icon="icon_user_data" maxsize="32" tooltip="@.user_data_tooltip" />
diff --git a/data/widgets/sprite_properties.xml b/data/widgets/sprite_properties.xml
index 19d4c5553..13fd1dd9e 100644
--- a/data/widgets/sprite_properties.xml
+++ b/data/widgets/sprite_properties.xml
@@ -4,7 +4,7 @@
 <gui>
 <window id="sprite_properties" text="@.title" help="sprite-properties">
   <vbox>
-    <grid id="properties_grid" columns="3">
+    <grid id="properties_grid" columns="3" expansive="true">
       <label text="@.filename" />
       <entry text="" id="name" maxsize="256" minwidth="64" readonly="true" cell_align="horizontal" />
       <button id="user_data" icon="icon_user_data" maxsize="32" tooltip="@general.user_data" />
```

src/ui/theme.h
```diff
@@ -132,6 +133,7 @@ class Theme {
   virtual gfx::Color calcBgColor(const Widget* widget, const Style* style);
   virtual gfx::Size calcMinSize(const Widget* widget, const Style* style);
   virtual gfx::Size calcMaxSize(const Widget* widget, const Style* style);
+  virtual gfx::Color getColorById(const std::string& id) const { return gfx::ColorNone; };
```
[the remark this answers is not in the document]
ckaiser: I wasn't exactly sure how to do this at first but then I had the brilliant idea of just accessing the theme layers manually to get the colors, which... I don't think any other widget does so I'm pretty sure it's not right but is there any example of ui-lib stuff having multiple styles and getting the colors that way? I'm assuming we can't count on the order of the layers to be always consistent across user themes but maybe that's a requirement and this is fine.

dacap:
I'll rebase this PR and continue this work in other branch/PR.

dacap:
We'll continue in https://github.com/aseprite/aseprite/pull/5455 (branch https://github.com/aseprite/aseprite/tree/multiple-lines-textbox ), I'll try to fix some problems with the selection with TTF fonts, etc.

PR #4980, opened 2025-02-07
Dithering bug in RGBA->Indexed conversion with 2 colors (fix #4975)

### LINKED ISSUE #4975: Dithering bug during pixel format conversion RGBA->Indexed + only 2 colors in the palette
Image to convert:

Image

Result (1.3.12.beta1)
Image

Expected results:
Image

Related internal report: https://igarastudio.zendesk.com/agent/tickets/4149

Aseprite version: 1.3.12.beta1

Gasparoken:
fix #4975


martincapello (commented):
I didn't test this yet, but I wanted to say that while reproducing the issue (without the PR applied) I've noted that the dithering quality varies depending on which two colors are available in the palette. It seems that for some pair of colors the dithering works better than for others.
For instance:
image

image

So now I'm wondering if actually there is something that needs to be fixed in the current algorithm instead of adding a specific case for a palette with 2 colors.

I would love to know if this is the case, because if it is, then we should approach the solution in a different way. But I'm not sure if there is another way to know this than to try to fix the issue...or maybe asking David what does he think.

PR #4994, opened 2025-02-11
Fix onionskin visiblity is not updated if cel is empty (fix #4969)

### LINKED ISSUE #4969: Onionskin Visiblity Is Not Updated If Cell Is Empty
# Issue
The visibility of a layers onionskin is not updated when the layers visibility is toggled, if the layer does not have a cell on the current frame.
It is updated as soon as one performs an action such as zooming or editing the layer.

Image
*Video example of issue*

# Reproduce
Open the attached aseprite file, go to frame 2, enable onoinskinning, and toggle visibility of the layer to observe the issue.

[OnionSkinVisibilityRepo.zip](https://github.com/user-attachments/files/18630470/OnionSkinVisibilityRepo.zip)

# Aseprite and System version
- Aseprite version: 1.3.13-beta1 and 1.3.12
- System: Windows 10.0.19045

Gasparoken:
fix #4969

### DIFF
--- a/src/app/ui/timeline/timeline.cpp
+++ b/src/app/ui/timeline/timeline.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2024  Igara Studio S.A.
+// Copyright (C) 2018-2025  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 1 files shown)


src/app/ui/timeline/timeline.cpp
```diff
@@ -1953,6 +1953,8 @@ void Timeline::onAfterLayerVisibilityChange(DocEvent& ev)
   layer_t layerIdx = getLayerIndex(ev.layer());
   if (layerIdx >= 0)
     invalidateRect(getPartBounds(Hit(PART_ROW_EYE_ICON, layerIdx)).offset(origin()));
+  if (docPref().onionskin.active())
+    docPref().onionskin.AfterChange();
```
martincapello, line 1957: This works because the editor is connected to the `docPref().onionskin.AfterChange` signal and calls its `invalidate()` method. I believe that it would be better if we just call the editor's invalidate method directly as:

```suggestion
    m_editor->invalidate();
```

martincapello (changes requested):

martincapello (approved):

src/app/ui/timeline/timeline.cpp
dacap: The only problem using an invalidate for the current editor only is that it will not update other views/editors:

image

We can do something like this:
```diff
    m_document->notifyGeneralUpdate();
```

PR #4996, opened 2025-02-12
Fix crash canceling "Apply" filter and then using "OK"

### LINKED ISSUE #1610: Adjust Hue/Saturation on a locked layer causes Aseprite to crash
It is possible to open the Adjust Hue/Saturation... window on a locked layer, and applying the changes results in the crash of the application. 

Aseprite version: 1.2.4-x64
System: Windows 10 Home x64

dacap:
This fixes the `ASSERT(!m_tx)` in `FilterManagerImpl::initTransaction()` in Debug mode, and a crash in Release that happens in any filter if we cancel a filter progress in the middle, and then press the "OK" button.

The removed string was removed because it looks like it's not displayed anymore (without this patch, with this patch it's displayed but doesn't make sense). Probably it's not displayed now because something changed in the middle from its introduction from ab51f02711eb30dc6fd5e8eb5ed4902e111c3d90 to fix #1610 and the current `main`/`beta` branch.

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -1,5 +1,5 @@
 # Aseprite
-# Copyright (C) 2018-2024  Igara Studio S.A.
+# Copyright (C) 2018-2025  Igara Studio S.A.
 # Copyright (C) 2016-2018  David Capello
 #
 # This work is licensed under the Creative Commons Attribution 4.0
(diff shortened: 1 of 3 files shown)


martincapello (approved):
LGTM!

PR #4998, opened 2025-02-12
Use local deps directory in build script

cybardev:
Rather than polluting outside the cloned repository,
keep the skia dependency within the repo,
similar to how `.build` is used for build configs.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,5 @@
 tests/_test*
 build
 .build
+.deps
 CMakeSettings.json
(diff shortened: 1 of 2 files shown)


cybardev:
> I think we should check the existence of that directory first, then the `.deps` one, and by default creating it inside `.deps` if none exists.

Makes sense. I'll add the checks.

cybardev:
@dacap Hi, I put back the checks and added `.deps` as default. Requesting re-review. Thanks.

cybardev:
Should be ok now.

dacap:
I'll try to merge a modified version of this, making the `.deps` directory the default one only for users. For developers it makes sense to have the dependencies outside the repository (because it's shared between different clones).

dacap:
Thanks @cybardev, I've just merged it squashed with a condition to allow this only for users:
https://github.com/aseprite/aseprite/commit/90be6aac30eabdc12db4cd59bee5f9690f0d264c

PR #5002, opened 2025-02-14
Fix set_sat() blend function

### LINKED ISSUE #4593: Saturation blend mode on hue 0
When using a color with hue 0, the Saturation blend mode doesn't correctly set the saturation value of said color

How to recreate:
Use a color with a hue of 0, for instance #ff8080 seems to work well.
In a layer above, overlap the red color with another (or even the same) color and set the blend mode of that layer to Saturation.

image

Liebranca:
Fix #4593. With colors where `min` and `mid` resolve to the same value, such as `#ff8080`, the `set_sat()` function would pick a reference to the same channel twice when selecting `min`, `max` and `mid`, so I tweaked the logic to account for this edge case.

I did this with inline functions because the linter doesn't seem to like nested ternaries, but both ways should compile to the same thing.

Side note: I'm leaving the comment on top of `set_sat()` as I'm not sure if the way saturation is being set is correct either, though the results *look* just about right to me. I tried a few alternate formulas to check this, but they were much more computationally expensive, and the difference was not very noticeable, so in the end I just left the calculation untouched.

### DIFF
--- a/src/doc/blend_funcs.cpp
+++ b/src/doc/blend_funcs.cpp
@@ -397,6 +397,43 @@ static void set_lum(double& r, double& g, double& b, double l)
   clip_color(r, g, b);
 }
 
+static inline uint8_t get_imin_channel(const double r, const double g, const double b)
+{
+  // We use '<=' to get min so as to catch two channels being equal
+  if (r <= g && r <= b)
+    return 0b001;
+  if (g <= r && g <= b)
+    return 0b010;
+
+  return 0b100;
+}
+
+static inline uint8_t get_imax_channel(const double r, const double g, const double b)
+{
+  if (r > g && r > b)
+    return 0b001;
+  if (g > r && g > b)
+    return 0b010;
+
+  return 0b100;
+}
+
+static inline uint8_t get_imid_channel(uint8_t imin, uint8_t imax)
+{
+  // Getting the remaining channel through exclusion guarantees that it is neither min nor max
+  return (~(imax | imin)) & 0b111;
+}
+
+static inline double& index_to_ref(double& r, double& g, double& b, uint8_t i)
+{
+  if (i == 0b001)
+    return r;
+  if (i == 0b010)
+    return g;
+
+  return b;
+}
+
 // TODO replace this with a better impl (and test this, not sure if it's correct)
 static void set_sat(double& r, double& g, double& b, double s)
 {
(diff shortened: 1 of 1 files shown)


martincapello (commented):
I must admit, this is a clever implementation. However it seems a bit hard to read. I think that I would have created an array with the three components and order it from min to max (some ifs and a swap function should be enough), and then just take the ordered elements from it.
Although, I'm not sure which would be the most performant approach.

EDIT: Forgot to say that this approach works, so if @dacap is fine with the code, it can be approved.

src/doc/blend_funcs.cpp
```diff
@@ -397,6 +397,43 @@ static void set_lum(double& r, double& g, double& b, double l)
   clip_color(r, g, b);
 }
 
+static inline uint8_t get_imin_channel(double& r, double& g, double& b)
```
dacap, line 400: It looks like this function doesn't require references:
```suggestion
static inline uint8_t get_imin_channel(const double r, const double g, const double b)
```

src/doc/blend_funcs.cpp
```diff
@@ -397,6 +397,43 @@ static void set_lum(double& r, double& g, double& b, double l)
...
+  // We use '<=' to get min so as to catch two channels being equal
+  if (r <= g && r <= b)
+    return 0b001;
+  if (g <= r && g <= b)
+    return 0b010;
+
+  return 0b100;
+}
+
+static inline uint8_t get_imax_channel(double& r, double& g, double& b)
```
dacap, line 411: Same here:
```suggestion
static inline uint8_t get_imax_channel(const double r, const double g, const double b)
```

dacap (changes requested):
I'm not a big fan of the original implementation using references (probably it generates super slow code), but this PR is not about performance but fixing #4593 bug.

src/doc/blend_funcs.cpp
```diff
@@ -409,9 +446,15 @@ static void set_sat(double& r, double& g, double& b, double s)
...
-  double& max = MAX(r, MAX(g, b));
+  // Fetch channel indices
+  const uint8_t imin = get_imin_channel(r, g, b);
+  const uint8_t imax = get_imax_channel(r, g, b);
+  const uint8_t imid = get_imid_channel(imin, imax);
+
+  // Map the indices for each channel to references
+  double& min = index_to_ref(r, g, b, imin);
+  double& max = index_to_ref(r, g, b, imax);
+  double& mid = index_to_ref(r, g, b, imid);
```
martincapello, line 457: Just wanted to suggest the alternative implementation I was talking about
```suggestion
  // Order components from min to max
  if (r > g)
    std::swap(r, g);
  if (g > b)
    std::swap(g, b);
  if (r > g)
    std::swap(r, g);

  double& min = r;
  double& mid = g;
  double& max = b;
```

Liebranca: I had to use pointers instead of references for the ordering to work, but it's working nicely. If rewriting the function to use pointers is no problem, then I'd go with this one for readability.

To briefly touch on performance: it varies from platform to platform. This alternative is likely the more performant approach on x86_64 when compiling with clang; glancing at the disassembly I'm seeing that it's generating much better code (branchless and around half as many instructions), but for other platforms I couldn't tell.

dacap: I'll be testing the solution of this PR as it is in e359cb70105fa05da7bbf1af28445f5f447a1851 and trying to merge it.

In case someone wants to compare performance of different solutions, `src/doc/blend_benchmark.cpp` is a good place to start.

dacap:
It works great! With this patch we offer a result similar to Affinity Photo: https://github.com/aseprite/aseprite/issues/4593#issuecomment-2681962770

PR #5004, opened 2025-02-14
Fix image:drawImage+BlendMode.SRC doesn't draw mask pixels (fix #5001)

### LINKED ISSUE #5001: `Image:drawImage(image, , , BlendMode.SRC)` does not draw transparent pixels
The documentation for `BlendMode` and [`BlendMode.SRC`](https://www.aseprite.org/api/blendmode#blendmodesrc) seems to indicate that the `sourceImage` will completely REPLACE the `destinationImage` for the overlapping pixels, including when those are transparent pixels.

Image

However, when calling the following script, this doesn't seem to work at all. The output is exactly the same as when you use `BlendMode.NORMAL`.

```
	local sourceImage = app.layer:cel(app.cel.frame.frameNumber + 1).image
	app.cel.image:drawImage(sourceImage, Point(0,0), 255, BlendMode.SRC)
```

These are the example images I've used, with the output in the third cel. 
Image

Based on the documentation, the output should look like this instead:

Image

Worth noting is that this DOES work if the pixels aren't fully transparent, but are a color with an alpha above 0.

Image

**Aseprite 1.3.12x64** and **Aseprite 1.3.13-beta1x64**
Windows 10

Gasparoken:
fix #5001

### DIFF
--- a/src/doc/blend_internals.h
+++ b/src/doc/blend_internals.h
@@ -1,5 +1,5 @@
 // Aseprite Document Library
-// Copyright (c) 2024 Igara Studio S.A.
+// Copyright (c) 2024-2025 Igara Studio S.A.
 // Copyright (c) 2001-2015 David Capello
 //
 // This file is released under the terms of the MIT license.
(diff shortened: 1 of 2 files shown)


dacap:
It works great :+1:

PR #5011, opened 2025-02-17
Fix #4999

### LINKED ISSUE #4999: Straight line via SHIFT + Tiled mode starts at wrong position
Steps to reproduce it:
- Draw a point on the canvas with the 'pencil' tool.
- Activate Tiled Mode in both axes.
- Press and hold SHIFT key.
- The straight line starts from the most left-top tile, instead of the centered tile.

Aseprite version 1.3.12

Jack-a-Lope:
Fix for tile_mode header file
made main tile position lines up with where the 
SHIFT + pencil line feature draws from when 
tiled mode is enabled

I have signed the CLA
following the steps given in 
https://github.com/igarastudio/cla#signing

### DIFF
--- a/src/app/util/tiled_mode.h
+++ b/src/app/util/tiled_mode.h
@@ -41,12 +41,12 @@ public:
   gfx::Point mainTilePosition() const
   {
     gfx::Point pt(0, 0);
-    if (int(m_mode) & int(filters::TiledMode::X_AXIS)) {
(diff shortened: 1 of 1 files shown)


Gasparoken:
Hi, thanks for your help. Sorry, I tried the solution but the error is still there. Please do not repeat the open/close issue action without reason.

PR #5026, opened 2025-02-21
Fix symmetry button is kept pressed when we didn't pressed (fix #4760)

### LINKED ISSUE #4760: One symmetry button is kept pressed even when we didn't pressed
From #4669 one minor UI thing: depending the order of symmetry buttons we select/unselect one of the non-clicked buttons/symmetries might be selected at the end. In this video I'm only trying to click the 1st and 3rd buttons, but then the 2nd or the 4th will be kept pressed:

symmetry

I know that symmetries are auto-selected because they are enabled implicitly by the current ones (the "non sense symmetries filter"), but probably one thing we could make is to avoid selecting buttons automatically (and just keep this an internal flag), and auto-enable symmetries could be semi-transparent in the canvas, but I'm not sure if we need too much detail:

image

Gasparoken:
I need some feedback about:

1. Moved the drawing process for symmetry axes from 'Editor::drawSpriteUnclippedRect' to 'Editor::drawOneSpriteUnclippedRect' to allow semi-transparent axes. (is there some alternative?)
2. Axes on every tile in tile mode.
3. Added functions to 'projection.h' to consider pixel ratios other than 1:1

During the process, I saw that the Slices have an error when the color of them is semi-transparent. Since I am on the subject and I see that it would not be difficult to correct it, I will push a solution to solve this (EDIT: https://github.com/aseprite/aseprite/pull/5027).

fix #4760

### DIFF
--- a/src/app/tools/symmetry.cpp
+++ b/src/app/tools/symmetry.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2021-2024  Igara Studio S.A.
+// Copyright (C) 2021-2025  Igara Studio S.A.
 //
 // This program is distributed under the terms of
 // the End-User License Agreement for Aseprite.
@@ -20,7 +20,7 @@ void Symmetry::generateStrokes(const Stroke& stroke, Strokes& strokes, ToolLoop*
 {
   Stroke stroke2;
   strokes.push_back(stroke);
-  gen::SymmetryMode symmetryMode = loop->getSymmetry()->mode();
+  const gen::SymmetryMode symmetryMode = tools::Symmetry::resolveMode(loop->getSymmetry()->mode());
   switch (symmetryMode) {
     case gen::SymmetryMode::NONE: ASSERT(false); break;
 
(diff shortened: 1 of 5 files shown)


dacap:
Some questions about this PR: why do you need this new `button_flags` preference? Is not the current `mode` enough? Shouldn't we be able to calculate the full `mode` on-the-fly from the user preference (without touching the original `mode` option)?

Gasparoken:
I think I can do without the flags. I added flags for the buttons to avoid having the ((A || B) && (C || D)) logic present in every 'if' that should detect ALL symmetry mode, however I see that it is only used once in 'symmetry.cpp'.
Without the flags, ALL symmetry mode should never be detected using 'if (mode == SymmetryMode::ALL)' , instead the following logic should be implemented: 'if (((mode & SymmetryMode::HORIZONTAL) || (mode & SymmetryMode::VERTICAL)) && ((mode & SymmetryMode::RIGHT_DIAG) || (mode & SymmetryMode::LEFT_DIAG)))' (Readability could be increased by making an ALL mode detection function)

Gasparoken:
It turns out that if I save the `SymmetryMode::ALL` mode in `docPref`, and then call the symmetry mode from `docPref` I lose the information of the pressed buttons (for example: horizontal symm + right diagonal symm).
So I find it more convenient to save in docPref some flags that represent the pressed symmetry buttons, instead of the mode itself (if you can confirm this, I can change the name SymmetryMode to SymmetryFlags or another name of your liking). In this last approach, the real mode should be calculated before using 'if' conditions or 'switch/case' statements.

[7ef4fa0](https://github.com/aseprite/aseprite/pull/5026/commits/7ef4fa0ad0b49fc309ba635ee63e4a6db1e14df4) preserves the original 'symmetry.mode()' name on `docPref` to bring the flags, But I need your opinion about the approach and variable names.

dacap:
I'll try to rebase and merge this, I'm not 100% happy that there are 3 completely different issues fixed by one commit, but I'll try to merge is at it is (rebasing and changing the commit message to reference all issues + this PR).

dacap:
Merged in fdc9b2f0001d492bc90bd5816ab5cfb12fabee9d resolving a conflict with old code (drawSlices(g) was moved inside a clip).

PR #5029, opened 2025-02-22
Fix Straight line via SHIFT + Tiled mode starts at wrong position

lightovernight:
The issue id is #4999.The strokes are drawn based on the m_lastDrawingPoint stored in the Doc class.However,when the tile mode is modified,this coordinate reference does not update accordingly. This results in strokes being rendered at incorrect positions relative to the updated tile layout.

### DIFF
--- a/src/app/ui/editor/editor.cpp
+++ b/src/app/ui/editor/editor.cpp
@@ -2364,6 +2364,10 @@ void Editor::onTiledModeChange()
   spritePos += mainTilePosition();
   screenPos = editorToScreen(spritePos);
 
+  auto lastPoint = document()->lastDrawingPoint();
+  lastPoint += mainTilePosition() - m_oldMainTilePos;
+  document()->setLastDrawingPoint(lastPoint);
+
   centerInSpritePoint(spritePos);
 }
 


.gitignore
```diff
@@ -12,6 +12,7 @@
 *.res
 .DS_Store
 .vs
+.vscode
```
dacap, line 15: Remove this change:
```suggestion
```
(probably as other devs are using VSCode, we'll add this, but we can remove it from this PR)

src/app/ui/editor/editor.cpp
```diff
@@ -2363,6 +2363,10 @@ void Editor::onTiledModeChange()
 
   spritePos += mainTilePosition();
   screenPos = editorToScreen(spritePos);
+
```
dacap, line 2366: This line includes trailing whitespaces, check the pre-commit check:
https://github.com/aseprite/aseprite/actions/runs/13473873443/job/37752188196#step:4:117

Remember to configure the pre-commit hook: https://github.com/aseprite/aseprite/blob/main/CONTRIBUTING.md#pre-commit-hooks

dacap (changes requested):
Thanks @lightovernight for this patch! It works great, I'm commenting some minor changes that can be done to fix the PR so we can merge it.

dacap:
Thanks a lot @lightovernight! Going to merge it right now :+1:

PR #5037, opened 2025-02-27
Execute scripts in separate OS threads using lua threads

### LINKED ISSUE #4425: Stop button for running scripts / progress icon
This is planned from a long time ago (related to #1967) but never created the issue.

We should have a button to stop the running script (e.g. if a script has an infinite for loop or a long operation running), and an icon in the status bar that some progress is being made by the script.

martincapello:
This PR demonstrates the possibility to use lua threads along OS threads (using base::task and base::thread_pool) to handle scripts execution. This is at a very draft stage, in its current form it doesn't execute most of the scripts because it lacks of the necessary adjustments. 
The idea of this work is just to evaluate the possibility and viability of the approach. One thing I had in mind while doing this is that maybe the extensions should keep running as is, or at least at the beginning (if we decide to implement this).

The issue with the current script execution mechanism is that it makes impossible to avoid getting the UI blocked by a script running in an infinite loop for instance. So the user cannot stop the script because of the unresponsiveness of the UI.
 
Once finished it should fix #4425

Here is a video demonstrating this PR in action:

https://github.com/user-attachments/assets/91f4428c-19d3-40ea-87e8-714beeca209b

This is the script used:
```lua
local function doSomething()
end

do
  local dialog = Dialog("Test")
  :button { id = "hang", text = "To the infinity and beyond",
            onclick=function()
                     local a = 0
                     local b = 0
                     while true do
                      a = a + 1
                      b = b + 1
                      doSomething()
                     end
                   end
  }

  dialog:show{wait = false}
end
```
DISCLAIMER: I didn't test how all this works with the current debugger, so crashes are expected.

### DIFF
--- a/src/app/commands/cmd_grid.cpp
+++ b/src/app/commands/cmd_grid.cpp
@@ -52,6 +52,7 @@ protected:
     docPref.grid.snap(newValue);
 
     StatusBar::instance()->showSnapToGridWarning(newValue);
+    StatusBar::instance()->showRunningScriptsWindow(newValue);
   }
 };
 
--- a/src/app/commands/cmd_grid.cpp
+++ b/src/app/commands/cmd_grid.cpp
@@ -52,7 +52,9 @@ protected:
     docPref.grid.snap(newValue);
 
     StatusBar::instance()->showSnapToGridWarning(newValue);
+#ifdef ENABLE_SCRIPTING
     StatusBar::instance()->showRunningScriptsWindow(newValue);
+#endif
   }
 };
 
--- a/src/app/commands/cmd_run_script.cpp
+++ b/src/app/commands/cmd_run_script.cpp
@@ -73,7 +73,7 @@ void RunScriptCommand::onExecute(Context* context)
       return;
   }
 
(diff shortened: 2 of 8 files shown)


martincapello:
Maybe we could separate all this in another script engine, and add an "Experimental" switch to let the user decide to go back to the old scripting engine.

dacap:
@martincapello did you try a script that access a document or tries to show a dialog with this new engine? the entire scripting engine is designed to being executed in the main thread.

martincapello:
Yes, in the video I'm creating a dialog and showing it. Then its onclick handler calls a function that enters an infinite loop. I can make the necessary changes to try to access a document. You can see the script I tried in the PR description

martincapello:
Also, I know that the current scripting engine is designed to run in the main thread. I've added some changes to be able to run scripts in separate threads with as minimum impact as possible. At least to try the concept and validate that it could be done if we wanted to

dacap:
I'm assigning this PR to @ckaiser so he can review this and if this approach makes sense to stop running scripts.

PR #5038, opened 2025-02-27
Fix in Export As dialog the '-export' string never included (fix #5035)

### LINKED ISSUE #5035: [Beta] In the 'Export As...' dialog the '-export' string is never included
In the "Export As" dialog when the output filename is detected to be the same as the current sprite filename, the "-export" string is not included in the filename. This works as expected in the stable version 1.3.12

Internal report: https://igarastudio.zendesk.com/agent/tickets/7354

Regression introduced in: 0fbaaf3d821d13d8374c42300495f1b6d9b45c36

Aseprite version: 1.3.13beta1

Gasparoken:
fix #5035


ckaiser (approved):
LGTM!

PR #5048, opened 2025-03-07
Fix Unreplaceable transparent pixels created by resize

### LINKED ISSUE #5032: Un-replaceable transparent pixels created by resize
When resizing an image, some transparent pixels are created outlining the opaque pixels which cannot be replaced with the Replace Color function.
(They can still be manually filled or selected and deleted, but these actions need to be done on every cell manually, so not as convenient as Replace Color across multiple cells). 

Image

lightovernight:
# Description
This PR fixed #5032  by adding skip_for_nearest_neighbor parameter to fixup_image_transparent_colors.
When using nearest neighbor algorithm for image resizing, we should not modify the color values of transparent pixels. This is because:
1. Nearest neighbor algorithm directly picks the closest source pixel without interpolation
2. The original function would modify transparent pixels' color values based on their neighbors
3. This modification can cause issues when later applying color replacement to transparent pixels, as some pixels might use interpolated colors instead of the original transparent color
The new parameter allows us to skip this color fixing process when using nearest neighbor algorithm, maintaining the original transparent pixel values and ensuring correct color replacement behavior.


src/app/util/resize_image.cpp
```diff
@@ -38,7 +38,9 @@ doc::Image* resize_image(const doc::Image* image,
   std::unique_ptr<doc::Image> newImage(doc::Image::create(spec));
   newImage->setMaskColor(image->maskColor());
 
-  doc::algorithm::fixup_image_transparent_colors(newImage.get());
+  doc::algorithm::fixup_image_transparent_colors(
+    newImage.get(),
+    method == doc::algorithm::RESIZE_METHOD_NEAREST_NEIGHBOR);
```
dacap, line 43: As the argument is used to skip the function call, I think we have some other options:

1. One option would be to skip the call at the function call site:
```suggestion
  if (doc::algorithm::does_resize_method_require_fixup(method))
    doc::algorithm::fixup_image_transparent_colors(newImage.get());
```

2. Other option would be to pass the `method` to the function:
```suggestion
  doc::algorithm::fixup_image_transparent_colors(newImage.get(), method);
```

3. But I was thinking that probably the `fixup_image_transparent_colors` can be called inside `resize_image` when the `method` is not `RESIZE_METHOD_NEAREST_NEIGHBOR`.

Probably the third option is the best one, as the `resize_image()` already requires the fixup for other resize methods.

lightovernight:
Thanks for the suggestions! However,  the third option requires changing the  first param of `resize_image()` to non-const pointer, which might lead to reduced performance. One workaround is to declare a new const pointer after calling `fixup_image_transparent_colors`, like this
```
void resize_image(Image* src,
                  Image* dst,
                  const ResizeMethod method,
                  const Palette* pal,
                  const RgbMap* rgbmap,
                  const color_t maskColor)
{
  if(RESIZE_METHOD_NEAREST_NEIGHBOR != method)
    doc::algorithm::fixup_image_transparent_colors(src);
  const Image* csrc = src;
  //other code...
}
```
But I`m not sure if there is a better solution. Personally, I think the first option is the simplest and safest way to fix this bug, even though it might not be the perfect one.

src/app/util/resize_image.cpp
lightovernight: Thanks for the suggestions! However,  the third option requires changing the  first param of `resize_image()` to non-const pointer, which might lead to reduced performance. One workaround is to declare a new const pointer after calling `fixup_image_transparent_colors`, like this
```
void resize_image(Image* src,
                  Image* dst,
                  const ResizeMethod method,
                  const Palette* pal,
                  const RgbMap* rgbmap,
                  const color_t maskColor)
{
  if(RESIZE_METHOD_NEAREST_NEIGHBOR != method)
    doc::algorithm::fixup_image_transparent_colors(src);
  const Image* csrc = src;
  //other code...
}
```
But I`m not sure if there is a better solution. Personally, I think the first option is the simplest and safest way to fix this bug, even though it might not be the perfect one.

dacap:
Thanks @lightovernight, I've just merged your commit in 6aabfef0b846f73afa07135ccc7256c79ed851fe but then made some changes to integrate the fixup call inside resize_image() as it brought some extra fixes: 06d3bbf953b9bb2ff08d33c7f4e8198f4885af57

PR #5049, opened 2025-03-07
Duplicate and copy & paste slices

### LINKED ISSUE #4466: Allow copy/paste of one or multiple slices
This could be done via right click context dialog, _Menu > Edit > Copy_ or via shortcut (holding 'Ctrl' key while Slice tool is used)

* Aseprite version: 1.3.6
* System: All

martincapello:
Fix #4466

### DIFF
--- a/src/app/commands/cmd_duplicate_slice.cpp
+++ b/src/app/commands/cmd_duplicate_slice.cpp
@@ -1,0 +1,92 @@
+// Aseprite
+// Copyright (C) 2025  Igara Studio S.A.
+//
+// This program is distributed under the terms of
+// the End-User License Agreement for Aseprite.
+
+#ifdef HAVE_CONFIG_H
+  #include "config.h"
+#endif
+
+#include "app/cmd/add_slice.h"
+#include "app/commands/command.h"
+#include "app/context.h"
+#include "app/context_access.h"
+#include "app/context_flags.h"
+#include "app/site.h"
+#include "app/tx.h"
+#include "app/ui/editor/editor.h"
+#include "app/util/slice_utils.h"
+#include "base/convert_to.h"
+#include "doc/object_id.h"
+#include "doc/slice.h"
+
+namespace app {
+
+class DuplicateSliceCommand : public Command {
+public:
+  DuplicateSliceCommand();
+
+protected:
+  void onLoadParams(const Params& params) override;
+  bool onEnabled(Context* context) override;
+  void onExecute(Context* context) override;
+
+private:
+  ObjectId m_sliceId;
+};
+
+DuplicateSliceCommand::DuplicateSliceCommand()
+  : Command(CommandId::DuplicateSlice(), CmdRecordableFlag)
+{
+}
+
+void DuplicateSliceCommand::onLoadParams(const Params& params)
+{
+  std::string id = params.get("id");
+  if (!id.empty())
+    m_sliceId = ObjectId(base::convert_to<doc::ObjectId>(id));
+  else
+    m_sliceId = NullId;
+}
+
+bool DuplicateSliceCommand::onEnabled(Context* context)
+{
+  return context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
+                             ContextFlags::HasActiveSprite | ContextFlags::HasActiveLayer);
+}
+
+void DuplicateSliceCommand::onExecute(Context* context)
+{
+  std::vector<Slice*> selectedSlices;
+  {
+    const ContextReader reader(context);
+    if (m_sliceId == NullId) {
+      selectedSlices = get_selected_slices(reader.site());
+      if (selectedSlices.empty())
+        return;
+    }
+    else
+      selectedSlices.push_back(reader.sprite()->slices().getById(m_sliceId));
+  }
+
+  Editor* editor = Editor::activeEditor();
+  ContextWriter writer(context);
+  Tx tx(writer, "Duplicate Slice");
+  Sprite* sprite = writer.site().sprite();
+  editor->clearSlicesSelection();
+  for (auto* s : selectedSlices) {
+    Slice* slice = new Slice(*s);
+    slice->setName(get_unique_slice_name(sprite, s->name()));
+    tx(new cmd::AddSlice(sprite, slice));
+    editor->selectSlice(slice);
+  }
+  tx.commit();
+}
+
+Command* CommandFactory::createDuplicateSliceCommand()
+{
+  return new DuplicateSliceCommand;
+}
+
+} // namespace app
--- a/src/app/ui/doc_view.cpp
+++ b/src/app/ui/doc_view.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2024  Igara Studio S.A.
+// Copyright (C) 2018-2025  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -36,9 +36,11 @@
 #include "app/ui/workspace.h"
 #include "app/ui_context.h"
 #include "app/util/clipboard.h"
+#include "app/util/slice_utils.h"
 #include "base/fs.h"
 #include "doc/color.h"
 #include "doc/layer.h"
+#include "doc/slice.h"
 #include "doc/sprite.h"
 #include "fmt/format.h"
 #include "ui/accelerator.h"
@@ -510,6 +512,8 @@ bool DocView::onCanCopy(Context* ctx)
     return true;
   else if (m_editor->isMovingPixels())
     return true;
+  else if (m_editor->hasSelectedSlices())
+    return true;
   else
     return false;
 }
@@ -527,6 +531,11 @@ bool DocView::onCanPaste(Context* ctx)
              ctx->checkFlags(ContextFlags::ActiveLayerIsTilemap)) {
       return true;
     }
+  }
+
+  if (ctx->checkFlags(ContextFlags::ActiveDocumentIsWritable) &&
+      ctx->clipboard()->format() == ClipboardFormat::Slices) {
+    return true;
   }
   return false;
 }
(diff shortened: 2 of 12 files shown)


src/app/commands/cmd_duplicate_slice.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+bool DuplicateSliceCommand::onEnabled(Context* context)
+{
+  return context->checkFlags(ContextFlags::ActiveDocumentIsWritable |
+                             ContextFlags::HasActiveSprite | ContextFlags::HasActiveLayer);
+}
+
+void DuplicateSliceCommand::onExecute(Context* context)
+{
+  std::vector<Slice*> selectedSlices;
```
Gasparoken, line 62: There is already a `Slices` class which is exactly `std::vector<Slice*>`

src/app/ui/doc_view.cpp
```diff
@@ -560,15 +569,22 @@ bool DocView::onCopy(Context* ctx)
     ctx->clipboard()->copy(reader);
     return true;
   }
-  else
-    return false;
+
+  std::vector<Slice*> selectedSlices = get_selected_slices(reader.site());
```
Gasparoken, line 573: Reeplace `std::vector<Slice*>` with `Slices` here too.

Gasparoken:
It works great! Just a couple of subjetive design comments (i.e. optional):
- Would be great (IMO) if the position of the new Slices were displaced a few pixel to right and down to give a simple feedback to the artist that shows the duplication was actually done.
- Perhaps would be better to add "Copy" as suffix in the name of the new Slices to preserve consistency with the Layer duplication and to express better the copy.

src/app/commands/cmd_duplicate_slice.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+  ContextWriter writer(context);
+  Tx tx(writer, "Duplicate Slice");
+  Sprite* sprite = writer.site().sprite();
+
+  Doc* doc = static_cast<Doc*>(sprite->document());
+  doc->notifyBeforeSlicesDuplication();
+  for (auto* s : selectedSlices) {
+    Slice* slice = new Slice(*s);
+    slice->setName(slice->name() + " Copy");
```
ckaiser, line 93: This should probably be localized with "{} Copy"

src/app/util/clipboard.cpp
```diff
@@ -782,6 +815,26 @@ void Clipboard::paste(Context* ctx, const bool interactive, const gfx::Point* po
...
+
+      if (slices.empty())
+        return;
+
+      ContextWriter writer(ctx);
+      Tx tx(writer, "Paste Slices");
+      editor->clearSlicesSelection();
+      for (auto& s : slices) {
+        Slice* slice = new Slice(s);
+        slice->setName(slice->name() + " Copy");
```
ckaiser, line 830: Localization here as well.

src/app/ui/editor/tool_loop_impl.cpp
```diff
@@ -693,7 +694,7 @@ class ToolLoopImpl final : public ToolLoopBase,
       // popup menu to create a new one.
       if (!m_editor->selectSliceBox(bounds) && (bounds.w > 1 || bounds.h > 1)) {
         Slice* slice = new Slice;
-        slice->setName(getUniqueSliceName());
+        slice->setName(get_unique_slice_name(m_sprite));
```
ckaiser, line 697: In the end now this ends up being only used here, seems like it could still be just a private member instead of a part of the utils, which might make the addition of the extra file and translation unit harder to justify since it could also just be a static function like `Slice::getSelectedFromSite` but that's just a style thing.

ckaiser (changes requested):
Duplicating with the right click menu works, but trying to use Ctrl+C or the copy menu does not for me, from what I can see, `Site::selectedSlices()` ends up being 0 when called in `get_selected_slices` - not sure if it's calling the wrong Site or before it gets populated or what exactly is going on there.

Left some comments over some localization/style stuff as well.

src/app/ui/editor/tool_loop_impl.cpp
martincapello: About `get_unique_slice_name`, I think I used it in other places at some point, but maybe after some changes it ended up being used in just that place...so yeah, I will move that function back to the private scope.
About replacing `get_selected_slices` with a static `Slice::getSelectedFromSite` function, sure, I'll do it.

martincapello:
> trying to use Ctrl+C or the copy menu does not for me ... 

I just tried in macOS and it works both ways, will try in Windows to see if there is an issue there. Take into account that when using the Ctrl+C and Ctrl+V or copy menu and then paste menu, the copied slices are pasted in the same position as the original ones, not displaced as it happens when using the `Duplicate slice` context menu. However, the new slices are selected after pasting them, so you can easily move them around to displace them from the original position.

martincapello:
Force pushed a rebased version with the latest from beta. So, there are no changes yet.

src/app/ui/editor/tool_loop_impl.cpp
martincapello: Okay, I've just recalled why I moved the `getUniqueSliceName` to an utility function. It didn't make sense to me that  the ToolLoopImpl class provided that function. So, my reasoning was that moving it to another translation unit would remove some noise from the ToolLoopImpl and reduce its size (speaking of lines of code) a bit.
So now I would prefer to keep the `get_unique_slice_name` function as an utility function in a separate translation unit. But I'm open to change my mind if you feel strongly about this.

martincapello: I might consider moving it as another static member of Slice, something like: `Slice::makeUniqueName()`maybe?

martincapello:
I can confirm that in Windows Ctrl+C doesn't copy the slices. However, the copy menu and paste menu do work as expected. Also, I've tried to copy the slices with the menu and paste them with Ctrl+V and it worked as well. 
So, I'm guessing that there is some collision with the keys combo in Windows. Will investigate.

src/app/ui/editor/tool_loop_impl.cpp
martincapello: I cannot convert the `get_selected_slices` function into `Slice::getSelectedFromSite` because the `Site` class belongs to the `app` module (src/app/site.h) and the `Slice` class belongs to the `doc` module (src/doc/slice.h), and the `doc` module doesn't depend on the `app` module. So I think I will leave both functions as is.
I have also found that the `doc` module doesn't depend on the `fmt` lib, which the `get_unique_slice_name` uses (I know I could easily avoid using the `fmt` library but I prefer to leave these functions in the `app` module).

dacap:
I'm going to test this PR. About the i18n of `{} Copy` text, I don't know if it's possible to create a generic string for that. E.g. Portuguese might need a `Cópia da {}` or `Cópia do {}`, or French `Copie du {}` or `Copie de {}`.

Dev-Re2906 (commented):
deploy

martincapello:
> I'm going to test this PR. About the i18n of `{} Copy` text, I don't know if it's possible to create a generic string for that. E.g. Portuguese might need a `Cópia da {}` or `Cópia do {}`, or French `Copie du {}` or `Copie de {}`.

In this particular case I'm using the message for when a slice is copied. And the parameter is the name of the slice. Obviously we are not going to determine the name's gender to show the correct variant. So I guess you mean in the case the `{} Copy` text is used for some program entity, like the slice, or a frame, then in english we would have `Slice Copy` or `Frame Copy`. But in Spanish they could be `Copia de la rebanada` and `Copia del cuadro`, and hence we would need two templates, one for the feminine case, and another for the masculine case as you said.
Then if we define templates by gender I see another issue: we don't know which gender will have the word used as the parameter of the template in the current language, then we cannot use always one of the templates neither. 

However, if we use `Copia de {}`in spanish, it would be `Copia de rebanada`, `Copia de cuadro` which works well enough IMO, and also works when using names of entities, like `Copia de 'Rebanada 1'`. Not sure if this applies to all languages though.

Let me know what do you think we should do about this. I have a fix in mind for using the gender based templates, but I think it doesn't worth the additional complexity.

ckaiser:
> Let me know what do you think we should do about this. I have a fix in mind for using the gender based templates, but I think it doesn't worth the additional complexity.

Doing this kind of work ourselves is  pretty intense, it's why projects like [Fluent](https://projectfluent.org/) exist. I don't think we have enough text to justify it, could just be a matter of not using variables for this kind of thing and instead just having different strings for each (`Copy of slice` vs `Copy of {}`), only reserving template variables for stuff we know is a proper name or a user input that will usually not need a translation. It makes things more tedious but it ends up being simpler.

dacap:
As @ckaiser, for our case, we can have the string for each case. If in a future we consider that it’s getting out of hand, we could adopt some library like Fluent (didn’t know about that project, I remember some pluralization features from the old gettext library but never used it).

dacap:
> In this particular case I'm using the message for when a slice is copied. And the parameter is the name of the slice. 

Probably I was confused by the other PR, but as you said, it might work if it is the copy of a "name". Probably I got confused by the default name of things (later, slice, etc). Not sure if it can work in a generic form.

martincapello:
@ckaiser exactly, that is the easiest way when we don't have to parameterize the text. Unfortunately, I have to.

> Probably I was confused by the other PR, but as you said, it might work if it is the copy of a "name". Probably I got confused by the default name of things (later, slice, etc). Not sure if it can work in a generic form.

No worries. I'm not sure either, but I think it is the only way to say that we copied something with a name. In the other PR (https://github.com/aseprite/aseprite/pull/5167) I was using it for when the user duplicates a sprite (the parameter is the name of the file) and when the user duplicates a layer (the parameter is the name of the layer).

Is it okay to keep this as is then? Or should I revert the https://github.com/aseprite/aseprite/pull/5049/commits/000f163df46d5a9e1375db29651b9accd131fd52 change?

dacap:
> Is it okay to keep this as is then? Or should I revert the https://github.com/aseprite/aseprite/commit/000f163df46d5a9e1375db29651b9accd131fd52 change?

We can keep it, but you should amend that commit, as the string is in the `[general_text]` section which are strings used for the `ui::Entry` widget. You can move it to `[general]`.

martincapello:
I've just amended the last commit and rebased my branch.

martincapello:
@dacap Should I rebase this against main branch now?

dacap:
> @dacap Should I rebase this against main branch now?

Please rebase with main 👍

PR #5054, opened 2025-03-11
Add uuid field to layers

### LINKED ISSUE #5033: Add uuid field to layers in the scripting API
We should be able to access the UUID of a layer using `layer.uuid`.

martincapello:
Fix #5033 by adding access to the layer's uuid field and the sprite's uuidsForLayers field.


dacap:
The only thing I don't like of this public API is the "Sprite.uuidsForLayers" property name (the double `s`), could we choose some other alternative? some options:

* layerUuids
* useLayerUuids
* enableUuids
* enableLayerUuids
etc.

PR #5068, opened 2025-03-18
Fix 'Export As' output filename does not work with absolute paths (fix #5066)

### LINKED ISSUE #5066: The 'Export As' output file name does not work with absolute paths
Internal report: https://igarastudio.zendesk.com/agent/tickets/7432

Steps to reproduce it:
1. New Sprite
2. Save As `/Users/<username>/Sprite-0001.aseprite`
3. 'Export As' with Output File:`/Users/<username>/Sprite-0001.png`

Result: a new file is created in `/Users/<username>/Users/<username>/Sprite-0001.png`

Expected:  new file is created in `/Users/<username>/Sprite-0001.png`

Aseprite version 1.3.13

Gasparoken:
fix #5066

For it to work, you must first merge: https://github.com/aseprite/laf/pull/133

### DIFF
--- a/src/app/ui/export_file_window.cpp
+++ b/src/app/ui/export_file_window.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-2025  Igara Studio S.A.
 // Copyright (C) 2018  David Capello
 //
 // This program is distributed under the terms of
@@ -123,6 +123,12 @@ void ExportFileWindow::savePref()
 
 std::string ExportFileWindow::outputFilenameValue() const
 {
+  if (base::is_absolute_path(m_outputFilename)) {
+    // Although the output filename already contains an absolute path,
+    // calling get_absolute_path() normalizes it (removes any extra component)
+    // which is useful to hit the same recent filename/path.
+    return base::get_absolute_path(m_outputFilename);
+  }
   return base::get_absolute_path(base::join_path(m_outputPath, m_outputFilename));
 }
 


src/app/ui/export_file_window.cpp
```diff
@@ -123,7 +123,9 @@ void ExportFileWindow::savePref()
 
 std::string ExportFileWindow::outputFilenameValue() const
 {
-  return base::get_absolute_path(base::join_path(m_outputPath, m_outputFilename));
+  return (base::is_absolute_path(m_outputFilename) ?
+            base::get_absolute_path(m_outputFilename) :
+            base::get_absolute_path(base::join_path(m_outputPath, m_outputFilename)));
```
dacap, line 128: In this case I'd prefer to comment about why we are calling `get_absolute_path()` even when the condition says that the path is already absolute (which is a confusing), this is an example:
```suggestion
  if (base::is_absolute_path(m_outputFilename)) {
    // Although the output filename already contains an absolute path,
    // calling get_absolute_path() normalizes it (removes any extra component)
    // which is useful to hit the same recent filename/path.
    return base::get_absolute_path(m_outputFilename);
  }
  return base::get_absolute_path(base::join_path(m_outputPath, m_outputFilename)));
```

dacap (changes requested):
It'd be nice to test if switching to the absolute path option *automatically* in the UI would be beneficial in this case:

image

But probably it's better to avoid doing so, as it might be even more confusing (although the user wrote the path as absolute).

dacap:
As beta was merged with main, this was already fixed with #4873. Closing the PR (the `ExportFileWindow::outputFilenameValue()` function doesn't exist anymore).

PR #5069, opened 2025-03-19
Palette picker pinning, loading speed, misc improvements (fix #2365)

### LINKED ISSUE #2365: Improve Palette Picker
Just a couple of simple-ish UI things I that I think would make the Palette Picker more robust.

image
 🔼 _This bad boy._

 - **Palette Groups:** Lets users sort their palettes into named categories by dragging and dropping palettes into collapsable/expandable folders, so they do not need to see every one of them every time. May or may not allow one palette to appear in multiple groups; that is neither here nor there for me.

 - **Recent Palettes List:** It would be helpful if there was a list of recently-loaded or saved palettes, in the same way that _Open Recent_ does for saved and loaded image files.
Also consider adding a "Add to Palette Picker" option to the Load pane (when loading palettes) that would list the loaded palette if it is not already there, maybe!

Thanks for considering it! :D 

..using ASEprite 1.2.18-x86 on Win10.

ckaiser:
Implements #2365:

 - Adds pinning to the palette resource list, stored in the ini file.
 - Restores the broken functionality of the "Loading" indicator inside resource lists.
 - Improved loading performance.
 - Makes it so transparent IconButton can be more easily used inside these kinds of widgets by not drawing the background.

Result:

https://github.com/user-attachments/assets/0587a1a3-cc13-411d-86a5-c91bc2bcaf77

### DIFF
--- a/src/app/ui/icon_button.cpp
+++ b/src/app/ui/icon_button.cpp
@@ -28,6 +28,12 @@ IconButton::IconButton(const SkinPartPtr& part) : Button(""), m_part(part)
   initTheme();
 }
 
+void IconButton::setIcon(const skin::SkinPartPtr& part)
+{
+  m_part = part;
+  invalidate();
+}
+
 void IconButton::onInitTheme(InitThemeEvent& ev)
 {
   Button::onInitTheme(ev);
@@ -44,7 +50,7 @@ void IconButton::onSizeHint(SizeHintEvent& ev)
 
 void IconButton::onPaint(PaintEvent& ev)
 {
-  auto theme = SkinTheme::get(this);
+  const auto* theme = SkinTheme::get(this);
   Graphics* g = ev.graphics();
   gfx::Color fg, bg;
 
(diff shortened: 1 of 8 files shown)


hacker30083 (approved):
During a quick review of the code, I didn't see anything that seemed off. I am also surprised that no one has already added it seeing as it is a good and useful feature. In the end I would definitely approve of the addition.

src/app/ui/icon_button.cpp
```diff
@@ -61,14 +61,16 @@ void IconButton::onPaint(PaintEvent& ev)
...
-  gfx::Rect bounds = clientBounds();
+  const gfx::Rect bounds = clientBounds();
   os::Surface* icon = m_part->bitmap(0);
   g->drawColoredRgbaSurface(icon,
                             fg,
-                            bounds.x + bounds.w / 2 - icon->width() / 2,
-                            bounds.y + bounds.h / 2 - icon->height() / 2);
+                            bounds.x + (bounds.w / 2) - (icon->width() / 2),
+                            bounds.y + (bounds.h / 2) - (icon->height() / 2));
 }
```
dacap, line 74: These changes are not needed, even more, they generate conflicts (see below).

src/app/ui/resources_listbox.cpp
```diff
@@ -67,7 +68,7 @@ void ResourceListItem::onPaint(PaintEvent& ev)
     text(),
     fgcolor,
     gfx::ColorNone,
-    gfx::Point(bounds.x + 2 * guiscale(), bounds.y + bounds.h / 2 - g->font()->height() / 2));
+    gfx::Point(bounds.x + (2 * guiscale()), bounds.y + (bounds.h / 2) - (g->font()->height() / 2)));
```
dacap, line 71: This change generates a conflict and is not needed in the PR.

Please remove everything that is superfluous from the PR and commit just the required changes to fix the issue.

dacap (changes requested):

dacap:
It works great! 👍

PR #5082, opened 2025-03-30
fix build.sh input acceptance for build type

AGitmit:
Disclaimer:
------------

Description of the fix:
----------------------
While attempting to build an executable using the `build.sh` script included in the repo I found that it crashed constantly during installation because it couldn't assign the input I've provided at the release type prompt for the user.
I looked into the script and quickly found that line (see '_File changed_' for reference) that didn't look right to me - I modified it and tried again - worked perfectly! like it should.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -265,7 +265,7 @@ else
     # New build
     if [[ "$build_n" == "n" || "$build_n" == "N" ]] ; then
         read -p "Select build type [RELEASE/debug]? "
-        if [[ "${REPLY,,}" == "debug" ]] ; then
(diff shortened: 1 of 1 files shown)


dacap:
Hi @AGitmit, are you using `sh` shell? Or `bash`?

It looks like the `${var,,}` syntax to convert a value to lowercase is a bash extension: https://unix.stackexchange.com/a/500275/49338

Anyway we could try to find other solution if this is the only incompatibility with `sh`, e.g. using

```
        if [[ "$(echo $REPLY | tr '[:upper:]' '[:lower:]')" == "debug" ]] ; then
```

Issue #1485, opened 2017-05-13
Duplicate Timeline Position Option in Preferences


KasumiArai:
I would like the position option that appears here: 
![image](https://cloud.githubusercontent.com/assets/8582123/26022073/c0bfbca8-3768-11e7-879b-ad8556ae9ff4.png)
(Left, right, bottom)
To be duplicated iin the Timeline preferences here.
![image](https://cloud.githubusercontent.com/assets/8582123/26022090/3427dd60-3769-11e7-92d1-f51eed6b3b35.png)
Reason: Someone asked me how to move the timeline, I remembered that you could, but I could not find it in the first place I looked. (Preferences)

dacap:
Referenced from: https://community.aseprite.org/t/a-way-to-change-the-default-size-of-the-layer-thumbnail/19863/4

dacap:
I'll try to add some timeline options in the preferences dialog. Although the timeline position will be changed soon with #3485

dacap:
At the moment the easiest path will be a "Set as Defaults" button:

![Image](https://github.com/user-attachments/assets/d7fc1f14-1f63-42f9-b826-a3a784071de1)

Although this doesn't fix this issue, which is aimed to simplify the discoverability of these settings.

PR #5086, opened 2025-03-31
Add option to scale up timeline thumbnails to fill the entire cel (fix #5083)

### LINKED ISSUE #5083: Fix regression with timeline thumbnails
We should revert #4974 and make the option configurable. From: https://community.aseprite.org/t/images-in-the-time-are-scale-odd/25097

Gasparoken:
fix #5083


dacap:
I'll merge this as it is. Unfortunately there is no way to set these options as global defaults at the moment. I'll try to implement part of #1485 and https://community.aseprite.org/t/a-way-to-change-the-default-size-of-the-layer-thumbnail/19863/4 to fix this.

Issue #5018, opened 2025-02-19
When Aseprite is launched on Windows, a white screen opens before the main application is displayed


Gasparoken:
This detail can be very annoying for the artist with photophobia.
Internal report: https://igarastudio.zendesk.com/agent/tickets/7324

Aseprite version 1.3.12
Windows OS

PR #5090, opened 2025-04-01
Startup optimizations

ckaiser:
When investigating #5018 I came across some opportunities to do some optimization to startup times, in most cases by delaying more expensive operations, which results in better perceived performance (time before seeing the window).

Haven't measured exact impact yet (hence the draft), but the DitheringSelector especially was especially taking up most of the initialization time of the main window:
image

Delaying initialization until the widget is visible removes even the initialize() function from the flamegraph. In theory we could apply more of these delayed initializations to other widgets and tools on a case-by-case basis, especially things that involve going through files.

### DIFF
--- a/src/app/ui/main_menu_bar.cpp
+++ b/src/app/ui/main_menu_bar.cpp
@@ -14,6 +14,8 @@
 #include "app/app.h"
 #include "app/app_menus.h"
 #include "app/extensions.h"
+#include "ui/manager.h"
+#include "ui/message.h"
 
 namespace app {
 
@@ -26,13 +28,26 @@ MainMenuBar::MainMenuBar()
 
   // Reload the main menu if there are changes in keyboard shortcuts
   // or scripts when extensions are installed/uninstalled or
-  // enabled/disabled.
-  m_extKeys = extensions.KeysChange.connect([this] { reload(); });
-  m_extScripts = extensions.ScriptsChange.connect([this] { reload(); });
+  // enabled/disabled, enqueued to avoid multiple reloads.
+  m_extKeys = extensions.KeysChange.connect([&] { queueReload(); });
+  m_extScripts = extensions.ScriptsChange.connect([&] { queueReload(); });
+}
+
+void MainMenuBar::queueReload()
+{
+  if (m_queuedReload)
+    return;
+
+  m_queuedReload = true;
+
+  auto* msg = new CallbackMessage([this] { reload(); });
+  msg->setRecipient(this);
+  Manager::getDefault()->enqueueMessage(msg);
 }
 
 void MainMenuBar::reload()
 {
+  m_queuedReload = false;
   setMenu(nullptr);
 
   // Reload all menus.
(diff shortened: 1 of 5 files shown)


src/app/ui/main_menu_bar.cpp
```diff
@@ -26,13 +28,23 @@ MainMenuBar::MainMenuBar()
...
+    if (m_queuedReload)
+      return;
+    m_queuedReload = true;
+
+    auto* msg = new CallbackMessage([this] { reload(); });
+    msg->setRecipient(this);
+    Manager::getDefault()->enqueueMessage(msg);
+  };
+  m_extKeys = extensions.KeysChange.connect(reloadCbk);
+  m_extScripts = extensions.ScriptsChange.connect(reloadCbk);
```
dacap, line 42: Probably I'd move this `reloadCbk` code into a `delayedReload()` member function.

src/ui/manager.cpp
```diff
@@ -196,12 +196,6 @@ Manager::Manager(const os::WindowRef& nativeWindow)
   , m_lockedWindow(nullptr)
   , m_mouseButton(kButtonNone)
 {
-  // The native window can be nullptr when running tests
-  if (nativeWindow) {
-    nativeWindow->setUserData(&m_display);
```
dacap, line 201: I'd do this anyway at the very beginning and move only the `setDragTarget()` for later.

src/ui/manager.cpp
```diff
@@ -242,6 +236,17 @@ Manager::Manager(const os::WindowRef& nativeWindow)
...
   // TODO check if this is needed
   onNewDisplayConfiguration(&m_display);
+
+  // The native window can be nullptr when running tests
+  if (nativeWindow) {
+    nativeWindow->setUserData(&m_display);
+
+    // Setting the drag target has a slight performance cost that we can offset by running it later.
+    auto* callbackMessage = new CallbackMessage(
+      [&, nativeWindow] { nativeWindow->setDragTarget(this); });
```
dacap, line 246:
```suggestion
      [this, nativeWindow] { nativeWindow->setDragTarget(this); });
```

dacap (changes requested):
These changes look good, I've made some minor comments (you can rebase editing the commits).

One function that I know is giving some performance issues on Windows is `os::win_is_key_pressed()` which is used to get the keyboard state for every single `ui::Message` that is not generated directly from a keyboard message (and probably it's not necessary to call it so frequently).

Now, after applying your changes that is the function where most of the time is spent in the initialization:

image

Something to see in a future.

dacap:
I've just created a PR to avoid calling `os::win_is_key_pressed()` so many times https://github.com/aseprite/aseprite/pull/5092

ckaiser:
Just did some crude measurements with [PassMark AppTimer](https://www.passmark.com/products/apptimer/) (which felt like going back in time to VB6 applications) and startup time is around 12% faster 🥳

PR #5091, opened 2025-04-01
Fix crash when undoing drag & drop of single image into a cel

martincapello:
The issue was that we were using the MoveCel command to move the cel from a temporal source document (the image dragged) to our destination cel in the current editor. Then when trying to undo, the source document didn't exist any more, so there is no document to put the moved cel back, thus getting a null layer when "AddCel::onRedo()" is executed at undo time.

### DIFF
--- a/src/app/cmd/drop_on_timeline.cpp
+++ b/src/app/cmd/drop_on_timeline.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2024  Igara Studio S.A.
+// Copyright (C) 2025  Igara Studio S.A.
 //
 // This program is distributed under the terms of
 // the End-User License Agreement for Aseprite.
@@ -11,7 +11,7 @@
 #include "app/cmd/drop_on_timeline.h"
 
 #include "app/cmd/add_layer.h"
-#include "app/cmd/move_cel.h"
+#include "app/cmd/copy_cel.h"
 #include "app/cmd/set_pixel_format.h"
 #include "app/console.h"
 #include "app/context_flags.h"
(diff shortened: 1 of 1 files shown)


src/app/cmd/drop_on_timeline.cpp
```diff
@@ -107,13 +107,13 @@ void DropOnTimeline::onExecute()
       }
 
       // If there is only one source document to process and it has a cel that
-      // can be moved, then move the cel from the source doc into the
+      // can be moved, then copy the cel from the source doc into the
       // destination doc's selected frame.
       const bool isJustOneDoc = (docsProcessed == 1 && !hasPendingWork());
       if (isJustOneDoc && canMoveCelFrom(srcDoc.get())) {
```
dacap, line 113: Probably we have to update the comment that says "it has a cel that can be moved" and the `canMoveCelFrom()` member function name.

martincapello: Sure! I will apply this to https://github.com/aseprite/aseprite/pull/5106 as well, because that PR will actually remove the DropOnTimeline command in favor of using the DocApi class.

martincapello: Hmmm, what if we just left this as is and I just put the changes on my draft? Because this changes will be lost anyway when that PR got merged.

martincapello: This is the commit in the other PR: https://github.com/aseprite/aseprite/pull/5106/commits/59dc18213f1d5d93d11caf1fa9d9add70b1311fd

martincapello:
This is not needed because it is included in #5106

PR #5092, opened 2025-04-02
Prevent polling keyboard state for each created ui::Message

dacap:
Just to avoid calling `os::win_is_key_pressed()` for each created `ui::Message`.


ckaiser (approved):
Looks good, and when I cherry-picked it into #5090 gave me an extra 1% startup speed improvement 🚀

PR #5093, opened 2025-04-02
Updated libpng from 1.6.40 to 1.6.42

bjorn:
Maybe we should update to the latest version (1.6.47) but my main need for updating this submodule is to raise the minimum CMake version to at least 3.5 since recent CMake versions no longer support older minimum versions.

In libpng 1.6.41, the minimum CMake version was raised to 3.6. The latest version of libpng needs CMake 3.14.

---


bjorn:
Ah, I see this is a duplicate of #5087 (or basically, a little more conservative version since it does the minimum update).

bjorn:
Urgh, and TinyEXIF has the same issue, heh.

dacap:
Thanks @bjorn, I'll take a look to these issues with cmake. There are some deprecation warnings too (e.g. for giflib), I'll try to fix all of them.

PR #5099, opened 2025-04-04
Fix timeline's vertical scrolling behavior

### LINKED ISSUE #4930: Don't touch timeline vertical scroll when we add a new layer and this layer will be visible w/the current scroll pos
Reported here: https://community.aseprite.org/t/bug-with-layer-window-and-scroll/2408

martincapello:
Fixes #4930 by avoiding re-stacking of layers. 
The issue was that the new layer was being added as the last layer first, which fired some notifications and made the timeline update its vertical scroll to make the new layer visible, then a re-stack of layers was being done to put the new layer in the correct position inside the list of layers, firing some notifications again making the timeline to update its vertical scroll again. 

The solution in this PR introduced some new methods in DocApi to allow creating a layer after a specified layer directly, thus avoiding the steps of adding it as the last layer first and re-stack the layers then.

### DIFF
--- a/src/app/commands/cmd_new_layer.cpp
+++ b/src/app/commands/cmd_new_layer.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-2025  Igara Studio S.A.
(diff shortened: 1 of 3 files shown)


Gasparoken:
Works great! LGTM.

PR #5106, opened 2025-04-09
Refactor drop on timeline command

### LINKED ISSUE #4777: Some refactorings due to new "drop on timeline" behavior
Some changes to do after #4636 was merged:

- [x] To be sincere I'm not a big fan about how the `cmd::DropOnTimeline` is implemented: All the `m_paths`/`getNextDoc()`/etc. looks like temporal data used only to move params from `DropOnTimeline()` constructor to its `onExecute()`, and then never used again for undo/redo purposes (where the `CmdSequence` is used). Probably a lot of this logic could be moved to a `DocApi` function and simplify the `DropOnTimeline` cmd if needed at all.
- [x] One problem might be creating a `Tx` inside `DropOnTimeline::onExecute()`, and as the `srcDoc` is deleted anyway, we don't need to "transact" the color conversion. We can just convert the sprite directly without undo information.
- [x] As there is a `delete srcDoc` there, probably the function could receive a `unique_ptr<Doc>` or something like that.
- [x] Try to avoid double pointers as much as possible (`Doc** srcDoc`, `doc::Layer**`, `doc::LayerGroup**`) as it makes things harder to maintain and more error-prone (they come from a "legacy" C era, and it'd be better if we avoid them). As alternatives:
   * We can use references to a pointer (e.g. `Doc*&`), or
   * Return the pointer if possible (e.g. for functions that return true/false if the pointer is set),
   * If you have to use them, use the parameter just one time, e.g.:
     ```c++
     bool getNextDoc(Doc** outDoc) {
       Doc* doc = ...;
       ...
       if (outDoc)
         *outDoc = doc;
       return (doc ? true: false);
     }
     ```
     or
     ```c++
     Doc* getNextDoc() {
       Doc* doc = ...;
       ...
       return doc;
     }
     ```
- [x] Remove the `ui::Manager` as a friend of `ui::Widget` (we should try to avoid "friend" between big classes). Probably creating a new kind of message and using `sendMessage()` (I know that we cannot enqueue the message as it must be processed ASAP so the OS gets a response). Or create a public API to call those virtual functions.
- [x] Show better feedback when dropping a cel, just by chance I noticed that the position of the mouse inside the cel will paste the image above/below the layer:
   [image]
- [x] All the `OpenFileJob` impl was moved to a header file, please create a new `.cpp` file to put the impl there.
- [x] Avoid creating/enqueing `os::Event` from the ui/app layers to the laf-os
- [x] Avoid creating a temporal surface in `static WEBP_CSP_MODE colorMode()` just to know the default surface color mode, and create the required laf-os API for this
- [x] I guess `copy_layer_with_sprite()` could use `clone = std::make_unique<T>(...)` several times instead of `clone.reset(new T(...))`

martincapello:
This PR fixes #4777 which involves several changes that aim to improve code consistency and remove unnecessary code


src/app/doc_api_dnd_helper.cpp
```diff
@@ -0,0 +1,142 @@
...
+// Returns true if the document srcDoc has a cel that can be copied into a
+// destDoc's cel.
+// The cel from srcDoc can be copied only when all of the following conditions
+// are met:
+// * Drop took place in a cel.
+// * Source doc has only one layer with just one frame.
+// * The layer from the source doc and the destination cel's layer are both
+// Image layers.
+// Otherwise this function returns false.
+bool canCopyCel(Doc* srcDoc, Doc* destDoc, DroppedOn droppedOn, doc::layer_t layerIndex)
```
dacap, line 34: We use `snake_case` for standalone functions.

dacap (approved):
I'll merge this as it's a make the changes in the naming functions, I was't able to give a proper testing on all platforms.

Issue #2019, opened 2019-02-21
mousewheel zooming (quick) bug


golddotasksquestions:
Mousewheel zooming works fine with Hand tool as well as Zoom tool selected.
It does not work with while in zoom (quick) however.
Would be nice if it could be fixed. Thanks.

### Aseprite and System version

* Aseprite version: 1.2.9
* System: Windows

PR #5118, opened 2025-04-13
Prioritize mouse wheel zooming with quick zoom tool

hwabis:

(Not signed yet but submitted the form)

Fixes https://github.com/aseprite/aseprite/issues/2019

Before this PR: scrolling with "Zoom Tool (quick)" held down works if you assign it to a keyboard shortcut that's not used by anything else (e.g. "7"). However, scrolling if it's assigned to something like "Ctrl+Space" does not work, I think because this was line eating it:
https://github.com/aseprite/aseprite/blob/b3f4e37b69eb0ef907452624568f3bfc1f60d52d/src/app/ui/editor/state_with_wheel_behavior.cpp#L98-L101


dacap:
Thanks @hwabis 👍 we're going to include this fix in the next release.

PR #5134, opened 2025-04-19
Allow the app to be closed when a PopupWindow is open in the foreground

### LINKED ISSUE #5111: [Beta branch] Program does not close when Text Tool "..." menu is open
I was playing with the new font hinting feature (https://github.com/dacap/aseprite/commit/b3f4e37b69eb0ef907452624568f3bfc1f60d52d) and found that when trying to quit the program with the "X" on the Windows top bar while the Text Tool "..." menu is open, the prompt to save before quitting does not appear. I assume it is not intended because every other menu allows the save prompt to appear.

OS: Windows

ckaiser:
Adds a check when processing the `kCloseDisplayMessage` message that checks if the foreground window is a PopupWindow - specifically one that if interacted with would close (to preserve a way to make a fully modal popup window if necessary in future, but we could remove this if it's not desirable).

Fixes #5111

### DIFF
--- a/src/app/modules/gui.cpp
+++ b/src/app/modules/gui.cpp
@@ -481,18 +481,27 @@ bool CustomizedGuiManager::onProcessMessage(Message* msg)
 #endif
 
   switch (msg->type()) {
-    case kCloseDisplayMessage:
+    case kCloseDisplayMessage: {
       // Only call the exit command/close the app when the the main
       // display is the closed window in this kCloseDisplayMessage
-      // message and it's the current running foreground window.
-      if (msg->display() == this->display() &&
-          getForegroundWindow() == App::instance()->mainWindow()) {
+      // message and it's the current running foreground window, unless
+      // the foreground window is a popup that is expected to be closed.
+
+      auto* foregroundPopup = dynamic_cast<PopupWindow*>(getForegroundWindow());
+      if (foregroundPopup != nullptr &&
+          foregroundPopup->getClickBehavior() == PopupWindow::ClickBehavior::DoNothingOnClick)
+        break; // Valid PopupWindows that do nothing on click are still modal.
+
+      if (getForegroundWindow() != App::instance()->mainWindow() && !foregroundPopup)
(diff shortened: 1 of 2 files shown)


Gasparoken:
Thanks! It works ok. There is an issue when `UI with multiple window` is active: the Warning dialog is centered in the `PopUpWindow` instead the `MainWindow`.

I was looking at https://github.com/aseprite/aseprite/issues/5111 and the behavior intended by the artist is not the regular behavior according to the comment embedded in the code:
```
      // Only call the exit command/close the app when the the main
      // display is the closed window in this kCloseDisplayMessage
      // message and it's the current running foreground window.
```

Anyway, if this new way to allow closing Aseprite by using the X button the code could be simpler (but I'm sure I'm missing something):
```
    case kCloseDisplayMessage: {
      // Execute the "Exit" command.
      Command* command = Commands::instance()->byId(CommandId::Exit());
      UIContext::instance()->executeCommandFromMenuOrShortcut(command);
      return true;
    }
```

dacap:
> There is an issue when UI with multiple window is active: the Warning dialog is centered in the PopUpWindow instead the MainWindow.

👍 This is a big problem, the alert should be centered to the main window, not in the popup:

image

> Anyway, if this new way to allow closing Aseprite by using the X button, the code could be simpler and more general

Tested and both solutions (the one in this PR and the your code) have an issue where if we click outside the "save changes" alert, all the UI became unresponsible until we press the Esc key (quite sure the PopupWindow is filtering all mouse clicks because it wasn't properly closed yet).

I think clicking the X should:
1. behave exactly like clicking outside the hot region (closing the popup), and then
2. trigger the Exit command (both actions).

I'm not sure if a possible solution would be to filter the kCloseDisplayMessage too from PopupWindow (just like kMouseDownMessage, etc. are filtered now). So the `CustomizedGuiManager` doesn't have to deal with the specific details of the PopupWindow behavior.

dacap:
I'll take this issue, I've found a fix.

PR #5135, opened 2025-04-19
Re-add os.rename and os.remove support

### LINKED ISSUE #5107: In scripting, there is no* way to delete files or non-empty directories
I'm writing an export script, and the first step is to clear the old files before saving new ones. Overwriting isn't enough, because the file name can change based on the sprite contents.

Currently, **there is no\* way for an Aseprite script to delete files or non-empty directories.** Aseprite's `app.fs.removeDirectory()` only supports empty directories, and Lua's `os.remove()` is not supported by Aseprite.

\* The only exception is using `os.execute()` to delete files with a terminal command, but using `os.execute()` in any capacity causes Aseprite to **hang for approximately 5 seconds**, which severely hurts the workflow. This seemingly happens any time `os.execute()` is used (even with something like `echo hello world`) - it is not specific to removing a file/directory. (For reference, this is on Windows)

Ideally, there would be a way to **remove files and non-empty directories within an Aseprite script without using `os.execute()`**, such as allowing `app.fs.removeDirectory()` to work on non-empty directories or supporting `os.remove()`. Or `os.execute()`'s hanging issue could be fixed.

ckaiser:
Fixes #5107, I also fixed a typo and added a more specific string for when we're writing files vs "accessing" them, although with this there should also be specific strings for directories and in general the access prompt could do with a cleanup/refactor to make things a bit more clear.

The default Lua implementation also doesn't allow me to remove directories in general, could be worth it to replace it with something that supports just removing a non-empty directory (and prompt the user for it), but that's a more extensive implementation than just restoring the base Lua functionality.

### DIFF
--- a/src/app/script/security.cpp
+++ b/src/app/script/security.cpp
@@ -40,6 +40,8 @@ int secure_io_lines(lua_State* L);
 int secure_io_input(lua_State* L);
 int secure_io_output(lua_State* L);
 int secure_os_execute(lua_State* L);
+int secure_os_remove(lua_State* L);
+int secure_os_rename(lua_State* L);
 int secure_package_loadlib(lua_State* L);
 
 enum {
@@ -49,6 +51,8 @@ enum {
   io_input,
   io_output,
   os_execute,
+  os_remove,
+  os_rename,
   package_loadlib,
 };
 
@@ -64,6 +68,8 @@ static struct {
   { "io",      "input",   secure_io_input        },
   { "io",      "output",  secure_io_output       },
   { "os",      "execute", secure_os_execute      },
+  { "os",      "remove",  secure_os_remove       },
+  { "os",      "rename",  secure_os_rename       },
   { "package", "loadlib", secure_package_loadlib },
 };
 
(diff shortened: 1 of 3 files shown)


src/app/script/security.cpp
```diff
@@ -185,6 +191,24 @@ int secure_os_execute(lua_State* L)
   return replaced_functions[os_execute].origfunc(L);
 }
 
+int secure_os_remove(lua_State* L)
+{
+  const char* cmd = luaL_checkstring(L, 1);
```
martincapello, line 196: I would suggest changing this line for something like:

```suggestion
  std::string absFilename = base::get_absolute_path(luaL_checkstring(L, 1));
  lua_pop(L, 1);
  lua_pushstring(L, absFilename.c_str());
```
And then adjust the following lines to use the new variable.

This change is to improve the paths that are reported by errors when passing relative paths to the os.remove function.

These changes should be applied to the secure_os_rename() function as well.

martincapello (commented):
Looking good! The changes work as expected. I've suggested some changes just to improve how the functions behave in case of error when relative paths are passed to them.

src/app/script/security.cpp
ckaiser: Did the changes and noticed that for `os.rename` we'd have to check both source & destination path as well - additionally I added a QOL feature to `os.remove` (it was failing to delete empty directories with a "access denied" on Windows, and from googling it seems to be common that people just use `os.execute` but that's a whole different permission)

src/app/script/security.cpp
```diff
@@ -185,6 +191,46 @@ int secure_os_execute(lua_State* L)
...
+    }
+  }
+
+  return replaced_functions[os_remove].origfunc(L);
+}
+
+int secure_os_rename(lua_State* L)
+{
+  std::string absSourceFilename = base::get_absolute_path(luaL_checkstring(L, 1));
+  std::string absDestFilename = base::get_absolute_path(luaL_checkstring(L, 2));
```
martincapello, line 222: Nice!

martincapello (commented):
Almost there!

martincapello (approved):
LGTM! 🚀

dacap:
Oh sorry, I was testing the latest commit and it looks like the original `os.rename()` fails if the destination file exists (`os.rename` returns `nil, "File Exists"`):

```lua
local dialog = Dialog("rename")
dialog
  :entry{label="from", id="from"}
  :entry{label="to", id="to"}
  :button{text="rename", onclick=function()
            local data = dialog.data
            local a, b = os.rename(data.from, data.to)
            print(a, b)
         end}
  :show()
```

image

Also if the source file doesn't exist it fails anyway (but we are asking for access to the non-existent file):

image

Probably we could create some Lua tests to check the default/expected behavior of both of these functions (we cannot test the security dialog from scripts that are executed from the CLI, but probably we can do some tests).

After that we could add some tests with Unicode chars, and probably they will fail (mainly on Windows), so in some cases we'll need to replace the original impl with our own implementation of the function to support UTF-8 file names.

ckaiser:
Yeah I think the behavior as it exists now is a little broken, but also the default Lua behavior is a little broken and inconsistent itself, might be worth it to just use our file renaming and removing functions instead of relying on the Lua implementation, and just return what they would. It's what I was thinking of doing originally but I didn't want to over-complicate things but I think it's gonna be more complicated if we *don't* do the changes, especially wrt cross platform and unicode support.

ckaiser:
I ended up just implementing reasonable default functions instead of trying to rely in the specific Lua implementations, and added some simple tests.

dacap:
It looks like `os_remove` and `os_rename` returns 3 values in case of failure, could we just use `luaL_fileresult()` to fill the return values?

https://github.com/aseprite/lua/blob/2a00e6b0013f54ce80b6e3cefe6514e13229987a/loslib.c#L160-L170

E.g. of `os.remove()`:

```
~/t/luatest> lua
Lua 5.4.4  Copyright (C) 1994-2022 Lua.org, PUC-Rio
> os.remove("non-existent.txt")
nil	non-existent.txt: No such file or directory	2
```

src/app/script/security.cpp
```diff
@@ -185,6 +191,77 @@ int secure_os_execute(lua_State* L)
...
+      luaL_pushfail(L);
+      if (!absFilename.empty())
+        lua_pushfstring(L, "%s: %s", absFilename.c_str(), e.what());
+      else
+        lua_pushstring(L, e.what());
+      return 2;
+    }
+  }
+
+  base::delete_file(absFilename);
```
dacap: May throw a `std::runtime_error()` exception.

ckaiser:
Added my own helper since `luaL_fileresult()` uses `errorno` for the actual specific code, which is why avoided doing it in the first place but this makes it match now.

dacap:
Thanks for dedicating some extra time to see these error codes @ckaiser.

I saw that in Linux `ENOTEMPTY` is returned when `remove()` is called for a non-empty dir. But this number is 39 on Linux, and 41 on Windows: https://godbolt.org/z/a1vff5Pva And there is no way to access these values by name (`ENOTEMPTY`) from scripts. So they don't make too much sense at the moment.

I'll merge this PR :+1: and if in the future some users requires the error codes, we can add those constants.

PR #5140, opened 2025-04-23
Add a way to change default Timeline options

dacap:
Fixes https://community.aseprite.org/t/19863
Related to #5083 / https://community.aseprite.org/t/25097 and #1485 in some way, although we've opted to avoid moving these options to the Preferences dialog as it's quite a big refactor.

Image


Gasparoken:
It works great! LGTM. Since the timeline will change soon, I think your approach is ok and balanced (rather than making a big change right now).

dacap:
Merged in 206065fb80b46a2b623a6daaf80faf363b71169c

PR #5144, opened 2025-04-25
Allow to use the close button when a popup window/menu box is displayed

### LINKED ISSUE #5111: [Beta branch] Program does not close when Text Tool "..." menu is open

dacap:
Fix #5111, related to #5134

This fixes works even for menus (like ink or palette menus) or menus inside popups like the brush options:

image

### DIFF
--- a/src/ui/menu.cpp
+++ b/src/ui/menu.cpp
@@ -1,5 +1,5 @@
 // Aseprite UI Library
-// Copyright (C) 2018-2024  Igara Studio S.A.
+// Copyright (C) 2018-2025  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This file is released under the terms of the MIT license.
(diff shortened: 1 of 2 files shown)


src/ui/menu.cpp
```diff
@@ -487,6 +487,22 @@ bool MenuBox::onProcessMessage(Message* msg)
...
+    case kCloseDisplayMessage:
+      if (menu)
+        menu->closeAll();
+
+      // In case that this event is received when we are inside a
+      // modal loop to show this menu box
+      // (showPopup()/openWindowInForeground()) we redirect this
+      // kCloseDisplayMessage to the main window.
+      if (window()->isForeground()) {
+        auto* closeMsg2 = new Message(msg->type());
```
Gasparoken, line 499: Great! Just a minor aesthetic suggestion. You could change:
`auto* closeMsg2 = new Message(msg->type());`
to
`auto* closeMsg2 = new Message(kCloseDisplayMessage);`
Just to avoid the idea that `msg->type()` could have a value other than `kCloseDisplayMessage`.

Gasparoken (approved):

PR #5145, opened 2025-04-25
Fix marquee tool escape deselection

JoshuaL03:
This PR fixes an issue when pressing the escape key in the marquee tool. Instead of deselecting the selection, it should now undo the movement and match the functionality of the X button.


dacap:
Thanks @JoshuaL03, I think the best will be to merge this PR as it matches the tooltip (Esc = the `X` button). Anyway I'm worried that this behavior change might impact some current users, so I'll check if we can add an option (probably right-clicking the `X` button) to customize the behavior.

Issue #5164, opened 2025-05-12
New scripting event when a layer visibility (flag) is changed


dacap:
New event for [Sprite.events](https://www.aseprite.org/api/sprite#spriteevents) to listen when a layer visibility (or any layer flag?) is changed.

Arecher:
Was looking for this functionality, glad to find that it is planned!

PR #5165, opened 2025-05-13
Add "layerchange" event to Sprite, continuous event

ckaiser:
Implements #5164, first draft. Could use some feedback on the API, tried to keep it as simple (and consistent) as possible, so I added a DocObserver event for the continuous mode.

Adds a `'layerchange'` event to `Sprite` with a given `event.type` that can be one of `visibility, added, removed, editable, blendmode, collapsed, mergeddown, opacity, restacked, continuous`.

Some of my doubts:
 * Right now changing the layer visibility for all layers by clicking the header sends an event for each layer, but doing this for `editable` or `continuous` doesn't, should I go ahead and change the timeline/event code for that, or would that have any unintended consequences?
 * As commented in the code, right now we only `onBeforeLayerEditableChange`, do we want to add `onAfterLayerEditableChange` or just expose things as they are?

### DIFF
--- a/src/app/script/events_class.cpp
+++ b/src/app/script/events_class.cpp
@@ -358,6 +358,18 @@ private:
   obs::scoped_connection m_resizeConn;
 };
 
+// Layer Change events, TODO: Do we want this in a struct/script constant?
+const std::string kLayerChangeVisibility = "visibility";
+const std::string kLayerChangeAdded = "added";
+const std::string kLayerChangeRemoved = "removed";
+const std::string kLayerChangeEditable = "editable";
+const std::string kLayerChangeBlendMode = "blendmode";
+const std::string kLayerChangeCollapsed = "collapsed";
+const std::string kLayerChangeMergedDown = "mergeddown";
+const std::string kLayerChangeOpacity = "opacity";
+const std::string kLayerChangeRestacked = "restacked";
+const std::string kLayerChangeContinuous = "continuous";
+
 class SpriteEvents : public Events,
                      public DocUndoObserver,
                      public DocObserver {
@@ -370,6 +382,7 @@ public:
 #if ENABLE_REMAP_TILESET_EVENT
     RemapTileset,
 #endif
+    LayerChange,
   };
 
   SpriteEvents(const Sprite* sprite) : m_spriteId(sprite->id()) { doc()->add_observer(this); }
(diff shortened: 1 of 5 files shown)


src/app/script/events_class.cpp
```diff
@@ -358,6 +358,18 @@ class WindowEvents : public Events,
   obs::scoped_connection m_resizeConn;
 };
 
+// Layer Change events, TODO: Do we want this in a struct/script constant?
```
dacap, line 361: Although I'm not a big fan of the string type (I'd prefer an enum), I'm worried about having too many enums for events/subevents. I'm not sure what might be the best solution (one enum for all kind of observable events for any kind of doc object?).

ckaiser: These are only used right now on the function call, so it might just be enough to leave things like
```cpp
    call(LayerChange,
    {
           { "layer", ev.layer()  },
           { "type" , "collapsed" }
    });
```
and have it be well documented?

dacap:
Here some thoughts:
1. Generally I prefer enum for performance issues. A string must be always compared char by char to be matched, even for functions from plugins that were already compiled.
2. Anyway our Lua "enums" aren't converted to constants when scripts are compiled, so basically we're accessing the metatable field (specific enum name) to get the "constant" (metatable field) value (which might be pretty similar to accessing the string). So we don't have that performance gain.
3. As we are already using strings to identify an event names, I think it makes sense to continue using event names with strings.
4. The only advantage of our current event names in `Sprite.events:on/off()` is that the event name will be converted just once (or a couple of times) from a string to an internal C++ enum (`EventType`) using `strcmp()` (or we could use a map in the future).
5. In this PR, we are creating one new `Sprite` event (`LayerChange`) to handle a lot of new layer events, even if the user wants to listen only to the layer visibility changes, they must listen to all other layer events that are not interested in, and on each of those events they must ask/filter for the event subtype (e.g. `if ev.type == "visibility"` ...). And this is the main point why I don't like this implementation in particular.
6. We should reduce the scope of this PR to the issue, probably e.g. listening to visibility changes, opacity, and blendmode at the moment (probably name too?), and creating one EventType for each one, so we can just:
```
Sprite.events:on(‘layervisibility’, …)
Sprite.events:on(‘layerblendmode’, …)
Sprite.events:on(‘layeropacity’, …)
Sprite.events:on(‘layername’, …)
```
And include tests for all these events in `tests/scripts/events.lua`.

PR #5168, opened 2025-05-15
Fix user data not visible after resetting window sizes and placement

### LINKED ISSUE #4923: Reset window sizes and placement hides the enabled User Data
https://github.com/user-attachments/assets/21b9b6e6-3d8e-4fff-acfd-a8618a7b29ac

After resizing the windows, I have to double-click on the "User Data" button. I think that if User Data were enabled, there is no need to hide User Data.

JoshuaL03:
Fix #4923.

This PR fixes an issue where resetting window sizes and placement would revert the size of the layer properties window to its default even though the user data was expanded. Resetting the sizes in these conditions should now cause the layer properties window to open with the user data expanded.

### DIFF
--- a/src/app/commands/cmd_layer_properties.cpp
+++ b/src/app/commands/cmd_layer_properties.cpp
@@ -135,7 +135,15 @@ public:
 
     remapWindow();
     centerWindow();
+
+    gfx::Rect originalBounds = bounds();
+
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @JoshuaL03 for this fix, LGTM :+1:

PR #5171, opened 2025-05-17
Do not export every tag to JSON when there's one selected (#5085)

### LINKED ISSUE #5085: Tag metadata doesn't match frame metadata when ignoring certain frames
I needed to split my sprite sheet into two files so I added two tags dividing the frame range into a "primary" and "secondary" sheet. However the tag metadata in the JSON still gives frames relative to the full file, so the frame pack data will be offset by an amount not readily identified from the JSON data.

(In practice this can be mitigated by checking which tag has the same number of frames as entries in the frame array, but that also depends on the tags used for export having unique frame counts)

ckaiser:
Fixes #5085 by giving the exporter knowledge of a selected tag and filtering them there.

### DIFF
--- a/src/app/doc_exporter.cpp
+++ b/src/app/doc_exporter.cpp
@@ -1411,6 +1411,9 @@ void DocExporter::createDataFile(const Samples& samples, std::ostream& os, doc::
       includedSprites.insert(sprite->id());
 
       for (Tag* tag : sprite->tags()) {
+        if (!m_selectedTagName.empty() && m_selectedTagName != tag->name())
+          continue; // Ignore unselected tags if we picked a specific one.
+
         if (firstTag)
           firstTag = false;
         else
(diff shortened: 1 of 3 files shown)


src/app/doc_exporter.cpp
```diff
@@ -1411,6 +1411,9 @@ void DocExporter::createDataFile(const Samples& samples, std::ostream& os, doc::
       includedSprites.insert(sprite->id());
 
       for (Tag* tag : sprite->tags()) {
+        if (!m_selectedTagName.empty() && m_selectedTagName != tag->name())
+          continue; // Ignore unselected tags if we picked a specific one.
```
martincapello: Would you mind moving the comment above the if clause?

```suggestion
        // Ignore unselected tags if we picked a specific one.
        if (!m_selectedTagName.empty() && m_selectedTagName != tag->name())
          continue; 
```

src/app/doc_exporter.cpp
```diff
@@ -1421,12 +1424,21 @@ void DocExporter::createDataFile(const Samples& samples, std::ostream& os, doc::
           format = "{tag}";
         }
 
+        frame_t fromFrame = tag->fromFrame();
+        frame_t toFrame = tag->toFrame();
+        if (!m_selectedTagName.empty()) {
+          // Reset the frame offset when we're only exporting one frame tag so they're consistent
```
martincapello: The [same](https://github.com/aseprite/aseprite/pull/5171#discussion_r2135706982) for this comment.

martincapello (commented):
LGTM! Just a couple of nitpicks.

martincapello:
I've been testing this a bit more and realized that I'm not sure if this is working well for the case when the selected tag has subtags. I think I would expect the subtag be exported into the json data...but there is a catch, what if a subtag is only partly inside of the "parent" tag? Like this:

image

Here the `P+S` tag is covering one full tag and just part of another. Not sure how these cases should be handled. I guess we should include the full tag metadata in `frameTags`, but that is just my guessing.

dacap:
It would be nice to ask to @clydeshaffer (in #5085) how both sheets are being divided with the tag information. What is the input and the expected output, and add a test about it.

If the selected tag (primary/secondary) is used to divide a whole set of frames/subtags, probably we have to adjust all those subtags. But it would be nice to fix the user's case.

Issue #2606, opened 2021-01-04
Support for "Display file icon on thumbnail"


Nikoichu:
I love the beta thumbnailer, but I think this feature will make it even better by differentiating .ASE files from other image files that would open in a normal image viewer.

Basically, windows has this tickbox "Display file icon on thumbnail" in its folder options that shows small previews of the file type in the lower right corner of the thumbnail, so you know what extension the file has without reading its name (especially if you have extensions disabled). I searched everywhere for a way to turn it on for .ASE files, but it seems it can only be turned on by the thumbnailing shell extension (the aseprite thumbnailer, in this case, correct me if I'm wrong).

For examples, look at VLC's thumbnails (showing the little cone in the corner), .KRA files from Krita and the .SVG thumbnailer from Inkscape.

dacap:
We'll add some new options:

<img width="1274" height="603" alt="Image" src="https://github.com/user-attachments/assets/2596f66b-197a-4347-b7e4-6a2c103bd868" />

The Aseprite icon will be added to the thumbnails by default in the next version.

dacap:
Actually we'll have a new section for Windows-specific settings:

<img width="1299" height="936" alt="Image" src="https://github.com/user-attachments/assets/8e6db036-6764-455a-9a29-4b7871024718" />

PR #5174, opened 2025-05-20
Extension tests, add custom formats with saving and loading (#4950)

### LINKED ISSUE #4950: Load/save custom formats, e.g. `app.registerFileType{ filetypes="htm,html", onload=function..., onsave=function...}` that can call a callback function when a file of the given types are selected

ckaiser:
Implements #4950.

Test extension: 
[ppm-format.zip](https://github.com/user-attachments/files/25758964/ppm-format.zip)


src/app/commands/cmd_open_file.cpp
```diff
@@ -142,6 +145,17 @@ void OpenFileCommand::onExecute(Context* context)
       return;
 
     if (fop->hasError()) {
+// TODO: YIKES - add an error code to FOPs?
```
dacap: I think we have two options:
1. The simplest one: enable a flag after a failed `createLoadDocumentOperation()` where the operation can be delegated to a custom extension
2. `FileOp::createLoadDocumentOperation()` should return a valid operation to load the file. This will need a way to specify that the operation is executed in the same UI thread instead of a background thread. (And modify `OpenFileJob` to have both behaviors: run on a bg thread or run the FOP in the same UI thread.)

src/app/commands/cmd_save_file.cpp
```diff
@@ -222,11 +223,22 @@ void SaveFileBaseCommand::saveDocumentInBackground(const Context* context,
   if (!fop)
     return;
 
+  bool customFormat = false; // TODO: Unspaghettify
```
dacap: Same as in the load case, we might need a valid `FileOp::createSaveDocumentOperation()` with the plugin format.

src/app/commands/cmd_save_file.h
```diff
@@ -85,6 +88,13 @@ class SaveFileBaseCommand : public CommandWithNewParams<SaveFileParams> {
                                 const MarkAsSaved markAsSaved,
                                 const ResizeOnTheFly resizeOnTheFly = ResizeOnTheFly::Off,
                                 const gfx::PointF& scale = gfx::PointF(1.0, 1.0));
+#ifdef ENABLE_SCRIPTING
+  bool saveCustomFormat(const Context* context,
```
dacap: Probably
```suggestion
  bool savePluginFormat(const Context* context,
```
but not sure if we need this function anymore (if we create a valid FOP with the plugin callback).

src/app/extensions.cpp
```diff
@@ -358,6 +363,28 @@ void Extension::addMenuSeparator(ui::Widget* widget)
   m_plugin.items.push_back(item);
 }
 
+void Extension::addCustomFormat(const CustomFormat& format)
```
dacap: Replace "custom" with "plugin file" (and don't use "plugin" word if we are inside an extension):
```suggestion
void Extension::addFileFormat(const FileFormat& format)
```

src/app/extensions.cpp
```diff
@@ -358,6 +363,28 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+{
+  // Check for any duplicates in global extension list for both ID and file extensions.
+  for (const auto& extension : App::instance()->extensions()) {
+    if (m_customFormats.count(format.id) > 0)
+      throw base::Exception("Duplicated format ID: " + format.id);
+
+    for (const auto& [id, other] : extension->m_customFormats) {
+      for (const auto& otherExt : other.extensions) {
+        for (const auto& ourExt : format.extensions) {
+          if (base::string_to_lower(otherExt) == base::string_to_lower(ourExt))
```
dacap: Store extensions in lower case.

src/app/extensions.cpp
```diff
@@ -769,6 +796,148 @@ void Extension::addScript(const std::string& fn)
...
+      base::move_file(destTempFile, filename);
+      if (!mv.empty())
+        base::delete_file(mv);
+
+      return true;
+    }
+    catch (std::exception& e) {
+      Console().printf("Saving failed: '%s'", e.what());
+      return false;
+    }
```
dacap: Is this necessary? What we are exactly doing here?

The data recovery/backup feature already makes a backup of the files in case the plugin doesn't save everything, probably that should be enough.

src/app/extensions.h
```diff
@@ -83,6 +91,34 @@ class Extension {
     ThemeInfo(const std::string& path, const std::string& variant) : path(path), variant(variant) {}
   };
 
+#ifdef ENABLE_SCRIPTING
+  struct CustomFormat {
```
dacap:
```suggestion
  struct FileFormat {
```
or
```suggestion
  struct PluginFileFormat {
```

src/app/extensions.h
```diff
@@ -83,6 +91,34 @@ class Extension {
...
+    bool supports(const Support support) const
+    {
+      if (support == Load && onloadRef > -1)
+        return true;
+      if (support == Save && onsaveRef > -1)
+        return true;
+      return support == Full && onloadRef > -1 && onsaveRef > -1;
+    }
+  };
+  struct CustomFormatSaveOptions {
```
dacap:
```suggestion
  struct FileFormatSaveOptions {
```

src/app/extensions.h
```diff
@@ -122,6 +158,7 @@ class Extension {
   void removeMenuGroup(const std::string& id);
 
   void addMenuSeparator(ui::Widget* widget);
+  void addCustomFormat(const CustomFormat& format);
```
dacap:
```suggestion
  void addFileFormat(const FileFormat& format);
```

src/app/extensions.h
```diff
@@ -137,6 +174,15 @@ class Extension {
 #ifdef ENABLE_SCRIPTING
   bool hasScripts() const { return !m_plugin.scripts.empty(); }
   void addScript(const std::string& fn);
+  bool hasCustomFormats() const { return !m_customFormats.empty(); }
```
dacap:
```suggestion
  bool hasFileFormats() const { return !m_fileFormats.empty(); }
```

src/app/script/plugin_class.cpp
```diff
@@ -300,6 +300,63 @@ int Plugin_newMenuSeparator(lua_State* L)
   return 0;
 }
 
+int Plugin_registerFormat(lua_State* L)
```
dacap:
```suggestion
int Plugin_newFileFormat(lua_State* L)
```

src/app/script/plugin_class.cpp
```diff
@@ -353,6 +410,7 @@ const luaL_Reg Plugin_methods[] = {
   { "newMenuGroup",     Plugin_newMenuGroup     },
   { "deleteMenuGroup",  Plugin_deleteMenuGroup  },
   { "newMenuSeparator", Plugin_newMenuSeparator },
+  { "registerFormat",   Plugin_registerFormat   },
```
dacap:
```suggestion
  { "newFileFormat",    Plugin_newFileFormat   },
```

src/app/script/sprite_class.cpp
```diff
@@ -909,6 +909,20 @@ int Sprite_set_filename(lua_State* L)
   return 0;
 }
 
+int Sprite_set_isModified(lua_State* L)
+{
+  auto sprite = get_docobj<Sprite>(L, 1);
+  const bool isModified = lua_toboolean(L, 2);
+  Doc* doc = static_cast<Doc*>(sprite->document());
+  if (doc && !isModified)
+    doc->markAsSaved();
```
dacap: This should be done by the FOP code instead of leaving this to the script, e.g. if the `onsave` function doesn't fail, it means that the file was saved successfully, but if we have just exported the file to this plugin file format, we don't have to mark the sprite as saved. For example see

https://github.com/aseprite/aseprite/blob/3c350c3e671fe5755a9ca098a13e3bd26b86bf5b/src/app/commands/cmd_save_file.cpp#L249-L253

The sprite is marked as saved only when "markAsSaved" flag is `MarkAsSaved::On` (which is not the case for `SaveFileCopyAsCommand`).

dacap (changes requested):

src/app/extensions.cpp
ckaiser: So my original thinking for this was as an additional layer of protection when interacting with existing files, because extensions are user code and error handling is kind of sparse, I only wanted the actual file saves to be performed if everything went well on the extension side, hence the double-gloving we do here.

I can get rid of it, I'm not 100% sure of how the data recovery feature integrates with how we're handling saving with custom formats but if it's enough then this is definitely redundant.

ckaiser:
Had another go at this and implementing the suggestions, it's a bit cleaner now, still have some TODOs to look over though but I wanted to get feedback on the save protection stuff and the general idea.

One thing that's nagging at me right now as well is that as it stands now, the script itself is in charge of creating the Sprite itself, and that seems like something we might want to avoid. Originally I left it like that because with adding isModified and having the script have to set the filename, we could leave it open for scripts to open multiple sprites from a single file, but if that's not behavior we intend to support, then I think creating the sprite behind the scenes and then passing it through an argument to the script so it can manipulate the pixels and then we can mark it as saved/etc is gonna be the way to go. 

Trying to think if there's any use cases that this way of doing things might make impossible though.

ZeWaka:
Perhaps the new `File Explorer -> File Associations` menu added in #2606 is relevant? An API to add into there would be wonderful.

dacap:
There is an assert failing when running `ctest` in debug mode and gui available:
```
    Start 61: keyboard_shortcuts_tests
1/2 Test #61: keyboard_shortcuts_tests .........Subprocess aborted***Exception:   0.17 sec
FT: Loading font ... FAIL
FT: Loading font ... FAIL
src/app/extensions.cpp:995: Assertion failed: !item.widget->parent()

    Start 62: app_tests
2/2 Test #62: app_tests ........................Subprocess aborted***Exception:   0.15 sec
[==========] Running 3 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 3 tests from App
[ RUN      ] App.JustClose
FT: Loading font ... FAIL
FT: Loading font ... FAIL
src/app/extensions.cpp:995: Assertion failed: !item.widget->parent()
```

dacap:
About the API, I think we should be able to specify an `extension` argument to `newFileFormat()`, and avoid `name` and `extensions` if the user wish (probably `extension` or `extensions` could be used indifferently as a string or as a table, accepting both types and both argument names). The `name` by default could be the first extension specified.

src/app/extensions.cpp
```diff
@@ -371,6 +373,227 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+public:
+  explicit ExtensionCustomFormat(const Extension::CustomFormatDefinition& formatDefinition)
+    : m_definition(formatDefinition)
+  {
+    m_dioFormat = dio::register_custom_format();
+
+    for (const auto& ext : formatDefinition.extensions) {
+      if (!dio::register_custom_format_extension(ext, m_dioFormat))
+        throw base::Exception("Extension %s is already registered to a file format", ext.c_str());
+    }
```
dacap: I think some tests in Linux could be failing because this `formatDefinition.extensions` being a `std::unordered_set` and the order file formats are being registered to the list of formats in `ExtensionCustomFormat()`. Probably removing the `"format20"` from the conflicting `CustomFormatConflictExt` test could work, anyway I think we should solve how to handle these cases: What we should do if an extension register a new file format for two extensions, one is conflict but the other one is not.

If we'd like to avoid this case at all, we should first check that there are no conflicting extension at all, and then start calling `dio::register_custom_format_extension()`.

dacap: After further investigation, although there is an issue in `dio::register_custom_format_extension()` invocation for some formats and other not when an exception is thrown from `ExtensionCustomFormat()`. I think I've found what is the problem with the test failing on Linux.

`Extensions::Extensions()` uses `base::list_files()` to register extensions, so the order of extensions is given by what the file system is returning. Probably we should sort the list of files/extensions, e.g.:

```c++
Extensions::Extensions()
{
...
    auto list = base::list_files(extensionsDir, base::ItemType::Directories);
    std::sort(list.begin(), list.end());
    for (const auto& fn : list) {
      ...
```

In a future we should order the list by some kind of dependency order (in case we support dependencies between extensions).

ZeWaka:
> About the API, I think we should be able to specify an `extension` argument to `newFileFormat()`, and avoid `name` and `extensions` if the user wish (probably `extension` or `extensions` could be used indifferently as a string or as a table, accepting both types and both argument names). The `name` by default could be the first extension specified.

So there'd be both an `extension` and an `extensions` argument? Seems a bit confusing.

I do agree that you could allow `extensions` to accept a string for the simple case of one extension, though. Making name optional also makes sense.

ckaiser:
Fixed the test issues, although I couldn't reproduce the assert issue you were having @dacap, does this happen in a fresh profile? There could be an installed extension causing, maybe this check causes problems?

I've made `name` optional, and I changed away from `unoredered_set` for the extension list. Originally my idea was to use it just to get duplicate removal for free, but to make `name` optional I wanted to pick a stable name, so I might as well just use a standard vector and do a one-time check when registering the extension.

I also added the possibility to use a string or table in the `extensions` parameter and an `extension` parameter that takes a string (you can mix and match if... you wanna be weird about it)

Lastly, I removed the starting "." in some of the error messages regarding formats, so that they can be descriptive instead of just extension based, since this could potentially cause confusion (even with the aseprite format, if you were saving as `.aseprite`, the error message could mention `.ase`, or with `.jpg`/`.jpeg`. This could also help us eventually to have more descriptive names for formats which is helpful when you start introducing some extension-based ones.
Example:
image

dacap:
> There could be an installed extension causing, maybe this check causes problems?

You were right, it was related to this installed extensions: https://github.com/zarstensen/aselapse

Not sure if the test will always fail with an installed extension (or just with this extension), but probably we should be thinking about how to run these tests without the assert failing.

> I've made `name` optional, 

I have to check other `Plugin:new...` functions, and see if we should use `id` or `name`.

> I also added the possibility to use a string or table in the `extensions` parameter and an `extension` parameter that takes a string (you can mix and match if... you wanna be weird about it)

:+1: 

> Lastly, I removed the starting "." in some of the error messages [...] This could also help us eventually to have more descriptive names for formats which is helpful when you start introducing some extension-based ones. 

I'd include the file extension introduced by the user anyway in the error message in those cases, like: "Portable Pixman Format" (.ppm)

I'll continue reviewing this PR later.

ckaiser:
Moved the arguments for `onload` and `onsave` to a single table and added clarification for formats in the error windows. ([updated the extension for it](https://github.com/user-attachments/files/25758964/ppm-format.zip))

As for `id` vs `name`, I think the latter is better since it's not really an identifier and it's not used elsewhere, it's just to show the user and it's optional.

dacap:
I'm testing this right now, saw that the list of file formats is not sorted :eyes: 

image

ckaiser:
@dacap Fixed the sorting issue (did not realize we were using the FileFormatsManager order), and fixed the assert, the signals were not getting connected because AppMenus is initialized after extensions so we didn't have the instance available, changed that to a lambda instead of direct connection.

src/dio/file_format.h
```diff
@@ -12,31 +12,40 @@
...
-  PAL_PALETTE,
-  PCX_IMAGE,
-  PNG_IMAGE,
-  SVG_IMAGE,
-  TARGA_IMAGE,
-  WEBP_ANIMATION,
-  CSS_STYLE,
-  PSD_IMAGE,
-  QOI_IMAGE,
+  ERROR = 0,
```
dacap: I don't remember why this change `ERROR=-1` -> `0` was needed. Are we using FileFormat as an index in an array?

src/dio/file_format.h
```diff
@@ -12,31 +12,40 @@
...
+  PNG_IMAGE = 15,
+  SVG_IMAGE = 16,
+  TARGA_IMAGE = 17,
+  WEBP_ANIMATION = 18,
+  CSS_STYLE = 19,
+  PSD_IMAGE = 20,
+  QOI_IMAGE = 21,
+
+  FIRST_CUSTOM = 22,
+  LAST_CUSTOM = 0x7fffffff
```
dacap: I'd prefer to avoid specifying the number of each item and probably only leave `LAST_CUSTOM = 0x7fffffff` (all other items can be as they were).

In case you want to specify `FIRST_CUSTOM`, we could use something like:
```
  FIRST_CUSTOM = 1000,
  LAST_CUSTOM = 9999,
```

dacap (changes requested):
Sorry that I'm reviewing this in chunks, but I wasn't able to review all tests yet, it will take some time.

src/app/script/plugin_class.cpp
```diff
@@ -18,6 +18,8 @@
 #include "app/ui/app_menuitem.h"
 #include "base/version.h"
 
+#include <base/trim_string.h>
+
```
dacap: Remove these lines:
```suggestion
```

src/app/script/plugin_class.cpp
```diff
@@ -331,6 +333,83 @@ int Plugin_newMenuSeparator(lua_State* L)
...
+    format.extensions.push_back(base::string_to_lower(lua_tostring(L, -1)));
+  }
+  else if (type != LUA_TNIL)
+    return luaL_error(L, "format parameter \"extensions\" must be a list or string");
+  lua_pop(L, 1);
+
+  type = lua_getfield(L, 2, "extension");
+  if (type == LUA_TSTRING)
+    format.extensions.push_back(base::string_to_lower(lua_tostring(L, -1)));
+  lua_pop(L, 1);
```
dacap: I was thinking in something like:
```suggestion
  for (const char* field : { "extensions", "extension" }) {
    type = lua_getfield(L, 2, field);
    if (type == LUA_TTABLE) {
      lua_pushnil(L);
      while (lua_next(L, -2) != 0) {
        format.extensions.push_back(base::string_to_lower(luaL_tolstring(L, -1, nullptr)));
        lua_pop(L, 2);
      }
    }
    else if (type == LUA_TSTRING) {
      format.extensions.push_back(base::string_to_lower(lua_tostring(L, -1)));
    }
    else if (type != LUA_TNIL)
      return luaL_error(L, "format parameter \"%s\" must be a list or string", field);
    lua_pop(L, 1);
  }
```
In the docs we'll reference that you can use
```lua
  extension=string,
  extensions=table,
```
but in reality anything can be used to avoid errors.

src/dio/detect_format.cpp
```diff
@@ -26,8 +26,27 @@
...
 
+#include <unordered_map>
+
+static std::unordered_map<std::string, dio::FileFormat> g_customFormats;
+
 namespace dio {
 
+bool register_custom_format_extension(const std::string& ext, const FileFormat format)
+{
+  if (detect_format_by_file_extension("file." + ext) != FileFormat::UNKNOWN)
```
dacap, line 37: A minor thing, but probably we can add a `detect_format_by_extension()` and pass the `ext` directly. And implement `detect_format_by_file_extension()` using this new `detect_format_by_extension()`.

src/app/extensions.cpp
```diff
@@ -25,7 +25,14 @@
 #include "base/file_handle.h"
 #include "base/fs.h"
 #include "base/fstream_path.h"
+#include "context.h"
```
dacap: Always use the full path to include our `.h` files:
```suggestion
#include "app/context.h"
```

src/app/extensions.cpp
```diff
@@ -25,7 +25,14 @@
...
+#include "dio/detect_format.h"
+#include "doc_undo.h"
```
dacap:
```suggestion
#include "app/doc.h"
#include "app/doc_undo.h"
```

src/app/extensions.cpp
```diff
@@ -40,26 +47,31 @@
 
 #include "archive.h"
 #include "archive_entry.h"
+#include "doc.h"
```
dacap: Don't mix our header files with third party files:
```suggestion
```

src/app/extensions.cpp
```diff
@@ -40,26 +47,31 @@
...
 #include "json11.hpp"
 
 #include <cctype>
 #include <fstream>
-#include <queue>
 #include <sstream>
 #include <string>
 
 #include "base/log.h"
```
dacap: This came from other commit, but please move this `base/log.h` at the top with our header files.

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
 
+class ExtensionCustomFormat : public FileFormat {
+public:
+  explicit ExtensionCustomFormat(const Extension::CustomFormatDefinition& formatDefinition)
+    : m_definition(formatDefinition)
+  {
+    m_dioFormat = dio::register_custom_format();
+
+    for (const auto& ext : formatDefinition.extensions) {
+      if (!dio::register_custom_format_extension(ext, m_dioFormat))
```
dacap: What happens if a plugin registers one file format but fails with another?

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+        throw base::Exception("Extension %s is already registered to a file format", ext.c_str());
+    }
+
+    if (formatDefinition.flags > 0)
+      m_flags = formatDefinition.flags;
+    else // TODO: What should the defaults be?
+      m_flags = FILE_SUPPORT_RGB | FILE_SUPPORT_RGBA | FILE_SUPPORT_GRAY | FILE_SUPPORT_GRAYA |
+                FILE_SUPPORT_INDEXED;
```
dacap: Please add some extra `{}` when there are multiple lines:
```suggestion
    else {
      // TODO: What should the defaults be?
      m_flags = FILE_SUPPORT_RGB | FILE_SUPPORT_RGBA | FILE_SUPPORT_GRAY | FILE_SUPPORT_GRAYA |
                FILE_SUPPORT_INDEXED;
    }
```

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+    for (auto* doc : docsToRemove) {
+      App::instance()->context()->documents().remove(doc);
+      delete doc;
+    }
+    App::instance()->context()->setActiveDocument(m_originalActiveDoc);
+  }
+
+  bool onLoad(FileOp* fop) override
+  {
+    script::Engine* engine = App::instance()->scriptEngine();
```
dacap: Cache `App::instance()` in a local variable (instead of calling `App::instance()` multiple times in this function, same in `onSave`):
```suggestion
    App* app = App::instance();
    script::Engine* engine = app->scriptEngine();
```

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+
+    lua_pushvalue(L, -1);
+    const int fileHandle = luaL_ref(L, LUA_REGISTRYINDEX);
+
+    lua_newtable(L);
+    lua_pushstring(L, fop->filename().c_str());
+    lua_setfield(L, -2, "filename");
+    lua_pushvalue(L, -2);
+    lua_setfield(L, -2, "file");
+    lua_remove(L, -2);
```
dacap, line 497: Add a comment that we are creating a table here like:
```suggestion
    // Push the onload(ev) argument where ev = { filename, file }
    lua_newtable(L);
    lua_pushstring(L, fop->filename().c_str());
    lua_setfield(L, -2, "filename");
    lua_pushvalue(L, -2);
    lua_setfield(L, -2, "file");
    lua_remove(L, -2);
```

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+      return false;
+    }
+
+    auto* doc = static_cast<Doc*>(sprite->document());
+    doc->resetUndoHistory();
+    doc->setFilename(fop->filename());
+    doc->markAsSaved();
+    fop->setDocument(doc);
+    return true;
+  };
```
dacap: Remove `;`
```suggestion
  }
```

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+    for (auto* doc : App::instance()->context()->documents())
+      m_originalDocs.push_back(doc);
+
+    if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
+      fop->setError("Failed to open file '%s' for reading, %s\n",
+                    fop->filename().c_str(),
+                    lua_tostring(L, -1));
+      revertDocState();
+      return false;
+    }
```
dacap, line 486: This `lua_pcall()` to the original `io.open()` asks for one return value, but we are not `lua_pop()`ing it in the return path.
E.g.
```c++
    if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
      fop->setError("Failed to open file '%s' for reading, %s\n",
                    fop->filename().c_str(),
                    lua_tostring(L, -1));
      lua_pop(L, 1);
      revertDocState();
      return false;
    }
```

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+
+    auto* doc = static_cast<Doc*>(sprite->document());
+    doc->resetUndoHistory();
+    doc->setFilename(fop->filename());
+    doc->markAsSaved();
+    fop->setDocument(doc);
+    return true;
```
dacap, line 526: We have to pop the sprite from the stack:

```suggestion

    lua_pop(L, 1);              // Pop the sprite
    return true;
```

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+      if (const char* s = lua_tostring(L, -1))
+        App::instance()->scriptEngine()->consolePrint(s);
+      Console().printf("Could not save format '%s' to destination '%s'",
+                       m_definition.name.c_str(),
+                       fop->filename().c_str());
+      return false;
+    }
+
+    if (!lua_toboolean(L, -1))
+      return false;
```
dacap: We must pop the returned value that we requested from `lua_pcall()` above:
```suggestion
    if (!lua_toboolean(L, -1)) {
      lua_pop(L, 1);
      return false;
    }
    lua_pop(L, 1);
```

src/app/extensions.cpp
```diff
@@ -371,6 +372,239 @@ void Extension::addMenuSeparator(ui::Widget* widget)
...
+    const auto call = lua_pcall(L, 1, 1, 0);
+    closeFileHandle(L, fileHandle);
+
+    if (call != LUA_OK) {
+      if (const char* s = lua_tostring(L, -1))
+        fop->setError(s);
+      else
+        fop->setError("Format error");
+      revertDocState();
+      return false;
```
dacap, line 509: Missing pop:
```suggestion
      lua_pop(L, 1);
      revertDocState();
      return false;
```

src/app/extensions_tests.cpp
```diff
@@ -0,0 +1,865 @@
+// Aseprite
+// Copyright (C) 2025  Igara Studio S.A.
+// Copyright (C) 2001-2016  David Capello
```
dacap: Probably just
```suggestion
// Copyright (C) 2025-present  Igara Studio S.A.
```

dacap (changes requested):
I wasn't able to check all tests yet, but I've found several fixes that are required in onLoad/onSave mainly (and probably more tests are needed for coverage).

The main thing is that the Lua stack is not kept intact after a `onLoad/onSave()` on every path (happy path and exception paths). You can check with a helper like this:

```c++

  struct CheckLStack {
    lua_State* L;
    int top;
    CheckLStack(lua_State* L) : L(L), top(lua_gettop(L)) {}
    ~CheckLStack() {
      PRINTARGS("top", top, "lua_gettop", lua_gettop(L));
      ASSERT(top == lua_gettop(L));
    }
  };

  bool onLoad(FileOp* fop) override
  {
    script::Engine* engine = App::instance()->scriptEngine();
    lua_State* L = engine->luaState();
    CheckLStack check(L);
...
```

src/app/extensions.cpp
ckaiser: It would register the extensions to the`dio` format but the format object itself in FileFormatsManager would not be created or inserted into the list so when it came time to check for any formats that supported the extension it would assign the extension to the `dio::FileFormat::FIRST_CUSTOM+n` but then fail with a normal "aseprite can't load this file" error.

I've added code to unregister any successful extensions if we fail with one. We could also make it so that it only fails when no extensions manage to get registered , but that could be messy.

ckaiser:
Fixed all the the stack issues (I just kinda forgot about that huh, maybe we could add a debug-only macro with this stack-checking helper and sprinkle it all over the place so that I don't make that mistake again 🙈).

I've added some more tests for the format registration function and for general completeness and especially failure paths.

dacap:
> maybe we could add a debug-only macro with this stack-checking helper and sprinkle it all over the place so that I don't make that mistake again 🙈).

Yeah, I was thinking the same (in the past I tried but avoid it because it's too much work for functions that return values). I'll add something after merging this PR.

dacap:
Looks great! :clinking_glasses:

Issue #2324, opened 2020-04-02
Modify last command parameters


dacap:
There are several commands that could have special parameters to re-execute them with different user preferences (even with real-time preview in the editor). For example (this is an incomplete list):

- [ ] After a color mode conversion (e.g. RGB -> Indexed) we could select other dithering methods
- [ ] After pasting an image from the clipboard there should be some extra options of how to merge both images, e.g. if the image in the clipboard is RGB and we are pasting into an indexed image, we could choose to add the missing colors into the new image, or the quantization method, etc.
- [ ] After opening an image (and a sequence of images is detected to load the entire animation) show an option to load the animation (this would remove this dialog completely and fix #2321)

These parameters could appear in the context bar, or below the context bar at the right side, like a mini-popup which should disappear when other command is executed or the user start drawing.

PR #5184, opened 2025-05-28
Copy tags when copying frames from one sprite to another (#4979)

### LINKED ISSUE #4979: Copy tags when copy frames from one sprite to another
It'd be nice to copy tags when we copy frames from one file to another:

Image

But we should include options to choose if we want to keep those tags or just copy the frames (#2324)

ckaiser:
Fixes [#4979](https://github.com/aseprite/aseprite/issues/4979).

https://github.com/user-attachments/assets/333db7b2-a6e2-4620-829a-2e73b3152c34

There's still some edge-cases where things could get duplicated - and there's no way to disable the copying of tags, that'd come as a part of #2324 which I could tackle next.

I also took the opportunity to localize some of the transaction names, but I could put those in another PR with the rest of them, since most of them (but not all) are not translated.

### DIFF
--- a/src/app/doc_api.cpp
+++ b/src/app/doc_api.cpp
@@ -51,6 +51,7 @@
 #include "app/transaction.h"
 #include "app/util/autocrop.h"
 #include "app/util/layer_utils.h"
+#include "cmd/set_user_data.h"
 #include "doc/algorithm/flip_image.h"
 #include "doc/algorithm/shrink_bounds.h"
 #include "doc/cel.h"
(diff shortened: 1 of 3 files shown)


dacap:
I tried the "Resolve conflict" feature of GitHub PRs and it made merge with main (we prefer rebases). I'll review this PR and if it works rebase it to main (fixing the conflict).

dacap:
In this case it duplicated the tag for each frame:

https://github.com/user-attachments/assets/91d37826-0622-428e-af01-f48ba5186403

ckaiser:
Did some more workarounds to avoid duplicating tags, since the only other way would be to add some kind of uuid to them, but I _think_ this covers most cases?

dacap:
Needs to be rebased (e.g. https://github.com/dacap/aseprite/tree/copy-tags), needs more info in the commit (fix number of issue), and it still needs more testing to avoid a situation where several tags are "multiplied" (part 0:10):

https://github.com/user-attachments/assets/fcefd23d-55ae-4575-ada2-5ba22f89ede5

dacap:
It will conflict with https://github.com/aseprite/aseprite/pull/5888, probably will try to merge that one first, and move this one to beta.

ckaiser:
Rebased, improved the copying code to reduce the duplicates although it still feels a bit brittle especially when pasting things over each other, but I'm not sure how much we can improve that with the tag/frame information we have rn 🤔

PR #5185, opened 2025-05-29
Fix: Prevent 1px-shift in scrollbar when clicking without dragging

LukeBarcenas:
### Summary
Fixes a bug where clicking on the scrollbar without moving the mouse would cause the bar to shift by 1 pixel. This issue occurred because drag logic was triggered immediately upon a mouse click, even if no mouse movement was detected.

### Description of Changes
Introduced a new static flag 'm_dragging' to ensure that drag behavior only begins after mouse movement is detected. The logic checks whether the mouse position has changed since the initial click and skips scroll calculations if it hasn't.

### Effect on Application
Prevents unintended 1px-scroll shifts when interacting with scrollbars.

### DIFF
--- a/src/ui/scroll_bar.cpp
+++ b/src/ui/scroll_bar.cpp
@@ -23,6 +23,7 @@ using namespace gfx;
 // two scroll-bars at the same time).
 int ScrollBar::m_wherepos = 0;
 int ScrollBar::m_whereclick = 0;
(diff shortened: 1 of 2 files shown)


dacap:
Hi @LukeBarcenas, thanks a lot for this fix! I'll merge it right now 🤝

PR #5192, opened 2025-06-03
Fix gif files to apply global palette in particular cases

Gasparoken:
This fix was prompted by Discord users who noticed that certain GIFs saved with Aseprite and uploaded to Discord chats resulted in animations whose colors changed unintentionally throughout the animation. It was discovered that Discord does not correctly process GIFs with disposal=DO_NOT_DISPOSE + global palette + local palettes.

This fix essentially defines a global palette in all cases where the animation (with Color Mode RGBA/GRAYSCALE) can be made with up to 256 global colors. This fixes simple animations from a Discord perspective.

On the other hand, all other cases (color count > 256) continue to be processed as before and may present issues in Discord.


martincapello (approved):
LGTM! However it seems that Discord fixed how they display gif files, so this change is not actually needed, but seems like a nice optimization to introduce anyway.

dacap:
@Gasparoken should we still merge this fix? it would be nice to create some kind of test suite for the gif encoder/decoder from https://github.com/aseprite/test-images and add some options to the encoder so we might be able to introduce new gif options to the encoder.

PR #5202, opened 2025-06-07
Fix file dialog size not being loaded properly

### LINKED ISSUE #4460: Size of the file dialog window
### The feature request
Each time you save a file, the dialog window that opens is positioned at the last saved coordinates, but each time at the default size. It would be cool to also save the size of the dialog window so that it is always displayed where it should be. (As it can be done, for example, with the image preview window during drawing).

Bit annoying
Looks good, but...

Perfect
May be perfect

* Aseprite version:  Steam [beta] Aseprite 1.3.6-arm64
* System: macOS Sonoma

JoshuaL03:
Fix #4460.

This PR fixes an issue where the file selector dialog would not load its size in some cases when using the Aseprite file dialog. Since the minimum size set on the dialog is fairly large, the minimum limit on the loaded size has been removed to allow for users to adjust the size more to their liking.

### DIFF
--- a/src/app/ui/file_selector.cpp
+++ b/src/app/ui/file_selector.cpp
@@ -408,7 +408,8 @@ bool FileSelector::show(const std::string& title,
 
   remapWindow();
   centerWindow();
-  load_window_pos(this, kConfigSection);
+  // The minimum size is large, so do not limit the minimum loaded size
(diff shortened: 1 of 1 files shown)


dacap:
Thanks @JoshuaL03, it looks like an issue when multiple windows option is disabled. I'll merge this fix :+1:

PR #5222, opened 2025-06-11
Implement tool loop for selection tools

### LINKED ISSUE #2852: Improve selection tools feedback

Liebranca:
This adds a new tool loop for use with selection tools (and slice, too): it simply makes it so one can see their strokes no matter the zoom level. Fix #2852.

For testing purposes (comparing the old tool loop to the new one), I've added a bool to toggle between them: holding down the keys for a 'change brush size' action while in standby state with a selection tool will enable the old implementation, and 'deselect all' re-enables the new one.

I am unsure whether it would be best to use a solid color for drawing the mask or a marching ants pattern, and which color or colors to use for it. There are many possibilities with this, could be worth it to make it customizable later on. I was tempted to make the colors depend on whether one is adding, subtracting, replacing or intersecting, but I held off on it as it seemed like too much.

Anyway, for now I've made it so the new implementation looks almost identical to the original, so as to not overcomplicate things. I've left a few additional switches to quickly swap between some alternatives, again just for testing purposes:
- With `kReplaceSelection` modifier (default), the mask is drawn over the ink as a solid black, so it *almost* looks the same as the old implementation.
- With `kAddSelection` modifier, only the mask is used.
- With `kSubtractSelection` modifier, the mask is drawn over the ink using a marching ants pattern.
- With `kIntersectSelection` modifier, only the mask is used, and drawk with a marching ants pattern.

Side note: I think that for the defaut case the mask could be disabled when zoomed in, as it's not necessary, and activated only when zoomed out and the ink becomes less visible. I didn't add this just to keep things simple, but it's fairly easy to do.

Finally, there's only a few minor details left to handle:
- When selecting, the mask can draw over the right and bottom edges of the canvas, and these are not always cleared, so the mask leaves a 'trail' over them. I couldn't find a way to redraw these two edges, save for forcing an `updateEditor()`, but I can tell that's not quite right. Needless to say, this is actually unnoticeable if the mask is the same color as the edges.
- Because I needed to access the mask from both the tool loop instance and the editor itself, I wasn't entirely sure where it was best to store it. I've simply added it to the editor class for now, in the new `m_selectionToolMask` member variable, but maybe there's a more appropriate place for it.
- In order to avoid code duplication, I made a *second* base class for tool loops, from which the old and new implementation are derived. This one essentially just holds all the methods and member variables that I am certain are needed by both, however, I haven't yet removed logic specific to slice/selection tools from the old implementation: this is so it can still be used for testing/comparison purposes.
- Following from the last point, this second base class would have to be slightly reworked if the old implementation is to be stripped of logic specific to slice/selection tools. But this is mostly just moving code from the base class to the derived one.

### DIFF
--- a/src/app/doc.h
+++ b/src/app/doc.h
@@ -195,6 +195,7 @@ public:
   const MaskBoundaries& maskBoundaries() const { return m_maskBoundaries; }
 
   MaskBoundaries& maskBoundaries() { return m_maskBoundaries; }
+  void setMaskBoundaries(const MaskBoundaries& segs) { m_maskBoundaries = segs; }
 
   bool hasMaskBoundaries() const { return !m_maskBoundaries.isEmpty(); }
 
--- a/src/app/tools/inks.h
+++ b/src/app/tools/inks.h
@@ -292,8 +292,15 @@ public:
       m_maxBounds |= rc;
     else {
       rc &= loop->getDstImage()->bounds();
-      for (int v = rc.y; v < rc.y2(); ++v)
-        BaseInk::inkHline(rc.x, v, rc.x2() - 1, loop);
+      if (loop->isSelectionToolLoop())
+        loop->addSelectionToolPoint(rc);
+
+      // NOTE: this condition is here for drawing mode switches, remove/rework after testing
+      if (!loop->isSelectionToolLoop() ||
+          (loop->getPointShape()->isFloodFill() || !loop->getPreviewFilled())) {
+        for (int v = rc.y; v < rc.y2(); ++v)
+          BaseInk::inkHline(rc.x, v, rc.x2() - 1, loop);
+      }
     }
   }
 
@@ -457,8 +464,15 @@ public:
     }
     else {
       rc &= loop->getDstImage()->bounds();
-      for (int v = rc.y; v < rc.y2(); ++v)
-        BaseInk::inkHline(rc.x, v, rc.x2() - 1, loop);
+      if (loop->isSelectionToolLoop())
+        loop->addSelectionToolPoint(rc);
+
+      // NOTE: this condition is here for drawing mode switches, remove/rework after testing
+      if (!loop->isSelectionToolLoop() ||
+          (loop->getPointShape()->isFloodFill() || !loop->getPreviewFilled())) {
+        for (int v = rc.y; v < rc.y2(); ++v)
+          BaseInk::inkHline(rc.x, v, rc.x2() - 1, loop);
+      }
     }
   }
 
--- a/src/app/tools/tool_loop.h
+++ b/src/app/tools/tool_loop.h
@@ -264,6 +264,11 @@ public:
   virtual void onSliceRect(const gfx::Rect& bounds) = 0;
 
   virtual const app::TiledModeHelper& getTiledModeHelper() = 0;
+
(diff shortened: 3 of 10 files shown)


dacap:
> This adds a new tool loop for use with selection tools (and slice, too): it simply makes it so one can see their strokes no matter the zoom level. Fix #2852.

Hi @Liebranca, good first draft! :+1:

> For testing purposes (comparing the old tool loop to the new one), I've added a bool to toggle between them: holding down the keys for a 'change brush size' action while in standby state with a selection tool will enable the old implementation, and 'deselect all' re-enables the new one.

As I said in the email, probably we can add a new experimental flag for this, and thinking a little more, probably we should make the option available in *Edit > Preferences > Experimental*, it's highly probable that some users might prefer to keep the current behavior anyway.

> I am unsure whether it would be best to use a solid color for drawing the mask or a marching ants pattern, and which color or colors to use for it.

As a first implementation I'd use a static marching ant pattern with the same black & white (like the `kIntersectSelection` implementation but with black & white colors).

Anyway something *important*: We shouldn't use the ExpandCelCanvas in this new implementation, we should paint the feedback directly on the editor. ExpandCelCanvas creates a cel of the full sprite size and it's something we want to avoid for selection tools. Should improve the performance a lot when we select big images.

> Side note: I think that for the defaut case the mask could be disabled when zoomed in, as it's not necessary, and activated only when zoomed out and the ink becomes less visible. I didn't add this just to keep things simple, but it's fairly easy to do.

As we should remove the `ExpandCelCanvas`, the ink feedback will be removed and only the mask should be visible.

> * When selecting, the mask can draw over the right and bottom edges of the canvas, and these are not always cleared, so the mask leaves a 'trail' over them.

I didn't see this in detail, I think `Editor::drawBackground()` should do the job, although we can fix this later.

> * Because I needed to access the mask from both the tool loop instance and the editor itself, I wasn't entirely sure where it was best to store it. I've simply added it to the editor class for now, in the new `m_selectionToolMask` member variable, but maybe there's a more appropriate place for it.

We can have this variable, probably this could be a static member variable shared between all editors (and nullptr by default, only valid when a `SelectionToolLoopImpl` is alive).

> * In order to avoid code duplication, I made a _second_ base class for tool loops, from which the old and new implementation are derived. This one essentially just holds all the methods and member variables that I am certain are needed by both, however, I haven't yet removed logic specific to slice/selection tools from the old implementation: this is so it can still be used for testing/comparison purposes.

I think `m_expandCelCanvas` should be part of `ToolLoopImpl` only.

> * Following from the last point, this second base class would have to be slightly reworked if the old implementation is to be stripped of logic specific to slice/selection tools. But this is mostly just moving code from the base class to the derived one.

At the moment (and probably for several versions) we will keep the old implementation working (so the user can disable this new experimental implementation, that if it works ok, should be enabled by default in future versions).

dacap:
I think it's a pretty good first implementation to go forward. I'll merge it as it is in the `beta` branch and do some minor changes to the code after that.

PR #5224, opened 2025-06-12
Add five new types of diffusion dither and expose zig zag to user

dpt:
Screenshot 2025-06-12 at 15 47 37

This extends the existing dithering code to add Jarvis-Judice-Ninke, Stucki, Atkinson, Burkes and Sierra error diffused dithers. It also exposes the zig zag (serpentine) choice to the user. Diffusion is tweaked to use fixed point for the error values.


dpt:
Spotted that serpentine mode wasn't flipping the matrix as it should have - fixed.
We now avoid copying error buffers around when just changing index will do.
Checking continueTask per-pixel seemed excessive so I've made that less frequent now - per line.

dpt:
Spotted a problem where dither settings are being lost when OK is hit... [now fixed]

src/app/cli/cli_processor.cpp
```diff
@@ -481,7 +491,7 @@ int CliProcessor::process(Context* ctx)
               case render::DitheringAlgorithm::None:    params.set("dithering", "none"); break;
               case render::DitheringAlgorithm::Ordered: params.set("dithering", "ordered"); break;
               case render::DitheringAlgorithm::Old:     params.set("dithering", "old"); break;
-              case render::DitheringAlgorithm::ErrorDiffusion:
+              case render::DitheringAlgorithm::FloydSteinberg:
                 params.set("dithering", "error-diffusion");
                 break;
```
dpt: Oops, missed some cases here.

dpt: Fixed

Issue #4875, opened 2024-12-10
Update minimum platform to macOS 10.14.6


dacap:
We've found that Aseprite v1.3 is being used only since macOS 10.14.6.

Older Aseprite versions (like v1.2.8 / v1.2.17 / v1.2.25) are used in macOS 10.13.2 or 10.12.6. Aseprite v1.1.5.7 on macOS 10.11.5. There are some cases of v1.2.9 on macOS 10.10.5. But none on macOS 10.9 which is our default target platform.

We have several limitations using some C++ features, quoting our [coding style](https://github.com/aseprite/aseprite/blob/main/docs/CODING_STYLE.md) guide:

> * Use `std::optional` but taking care of some limitations from macOS 10.9:
>   * Use `std::optional::has_value()` instead of `std::optional::operator bool()` ([example](https://github.com/aseprite/laf/commit/81622fcbb9e4a0edc14a02250c387bd6fa878708))
>   * Use `std::optional::operator*()` instead of `std::optional::value()` ([example](https://github.com/aseprite/aseprite/commit/4471dab289cdd45762155ce0b16472e95a7f8642))
> * Use `std::variant` but taking care of some limitations from macOS 10.9:
>   * Use `T* p = std::get_if<T>(&value)` instead of `T v = std::get<T>(value)` or
>     create an auxiliary `get_value()` using `std::get_if` function ([example](https://github.com/aseprite/aseprite/commit/dc0e57728ae2b10cd8365ff0a50263daa8fcc9ac#diff-a59e14240d83bffc2ea917d7ddd7b2762576b0e9ab49bf823ba1a89c653ff978R98))
>   * Don't use `std::visit()`, use some alternative with switch-case and the `std::variant::index()` ([example](https://github.com/aseprite/aseprite/commit/574f58375332bb80ce5572fdedb1028617786e45))
> * Use `std::any` but taking care of some limitations from macOS 10.9:
>   * Use `T* p = std::any_cast<T>(&value)` instead of `T v = std::any_cast<T>(value)` ([example](https://github.com/aseprite/aseprite/commit/c8d4c60f07df27590381ef28001a40f8f785f50e))

So it'd be great to update this a little further.

dacap:
It looks like Skia m139 dropped support for macOS 10.15 (making macOS 11 as the minimum platform): https://skia.googlesource.com/skia/+/main/RELEASE_NOTES.md#milestone-139

[...]

PR #5237, opened 2025-06-19
Fix thumbnails for macOS Sequoia (fix #5009)

### LINKED ISSUE #5009: Fix thumbnails for macOS Sequoia
Use new thumbnail API on macOS: https://developer.apple.com/documentation/quicklookthumbnailing?language=objc

Gasparoken:
This fix is ​​preliminary.
- The background mask color has been replaced with white in thumbnails.
- The "distribution" code needs to be completed to integrate AsepriteThumbnailer into the bundle.
- Feedback is welcome.

fix #5009


src/desktop/osx/appex/thumbnails.mm
```diff
@@ -1,22 +1,18 @@
...
 #include "dio/decode_file.h"
 #include "dio/file_interface.h"
 #include "doc/sprite.h"
 #include "render/render.h"
 
 #include <Cocoa/Cocoa.h>
-
-#include <algorithm>
-
-namespace desktop {
```
martincapello, line 19: Not sure why this namespace got removed.

martincapello (commented):
Looking good so far! I'm just wondering about thumbnail support for older macOS versions (<= 10.14), because AFAIK this implementation is for macOS >= 10.15, right? But I'm not sure which is the oldest macOS supported by Aseprite...because in the INSTALL.md file I see `-DCMAKE_OSX_DEPLOYMENT_TARGET=10.9` but in Steam it says that macOS 10.15 is the minimum supported version.

dacap:
> I'm just wondering about thumbnail support for older macOS versions (<= 10.14)

For older macOS versions we can just remove the thumbnail support. In a near future we should upgrade at least to macOS 10.14 as minimum platform anyway: #4875

dacap:
Finally fixed with #5314

Issue #1338, opened 2016-12-01
Undo undoes rotation pivot point


kyrieru:
If you select something, move the pivot, and then rotate,
Pressing undo will undo the rotation _and_ the pivot. 
It would be nice if undoing back to a selection state also saved what the pivot was at that time.

1.1.9
portable

dacap:
Related to #424 and #1459

PR #5241, opened 2025-06-23
Fix do not move the pivot when dragging a selection with Alt key to snap to grid (fix #5238)

### LINKED ISSUE #5238: Do not move the pivot when dragging a selection with Alt key to snap to grid
Reported here: https://bsky.app/profile/torcado.com/post/3ls3q2y2tbc2s

This GIF shows a first drag of the selection without pressing the Alt key, and then pressing the Alt key (to snap to grid):

Image

Gasparoken:
fix #5238


Gasparoken:
I found some issues that could be improved:
- Visually, the pivot doesn't update during the scaling change (without modifier pressed) (#5242).
- When performing an "undo" after an selection move with a modified pivot position, the pivot change is undone first, then the selection move command is undone, when clearly the order should be reversed (#1338).

PR #5247, opened 2025-06-24
Recover Windows Vista compatibility

### LINKED ISSUE #5211: Broken compatibility with Windows 7
Probably something changed in MSVC, or Sentry, or some other dependency (or just the CRT impl), but it looks like now we depend on [CreateFile2](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfile2) which was introduced on Windows 8.

This is strange because we explicitly compile with Windows Vista support (`_WIN32_WINNT=0x0A00 WINVER=0x0A00`), but probably this is due some dependency.

We can find similar issues on GitHub: https://github.com/search?type=issues&q=CreateFile2

Internal report: https://igarastudio.zendesk.com/agent/tickets/7700

martincapello:
Fix #5211


dacap:
Yesterday I was able to test this in a Win7 VM (32-bit) and it worked 👍

PR #5258, opened 2025-07-01
Fix convert_image_to_surface function (fix #5257)

### LINKED ISSUE #5257: Wrong results with "convert_image_to_surface" when transparency is used
The following 'copy' routine resulted in incorrect colors when pixels have transparency in`convert_image_to_surface`
https://github.com/aseprite/aseprite/blob/3c350c3e671fe5755a9ca098a13e3bd26b86bf5b/src/app/util/conversion_to_surface.cpp#L259-L263

This affects timeline thumbnails and possibly other thumbnails used in Aseprite.

The next replacement works ok:
```cpp
        for (int y=0; y<h; y++) {
          for (int x=0; x<w; x++) {
            surface->putPixel(image->getPixel(x, y), x, y);
          }
        }
```

Aseprite 1.3.14.3 version

Gasparoken:


src/app/util/conversion_to_surface.cpp
```diff
@@ -250,6 +250,9 @@ void convert_image_to_surface(const doc::Image* image,
   os::SurfaceLock lockDst(surface);
   os::SurfaceFormatData fd;
   surface->getFormat(&fd);
+  // Needed because Skia surfaces work with premultiplied alpha and
+  // here we need to copy unpremultiplied alpha RGB values to the Skia surface.
+  ((os::SkiaSurface*)surface)->bitmap().setAlphaType(kUnpremul_SkAlphaType);
```
dacap, line 256: We need to enclose this code between `LAF_SKIA` so the compilation doesn't fail when `LAF_BACKEND=none`:
```suggestion

#if LAF_SKIA
  // Needed because Skia surfaces work with premultiplied alpha and
  // here we need to copy unpremultiplied alpha RGB values to the Skia surface.
  ((os::SkiaSurface*)surface)->bitmap().setAlphaType(kUnpremul_SkAlphaType);
#endif
```

PR #5264, opened 2025-07-03
Fix session 'isRunning' detection (fix #5252)

### LINKED ISSUE #5252: Sometimes files available for recovery aren't displayed
Sometimes an artist cannot recover its own files.

This may be due to a method that is not sufficiently deterministic to distinguish the running Aseprite instance from a previously launched Aseprite instance using the process ID (PID).
It turns out that the probability that the PID of a closed Aseprite instance matches the PID number of a subsequent Aseprite run is not zero, nor even low enough to leave it as is.

Report with a workaround: https://igarastudio.zendesk.com/agent/tickets/7783
When you go to the _Home tab > Recover Files > Refresh_ and nothing appears, it's recommended to:
1. Back up the desired session folder.
2. Go to the desired session folder.
3. Change the number in the `pid` file and save it.
4. Go to the _Home tab > Recover Files > Refresh_

Aseprite version 1.3.14.2

Gasparoken:
Note: The following function will be deprecated:
https://github.com/aseprite/aseprite/blob/b4555fc09876753fc4e74ac6957cce3c9b71b19b/src/app/crash/session.cpp#L131

fix #5252

### DIFF
--- a/src/app/crash/data_recovery.cpp
+++ b/src/app/crash/data_recovery.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2023  Igara Studio S.A.
+// Copyright (C) 2019-2025  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -139,6 +139,13 @@ DataRecovery::Sessions DataRecovery::sessions()
   return copy;
 }
 
+bool DataRecovery::isRunningSession(const SessionPtr& session) const
+{
+  ASSERT(session);
+  ASSERT(m_inProgress);
+  return session->path() == m_inProgress->path();
+}
+
 void DataRecovery::searchForSessions()
 {
   Sessions sessions;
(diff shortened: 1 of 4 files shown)


src/app/crash/session.h
```diff
@@ -47,6 +47,7 @@ class Session {
 
   std::string name() const;
   std::string version();
+  std::string path() const;
```
dacap: It can be defined as
```suggestion
  const std::string& path() const { return m_path; }
```
because this returns the same value of `m_path` (other member functions are calculating the return value).

src/app/crash/data_recovery.cpp
```diff
@@ -139,6 +140,12 @@ DataRecovery::Sessions DataRecovery::sessions()
   return copy;
 }
 
+bool DataRecovery::isRunningSession(const SessionPtr& session) const
+{
+  ASSERT(session);
+  return session->path() == m_currentSessionPath;
+}
```
dacap: Probably we can use `m_inProgress`:
```suggestion
bool DataRecovery::isRunningSession(const SessionPtr& session) const
{
  ASSERT(session);
  ASSERT(m_inProgress);
  return session->path() == m_inProgress->path();
}
```
and remove `m_currentSessionPath`.

src/app/crash/session.h
```diff
@@ -47,9 +47,9 @@ class Session {
 
   std::string name() const;
   std::string version();
+  std::string& path() { return m_path; };
```
dacap:
```suggestion
  std::string& path() { return m_path; }
```

PR #5269, opened 2025-07-11
Add 'Show hidden' check to file selector

### LINKED ISSUE #3079: Hard to navigate to dot-prefixed folders in file dialogs
File dialogs ("open", "save as", etc) don't show `.`-prefixed directories.
It seems, the only way to enter such a directory is to type its name from memory (as the file name), and press the open/save/... button.
Would be great to have a way to toggle the visibility of such directories (e.g. a checkbox in the dialogs).

* Aseprite version: v1.2.30-x64, Steam
* System: Xubuntu 20.04.3

Liebranca:
Adds a checkbox to control showing hidden files in file selector window. Fix #3079.

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -356,6 +356,7 @@
     <section id="file_selector">
       <option id="current_folder" type="std::string" default="&quot;&lt;empty&gt;&quot;" />
       <option id="zoom" type="double" default="1.0" />
+      <option id="show_hidden" type="bool" default="false" />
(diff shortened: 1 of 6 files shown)


dacap:
It works great @Liebranca :+1: I'll merge it as it is and then fix some issue on Windows to handle special hidden files.

Later we'll think in an alternative icon for this button, but I'm not sure what to use yet.

PR #5290, opened 2025-07-22
Fix slice transformations not updating editors

### LINKED ISSUE #5225: Preview window (and other windows) don't refresh when we're transforming slice pixels
Reported: https://community.aseprite.org/t/preview-doesnt-update-slicing-transformation/25733

Image

ckaiser:
Fixes #5225.

It's kind of an "inelegant" solution, I first attempted to use the bounds and `notifySpritePixelsModified` but it always ended up with some parts of the image not updating correctly, and since we were originally doing a full `invalidate()` I figured just using the equivalent of that but for all editors would be fine.


dacap:
LGTM! :+1:

PR #5292, opened 2025-07-22
Fix crash deselecting moved pixels when using certain extensions

### LINKED ISSUE #5280: Crash using timelapse extension after deselecting
There is a crash using the https://github.com/luciopaiva/aseprite-timelapse extension after deselecting. Probably the extension needs some patch, but we should avoid the crash from our side.

Internal: Issue discovered in ticket 7779 https://igarastudio.zendesk.com/agent/tickets/7779
Reproduced for Sentry: https://igara-studio.sentry.io/issues/6751083536/events/ad13620b4d3042dc8f474ceef4183d1d/
Reported in the same extension: https://github.com/luciopaiva/aseprite-timelapse/issues/1

dacap:
Fix #5280


Gasparoken:
It works great with Selections. The same should be added for `WritingTextState::onLeaveState` which has the same problem.

dacap:
> It works great with Selections. The same should be added for `WritingTextState::onLeaveState` which has the same problem.

Nice catch, it looks like `MovingSelectionState::onLeaveState` needs the same too.

PR #5293, opened 2025-07-22
Fix thumbnails for macOS Sequoia (fix #5009)

### LINKED ISSUE #5009: Fix thumbnails for macOS Sequoia

Gasparoken:
fix #5009


Gasparoken:
Pending: The background mask color is white in the thumbnails preview.

dacap:
I saw that this PR includes the `Info.plist` file from our build scripts with the `$VERSION` constant that is replaced in our build process. I'll move this file to `src/main/` (just like `resources_win32.rc` file for Windows) and this PR can update the file with the required properties. I personally don't like to have this file that is about Aseprite app inside the extension specifics.

src/CMakeLists.txt
```diff
@@ -187,6 +183,10 @@ if(ENABLE_ASEPRITE_EXE)
   add_executable(${main_target}
     main/main.cpp
     ${main_resources})
+
+  if(ENABLE_DESKTOP_INTEGRATION)
+    add_subdirectory(desktop)
+  endif()
```
dacap, line 189: It looks like this was moved here to use `add_dependencies(${main_target} ${extension_target})` but the main target (the `aseprite` binary) doesn't depend on the thumbnailer. The Aseprite.app depends on the thumbnailer, but our cmake files don't build the .app bundle (that's why the Info.plist/entitlements files are not included in the repo).

I'll include these files (Info.plist/entitlements) in the repo, and probably we could make a target to build the Aseprite.app (so users can build their local .app bundle), but this dependency/change shouldn't be required (the thumbnailer is built explicitly with `ninja AsepriteThumbnailer.qlgenerator` currently)

Issue #5205, opened 2025-06-10
Add an API for scripts that access the number of undo history


xiankuang:
I developed a script for creating a painting process GIF similar to Procreate's replay feature. However, there is no API to access the number of undo history, which results in having to manually input the number of frames for exporting the GIF.

PR #5294, opened 2025-07-23
[lua] Add sprite.undoHistory

ckaiser:
Implements #5205. ~I'm not 100% convinced by the API names, alternatively we can use something like `undoStatesCount` or call them `undoSteps` (or stepCount) -- I also figured `Site` was the place to put this but it could be moved without much issue as long as we have document access, it makes sense to me since it's a level above the sprite and scripting has no concept of a "document".~

### DIFF
--- a/tests/scripts/sprite.lua
+++ b/tests/scripts/sprite.lua
@@ -228,3 +228,69 @@ do
   c = app.open(fn)
   assert(c.tileManagementPlugin == nil)
 end
+
+-- Undo History
+
+function test_undo_history()
+  local sprite = Sprite(1, 1)
+
+  assert(sprite.undoHistory.undoSteps == 0)
+  assert(sprite.undoHistory.redoSteps == 0)
+
+  sprite:resize(10, 10)
+
+  assert(sprite.undoHistory.undoSteps == 1)
+  assert(sprite.undoHistory.redoSteps == 0)
+
+  sprite:resize(10, 15)
+
+  assert(sprite.undoHistory.undoSteps == 2)
+  assert(sprite.undoHistory.redoSteps == 0)
+
+  sprite:resize(10, 30)
+
+  assert(sprite.undoHistory.undoSteps == 3)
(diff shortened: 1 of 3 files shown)


dacap:
> Implements #5205, I'm not 100% convinced by the API names, alternatively we can use something like `undoStatesCount` or call them `undoSteps` (or stepCount)

I think `undoHistory.undoSteps` and `undoHistory.redoSteps` might be a good option.

In a future we could have `undoHistory.undoStates` (or `undoHistory.states`) for a future API to get/move to a specific position/state in the undo history. But that requires several new objects to make it useful.

> I also figured `Site` was the place to put this but it could be moved without much issue as long as we have document access, it makes sense to me since it's a level above the sprite and scripting has no concept of a "document".

`Site` is like a snapshot of the editor/timeline position in a specific time (`app.site` the current site), it's not updated when it's stored. Actually the best position for the undoHistory is the Sprite. So we can do `sprite.undoHistory`.

In some way the internal `app::Doc` + `doc::Sprite` is represented by the scripting `Sprite` (check `sprite_class.cpp` functions to see how the document is obtained from the sprite).

ckaiser:
Renamed and moved to Sprite

tests/scripts/sprite.lua
```diff
@@ -228,3 +228,49 @@ do
...
+
+  app.undo()
+  app.undo()
+
+  assert(sprite.undoHistory.undoSteps == 0)
+  assert(sprite.undoHistory.redoSteps == 3)
+
+  sprite:resize(10, 30)
+
+  assert(sprite.undoHistory.undoSteps == 1)
```
dacap: LGTM, there is only a little issue in this line if we enable `app.preferences.undo.allow_nonlinear_history = true`, where it fails. We might convert this `do/end` block in a function to test both options (`allow_nonlinear_history = false` and `true`), the expected value here is different (as redo info is not deleted and the action is just added at the end).

dacap:
Looks great! 👍

Issue #3952, opened 2023-07-06
Review exception/error handling


dacap:
There are several situations where we would like to avoid exceptions and try to use just a `bool` return value to check if an operation failed or not (because the operation is quite common to fail, so there is nothing exceptional about it).

This include locking/unlocking context/document access, removing files, parsing files, etc.

Probably related to #528

dacap:
To read https://abseil.io/tips/76

dacap:
To watch:
* https://www.youtube.com/watch?v=BGmzMuSDt-Y
* https://www.youtube.com/watch?v=LorcxyJ9zr4

PR #5303, opened 2025-07-27
[WIP] Command refresh

### LINKED ISSUE #3025: Add ui parameter to all commands to disable dialog/user interaction on each command
For scripting purposes we should be able to call every command from the Lua API with:

```lua
app.command.CommandName{ ui=false }
```

to run it without UI / user interaction.

Related issues:
* https://github.com/aseprite/api/issues/44

ckaiser:
Fixes #3025. Companion PR to [api/#124](https://github.com/aseprite/api/pull/124)

I've gone through all of the commands testing them in batch and UI mode and this PR is the resulting cleanup which:

 * Makes them behave consistently and return `false` in `onEnabled` when they will not work in batch mode, instead of failing silently or crashing.
 * Adds `ui` parameters to `LoadMask`, `LoadPalette`, `ModifySelection`, `PaletteSize`, `SaveMask` and `SavePalette`.
 * Modified other commands to respond better to not having the UI available.
 * Removes the `CmdUIOnlyFlag` and `CmdRecordableFlag` flags, which did nothing.

Now that `Command::onEnabled` is more consistent, I think we should consider adding a way for scripts to know that the command they attempted to execute was disabled by something, and ideally also for a way for commands to report errors. Right now, the command return value doesn't really do much.

One way to do this would be adding something like `app.getLastCommandError()` or something like that, that we could populate after every execution? Maybe with some common error codes or strings, that way scripts could communicate in their own way what errors they stumble into.


src/app/commands/cmd_frame_properties.cpp
```diff
@@ -62,15 +61,19 @@ void FramePropertiesCommand::onLoadParams(const Params& params)
...
   }
-  else {
-    m_target = SPECIFIC_FRAME;
-    m_frame = frame_t(base::convert_to<int>(frame));
-  }
+
+  const int frameNumber = frame_t(base::convert_to<int>(frame));
+  // Will open with the default target (CURRENT_RANGE) if the frame number cannot be parsed.
+  if (frameNumber < 1)
+    return;
```
dacap: We've to set the values of the member variables after `onLoadParams()`, as the same instance is used each time the command is used, i.e. there is one instance of each command in the program, and the same instance is used (onLoadParams -> onExecute, etc.) for the same command execution. So the constructor of `FrameProperties` is called just once at the very beginning of Aseprite.

With this change, if we do:

```lua
app.command.FrameProperties{ frame="all" }
app.command.FrameProperties()
```
The second call will do the same (show the Frame properties for all frames).

In this case I'd recommend you to keep the code as it was or use the `has_param()` to check if the `frame` argument exists. If it does exist, we show the properties of the specific frame, in other case the current frame (this will change the behavior of `app.command.FrameProperties()` but I think it does make sense to show the current frame properties by default).

Remember: we must initialize/set all member variables in `onLoadParams()`.

src/app/commands/cmd_launch.cpp
```diff
@@ -34,18 +34,15 @@ class LaunchCommand : public Command {
...
+LaunchCommand::LaunchCommand() : Command(CommandId::Launch()), m_type(Url), m_path("")
 {
 }
 
 void LaunchCommand::onLoadParams(const Params& params)
 {
   m_path = params.get("path");
 
-  if (m_type == Url && !m_path.empty() && m_path[0] == '/') {
+  if (m_type == Url && !m_path.empty() && m_path[0] == '/' && !base::is_absolute_path(m_path)) {
```
dacap: In this case `m_path` is an URL and `base::is_absolute_path()` works for file paths (not URLs).

src/app/commands/cmd_load_mask.cpp
```diff
@@ -36,13 +37,14 @@ class LoadMaskCommand : public Command {
...
-LoadMaskCommand::LoadMaskCommand() : Command(CommandId::LoadMask(), CmdRecordableFlag)
+LoadMaskCommand::LoadMaskCommand() : Command(CommandId::LoadMask())
 {
-  m_filename = "";
 }
 
 void LoadMaskCommand::onLoadParams(const Params& params)
 {
+  if (params.has_param("ui"))
+    m_ui = params.get_as<bool>("ui");
```
dacap, line 47: Set `m_ui` value to `true` if it's not present (review other commands).

dacap (changes requested):
I didn't make a full review, but just some comments so far in some changes I was able to see.

dacap:
> One way to do this would be adding something like `app.getLastCommandError()` or something like that, that we could populate after every execution? Maybe with some common error codes or strings, that way scripts could communicate in their own way what errors they stumble into.

Actually [Command_call](https://github.com/aseprite/aseprite/blob/cef92c1a3810c867ded72dad59e03543feb5460f/src/app/script/app_command_object.cpp#L55) should [return false](https://github.com/aseprite/aseprite/blob/cef92c1a3810c867ded72dad59e03543feb5460f/src/app/script/app_command_object.cpp#L61) if the command fails ([an exception is thrown](https://github.com/aseprite/aseprite/blob/cef92c1a3810c867ded72dad59e03543feb5460f/src/app/context.cpp#L223)).

Anyway there is not too much information about the error, and for regular errors (non-exceptional, like the UI is not available) we could try to create a new method to return command results (probably related to #3952).

I'd leave this error handling for the future (not in this PR).

src/app/commands/cmd_frame_properties.cpp
ckaiser: Ohh, I was not aware of the command instance reuse, will do another pass then!

src/app/commands/cmd_launch.cpp
ckaiser: That's why I'm using it here, since Launch can also work with file paths, if it detects that it's an absolute path it doesn't append the aseprite.org url. It can't really happen on windows because of C:// but on Mac/Linux if I want to open a file by path, It'd instead open the aseprite website with a path. 
Now granted, I did this fix before I did the other "Open" command, might just be a matter of making sure this one is strictly only for URLs?

ckaiser:
Did the changes, I'm also gonna see about documenting the fact that we reuse instances because it can definitely be a gotcha for script authors as well.

> Actually [Command_call](https://github.com/aseprite/aseprite/blob/cef92c1a3810c867ded72dad59e03543feb5460f/src/app/script/app_command_object.cpp#L55) should [return false](https://github.com/aseprite/aseprite/blob/cef92c1a3810c867ded72dad59e03543feb5460f/src/app/script/app_command_object.cpp#L61) if the command fails ([an exception is thrown](https://github.com/aseprite/aseprite/blob/cef92c1a3810c867ded72dad59e03543feb5460f/src/app/context.cpp#L223)).

An exception yes, but the command being disabled still returns `true` when the command hasn't been run, which I think we should change especially now that it's more consistently indicating stuff.

Edit: Added `CommandResult::kDisabled` and in testing it seems to work well for this kind of thing, let me know what you think!

src/app/commands/cmd_launch.cpp
dacap: At the moment the `Launch` command is only for URLs (see `m_type == Url`), and `OpenWithApp` is used to open files with the associated app in the OS (And `OpenInFolder` to open the associated folder). It's used in the Help menu mainly. Probably the name should be `OpenWebsite`, but that's too late (there are future plans to add aliases for commands, but for backward compatibility we should keep the `Launch` name too).

I'd say this change should be reverted as it is confusing to use the URL with file-related functions.

dacap:
> I'm also gonna see about documenting the fact that we reuse instances because it can definitely be a gotcha for script authors as well.

I think this internal detail shouldn't be documented in the API. Shouldn't affect the behavior of scripts, at least commands should behave like an action in itself that doesn't depend on previous calls of the same command.

> An exception yes, but the command being disabled still returns `true` when the command hasn't been run, which I think we should change especially now that it's more consistently indicating stuff.
> 
> Edit: Added `CommandResult::kDisabled` and in testing it seems to work well for this kind of thing, let me know what you think!

Yeah, we can add a new kind of result code 👍  Anyway we might need some way to check if the command is enabled before executing it, e.g.:

```lua
if app.command.SaveFile.enabled then
  ...
end
```

ckaiser:
Added the `.enabled` property and added some more `isUIEnabled` checks.

~Edit: Looking into the test failures.~

src/app/commands/cmd_goto_layer.cpp
```diff
@@ -88,12 +84,12 @@ class GotoLayerCommand : public Command {
 
 class GotoPreviousLayerCommand : public GotoLayerCommand {
 public:
-  GotoPreviousLayerCommand() : GotoLayerCommand(-1, "GotoPreviousLayer", CmdUIOnlyFlag) {}
+  GotoPreviousLayerCommand() : GotoLayerCommand(-1, "GotoPreviousLayer") {}
```
dacap: We can remove the string literal here too:
```suggestion
  GotoPreviousLayerCommand() : GotoLayerCommand(-1, CommandId::GotoPreviousLayer()) {}
```

src/app/commands/cmd_goto_layer.cpp
```diff
@@ -88,12 +84,12 @@ class GotoLayerCommand : public Command {
...
 };
 
 class GotoNextLayerCommand : public GotoLayerCommand {
 public:
-  GotoNextLayerCommand() : GotoLayerCommand(+1, "GotoNextLayer", CmdUIOnlyFlag) {}
+  GotoNextLayerCommand() : GotoLayerCommand(+1, "GotoNextLayer") {}
```
dacap: Same here:
```suggestion
  GotoNextLayerCommand() : GotoLayerCommand(+1, CommandId::GotoNextLayer()) {}
```

src/app/commands/cmd_rotate.cpp
```diff
@@ -190,6 +190,10 @@ void RotateCommand::onLoadParams(const Params& params)
 
 bool RotateCommand::onEnabled(Context* context)
 {
+  // Because we use the toolbox & editor to transform the selection, this won't work without a UI
```
dacap:
```suggestion
  // Because we use the toolbar & editor to transform the selection, this won't work without a UI
```
Actually the `Toolbox` represents the available tools configured in the `gui.xml` file.

src/app/commands/cmd_frame_properties.cpp
```diff
@@ -62,15 +61,20 @@ void FramePropertiesCommand::onLoadParams(const Params& params)
   else if (frame == "current") {
```
dacap:
```suggestion
  // If "frame" param is not specified, we show the properties of the current frame.
  else if (frame == "current" || !params.has_param("frame")) {
```

src/app/commands/cmd_frame_properties.cpp
```diff
@@ -62,15 +61,20 @@ void FramePropertiesCommand::onLoadParams(const Params& params)
...
-    m_frame = frame_t(base::convert_to<int>(frame));
+
+  const auto frameNumber = base::convert_to<frame_t>(frame);
+  if (frameNumber < 1) {
+    // Will open with the default target (CURRENT_RANGE) if the frame number cannot be parsed.
+    m_frame = CURRENT_RANGE;
   }
+
+  m_target = SPECIFIC_FRAME;
+  m_frame = frameNumber;
```
dacap: Cannot remove the `else` (as we'll overwrite the `m_target` value) and `CURRENT_RANGE` is for `m_target`, not for `m_frame` (`m_frame = CURRENT_RANGE` is not valid).

I think we can leave this function as:
```c++
void FramePropertiesCommand::onLoadParams(const Params& params)
{
  std::string frame = params.get("frame");
  if (frame == "all") {
    m_target = ALL_FRAMES;
  }
  else if (frame == "current" || !params.has_param("frame")) {
    m_target = CURRENT_RANGE;
  }
  else {
    m_target = SPECIFIC_FRAME;
    m_frame = frame_t(base::convert_to<int>(frame));
  }
}
```

src/app/commands/cmd_save_mask.cpp
```diff
@@ -15,13 +15,20 @@
 #include "app/i18n/strings.h"
 #include "app/util/msk_file.h"
 #include "base/fs.h"
+#include "doc/file/act_file.h"
```
dacap: Not required
```suggestion
```

src/app/commands/cmd_save_mask.cpp
```diff
@@ -41,21 +48,27 @@ bool SaveMaskCommand::onEnabled(Context* context)
...
+                                 selFilename))
+      return;
 
-  std::string filename = selFilename.front();
+    filename = selFilename.front();
+  }
 
-  if (save_msk_file(document->mask(), filename.c_str()) != 0)
+  const bool result = save_msk_file(document->mask(), filename.c_str()) != 0;
+  if (result && ui)
```
dacap: A little confusing logic here, `result` if true when `save_msk_file()` fails (even the `!= 0` is quite hidden because there is no parenthesis in this expression).

It would be better something like:

```suggestion
  const bool result = (save_msk_file(document->mask(), filename.c_str()) == 0);
  if (!result && ui)
```

src/app/commands/cmd_save_palette.cpp
```diff
@@ -84,28 +90,31 @@ void SavePaletteCommand::onExecute(Context* ctx)
     }
   }
   gfx::ColorSpaceRef colorSpace = nullptr;
-  auto activeDoc = ctx->activeDocument();
+  auto* activeDoc = ctx->activeDocument();
   if (activeDoc)
     colorSpace = activeDoc->sprite()->colorSpace();
 
-  if (!save_palette(filename.c_str(), palette, 16, colorSpace)) // TODO 16 should be configurable
+  const bool result = save_palette(filename.c_str(), palette, params().columns(), colorSpace);
```
dacap, line 95: Nice to have the `filename` and `columns` as params too for the non-UI version 👍

dacap (changes requested):
I've fully review the PR now, some extra changes 🔍

src/app/commands/cmd_save_mask.cpp
ckaiser: Since we're touching basically every single file anyhow, I did a pass and removed all the unused includes in the commands 😁

src/app/commands/cmd_copy_cel.cpp
```diff
@@ -25,13 +25,13 @@ class CopyCelCommand : public Command {
...
 
-CopyCelCommand::CopyCelCommand() : Command(CommandId::CopyCel(), CmdUIOnlyFlag)
+CopyCelCommand::CopyCelCommand() : Command(CommandId::CopyCel())
 {
 }
 
 bool CopyCelCommand::onEnabled(Context* context)
 {
-  return App::instance()->timeline()->isMovingCel();
+  return App::instance()->timeline() && App::instance()->timeline()->isMovingCel();
```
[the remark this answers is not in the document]
dacap: I'll disable `readability-implicit-bool-conversion`.

src/app/commands/cmd_save_palette.cpp
```diff
@@ -20,57 +19,62 @@
 #include "app/i18n/strings.h"
 #include "app/modules/palettes.h"
 #include "base/fs.h"
-#include "doc/palette.h"
+#include "new_params.h"
```
dacap: Use the full path from `src/` dir:
```suggestion
#include "app/commands/new_params.h"
```

ckaiser: Automatic includes strike again

ckaiser:
Added parameters for `PasteTextCommand`, this involved basically writing a different version that didn't use `Editor::pasteImage`, so an extra set of eyes on the implementation is appreciated since pixel pushing is not my forte (yet?)

PR #5304, opened 2025-07-28
[win] Fix ResourceFinder::includeDesktopDir returning the default path

### LINKED ISSUE #5302: `-debug` does not produce log file if Desktop folder is synced to OneDrive
if the Desktop folder is synced to OneDrive, it will be located at `C:\Users\USERNAME\OneDrive\Desktop`, and the `C:\Users\USERNAME\Desktop` folder will not exist.

When running Aseprite with `-debug` and a synced Desktop folder, no log files will be generated. If you manually create the `C:\Users\USERNAME\Desktop` folder, then it will place them there.

OS: Microsoft Windows 11
Aseprite: v1.3.14.4

ckaiser:
Fixes #5302

Uses `SHGFP_TYPE_CURRENT` which returns the Desktop that the user has configured instead of the default. Fixes Windows 11's OneDrive-by-default Desktop folder and any customizable/moved Desktop in general. Have not tested on Windows versions below 11 but we're still using the old API so it shouldn't really change anything.

### DIFF
--- a/src/app/resource_finder.cpp
+++ b/src/app/resource_finder.cpp
@@ -213,7 +213,7 @@ void ResourceFinder::includeDesktopDir(const char* filename)
 #ifdef _WIN32
 
   std::vector<wchar_t> buf(MAX_PATH);
-  HRESULT hr = SHGetFolderPath(NULL, CSIDL_DESKTOPDIRECTORY, NULL, SHGFP_TYPE_DEFAULT, &buf[0]);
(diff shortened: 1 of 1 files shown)


dacap:
LGTM 👍 this function is used for the debug .txt file and for screenshots (ScreenshotCommand), and it looks like it matches the `SHGetSpecialFolderLocation(NULL, CSIDL_DESKTOP, &pidl)` return value too (used in `file_system.cpp` / the Aseprite file dialog).

Issue #5017, opened 2025-02-18
Add `isValid(Object)` to detect deleted objects


Arecher:
While creating scripts and especially plugins/extensions, I often find myself wishing the API had something I could call to **confirm that an Object is valid.** Saving references to Objects such as sprites, cels or images is extremely useful in quickly being able to access an object and it's properties, but it's pretty hard to detect that they have been deleted. They don't become nil, and attempting to access them in any way just results in warnings being printed to the log.

Take for example this list of sprites 

```
local spritelist = {spritename1 = SpriteObj1, spritename2 = SpriteObj2}
```

The only way to make sure these Objects are still valid, is to check that they still exist in `app.sprites`, keep track of that, and then remove the ones that aren't found.

```
local existingsprites = {}

for index, sprite in ipairs(app.sprites) do
	if spritelist[sprite.filename] then
		-- Exists
		existingsprites[sprite.filename] = true
end end 	

for spritename, sprite in pairs(spritelist) do
	if not existingsprites[spritename] then
		-- Remove the invalid data
		spritelist[spritename] = nil
end end 
```

That's quite the hassle. It'd love to be able to simply call (for example)
```
for spritename, sprite in pairs(spritelist) do
	if not isValid(sprite) then
		spritelist[spritename] = nil
end end 
```

This could take form in a bunch of different ways. Any way to detect deleted objects, so I can remove them, work around them or recalculate more easily would be welcome!

dacap:
Just in case, by default Aseprite keeps sprites valid for 15 minutes after they are closed:

![Image](https://github.com/user-attachments/assets/75d6126d-4c85-4de2-b31d-1b9714a2a53c)

If you disable/uncheck that you can test the script quickly after closing the sprite.

PR #5308, opened 2025-07-31
[lua] Add app.isValid() to check Object validity

ckaiser:
Implements #5017 

Creating this a draft for API discussion. I was originally going to implement this as Sprite.isValid, but since we'd also need an .isValid for all the other things that can get stored in scripts and would also need a validation check, I settled for `app.isValid`, although the naming is a bit generic, we could make it more specific.

I find this solution to be the cleaner one and that wouldn't need us to add properties to a bunch of objects and would scale to anything else derived from doc::Object.

Left some `TODO`s in the code with how we want to handle the function being given a regular Lua type or something else entirely, and depending on how we want to do that, I'll add some tests.

### DIFF
--- a/src/app/script/app_object.cpp
+++ b/src/app/script/app_object.cpp
@@ -498,6 +498,35 @@ int App_useTool(lua_State* L)
   return 0;
 }
 
+int App_isValid(lua_State* L)
+{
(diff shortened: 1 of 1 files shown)


ckaiser:
Closing in favor of a different approach (#5573)

PR #5316, opened 2025-07-31
Improve StatusBar tooltips, add app.tip

ckaiser:
Makes some changes to how StatusBar::showTip behave:

 * Now they stay for the duration even if the user clicks elsewhere.
 * When the user hovers over them, their duration timer stops until the mouse leaves. They can be closed with any click.
 * If another tooltip is shown while there's already one visible, the text is added onto the tooltip. This doesn't look perfect but it's definitely easier than trying to stack multiple tooltips, could use some visual improvement though wrt margins.
 * Added `app.tip(message, delay)` for scripts to show their own tooltips, this is what sparked the modifications.

Marking it as a draft for UX feedback, I wasn't originally planning on working on this, it kinda just happened while I was looking at some of the commands which use tooltips to relay information to the user. It's possible that some things are also relying on `showTip` only showing one thing at a time and that we might spam the user with a massive list of messages?

### DIFF
--- a/src/app/script/app_object.cpp
+++ b/src/app/script/app_object.cpp
@@ -34,10 +34,10 @@
 #include "app/tools/tool_loop_manager.h"
 #include "app/tx.h"
 #include "app/ui/context_bar.h"
-#include "app/ui/doc_view.h"
 #include "app/ui/editor/editor.h"
 #include "app/ui/editor/tool_loop_impl.h"
 #include "app/ui/main_window.h"
+#include "app/ui/status_bar.h"
 #include "app/ui/timeline/timeline.h"
 #include "app/ui_context.h"
 #include "base/fs.h"
@@ -498,6 +498,30 @@ int App_useTool(lua_State* L)
   return 0;
 }
 
+int App_tip(lua_State* L)
+{
+  const auto* ctx = App::instance()->context();
+  if (!ctx || !ctx->isUIAvailable() || !StatusBar::instance())
+    return 0; // No UI to show the tooltip
+
+  if (!lua_isstring(L, 1))
+    return luaL_error(L, "app.tip() message parameter must be a string");
+
+  const std::string message = lua_tostring(L, 1);
+  if (message.empty())
+    return luaL_error(L, "app.tip() message cannot be empty");
+
+  int duration = 2000;
+  if (lua_isinteger(L, 2)) {
+    const int durationInput = lua_tointeger(L, 2);
+    if (durationInput > 0 && durationInput < 30000) // Reasonable limits
+      duration = lua_tointeger(L, 2);
+  }
+
+  StatusBar::instance()->showTip(duration, message);
+  return 0;
+}
+
 int App_get_events(lua_State* L)
 {
   push_app_events(L);
@@ -820,6 +844,7 @@ const luaL_Reg App_methods[] = {
   { "alert",       App_alert       },
   { "refresh",     App_refresh     },
   { "useTool",     App_useTool     },
+  { "tip",         App_tip         },
   { nullptr,       nullptr         }
 };
 
--- a/src/app/ui/status_bar.cpp
+++ b/src/app/ui/status_bar.cpp
@@ -549,7 +549,10 @@ private:
 
 class StatusBar::CustomizedTipWindow : public ui::TipWindow {
 public:
-  CustomizedTipWindow(const std::string& text) : ui::TipWindow(text) {}
+  CustomizedTipWindow(const std::string& text) : ui::TipWindow(text)
+  {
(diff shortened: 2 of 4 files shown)


martincapello:
For being a first draft it looks good! However, I'm wondering if this is something that the users will appreciate or not. I'm saying this because maybe some users are happy with these notifications as is.

In my opinion, it would be nice to handle notifications (or tips) in a more "traditional" way...I mean, by stacking the messages (despite being harder to implement as you said), and also by introducing a configuration parameter that would let the user limit how many messages can be stacked, then it should show only the last N messages by removing the oldest when a new one must be displayed. This way we could keep happy any user that is fine with just showing the last message, and have an alternative for the rest.
BUT...it feels like too much work for little value...so, this is a dilemma for me. Again, in my opinion, the current implementation of this draft looks good, but feels a bit odd to me, or not in par with other notification UXs.

In summary, if we go for having stacking notifications I would prefer a more "traditional" UX, despite requiring a harder implementation. Otherwise I think it doesn't actually worth changing the current behavior.

However, if @dacap thinks that it actually worth giving a shot to this, then I would suggest the following:
- Don't skip messages that are the same text, just keep adding them.
- Add a limit to the amount of messages that can be displayed.
- When executing the `redo` command the messages are not being stacked. This should be fixed.
- It seems that when undoing, the last undo command makes the tip disappear right away instead of waiting for the timeout.

ckaiser:
> In my opinion, it would be nice to handle notifications (or tips) in a more "traditional" way...I mean, by stacking the messages (despite being harder to implement as you said), and also by introducing a configuration parameter that would let the user limit how many messages can be stacked, then it should show only the last N messages by removing the oldest when a new one must be displayed. This way we could keep happy any user that is fine with just showing the last message, and have an alternative for the rest. BUT...it feels like too much work for little value...so, this is a dilemma for me. Again, in my opinion, the current implementation of this draft looks good, but feels a bit odd to me, or not in par with other notification UXs.
> 
> In summary, if we go for having stacking notifications I would prefer a more "traditional" UX, despite requiring a harder implementation. Otherwise I think it doesn't actually worth changing the current behavior.

The main reason why I didn't go with a stacking implementation and why I decided to implement the "if the text is the same then skip the message" was to mostly retain the current behavior, where in normal use having stacking notifications can be pretty rare, and to avoid bloating things with configuration options and whatnot.

>     * Don't skip messages that are the same text, just keep adding them.

I'm not a big fan of this because it can happen pretty randomly, the reason I implemented this in the first place was because I tried to do stuff on a locked layer and got spammed with a million notifications. A compromise can be shaking the current notification if the text repeats, or adding a number (or a personal favorite, adding exclamation marks at the end for every repetition).

>     * Add a limit to the amount of messages that can be displayed.

A compromise for the current implementation can be just giving the big tooltip a maximum size, if it's gonna exceed that then just show the latest notification.

>     * When executing the `redo` command the messages are not being stacked. This should be fixed.

I think this is a good candidate for grabbing the attention of the user in another way rather than stacking them, right now what happens is that you notice that there were two popups because the tooltip flickers since it's opening and closing. Re-creating something eye-catching like a color tween or a slight shake or something like that could be friendlier. 

>     * It seems that when undoing, the last undo command makes the tip disappear right away instead of waiting for the timeout.

This seems to be a side-effect of how the current undo-redo implementations used the tips, fixed.

src/app/commands/cmd_undo.cpp
```diff
@@ -106,8 +106,6 @@ void UndoCommand::onExecute(Context* context)
       msg = "Redid " + undo->nextRedoLabel();
     if (Preferences::instance().undo.showTooltip())
       statusbar->showTip(1000, msg);
-    else
-      statusbar->setStatusText(0, msg);
```
dacap, line 110: If we disable the undo tooltips (Edit > Preferences > Undo > Show Undo Tooltip) with this change we lost the feedback in the status bar. These lines should be restored.

src/app/script/app_object.cpp
```diff
@@ -498,6 +498,30 @@ int App_useTool(lua_State* L)
...
+  const std::string message = lua_tostring(L, 1);
+  if (message.empty())
+    return luaL_error(L, "app.tip() message cannot be empty");
+
+  int duration = 2000;
+  if (lua_isinteger(L, 2)) {
+    const int durationInput = lua_tointeger(L, 2);
+    if (durationInput > 0 && durationInput < 30000) // Reasonable limits
+      duration = lua_tointeger(L, 2);
+  }
```
dacap, line 519: We can use a number (floating point) that indicates the seconds. Check existent functions in the API like `Frame_set_duration` or `Timer_set_interval`.

Issue #5177, opened 2025-05-21
Add Dialog property for disabling user resizing


MapleGecko:
(Made for [v1.3.x (features for extensions)](https://github.com/aseprite/aseprite/milestone/59) milestone)

It would be nice to have an option to disable manual dialog resizing from the user. 

At the moment, being able to resize every dialog feels uneeded, especially for smaller ones: i.e. small menus, validation prompts and "notitlebar" dialogs used as pseudo info-bubbles.

In those cases, it would be better the let the script's author handle it. Complements #5176.

**Suggested implementation:**

``` Lua
Dialog{userResize=false}; 
```

### Aseprite and System version

* Aseprite version: v1.3.13
* System: Windows 11

PR #5317, opened 2025-07-31
[lua] Add `resizeable` property to Dialog constructor

ckaiser:
Implements #5177


dacap:
It looks great 👍 the only thing I'd add is a ` (fix #5177)` at the end of the commit line.

PR #5318, opened 2025-08-01
Activate the native window before asking the user to save sprite changes

### LINKED ISSUE #3542: Quitting the app doesn't focus on the window if there's a warning about saving changes
Title says it all really. When I quit the app sometimes I don't notice the warning popup asking me if I want to save changes as the window is behind others. The window should appear on top of all others so I'm aware of the warning.

I've only checked this on macOS.

ckaiser:
Fixes #3542

Comes with some changes to LAF  (#153) so that the behavior of `activate()` matches the other platforms, although for some reason testing this on Windows I'm having an entirely different bug where the no events are getting through to the app when it's minimized and I can't reproduce it on the stable version. This is unrelated to these changes and needs some more testing.

### DIFF
--- a/src/app/ui/doc_view.cpp
+++ b/src/app/ui/doc_view.cpp
@@ -42,6 +42,7 @@
 #include "doc/sprite.h"
 #include "fmt/format.h"
 #include "ui/alert.h"
+#include "ui/display.h"
 #include "ui/menu.h"
 #include "ui/message.h"
 #include "ui/shortcut.h"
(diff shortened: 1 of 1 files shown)


Gasparoken:
Works great! LGTM👍 Before seeing anything, I thought it might interfere with the floating windows, but no. From what I've tested, they work fine.

dacap:
Cherry picked as 49fa35237a2dc3ed05983854bdc6bf551ca91314 (updated laf module + added issue reference in the commit message)

PR #5320, opened 2025-08-01
Allow gif encoding to be stopped (fix #2619)

### LINKED ISSUE #2619: Cancel button doesn't stop the animation export
Pressing Cancel during animation export will not actually stop the operation, only freeze the UI until it completes. The output file has all frames regardless of when export was cancelled. Happens to both gif and png seq as far as I tested.

To reproduce
- Create a file with 100+ non-empty frames to be able to press the button before the export completes
- Export with any settings
- Press Cancel before the window disappears
- Compare the output file to the uninterrupted export

* Aseprite version: 1.2.25-x64 Steam
* System: Windows 10

ckaiser:
Fixes #2619, here's a test file with a lot of frames to try it: [f500.zip](https://github.com/user-attachments/files/21555653/f500.zip)

### DIFF
--- a/src/app/file/gif_format.cpp
+++ b/src/app/file/gif_format.cpp
@@ -1095,6 +1095,9 @@ public:
     gifframe_t nframes = totalFrames();
     for (gifframe_t gifFrame = 0; gifFrame < nframes; ++gifFrame) {
       ASSERT(frame_it != frame_end);
+      if (m_fop->isStop())
+        break;
+
       frame_t frame = *frame_it;
       ++frame_it;
 


src/app/file/gif_format.cpp
```diff
@@ -1095,6 +1095,9 @@ class GifEncoder {
     gifframe_t nframes = totalFrames();
     for (gifframe_t gifFrame = 0; gifFrame < nframes; ++gifFrame) {
       ASSERT(frame_it != frame_end);
+      if (m_fop->isStop())
+        break;
```
dacap, line 1099: Shouldn't return `false`?

ckaiser: This way you end up with a valid GIF, even if it doesn't have the full frames. Better than a broken file IMO since we're already past the point where it's started writing to disk.

dacap: Oh it's fine in this way, other encoders do the same, and the issue is handled later checking if the operation was canceled so the sprite is marked as "modified", so the user has a second chance to save the file later.

PR #5321, opened 2025-08-03
[lua] Added `autofit` and `sizeHint` properties to Dialog (fix #5176)

### LINKED ISSUE #5176: Add Dialog:autofitBounds() to automatically resize dialogs from their content

At the moment, there is no way to automatically resize a dialog from its visible content after using Dialog:modify(), without having to use close/show functions every time. 

It would be nice to have this new option to handle cases when the modified contents of a dialog don't fit in the initial bounds.

Example script:

``` Lua

LabelIsVisible = false
function generateDialogFit()
    if DialogFit~=nil then DialogFit:close() end
    DialogFit = Dialog{
    id="DialogFit",
    title="Dialog",
    onclose=function()
        DialogFit = nil
    end
    }
    DialogFit:button{text="Show label  ", onclick=function() 
        LabelIsVisible = not LabelIsVisible
        DialogFit:modify{id="DialogFit_demolabel", visible=LabelIsVisible}
    end}
    DialogFit:label{id="DialogFit_demolabel", visible=LabelIsVisible, text="Label is too long to be displayed"}
    DialogFit:show{wait = false}
end
generateDialogFit()

```

Suggested implementation:

``` Lua
DialogFit:autofitBounds{align=(Align.CENTER | Align.TOP)}; 
```

* Aseprite version: v1.3.13
* System: Windows 11

ckaiser:
Fixes  #5176, makes automatic fitting the default, with `autofit` set to `Align.LEFT | Align.TOP`. (Note: Could've used `ui::HORIZONTAL` and `ui::VERTICAL` since those are more accurate for what we're doing, but they're not exposed to scripting at the moment - I could just add them, though).

This property can be changed via Lua, and I've added a new read-only sizeHint property to know the size the dialog wants to be to fit all the elements.

I've also while testing encountered a bug that caused old drawn pieces of dialogs linger behind them when the bounds are changed to something smaller, so I fixed that by invalidating the old rect when resizing.

Testing script, adapted from the one by @MapleGecko 
```lua
local labelVisible = false
local dialog

local function generateDialog()
    if dialog ~= nil then dialog:close() end
    dialog = Dialog {
        id = "dialog",
        title = "Dialog",
        autofit = 0,
        onclose = function()
            dialog = nil
        end
    }

    dialog:label { id = "dialog_autofitlabel", text = "autofit = 0" }

    dialog:button { text = "Show label", onclick = function()
        labelVisible = not labelVisible
        dialog:modify { id = "dialog_demolabel", visible = labelVisible }
        -- Manual bounds with sizeHint
        --dialog.bounds = Rectangle(dialog.bounds.origin, dialog.sizeHint )
    end }

    dialog:button {
        text = "Switch fit", onclick = function()
        if (dialog.autofit == 0) then
            dialog.autofit = Align.LEFT
        elseif (dialog.autofit == Align.LEFT) then
            dialog.autofit = Align.TOP
        elseif (dialog.autofit == Align.TOP) then
            dialog.autofit = Align.LEFT | Align.TOP
        else
            dialog.autofit = 0
        end

        dialog:modify { id = "dialog_autofitlabel", text = "autofit = " .. dialog.autofit }
    end }

    dialog:label { id = "dialog_demolabel", visible = labelVisible, text = "Label is too long to be displayed" }
    dialog:show { wait = false }
end

generateDialog()
```


dacap:
Hi @ckaiser, from the conversation we had, the idea I was thinking about the `autofit` was to specify a point of anchor to resize, so `TOP|LEFT` is the default (as the origin of the window is kept), but we could specify `CENTER|MIDDLE` (we have to add `Align.MIDDLE` to Lua) to resize the window from the center, or `RIGHT|BOTTOM`, etc, or just `0` to avoid resizing.

ckaiser:
Updated with the top/right/bottom/left functionality, also updated the script with a dropdown for easier testing.

dacap:
It works great :+1: I think it needs a minor fix in the `Dialog_new` function (to avoid setting autofit to -1), going to fix it after merging this.

Issue #5218, opened 2025-06-11
Add semibold and other weights to the font selector


dacap:
From: https://community.aseprite.org/t/using-other-font-weights-with-text-tool/25709/2

This is a known issue and a nice thing to have in a near future in the font selector:

![Image](https://github.com/user-attachments/assets/587951aa-6e38-4570-bb3f-08b28e3ef50f)

PR #5323, opened 2025-08-03
Add a font weight selector to the font entry

ckaiser:
Implements #5218. Adds a ComboBox to select the font's weight:
Recording 2025-08-03 022925

Two concerns:
 * I had to implement an ugly hack to prevent the ComboBox's window from going over the popup's size, since you couldn't click on items there, @dacap is there an easy fix for this that I'm missing? Holding the mouse button and moving down works fine, but individual clicks just fall through. I'm assuming this is because of the popup's clicking-outside-bounds-closes-it, but I don't know if we wanna get rid of that behavior. Originally I intended moving everything to a Menu and having the weights be in a submenu but those don't have checkbox support and I didn't feel like going down that particular rabbit hole... unless?
 * Right now the *B* bold button is selected when the weight is equal or above to SemiBold, now that we have granular control over font weights, do we want to keep this as-is or change it?

### DIFF
--- a/src/app/ui/font_entry.cpp
+++ b/src/app/ui/font_entry.cpp
@@ -20,9 +20,11 @@
 #include "base/convert_to.h"
 #include "base/scoped_value.h"
 #include "fmt/format.h"
+#include "text/font_style_set.h"
 #include "ui/display.h"
 #include "ui/fit_bounds.h"
 #include "ui/manager.h"
+#include "ui/menu.h"
 #include "ui/message.h"
 #include "ui/scale.h"
 
@@ -271,7 +273,7 @@ FontEntry::FontStyle::FontStyle(ui::TooltipManager* tooltips) : ButtonSet(3, tru
   addItem("...");
   setMultiMode(MultiMode::Set);
 
-  tooltips->addTooltipFor(getItem(0), Strings::text_tool_bold(), BOTTOM);
+  tooltips->addTooltipFor(getItem(0), Strings::font_style_font_weight(), BOTTOM);
   tooltips->addTooltipFor(getItem(1), Strings::text_tool_italic(), BOTTOM);
   tooltips->addTooltipFor(getItem(2), Strings::text_tool_more_options(), BOTTOM);
 }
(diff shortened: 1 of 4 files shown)


dacap:
> * Originally I intended moving everything to a Menu and having the weights be in a submenu but those don't have checkbox support and I didn't feel like going down that particular rabbit hole... unless?

I was thinking about the possibility to show a popup menu pressing the `B` button showing the available weights for the specific font (not all fonts have all weights, it's something we have to navigate for each typeface to show proper/available items/weights). You can use `MenuItem::setSelected()` to check the selected item ([for example](https://github.com/aseprite/aseprite/blob/4bb9239f50e3c8f7d3cdb00fa1f7713b440429e9/src/app/ui/context_bar.cpp#L427)).

This change requires to save the weight information in the `FontInfo` string (convert_to [from string](https://github.com/aseprite/aseprite/blob/4bb9239f50e3c8f7d3cdb00fa1f7713b440429e9/src/app/fonts/font_info.cpp#L168) and [to string](https://github.com/aseprite/aseprite/blob/4bb9239f50e3c8f7d3cdb00fa1f7713b440429e9/src/app/fonts/font_info.cpp#L232)). There is a shortcut called `bold` but probably we'll need an extra property called `weight=...` for other kind of weights (when it's not exactly `text::FontStyle::Weight::Bold`).

ckaiser:
> I was thinking about the possibility to show a popup menu pressing the B button showing the available weights for the specific font (not all fonts have all weights, it's something we have to navigate for each typeface to show proper/available items/weights). 

I thought about the popup option as well, but I figured something inside the "..." button would be more discoverable, unless we had some kind of indication that the B could be right clicked/long-pressed for the popup, unless we wanted to show it.

Is there a generic way we can pull the available weights? Would be nice to disable the options when they do nothing.

> You can use MenuItem::setSelected() to check the selected item

I got mixed up by looking at `AppMenuItem` and forgot all about `isSelected`, whoops.

---

I'll work on the menu and the to_string saving stuff, then we can decide if we want it as a popup on the bold button or a "more" submenu.

dacap:
> I figured something inside the "..." button would be more discoverable, unless we had some kind of indication that the B could be right clicked/long-pressed for the popup, unless we wanted to show it.

At first I thought that just clicking the `B` button would switch the bold weight and show the popup, but there is no easy way to do both with the same click. Other possibility would be to:

1) If there is just one variant (normal / bold) the `B` button acts like a checkbox
2) If there are more variants, but button shows a popup menu to select the specific weight
3) If there are no weight variant, the button could be disabled

It's quite important that for option 2) we can press the mouse button, drag the mouse, and release it in the specific option (e.g. exactly like the [Ink selector](https://aseprite.org/docs/ink)).

> Is there a generic way we can pull the available weights? Would be nice to disable the options when they do nothing.

You can use the `FontMgr::familyStyleSet()` or `matchFamily()` which returns a `FontStyleSet`, probably using `fontMgr->matchFamily(font->typeface()->familyName())`

ckaiser:
Iteration 2:
Recording 2025-08-04 185753

The implementation is a bit more contrived, but I think it's mostly there -- I do have some concerns though:

 * What used to take 1 click (setting the font to bold), now takes 2 clicks. I thought maybe pre-highlighting the "Bold" option when coming from normal weight would help, but idk that it does.
 * I added the "-" and "+" indicators to have a visual idea of when we're in weights beyond normal and bold, but I'm not totally happy with it.

src/app/ui/font_entry.cpp
```diff
@@ -430,14 +476,45 @@ void FontEntry::onStyleItemClick(ButtonSet::Item* item)
   switch (m_style.getItemIndex(item)) {
     // Bold button changed
     case 0: {
-      const bool bold = m_style.getItem(0)->isSelected();
-      style = text::FontStyle(
-        bold ? text::FontStyle::Weight::Bold : text::FontStyle::Weight::Normal,
-        style.width(),
-        style.slant());
+      if (m_availableWeights.size() > 1) {
```
martincapello: > What used to take 1 click (setting the font to bold), now takes 2 clicks. 

To get the 1 click back in the case of having just 2 weights available:

```suggestion
      if (m_availableWeights.size() > 2) {
```

src/app/ui/font_entry.cpp
```diff
@@ -384,16 +386,60 @@ void FontEntry::setInfo(const FontInfo& info, const From fromField)
 {
   m_info = info;
 
-  if (fromField != From::Face)
+  const auto& style = m_info.style();
```
martincapello: We can remove this line, the style variable is not used.

dacap:
> So far so good! But there are still some corners to polish: [...]

@martincapello @ckaiser just in case keep this PR to what it should do, implementing #5218, if there is any other problem a new issue (and new PR) can be created.

martincapello:
> @martincapello @ckaiser just in case keep this PR to what it should do, implementing #5218, if there is any other problem a new issue (and new PR) can be created.

Sure, I'was just mentioning some glitches I found...if both of you find more convenient to fix them in a separate issue, I'm fine with that.

ckaiser:
Addressed the button inconsistencies, I think the other stuff should go in another issue/PR (I'm creating it right now), the main blocker is this:

>     * When changing the weight of some fonts, it seems to not have any effect. For instance, I've been testing the `Arial Hebrew` font on macOS which has 3 weights (Light, Normal, Bold) and none of them changes the face of the font. Tested the same font in other program and it does change. But for other fonts it works as expected...so I'm not sure what is going on.

Since it directly affects the usability of this new interface, but it's not related to the UI and it seems to be an issue with text rendering framework itself. I'm gonna look into while the rest gets reviewed.

martincapello (approved):
LGTM!

dacap:
It looks like the pinning feature doesn't work correctly (fonts are not being matched to unpin the correct variation), i.e. we cannot unpin some fonts:

image

Also some information is needed to identify the specific weight variation in the popup list.

ckaiser:
Updated the main popup to reflect the selected weight and now we check if the font supports the currently selected weight when changing to them and I _think_ that fixed the unpinning bug since I can't reproduce it anymore but I'm not 100% sure.

dacap:
It's working now :rocket: Going to merge it as it is. A little fix for the "no hinting" label is needed but I'm going to add that later.

PR #5331, opened 2025-08-06
Unify ContextBar updates when moving pixels, tooltips (fix #5329)

### LINKED ISSUE #5329: Using arrow keys to move a selection doesn't update contextbar position values
When using the arrow keys to move a selection, the context bar position values don't update. 

When moving the selection with the mouse, it updates in real time, but with the arrow keys (including arrow + shift) it doesn't update at all. It only re-updates when you click on the selection.

Attached is a gif of the issue:

Image

ckaiser:
Fixes #5329, and I took an extra second to add some tooltips & localization for consistency while I was there.

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -621,6 +621,14 @@ current_layer = Current Layer
 first_ref_layer = First Reference Layer
 pick = Pick:
 sample = Sample:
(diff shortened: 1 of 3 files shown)


dacap:
Works great! :+1:

PR #5346, opened 2025-08-13
Fix palette popup sizing in multi display mode

ckaiser:
Noticed that the palette popup had strange sizing when in multi-display mode, turns out it was using my display size to calculate the size instead of the window size, fixed it so that both use the same consistent sizing.

<table>
<tr>
 <td>Before (Single Display)
 <td>Before (Multi Display)
 <td>After (Multi Display)
<tr>
 <td> palette_popup_before_singlewindow
 <td> palette_popup_before_multiwindow
 <td> palette_popup_after_multiwindow
</table>


dacap:
I have to purchase a 2nd monitor to start testing these multiple monitor issues :sweat: I'll try this week :pray:

ckaiser:
> I have to purchase a 2nd monitor to start testing these multiple monitor issues :sweat: I'll try this week :pray: 

Oh my brain got scrambled trying to find a way to describe, like, aseprite's one native window per local window option. I guess I meant display as in ui::Display? Point being: No need to buy a monitor to reproduce this, just changing the window pref, I'll change the description

dacap:
I'm not sure about merging this because it makes the palette popup too small if the window is resized (even menus are bigger / use more available screen space), e.g.

Peek 2025-11-19 18-55

ckaiser:
Yeah, I think originally I was just trying get them to behave consistently but my solution ignores the advantages of actually having multiple windows, a possible approach may be to have a ideal size for the widget (maybe something where there's at least a certain amount of palettes visible) and try to get it to strive for that or keep expanding vertically if it's inside the workspace.

ckaiser:
Experiment 1
The simplest, just a minimum size when in multiple window mode, otherwise behaves just like the normal palette
image

Caveats: Hardcoded, but since it's a pretty dynamic popup idk how I'd do it differently in that regard.

Issue #5081, opened 2025-03-28
FX Layers


dacap:
Referenced as #1152 in the past, but here we make it explicit: We should be able to create a effect layer to be applied to a specific regular layer (or composed group of layers) in **real-time**. The fx could be several things: any filter from _Edit > FX_ or _Edit > Adjustments_.

Should work in a similar fashion to mask layers (#459).

LazyEti:
I noticed that I and many other devs love aseprite but are unfortunately forced to use photoshop because our workflow / style requires an immediate visualization of outlines.

For example a thick anti-aliased via LayerStyle -> Stroke:
<img width="625" height="492" alt="Image" src="https://github.com/user-attachments/assets/87ade5ac-47c4-4b30-8648-5690a2e4cc7f" />

@dacap is there any chance this feature gets bumped up in priority slightly or at least gets added to the v1.4 "First features" list?

Issue #5012, opened 2025-02-17
Nested Aseprite files i.e. Flash MovieClips


chosencharacters:
Hi!

I do quite a bit of animation and UI work with aseprite. What would be ideal is the ability to "nest" aseprite files, which I realize may be complicated depending on how re-usable the Reference layer code is and the underlying structure is, but would be a game changer.

What this would do is be able to designate an aseprite file layer, sort of like the reference layer, and play it as a gif within a gif when run. The nested aseprite file would use the frame rate of the parent's frame.

A basic, but still incredibly useful, version would just use the first frame without animation.

Regardless this would allow "change it once, change it everywhere" like for example if you have a character with a hat that you want to change, or want to make a sprite sheet with several hats. Being able to designate "this is the position of every hat, regardless of the changes I make" and have that live updated would be a huge work flow changer.

dacap:
Hi @chosencharacters, actually this is duplicate of #593 and #184.

Some work was done in the past as an extension (https://github.com/aseprite/Attachment-System) but it'd be nice to integrate these concepts of sub-sprites in the future.

Issue #4617, opened 2024-08-27
Future work for the text tool


dacap:
The first version of the new text tool (#28, #4383) is almost ready for the beta version. Some work to do in future iterations:

- [ ] better handles to resize the bounding box
- [ ] a fixed bounding box to limit the width (and create multiple lines of text)
- [ ] better bold/italic selection (probably with weight or more styles of the same font, disabling unavailable options)
- [ ] options to save/select presets quickly (font family + size + style + antialias option/etc.)

We'd prefer to get some feedback about this to continue improving the tool.

dndaniell16:
Hi, just started using the new beta including the text tool. Has the algo to perform antialiasing or font rendering changed? I've found that the quality of the text is much worse now. The antialiasing and kerning of the text has changed drastically. See the screenshot below. The antialiasing is even incomplete in some areas. Text is supposedly the same size and font, Lora-Medium size 24. The top text uses the beta text box tool. The bottom text uses the original text tool without text boxes. I absolutely love this feature and was looking forward to it, but I unfortunately can't use it because of this. Is there any way that I can fix this issue? Thanks for your hard work.

<img width="796" alt="TextExample" src="https://github.com/user-attachments/assets/f0876e6f-940b-4afd-bce7-a835f457dfd6" />

PR #5347, opened 2025-08-13
Refactors needed for new layer types

dacap:
This is a draft for an initial refactor to start working in several new layer types:

* Audio layer: #70
* Mask layer: #459
* FX layer: #5081
* Text layer: #4617 
* Vector layer: #814 / #2704 / #2976
* Event & gizmo layer (abstracts/data only layers): #722 #3511
* Sub-sprites / external sprite layer: #184 / #593 / #1321 / #5012
* Tweening: #96
* #1303
* #3904 
* Virtual slice layer to show slice temporal data #721

With this PR we'll try to incorporate the required concept for these layers, not to implement all these kind of layers at once, but it will be a good starting point to continue working after this is merged.

### DIFF
--- a/src/doc/object_type.h
+++ b/src/doc/object_type.h
@@ -1,5 +1,5 @@
 // Aseprite Document Library
-// Copyright (C) 2019-2020  Igara Studio S.A.
+// Copyright (C) 2019-2026  Igara Studio S.A.
 // Copyright (C) 2001-2017  David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -9,9 +9,11 @@
 #define DOC_OBJECT_TYPE_H_INCLUDED
 #pragma once
 
+#include "base/ints.h"
+
 namespace doc {
 
-enum class ObjectType {
+enum class ObjectType : uint16_t {
   Unknown = 0,
   Image = 1,
   Palette = 2,


src/doc/layer.cpp
```diff
@@ -32,19 +32,59 @@ Layer::Layer(ObjectType type, Sprite* sprite)
   , m_blendmode(BlendMode::NORMAL)
   , m_opacity(255)
 {
-  ASSERT(type == ObjectType::LayerImage || type == ObjectType::LayerGroup ||
-         type == ObjectType::LayerTilemap);
```
ckaiser, line 36: new-layer-type.md mentions adding the type to this ASSERT, should probably either keep the assert or remove the mention in the documentation.

src/doc/layer.h
```diff
@@ -85,9 +91,22 @@ class Layer : public WithUserData {
   {
     return (type() == ObjectType::LayerImage || type() == ObjectType::LayerTilemap);
   }
+  bool hasSublayers() const { return !m_layers.empty(); }
   bool isGroup() const { return type() == ObjectType::LayerGroup; }
   bool isTilemap() const { return type() == ObjectType::LayerTilemap; }
-  virtual bool isBrowsable() const { return false; }
+  bool isBrowsable() const { return isExpanded() && !m_layers.empty(); }
+  // If this layer is a Mask or FX
+  bool isMaskOrFx() const
```
ckaiser: Is there a more generic word that can encompass "mask or fx" that would allow for future expansion? It feels a little specific to have a check for something that is one of two things, and in future there may be another type of layer that requires the same kind of behavior. Overlay layer? Blanket layer? Modifier layer?

src/doc/layer.h
```diff
@@ -57,19 +57,25 @@ LAF_ENUM_FLAGS(LayerFlags);
 
 class Layer : public WithUserData {
 protected:
+  // Only constructured by derived classes
   Layer(ObjectType type, Sprite* sprite);
 
 public:
+  // Disable assigment
+  Layer& operator=(const Layer& other) = delete;
```
ckaiser, line 65: Probably more consistent to use the `DISABLE_COPYING` macro

src/doc/object_type.h
```diff
@@ -35,6 +37,16 @@ enum class ObjectType {
...
   Tileset = 15,
   Tilesets = 16,
+
+  LayerFill = 17,
+  LayerMask = 18,
+  LayerFx = 19,
+  LayerText = 20,
+  LayerVector = 21,
+  LayerAudio = 22,
+  LayerHitbox = 23,
```
ckaiser: I dunno about pre-defining all these layers types in this PR, I feel like it locks us in to keeping these as TODOs in the code for a long time before they're implemented, and maybe in the meantime we might change our mind about either their existence/necessity/naming, I'd rather we just leave things open for future expansion in the broadest, most generic way, and then in each implementation we can adjust to what we need.

src/doc/layer.cpp
dacap: I think I'll just remove the ASSERT from docs.

src/doc/layer.h
dacap: I don't like the name either, but at the moment might be `isEffect()` if in the future LayerMask is an effect too (both kind of layers have in common that don't make sense alone, and they need to be linked to another layer to apply the effect).

src/doc/layer.h
dacap: Actually the `DISABLE_COPYING` is from a pre-C++11 era. Someday we could get rid of it (at the moment we have a mix of both options).

src/doc/object_type.h
dacap: Right now I'll split the PR in two big changes: One with the refactors, and another new PR with the "new layer types" to continue working. So the refactors can be integrated, but the new types are not yet added to the codebase.

PR #5351, opened 2025-08-13
Improve statusbar tooltip behavior for readability

ckaiser:
Grabs the more accessibility-adjacent changes from #5316 (IMO), the ability to hover over a tooltip to be able to read it in full before it vanishes, and being able to click to dismiss it early.


dacap:
LGTM :+1:

Issue #4736, opened 2024-10-27
Add onFocus/onClick and onResize/onMove/onBoundsChange Dialog events


MapleGecko:
(Made for [v1.3.x (features for extensions)](https://github.com/aseprite/aseprite/milestone/59) milestone)

Similar to the already existing onClose event, it would be great to have more control when the user is interacting with a [Dialog](https://www.aseprite.org/api/dialog?l=Dialog&t=dialog#dialog) in various ways.

These two events would enable complex, multi-dialog interfaces in the most simple and optimized way.

### Aseprite and System version

* Aseprite version: v1.3.9.2
* System: Windows

PR #5354, opened 2025-08-14
[lua] Add Dialog onresize event

ckaiser:
Partially implements #4736, `onFocus` would need a LAF-level refactor to have an event for window activation, so I'm leaving that for later discussion.

Implementing this also noticed that the Resize event was being fired 5(!) times during Dialog creation, so I tried to trim that as much as possible by using setBoundsQuietly when necessary, I've tested most types of popups and windows with both multiple and single displays and so far it does not seem to affect things, but it might require extra testing to make sure there's no corner case that these changes can trigger.

### DIFF
--- a/src/ui/fit_bounds.cpp
+++ b/src/ui/fit_bounds.cpp
@@ -136,7 +136,7 @@ void fit_bounds(const Display* parentDisplay,
 
     // Set frame bounds directly
     pos = nativeWindow->pointFromScreen(frame.origin());
-    window->setBounds(gfx::Rect(pos.x, pos.y, frame.w / scale, frame.h / scale));
+    window->setBoundsQuietly(gfx::Rect(pos.x, pos.y, frame.w / scale, frame.h / scale));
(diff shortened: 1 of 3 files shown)


dacap:
I'll not be able to fully review this until next week, but please refer issues in commits when possible (makes a lot easier to create the release notes).

dacap:
I'll cherry-pick 2dd165dc6e39f7dcdc56a3e7aca067bbe3df52f5 with `onresize` as event name, and discard 2667c736c7cbb20e85e5a9e40f42149147f23f44 as it introduces too many issues/regressions. For example, with this script:

```lua
local dialog = Dialog("Test")
dialog:newrow{always=true}
for i=1,100 do
  dialog:button{ text="OK", onclick=function() app.alert "It works" end }
end
dialog:show{wait=false, bounds=Rectangle(30, 40, 100, 200), autoscrollbars=true}
```

We can get two different windows:

image

While it should always show this one:

image

PR #5355, opened 2025-08-15
Allow typing numbers into Slider widgets

ckaiser:
Adds the ability for the user to use the number keys change the value of slider widgets, including negative numbers. Pressing backspace also deletes a number like in a text input, serves as both a power user feature and an accessibility one, for fine grained control and keyboard navigation.

https://github.com/user-attachments/assets/2c247a63-cb91-4ccb-aff9-cfff3138896f

<sub>(I should've recorded this with some kind of input overlay, hopefully it's understandable enough)</sub>

### DIFF
--- a/src/ui/slider.cpp
+++ b/src/ui/slider.cpp
@@ -11,6 +11,7 @@
 
 #include "ui/slider.h"
 
+#include "base/time.h"
 #include "text/font.h"
 #include "ui/message.h"
 #include "ui/size_hint_event.h"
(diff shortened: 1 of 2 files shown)


src/ui/slider.cpp
```diff
@@ -189,19 +189,52 @@ bool Slider::onProcessMessage(Message* msg)
         int min, max, value, oldValue;
         getSliderThemeInfo(&min, &max, &value);
         oldValue = value;
+        if (m_keyTiming.elapsed() > 1.5) {
+          m_keyBuffer.clear();
+          m_keyTiming.reset();
```
dacap:
```suggestion
          m_keyTiming = base::current_tick();
```

src/ui/slider.cpp
```diff
@@ -189,19 +189,52 @@ bool Slider::onProcessMessage(Message* msg)
         int min, max, value, oldValue;
         getSliderThemeInfo(&min, &max, &value);
         oldValue = value;
+        if (m_keyTiming.elapsed() > 1.5) {
```
dacap:
```suggestion
        if (base::current_tick() - m_keyTiming > 1500) {
```

src/ui/slider.h
```diff
@@ -67,6 +68,9 @@ class Slider : public Widget {
   int m_value;
   bool m_readOnly;
   SliderDelegate* m_delegate;
+
+  base::Chrono m_keyTiming;
```
dacap: We can use `base::tick_t` which is smaller than `base::Chrono`.
```suggestion
  base::tick_t m_keyTiming = 0;
```

PR #5356, opened 2025-08-15
Implement Label buddies

ckaiser:
Implements Label "buddies", so that clicking on a label that's attached to another widget will focus said widget, similar to the HTML <label> `for` property.

The PR adds the implementation and also adds one to every label that benefits from it, this required minor changes in some places (adding `setId` calls when there were placeholder widgets and adjusting some hierarchies so the buddy was easy to find with `findChild`, which also simplifies markup).

Examples:

https://github.com/user-attachments/assets/20d61d53-8fd2-4a0a-8c5c-9f16b15eb23e

### DIFF
--- a/data/widgets/duplicate_sprite.xml
+++ b/data/widgets/duplicate_sprite.xml
@@ -3,16 +3,13 @@
 <gui>
 <window id="duplicate_sprite" text="@.title">
   <box vertical="true">
-    <box horizontal="true" expansive="true">
-      <box vertical="true" homogeneous="true">
-        <label text="@.duplicate" />
-        <label text="@.as" />
-      </box>
-      <box vertical="true" homogeneous="true" expansive="true">
-        <label text="" id="src_name" />
-        <entry maxsize="256" id="dst_name" magnet="true" />
-      </box>
-    </box>
+    <grid columns="2">
+      <label text="@.duplicate" />
+      <label text="" id="src_name" cell_align="horizontal" />
+
+      <label text="@.as" for="dst_name" />
+      <entry maxsize="256" id="dst_name" magnet="true" cell_align="horizontal" />
+    </grid>
     <check text="@.merged_layers" id="flatten" />
     <box horizontal="true" homogeneous="true">
       <button text="@general.ok" closewindow="true" id="ok" magnet="true" />
--- a/data/widgets/dynamics.xml
+++ b/data/widgets/dynamics.xml
@@ -32,29 +32,29 @@
     <grid id="options" columns="2" childspacing="0" expansive="true">
       <separator id="separator" text="@.max_point_value" horizontal="true" cell_hspan="2" />
 
-      <label id="size_label" text="@.size" style="mini_label" />
+      <label id="size_label" text="@.size" style="mini_label" for="min_size" />
       <slider id="min_size" value="1" min="1" max="64" cell_align="horizontal" style="mini_slider"
               tooltip="@.min_size_tooltip" tooltip_dir="left" />
       <boxfiller />
       <slider id="max_size" value="64" min="1" max="64" cell_align="horizontal" style="mini_slider"
               tooltip="@.max_size_tooltip" tooltip_dir="left" />
 
-      <label id="angle_label" text="@.angle" style="mini_label" />
+      <label id="angle_label" text="@.angle" style="mini_label" for="min_angle" />
       <slider id="min_angle" value="0" min="-180" max="+180" cell_align="horizontal" style="mini_slider"
               tooltip="@.min_angle_tooltip" tooltip_dir="left" />
       <boxfiller />
       <slider id="max_angle" value="0" min="-180" max="+180" cell_align="horizontal" style="mini_slider"
               tooltip="@.max_angle_tooltip" tooltip_dir="left" />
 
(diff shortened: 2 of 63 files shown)


data/widgets/duplicate_sprite.xml
```diff
@@ -3,16 +3,13 @@
...
-        <entry maxsize="256" id="dst_name" magnet="true" />
-      </box>
-    </box>
+    <grid columns="2">
+      <label text="@.duplicate" />
+      <label text="" id="src_name" />
+
+      <label text="@.as" for="dst_name" />
+      <entry maxsize="256" id="dst_name" magnet="true" />
+    </grid>
```
dacap: This breaks the dialog when resized, before:

Image

With this patch:

Image

data/widgets/dynamics.xml
```diff
@@ -32,29 +32,29 @@
     <grid id="options" columns="2" childspacing="0" expansive="true">
       <separator id="separator" text="@.max_point_value" horizontal="true" cell_hspan="2" />
 
-      <label id="size_label" text="@.size" style="mini_label" />
+      <label id="size_label" text="@.size" style="mini_label" for="min_size" />
```
dacap, line 35: I'm not sure if it should go to "min_size" or "max_size" (as the max is commonly changed). Same for the angle?

data/widgets/dynamics.xml
```diff
@@ -32,29 +32,29 @@
...
-      <label id="angle_label" text="@.angle" style="mini_label" />
+      <label id="angle_label" text="@.angle" style="mini_label" for="min_angle" />
       <slider id="min_angle" value="0" min="-180" max="+180" cell_align="horizontal" style="mini_slider"
               tooltip="@.min_angle_tooltip" tooltip_dir="left" />
       <boxfiller />
       <slider id="max_angle" value="0" min="-180" max="+180" cell_align="horizontal" style="mini_slider"
               tooltip="@.max_angle_tooltip" tooltip_dir="left" />
 
-      <label id="gradient_label" text="@.gradient" style="mini_label"  />
+      <label id="gradient_label" text="@.gradient" style="mini_label" for="gradient_selector"  />
```
dacap: It looks like this one is not working.

data/widgets/dynamics.xml
```diff
@@ -32,29 +32,29 @@
...
       <hbox id="gradient_placeholder" cell_vspan="2" />
       <link id="gradient_from_to" style="mini_label"  />
 
       <separator id="separator2" text="@.sensors_tweaks" horizontal="true" cell_hspan="2" />
 
-      <label id="pressure_label" text="@.pressure" style="mini_label" />
+      <label id="pressure_label" text="@.pressure" style="mini_label" for="pressure_slider" />
```
dacap, line 55: These labels are not working too. Anyway the threshold sliders are not quite keyboard-friendly  (they don't have a special "focused" style if I remember correctly).

src/app/commands/cmd_new_layer.cpp
```diff
@@ -227,6 +227,7 @@ void NewLayerCommand::onExecute(Context* context)
     if (isTilemap) {
       tilesetSelector = new TilesetSelector(sprite, tilesetInfo);
       window.tilesetOptions()->addChild(tilesetSelector);
+      window.tilesetLabel()->setBuddy(tilesetSelector);
```
dacap:
```suggestion
      window.tilesetLabel()->setBuddy(tilesetSelector->tilesets());
}
```

src/ui/label.cpp
```diff
@@ -9,19 +9,94 @@
...
+          case kGridWidget: {
+            // Iterate through containers until we find something worth focusing
+            for (auto* child : bud->children()) {
+              if (child->isVisible() && child->isEnabled() && !child->hasFlags(IGNORE_MOUSE)) {
+                child->requestFocus();
+                break;
+              }
+            }
+          } break;
+        }
```
dacap: Some app-specific widgets like color buttons require a special behavior too, probably we can move this logic to each widget `ui::kFocusEnterMessage` case, and add a new `fromBuddy()` flag to `ui::FocusMessage` and send that message from here to the label buddy.

ckaiser: I've added a `FocusMessage::Source`, similar to how Qt has a focus "reason", so we can have specific behavior based on how something is focused, and I've moved most of this code to the specific widgets, except for the one for box/grid widgets since that makes more sense being here where it's not really widget behavior based, but more focus based.

data/widgets/dynamics.xml
ckaiser: My logic was "go to whichever is closer to the label". Maybe for a future iteration we could add a comma separated list of widgets so you can keep clicking to switch between them 👀?

data/widgets/dynamics.xml
ckaiser: Turns out the threshold sliders had the code to show the focused style but they were missing `setFocusStop(true)` (and weren't invalidating)

data/widgets/dynamics.xml
dacap: For a first version we can leave as it is ("closer to the label"). And in a future we can review this if users find this useful.

PR #5357, opened 2025-08-15
Fix Color Management regression (fix #5333)

### LINKED ISSUE #5333: Color Management regression
Color differences may be noticeable in the editor compared to other image software and an older version of Aseprite.

For example, the same image was opened in three applications: Aseprite 1.3.14.4, Aseprite 1.2.40 and other software (the embeded Color Profile is "Display P3")

Image

Configuration in both versions of Aseprite:
Image

Version 1.2.40 looks fine, but 1.3.14.4 doesn't.

Aseprite version 1.3.14.4

Gasparoken:
fix #5333

### DIFF
--- a/src/ui/display.cpp
+++ b/src/ui/display.cpp
@@ -1,5 +1,5 @@
 // Aseprite UI Library
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-2025  Igara Studio S.A.
 //
 // This file is released under the terms of the MIT license.
 // Read LICENSE.txt for more information.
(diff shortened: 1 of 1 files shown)


Gasparoken:
I'll be doing a bit more research because there may still be similar bugs in other makeSurface() calls. However, the current PR, which would be the most critical fix, may be merged asap.

dacap:
Should we merge this PR as it is?

Gasparoken:
> Should we merge this PR as it is?

Yes. Just to let you know that this PR is half the solution. I discovered there's another part of the code that also converts the color space to sRGB (I'm still looking).
Anyway, by mixing this fix we'd be in a better position than before, so go ahead and merge it.

PR #5362, opened 2025-08-18
[lua] Prevent adding a mask without a bitmap to the clipboard (fix #5361)

### LINKED ISSUE #5361: Adding Empty `Selection` to `app.clipboard` causes Aseprite Crash on `Paste`
**Describe the bug**
Aseprite crashes, without errors or warnings when you paste, after having set `app.clipboard.contents = {image = Image(), selection = Selection()}

**To Reproduce**
1. Create a script with the following code:
```
app.clipboard.content = {image = Image(64,64), selection = Selection()}
app.command.Paste()
```
3. Run the script
4. Aseprite will hard crash

**Expected behavior**
Since it does not crash when adding an empty Image, I would expect it to be able to handle an empty selection as well. Most likely it would have to adjust the selection to 'nil' before setting it to the clipboard content.

Alternatively, an error that gets send rather than Aseprite crashing would have been fine too.

**Screenshots**
Image

**Aseprite & System:**
- Aseprite: v1.3.14.4
- System: Windows 10
- Extensions: A lot

**Additional context**
I encountered the crash while setting the clipboard content as following `app.clipboard.contents = {image = copiedimage, selection = app.sprite.selection}`. Since `app.sprite.selection` returns `Selection()` when there is no selection Aseprite completely crashed. Took a while to find the cause, since it gave no errors or information. After finding the cause it makes sense, although crashing without warning is always worthy of a report! Especially since `Selection()` is technically a valid `Selection`, so assigning it to the clipboard should not cause such a large issue.

ckaiser:
Fixes #5361, the pasting code expects a mask with a bitmap, at least for the initial one. Fixing this on the Lua side seemed more practical, but I'm not sure if there's a case where we'd want the user to add a selection like this to the clipboard? In that case it'd probably need a refactor of some kind to the pasting/pixelsmovement code.

### DIFF
--- a/src/app/script/app_clipboard_object.cpp
+++ b/src/app/script/app_clipboard_object.cpp
@@ -199,7 +199,7 @@ int Clipboard_set_content(lua_State* L)
   type = lua_getfield(L, 2, "selection");
   if (type != LUA_TNIL) {
     mask = get_mask_from_arg(L, -1);
-    if (!mask)
+    if (!mask || !mask->bitmap())
       return luaL_error(L, "invalid selection provided");
   }
   lua_pop(L, 1);


ckaiser:
Alright no, this wasn't the right approach (thanks, tests!) - gonna have another go at it on the pasting side.

PR #5367, opened 2025-08-19
[lua] Refresh timeline when changing layer collapsed status (fix #5366)

### LINKED ISSUE #5366: `Layer.isCollapsed/isExpanded` do not update Timeline visuals
**Describe the bug**
When calling `Layer.isCollapsed = true/false` or `Layer.isExpanded = true/false` the timeline is not updated, and the groups remain in their original state.

**To Reproduce**
1. Select a layer
2. Call `app.layer.isCollapsed = false` on an Expanded Group
3. The timeline folder icon will update properly, but the `group.layers` will still be visible in the timeline.
4. Clicking the expand/collapse button in the timeline will not do anything the first time you click, apart from change the icon. This is likely because, as far as Aseprite knows, the Collapse/Expand commands have happened, and clicking it again will turn it back to the state that it is already in.

**Expected behavior**
It collapses or expands the group depending on true/false.

**Screenshots**
Image

- Aseprite: v1.3.14.4
- System: Windows 10
- Extensions: :')

**Additional context**
It is not an `app.command.Refresh()` issue, that does not update the timeline.

ckaiser:
Fixes #5366, had to add a function to force the relayouting of rows, unclear if there's a way to do this through more "proper" event stuff but this seems logical.

### DIFF
--- a/src/app/script/layer_class.cpp
+++ b/src/app/script/layer_class.cpp
@@ -22,6 +22,7 @@
 #include "app/script/luacpp.h"
 #include "app/script/userdata.h"
 #include "app/tx.h"
+#include "app/ui/timeline/timeline.h"
 #include "doc/layer.h"
 #include "doc/layer_tilemap.h"
 #include "doc/sprite.h"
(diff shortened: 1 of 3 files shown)


dacap:
LGTM :+1: Probably `regenerateCols();` is not necessary and doing just the same as `Timeline::onLayerCollapsedChanged()` is enough, I'll fix this with another commit.

In a future we have to improve the timeline rows generation (to group several operations in one).

PR #5368, opened 2025-08-19
[lua] Fix new layers not going to the top when selected (fix #5364)

### LINKED ISSUE #5364: `app.command.NewLayer{top = true}` does not place layer at top of stack
**Describe the bug**
Calling `app.command.NewLayer{top = true}` does not place the layer at the top of the stack, as the [documentation ](https://www.aseprite.org/api/command/NewLayer#newlayer)says, but only places it **above the currently active layer**.

**To Reproduce**
1. Create a new sprite with a few layers.
2. Select a layer in the middle
3. Run a script with `app.command.NewLayer{name = "I am on top!", top = true}`
4. The new layer will appear above the currently selected layer

**Expected behavior**
The layer appears at the top of the timeline, or depending on how you read it, at the top of the parent the currently active layer is in. Which perhaps the documentation should clarify, since both are valid options to want!

**Screenshots**
Image

- Aseprite: Aseprite v1.3.14.4
- System: Windows 10
- Extensions: More each day!

**Additional context**
Can be worked around by setting `app.layer = toplayer` before calling `app.command.NewLayer{top = true}`

ckaiser:
Fixes #5364, this _seems_ like an oversight where the code just wasn't there, but it might be this way for legacy reasons.

### DIFF
--- a/src/app/commands/cmd_new_layer.cpp
+++ b/src/app/commands/cmd_new_layer.cpp
@@ -267,16 +267,8 @@ void NewLayerCommand::onExecute(Context* context)
     bool afterBackground = false;
 
     switch (m_type) {
-      case Type::Layer:
-
-        if (m_place == Place::BeforeActiveLayer) {
-          layer = api.newLayer(parent, name);
-          api.restackLayerBefore(layer, parent, activeLayer);
-        }
-        else
-          layer = api.newLayerAfter(parent, name, activeLayer);
-        break;
-      case Type::Group: layer = api.newGroupAfter(parent, name, activeLayer); break;
+      case Type::Layer: layer = api.newLayer(parent, name); break;
+      case Type::Group: layer = api.newGroup(parent, name); break;
       case Type::ReferenceLayer:
         layer = api.newLayer(parent, name);
         if (layer)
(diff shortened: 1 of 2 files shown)


src/app/commands/cmd_new_layer.cpp
```diff
@@ -273,6 +273,9 @@ void NewLayerCommand::onExecute(Context* context)
           layer = api.newLayer(parent, name);
           api.restackLayerBefore(layer, parent, activeLayer);
         }
+        else if (m_place == Place::Top) {
+          layer = api.newLayer(sprite->root(), name);
```
dacap: It looks like it works only for regular layers (not for groups, tilemaps, etc.).

Probably if `m_place` is `Top` we have to keep `parent = sprite->root()`. The same is true for `BeforeActiveLayer`, it looks like a bug in the current code, which works only for a regular layer (we should do the `api.restackLayerBefore(...)` for any kind of layer?). It might be a good idea to add several automated tests for all these scenarios.

ckaiser: Shuffled the logic so we do the restacking after creation and added a new test so make sure it's consistent across the different types (except for reference layers, those I haven't looked at the stacking logic they're supposed to have beyond "at the start of everything").

There's probably more complex tests we could do with nested layouts and such, but I think the logic should keep.

dacap:
Works great, just tested with:
```lua
local d = Dialog("layers")
d:button{ text="after", onclick=function() app.command.NewLayer{} end }
d:button{ text="before", onclick=function() app.command.NewLayer{before=true} end }
d:button{ text="top", onclick=function() app.command.NewLayer{top=true} end }
d:show{ wait=false }
```

Issue #2316, opened 2020-03-26
Rotation pivot is positioned incorrectly when "tiled mode" is enabled


HolyBlackCat:
It's easier to explain with a screenshot:

![1](https://user-images.githubusercontent.com/20289055/77683765-f3e13400-6fa9-11ea-92fa-654a6c259ccc.png)

Steps to reproduce:

* `View` ➜ `Tiled Mode` ➜ `✓ Tiled in Both Axes`
* Enable selection tool
* On the top bar, `Rotation Pivot` ➜ `✓ Display pivot by default`
* <kbd>Ctrl</kbd>+<kbd>A</kbd>

Now the pivot is in the middle of the top-left tile, instead of the middle tile.

If I disable the tiled mode, the pivot is correctly centered in the selection.

### Aseprite and System version

* Aseprite version: v1.2.17-x64, Steam
* System: Windows 7

PR #5369, opened 2025-08-19
Fix tiled mode when moving pixels 

### LINKED ISSUE #1457: Ability to "loop" when moving selection into tiled area
If you're using tiling in any direction, you can't move selected pixels outside of the original square. They'll just disappear instead of loop back on the other side. I don't know if this is a bug or just not a finished feature, but I'd like tiling to work like the extras that show up are the middle tile for all intents and purposes. Probably part of the same issue that makes the fill tool not work properly on the tiled parts.

martincapello:
Fix #1457

This PR doesn't include changes to make the fix work with tilemaps. Should this be addressed? If so, IMO it would be better to do it in a separate issue.

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -726,6 +726,7 @@ target_sources(app-lib PRIVATE
   util/shader_helpers.cpp
   util/slice_utils.cpp
   util/tile_flags_utils.cpp
+  util/tiled_mode.cpp
   util/tileset_utils.cpp
   util/wrap_point.cpp
   widget_loader.cpp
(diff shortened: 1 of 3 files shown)


Gasparoken:
Works great! A minor observation: Perhaps it would be better to use the new `hasModeFlag()` function in more places instead the expression `((int(m_mode) & int(filters::TiledMode::SOME_MODE)))`.

Throughout testing, I discovered that the pivot point isn't positioned correctly in tiled mode, but that's a different issue (https://github.com/aseprite/aseprite/issues/2316); I'll look into a fix for this in another PR.

Gasparoken (approved):

dacap:
It works great :+1: it would be nice to move the whole `TiledModeHelper` impl in its own `.cpp` file. Adjusting the required #includes (e.g. `#include "doc/algorithm/rotate.h"` can be moved to the `.cpp`)

dacap:
LGTM! :+1:

PR #5380, opened 2025-08-25
Fix bug in Saturation layer blend mode (fix #2661)

### LINKED ISSUE #2661: Bug in Saturation layer blend mode
I noticed that the [`set_sat`](https://github.com/aseprite/aseprite/blob/master/src/doc/blend_funcs.cpp#L400) function has a bug when the two lower components are the same value. This can be shown in the following test file:

[blend_saturation_bug.zip](https://github.com/aseprite/aseprite/files/6111637/blend_saturation_bug.zip)

It looks as follows:

blend_saturation_bug
blend_saturation_bug_4x

The top layer is all the same color and uses blend mode "Saturation".

The bottom contains three strips of slightly different blues. The middle has R value = G value. Note how the merged value suddenly jumps colors when they should all have the same hue?

* Aseprite version: 1.2.25-x64
* System: Windows

Gasparoken:
fix #2661

### DIFF
--- a/src/doc/blend_funcs.cpp
+++ b/src/doc/blend_funcs.cpp
@@ -1,5 +1,5 @@
 // Aseprite Document Library
-// Copyright (c) 2019-2022 Igara Studio S.A.
+// Copyright (c) 2019-2025 Igara Studio S.A.
 // Copyright (c) 2001-2017 David Capello
 //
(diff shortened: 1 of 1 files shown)


dacap:
LGTM :+1: going to include in v1.3.15.1 today.

PR #5381, opened 2025-08-26
Add placeholder property to Entry, default placeholder for SearchEntry

ckaiser:
Adds a "placeholder" property to Entry widgets, and a new default "Search" placeholder for SeachEntry widgets. This is an excerpt from the work I'm doing on the command runner, figured it's very self contained so it deserves its own PR.

image


dacap:
Nice idea :+1:

PR #5393, opened 2025-08-28
Add "Run Command" window

ckaiser:
Adds a command runner/"command palette" with the following features:

 * Search through all of the Aseprite `Command`s, including variations with parameters declared as keyboard shortcuts.
 * Math expressions
 * Inline Lua script evaluation.

An (outdated now) video showcase of everything it has to offer:

https://github.com/user-attachments/assets/46540590-caca-4a00-adb3-c9917daa0aa5

Some things needed in order to finish this:
- [x] Noticed a duplicate on the layer properties command, might be because of duplicated shortcuts?
- [x] ~~Decide on the shortcut, right now it's Ctrl+Shift+R~~ Ctrl+Space
- [ ] The sorting order tries to be smart but that might be more confusing than straight up matching
- [ ] Decide how many results we want to include, and if we want pagination/scrolling
- [x] When writing and deleting a word, we show the very first available commands in the sort order (and also it can sometimes show two highlighted commands when using kbm navigation)
- [ ] Code TODOs


ckaiser:
Added fuzzy string matching, we now check for duplicates and commands with `isChecked(...) == true` show up with a little checkbox.

@dacap AFAIK there's no way to check if a command is functions like a "toggle", you can call isChecked to know if it's checked, but there's nothing you can call to see if we should show an empty checkbox or maybe do the toggle inside the runner without leaving, might be worth seeing if we wanna add something like `Command::isToggleable()` to distinguish them?

image
image

dacap:
This looks great :rocket: 

> Added fuzzy string matching, we now check for duplicates
> and commands with `isChecked(...) == true` show up with a little checkbox.

Nice!

> @dacap AFAIK there's no way to check if a command is functions like a "toggle", you can call isChecked to know if it's checked, but there's nothing you can call to see if we should show an empty checkbox or maybe do the toggle inside the runner without leaving, might be worth seeing if we wanna add something like `Command::isToggleable()` to distinguish them?

We could change the `isChecked()` return value to `std::optional<bool>` probably, but I'm not sure if we should show an unchecked box for these togglable commands (as I think an empty box will be confusing for these cases). At the moment I'd avoid this.

Some notes:
1. probably the `=...` syntax is not needed, we can just evaluate the expression and if the expression works and is different to what is written, we show the result (+search for commands, just like google works).
2. We can set <kbd>Ctrl+Space</kbd> as the default shortcut for this.
3. Should we order things by menu order? For example, if we search for `open`
   image
   "Open sprite" is a good first candidate and "open browser" isn't (as it requires parameters too).

ckaiser:
Adjusted things to avoid listing commands that need parameters to function (added a few missing `onNeedParams()` implementations), tweaked the search score a bit so it prioritizes commands of similar length to what you're typing to mitigate your case, although at this point I might rethink the whole thing cuz it's getting a bit arbitrary maybe?

> probably the =... syntax is not needed, we can just evaluate the expression and if the expression works and is different to what is written, we show the result (+search for commands, just like google works).

Personally I like it better with the prefix because it clearly delineates things as separate "mode", like the Lua expression editor and more importantly makes the code simpler and more readable IMO, it's already a bit complicated as it is. We're also already doing a bunch of things on every `Entry::Change` so I'd like to avoid doing more, I picked "=" since it's what you to use to do math in Excel formulas so it might be familiar to some.

dacap:
I think it's working great :+1: we can merge it as it is (you can move this to "ready for review") and I'll merge it just in case something is missing.

After this is merged I'll make some little changes to the window title bar to reduce a little the translation/i18n work.

ckaiser:
Great! I've marked it ready for review and I'll get to writing the documentation so the "?" button doesn't link to a 404 page

PR #5405, opened 2025-09-01
Add rounded rectangle support

### LINKED ISSUE #2184: Rounded rectangle
Add the possibility to draw rounded rectangles (maybe just a parameter/key modifier for the regular rectangle tool?).

martincapello:
Fix #2184

Introduces a couple of algorithms to draw the rounded corners:
- Drawing of 4-sliced circle outlines and filled circles, used for drawing the rounded corners of non-rotated rectangles and filled rotated and non-rotated rectangles.
- Drawing of arcs, used for drawing the rounded corners of rotated rectangles.

How to use it:
Draw a regular rectangle and use the <kbd>C</kbd> key as the corner radius modifier. Move the mouse inside/outside the corner to change the radius. This also works for the rectangular marquee tool.

TODO: 
* Add parameter to set the corner radius in rectangle_class.cpp so it can be used from scripts. (Should we make this in a separate issue/PR?)

### DIFF
--- a/src/app/ui/context_bar.cpp
+++ b/src/app/ui/context_bar.cpp
@@ -406,6 +406,25 @@ public:
   void setValue(int value) { m_cornerRadius->setValue(value); }
 
 protected:
+  bool onProcessMessage(ui::Message* msg) override
+  {
+    switch (msg->type()) {
+      case ui::kMouseMoveMessage:
+        if (children()[0]->hasCapture()) {
+          MouseMessage* mouseMsg = static_cast<MouseMessage*>(msg);
+          Manager* mgr = manager();
+          Widget* pick = mgr->pickFromScreenPos(
+            display()->nativeWindow()->pointToScreen(mouseMsg->position()));
+          if (pick == m_cornerRadius) {
+            mgr->transferAsMouseDownMessage(children()[0], m_cornerRadius, mouseMsg);
+          }
+        }
+        break;
+    }
+
+    return ButtonSet::onProcessMessage(msg);
+  }
+
   virtual void onItemChange(Item* item) override
   {
     ButtonSet::onItemChange(item);


Gasparoken:
Great! At first glance it works very well.
However I noticed some details that could be improved:
1. When holding 'J', the radius should respond to horizontal and vertical gestures. Also, the action of enlarging/shrinking the radius should not be reversed when moving in one direction (apparently, it should be related to an absolute distance from a reference point).
2. I see that in successive repetitions of the rectangle + J + radius change command, the internal reference value accumulates with the mouse movement, regardless of whether the radius has changed (there is no radius change when we reach the maximum and minimum limits). This makes it seem like the radius modification stops working if too much "internal movement" has accumulated. This mouse movement accumulator should be reset for each new rectangle drawn.

Comments:
1. I think it's great that the radius is maintained for future rectangles.
2. Perhaps there should be an entry in the context bar to define the radius. However, as a first implementation, it seems fine to me.

martincapello:
@Gasparoken thanks for the feedback!

> When holding 'J', the radius should respond to horizontal and vertical gestures. 

Sure, no problem.

> Also, the action of enlarging/shrinking the radius should not be reversed when moving in one direction (apparently, it should be related to an absolute distance from a reference point).

👍 

> I see that in successive repetitions of the rectangle + J + radius change command, the internal reference value accumulates with the mouse movement, regardless of whether the radius has changed (there is no radius change when we reach the maximum and minimum limits). This makes it seem like the radius modification stops working if too much "internal movement" has accumulated. This mouse movement accumulator should be reset for each new rectangle drawn.

I'm not sure if I'm getting this right. The radius doesn't have a hard limit, basically it is visually limited by the size of the rectangle, but the radius can be bigger than the visible radius. If you set a big corner radius it will be adjusted according to rectangle size. So, are you proposing that the radius should not be increased once the max visible radius for the current rectangle size is reached? In which case, once the user reached the max visible radius they would be forced to resize the rectangle first and then increment the radius, right?

martincapello:
> When holding 'J', the radius should respond to horizontal and vertical gestures.

Been trying this...and it makes harder to set the radius because when you move the mouse pointer, usually it moves in both directions so it makes harder to get a one by one increment. Do we still implement this?

gasparcapello:
> The radius doesn't have a hard limit, basically it is visually limited by the size of the rectangle, but the radius can be bigger than the visible radius.

IMO, this is not convenient because you lose the visual feedback of the radius.

> So, are you proposing that the radius should not be increased once the max visible radius for the current rectangle size is reached?

Yes, that's right.

> In which case, once the user reached the max visible radius they would be forced to resize the rectangle first and then increment the radius, right?

No, IMO the correct behavior is:
1. You can modify the rectangle size when the 'J' key isn't pressed, then
2. When 'J' is pressed the radius size should be modified only

I think it's best for the artist to modify one parameter at a time, rather than enabling a double modification (radius + rectangle size ).

I can imagine your approach, but I found one case that might be annoying for the artist:
When the radius size is near to the maximum, it's hard to know if the radious is actually the max. We'll notice of it when the size of the rectangle increases, then the original rectangle size is loss., then you can decrease the size of the rectangle until the original rectangle size.
It doesn't sound bad, but I prefer the two-step approach.

Anyway, maybe I misinterpreted your proposal, or you've come up with something better. Let me know if that's the case.

Gasparoken:
Modifying the radius is now much more intuitive, and I like it.
Going into a little more detail, I saw that a square with the maximum radius doesn't match the shape of a circle of the same size (tested with a 64x64 figure). When testing, I discovered that the arcs of the quadrants correspond to a 63px circumference, so I deduce that the algorithm adds 1 extra pixel to the square halves to form the 64x64px rounded rectangle. This should be fixed (no extra pixels should be added, and the circumference should correspond to a 64px diameter).

Gasparoken:
I just finished testing it. I love it 🎉
After using it a little more, I realize it would be nice to separate the rounding radius of the Selection tool from the Rectangle tool. I found that I generally want to select a rectangular area, as opposed to the Rectangle tool, which is more likely to want to make multiple rectangles with the same rounding radius.
I feel like it could be left as is and then the "roundness radius" could be made independent of the tools described. I'll leave that up to you.

dacap:
It works great! 🚀 

We could merge this as it is. Anyway I'd like to quote @Gasparoken:

> Perhaps there should be an entry in the context bar to define the radius.

I think we should have the corner radius in the context bar in some kind of field and show a tooltip including its shortcut/modifier.

About the <kbd>J</kbd> shortcut, I'm not quite sure. I think it makes sense because it is below the <kbd>U</kbd> key and it looks like a rounded corner, but it has some problems if we want to combine other modifiers (Shift, Alt, Ctrl, Space). Probably the <kbd>C</kbd> key might do the job and it's in the left side of the keyboard (closer to other modifiers).

martincapello:
I've made the corner radius field's slider optional because I set a maximum value of 999 and it doesn't make sense to have a slider for such range, it doesn't even let me slide unit by unit because of the width of the slider's popup being less than 999. 
Also there could be no actual limit to the radius entered by the user, it is automatically limited by the size of the rectangle while the user is changing its size. So the maximum value could be unbounded, let me know if we should use a bigger (or the biggest) maximum value.

dacap:
Several points here:

1. I think the corner radius field could show a slider for a smaller range (e.g. 0-32) but we could be able to write down a bigger radius (so we have the slider + the entry without limit)
2. there is a bug where the corner radius value is reduced just as we use the rectangle tool:

   rounded-corner

3. as in the spray tool, it's a little confusing to have two entry fields without any label, could we add something like:

   image

   In the best case the corner radius text entry shouldn't be visible by default, not sure how we could do it: 

   image

martincapello:
Added a corner radius button that shows the corner radius entry in a popup.

dacap:
It works great. Some minor things:

1. there is a conflict in `src/app/tools/intertwiners.h` so I cannot merge this PR, you can resolve it leaving the include section as:
```c++
#include "app/tools/controller.h"
#include "app/tools/intertwine.h"
#include "app/tools/point_shape.h"
#include "app/tools/tool_loop.h"
#include "app/tools/tool_loop_modifiers.h"
#include "base/pi.h"
#include "doc/algo.h"
#include "doc/algorithm/polygon.h"
#include "doc/layer_tilemap.h"
#include "gfx/point.h"

#include <algorithm>
```

2. The "corner radius" button is visible in the "text tool":

image

3. Using the shortcut (C key) doesn't change the tool preference, I think it should change it.

4. There is a minor thing that would be nice to have: dragging the mouse from `CornerRadiusField` could `offerCapture` to `CornerRadiusEntry`.

martincapello:
Something has changed...now the slider doesn't show up anymore...I will be fixing that along the other things.

EDIT: Alright, maybe nothing changed...it turns out that in single window UI mode it works, but with multiple windows it doesn't...not sure if it worked before the rebase.

martincapello:
@dacap Last commit adds the item 4 you mentioned in your last review, but it doesn't work on macOS (using multiple windows). However, it does work on Windows (I didn't test it on Linux yet). 

The issue on macOS occurs because when the user starts dragging something the mouseMove events received are coming from the window that received the mouseDown. Then the position of the mouse is relative to that native window, and not to the popup window that has the widget to which I'm trying to offer the capture, hence the windows manager is not able to pick the slider. It seems that this is due to the fact that in macOS there is no captureMouse() and releaseMouse() implementations. So I'll see if I can do something about this to fix it.

martincapello:
I couldn't find any fix today. I think it would be better to leave it as is and file a new issue for macOS later. Thus we don't delay the rounded rectangle support anymore.

I've tried:
- Using a global variable to store the view that captured the mouse and then use it to adjust the mouse position returned in the mouse events...this almost worked.
- Using events to simulate a mouse up and a mouse down when the mouse capture changes. I've tried with NSEvents and CGEvents...couldn't make them work.

dacap:
I've tested on Linux and it works great! I'll merge it as it is, we can take a look to the macOS issue later. I'm not sure if the UI will be kept in this way, or a text field that is expanded in the same context bar might be a better option in the future (so the corner radius button acts like a checkbox to make show/hide the text field in the same context bar).

PR #5409, opened 2025-09-02
Fix user-defined shortcuts priority and key context matching

### LINKED ISSUE #5390: Cannot configure Left/right keys to Undo/Redo without removing the association with prev/next frame
Reported here: https://steamcommunity.com/app/431730/discussions/2/596284386693932356/ it looks like the key context matching is not working as before (mainly for keys with "Normal" key context). Probably a regression from: https://github.com/aseprite/aseprite/commit/9a1e92da3525037599d895ae02855322d802b059

It would be great to have some automated tests to check this kind of configurations and avoid breaking the key context matching.

dacap:
We're going to try to fix #5390 giving more priority to shortcuts defined by the user and fixing the KeyContext matching (which was broken in 9a1e92da3525037599d895ae02855322d802b059)


src/app/ui/key.cpp
```diff
@@ -297,6 +295,26 @@ namespace app {
...
+     thisShortcutContext == currentContext) ||
+    // Better key source/level: User-defined > Extension-defined > App-defined
+    (int(this->source) > int(otherShortcut.source) && (thisShortcutContext == currentContext ||
+                                                       // User-defined "Any" context overwrites all
+                                                       // app-defined context
+                                                       thisShortcutContext == KeyContext::Any)) ||
+    // Normal > SelectionTool > Transformation
+    ((currentContext == KeyContext::Transformation &&
+      otherShortcutContext != KeyContext::Transformation &&
+      thisShortcutContext == KeyContext::SelectionTool));
```
ckaiser: The logic seems good but this kind of complex one-liner is a bit hard to read, I feel like it could just be several ifs with early returns and it'd read nicer and generate basically the same code.

src/app/ui/key.cpp
```diff
@@ -439,48 +457,69 @@ void Key::add(const ui::Shortcut& shortcut, const KeySource source, KeyboardShor
...
+    // This key is for any context
+    (m_keycontext == KeyContext::Any) ||
+    // This key is for the same context
+    (m_keycontext == keyContext) ||
+    // Use Normal or SelectionTool keys in Transformation context
+    (keyContext == KeyContext::Transformation &&
+     (m_keycontext == KeyContext::SelectionTool || m_keycontext == KeyContext::Normal)) ||
+    // Use Normal keys in SelectionTool or FramesSelection contexts
+    ((keyContext == KeyContext::SelectionTool || keyContext == KeyContext::FramesSelection) &&
+     (m_keycontext == KeyContext::Normal));
```
ckaiser: This one's simpler but I'd still vote for returns

src/app/ui/key.h
```diff
@@ -103,12 +103,47 @@ inline KeyAction operator&(KeyAction a, KeyAction b)
...
+  AppShortcut(const KeySource source, const ui::Shortcut& shortcut)
+    : source(source)
+    , shortcut(shortcut)
+  {
+  }
+
+  bool isEmpty() const { return shortcut.isEmpty(); }
+  std::string toString() const { return shortcut.toString(); }
+  bool isPressed() const { return shortcut.isPressed(); }
+  bool isLooselyPressed() const { return shortcut.isLooselyPressed(); }
```
ckaiser: Is there a reason we're not just subclassing Shortcut?

src/app/ui/key_context.h
```diff
@@ -11,28 +11,61 @@
...
+//
+// This list should be sorted in such a way that more specific
+// contexts are below:
+//
+//   Normal > SelectionTool > Transformation
+//
+// This means that keys defined in SelectionTool are still valid in
+// Transformation context.
+//
+// Other key context are just temporal in the Editor, like a mouse
```
ckaiser:
```suggestion
// Other key context are just temporary in the Editor, like a mouse
```

src/app/ui/key_context.h
```diff
@@ -11,28 +11,61 @@
...
+// drag operation when we are transforming the
+// selection. E.g. TranslatingSelection, ScalingSelection, etc.
 enum class KeyContext {
+
+  // Special context to define keys. This is not be used as "current
```
ckaiser:
```suggestion
  // Special context to define keys. This is not used as a "current
```

src/app/ui/key_context.h
```diff
@@ -11,28 +11,61 @@
...
+  // key context", but just to define keys for the whole range of key
+  // context.
```
ckaiser:
```suggestion
  // contexts.
```

src/app/ui/key_context.h
```diff
@@ -11,28 +11,61 @@
...
   FreehandTool,
   ShapeTool,
+
+  // When the mouse wheel is used above the editor.
   MouseWheel,
+
+  // The user has a range of frames/cels selected in the timeline.
   FramesSelection,
+
+  // The user has moved the selection and didn't still dropped the
```
ckaiser:
```suggestion
  // The user has moved the selection but has yet to drop the
```

ckaiser (commented):
Looks great, having tests for this should reduce a bunch of headaches 🎉. I just had minor feedback and grammar stuff.

src/app/ui/key.h
dacap: I'll give a try. Actually it was a struct with several changes in the middle of the development of this PR (less functions, public access to shortcut) until I've needed to create the `ui::ShortcutsT` template and do several other refactors.

Issue #4423, opened 2024-04-18
Color management issues


dacap:
[When the color management (CM) was added](https://github.com/aseprite/aseprite/issues/1576), was mainly focused to render images with custom ICC correctly, and convert those profiles to sRGB. There are still issues handling monitor color profiles (because lack of time and hardware to fully test this in all platforms).

Here's a list of issues we've received so far about the current CM feature:

- One user reported that the issue was still persistent on macOS:
  - https://community.aseprite.org/t/color-display-wrong-on-mac/1970/9?u=dacap
  - https://community.aseprite.org/t/foreground-color-not-displaying-correctly/7985

- Some issues were solved but we don't know what the user exactly did
  (if enable CM or disable it):
  -  https://community.aseprite.org/t/colors-wrong-on-sidecar-ipad/6430

- Latest reports on Windows (color in color bar doesn't match color on canvas):
  - https://community.aseprite.org/t/color-on-palette-panel-didnt-same-as-drawing-on-the-canvas-when-use-color-management/15751
  - https://community.aseprite.org/t/wrong-colors-on-palette-selector-if-color-managed/21945

alwsmog:
Regarding latest reports on Windows:
[1] https://community.aseprite.org/t/wrong-colors-on-palette-selector-if-color-managed/21945
[2] https://community.aseprite.org/t/color-on-palette-panel-didnt-same-as-drawing-on-the-canvas-when-use-color-management/15751

I am no developer, so forgive me my naive considerations. Yet, I "understand" CM, thus hope my considerations may help

R E S U M E
- **Bug =** under **`Color Management (CM)` color in `Color Bar` * is _wrong_** & doesn't match _correct_ color in `Sprite Editor` (i.e. on `Canvas`)
  - *more specifically, in `Palette` & `Foreground/Background Color Boxes`
- **Bug Cause = color appearance for `Color Bar`** essentially undergoes **CM-adjustment procedure _twice_, instead of just once**
(see 5. for details)
- Bug is **significant**
  - sRGB-benchmarks and thus CM are required professionally
  - Bug partly ruins CM experience & limits features use
- Bug has **"simple" fix potential**
    - Correct rendering in `Sprite Editor` implies that incorporated `CM Module` is already realized properly with the only mentioned exception application

D E T A I L S
1. Applies to:
   - at least Windows 10, 11 (didn't test other OS)
   - all Aseprite v1.3+ (tested on v1.3.0, v1.3.5, v1.3.6)
[...]

PR #5414, opened 2025-09-03
Fix wrong color space in multiple UI elements

Gasparoken:


Gasparoken:
This fix is ​​complemented by https://github.com/aseprite/laf/pull/155

This solves:

1. https://community.aseprite.org/t/color-on-palette-panel-didnt-same-as-drawing-on-the-canvas-when-use-color-management/15751
2. https://community.aseprite.org/t/wrong-colors-on-palette-selector-if-color-managed/21945/6

Reported in https://github.com/aseprite/aseprite/issues/4423

dacap:
I was able to reproduce the bug in my old MacBook 2018 with the [webkit p3 test image](https://webkit.org/blog-files/color-gamut/comparison.html) (this is without this PR):

Screenshot 2025-09-10 at 18 26 46

I'll be testing your patch right now.

dacap:
It works great and fixes a lot of color issues :+1: I will merge this PR right now.

After this is merged I'll try to create a couple of PRs in laf & aseprite just to change the API a little and match the `SkPaint` API (e.g. having just one `os::Paint::color()` function where the `colorSpace = nullptr` by default).

dacap:
As an additional comment, I saw what you meant with the current color space (`get_current_color_space()`). I'll be creating other PR to change that to use the color space of the given displays (but I need to test this in multiple displays and in Windows/macOS/Linux).

dacap:
I'm working in a refactor to use the window color space, but I've noticed a new bug with a color difference between tabs and the workspace fill:

Screenshot 2025-09-12 at 14 57 16

I'll try to fix this too.

dacap:
Investigating a little more [this issue](https://github.com/aseprite/aseprite/pull/5414#issuecomment-3286339563) is because when we fill colors from the `theme.xml`, those colors are specified in sRGB color space (just like the `theme.png` file, which should be in sRGB). And when we fill colors for the palette, it should use the sprite color space. I'll see what we can do, but we need some changes in the `ui::Graphics` API to do this.

PR #5419, opened 2025-09-10
See defaults when the saved font info is empty or invalid (fix #5339)

### LINKED ISSUE #5339: Font Entry Issues
Some issues with the font entry widget that were found in #5323

- [ ] Stored pinned fonts can be empty or invalid, need error checking to ensure we're not saving or loading invalid strings.
- [ ] When the font search textbox is empty and the user selects the text tool, it doesn't display which font the user is using.
- [ ] When clearing the font size's editable combo box it doesn't show the current font size that the user is using.

ckaiser:
Attempts to fix the font selector showing an empty field and 0px when there's no text tool preferences saved (on first run)
image

Now it should show the font we're actually using, the theme one. I'm not sure if there's a cleaner way to do this, I spent a while going around the code to find a place where we could more cleanly get this but it seems like it's just using the widget font when there's nothing available, so I'm defaulting to that.

This also now checks at initialization time in the font selector if the font is available, otherwise deleting a font from the system would still show it as selected (and revert to the default theme font silently).

### DIFF
--- a/src/app/fonts/font_info.cpp
+++ b/src/app/fonts/font_info.cpp
@@ -10,9 +10,12 @@
 
 #include "app/fonts/font_info.h"
 
+#include "app/app.h"
 #include "app/fonts/font_data.h"
 #include "app/i18n/strings.h"
 #include "app/pref/preferences.h"
+#include "app/ui/main_window.h"
+#include "app/ui/skin/skin_theme.h"
 #include "base/fs.h"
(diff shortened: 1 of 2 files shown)


dacap:
It's not the more "elegant" solution but it works and fixes the issue :+1: The placeholder is a nice addition to show the font when we clean its name. I'm going to merge this.

PR #5421, opened 2025-09-11
Now Paint::color() can receive a color space w/any laf backend

dacap:
Little refactor of laf API for #5414


Gasparoken:
LGTM! Much clearer.

PR #5424, opened 2025-09-15
Fix painting UI theme colors as they are specified in sRGB color space

dacap:
Fixes a regression found after merging #5414:
https://github.com/aseprite/aseprite/pull/5414#issuecomment-3286339563

We always expected a sRGB color in ui::Graphics API and we can specify a color in another color space using the ui::Paint version of its member functions.

Several functions related to color spaces are now using a ui::Display to receive the specific display where we're going to paint, instead of using os::System::instance()->defaultWindow()->colorSpace().

### DIFF
--- a/src/app/color_spaces.cpp
+++ b/src/app/color_spaces.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2018-2024  Igara Studio S.A.
+// Copyright (C) 2018-2025  Igara Studio S.A.
 //
 // This program is distributed under the terms of
 // the End-User License Agreement for Aseprite.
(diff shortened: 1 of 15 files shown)


dacap:
Just in case, this doesn't fix other bug related to having multiple monitors with different color spaces. I'm working on other PR for that (after this one is merged).

Gasparoken:
It works much better! Just one note: the "fillRect" of the harmony swatches in the "ColorWheel" should somehow include the color space.
https://github.com/aseprite/aseprite/blob/5d5f3ec234f37760d1da06e7286ed7b43207a857/src/app/ui/color_wheel.cpp#L377

PR #5428, opened 2025-09-18
Avoid syntax errors reading extension preferences

### LINKED ISSUE #5412: Saving Unicode characters in `Preferences` causes Preferences File to be emptied on Aseprite Startup
**Describe the bug**
Trying to save a value in the `plugin.preferences` file with a value that contains a unicode character, or at the very least the charcter `°`, causes the preferences file to be completely emptied the next time Aseprite is started.

**To Reproduce**
1. Take this quick extension I made to demonstrate: [test_plugin.zip](https://github.com/user-attachments/files/22120182/test_plugin.zip)
```
local extension = {}

function init(plugin)
	plugin:newCommand{
		id="test_plugin",
		title="Run Debug Plugin",
		group= "file_scripts",
		onclick=function()
			 extension:run_debug(plugin)
			end,
		}
return end

function extension:run_debug(plugin)
	local testname = "degrees°"
	if plugin.preferences[testname] == nil then
		plugin.preferences[testname] = 90
		print("That value doesn't exist in the preference table!")
		print("Set _pref file value " .. testname .. " to " .. tostring(90))
	end

	print("Reading _pref file value " .. testname .. " as " .. tostring(plugin.preferences[testname]) )

return end

```

2. Run the plugin via `Scripts > Run Debug Plugin`
3. This will now create a preferences file, and attempt to save `degrees° = 90` to it.
4. Running the plugin again will show the value can still be read
5. Close Aseprite
6. Restart Aseprite
7. Run the plugin again
8. The value needs to be set again, because it is not loaded properly on startup. You can also see this by opening the preferences file after closing down Aseprite.

**Expected behavior**
I would expect that the characters and data would just be saved and loaded appropriately, since Aseprite seems to allow these characters in layernames, filenames, etc. And if it is not supported, it likely should alert the user that it could not save that data to the preferences file at all, rather than allowing it and subsequently making the file unparseable.

- Aseprite: **v1.3.15.2**
- System: Windows 10
- Extensions: A lot!

**Additional context**
I suspect JSON needs the unicode characters to be escaped properly to be able to read them during the loading. If it fails, it cannot parse, and it loses all the data in the preferences file. *(Which is real bad by the way, to the point that I have automatic daily backups for my Aseprite Plugin data, in case I manage to corrupt my JSON file, which has happened a LOT already, usually by me messing up though)*

I have only tested this with the ° symbol, but I suspect others might have the same issue.

martincapello:
Fix #5412 by using brackets and double quotes to enclose the name of the preferences' properties and avoid issues when they contain unicode chars

### DIFF
--- a/src/app/extensions.cpp
+++ b/src/app/extensions.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2020-2024  Igara Studio S.A.
+// Copyright (C) 2020-2025  Igara Studio S.A.
 // Copyright (C) 2017-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -555,7 +555,13 @@ static void serialize_table(lua_State* L, int idx, std::string& result)
     // Save key
     if (lua_type(L, -2) == LUA_TSTRING) {
       if (const char* k = lua_tostring(L, -2)) {
+        // Enclose the key name in brackets and double quotes to avoid syntax
+        // errors when the name has unicode characters.
+        result.push_back('[');
+        result.push_back('"');
         result += k;
+        result.push_back('"');
+        result.push_back(']');
         result.push_back('=');
       }
     }


dacap:
It works 👍 I'll merge this but make the `["..."]` conditional to add it only when it's required.

PR #5438, opened 2025-09-26
Fix move tool with Auto Select Layer isn't working (fix #4812)

### LINKED ISSUE #4812: Move tool with Auto Select Layer isn't working when composing groups separately
It looks like just the top layer in the group is selected.

Gasparoken:


martincapello (approved):
~~Code looks good and works like a charm! 🎖️~~

I'll review this a second time...just in case.

dacap:
The idea of the `ColorPicker` class was to pick colors and cels, i.e. to solve the logic side of eyedropper and move tools. Is there a way to fix the issue but keeping the `ColorPicker` usage? I'd prefer that instead of creating the `RenderPlan` and using `Sprite::pickCels` directly.

For example, I've reduced this PR (and #5439) to this patch:
```diff
diff --git a/src/app/color_picker.cpp b/src/app/color_picker.cpp
index b463a6fff..8e29386b4 100644
--- a/src/app/color_picker.cpp
+++ b/src/app/color_picker.cpp
@@ -104,7 +104,7 @@ void ColorPicker::pickColor(const Site& site,
   switch (mode) {
     // Pick from the composed image
     case FromComposition: {
-      doc::RenderPlan plan(pref.experimental.composeGroups());
+      doc::RenderPlan plan;
       plan.addLayer(sprite->root(), site.frame());
 
       doc::CelList cels;
```
But probably I'm missing something.

Gasparoken:
Hi @dacap. You're right about the solution proposed. I initially considered that solution, but I saw the code inside
`if (editor->isAutoSelectLayer())` in `StandbyState::onMouseDown` and the `color` obtained in the `picker` wasn't important (it's unnecessary to do a `render::get_sprite_pixel` in `ColorPicker::pickColor`). But if you're comfortable with that, of course I am too.
I think there are intermediate alternatives (like creating a new function in `ColorPicker` or passing an argument to `ColorPicker::pickColor` to skip `render::get_sprite_pixel`), but I think we're done with the modification you recommended. If there's anything that needs to be optimized, I could do it in the PR https://github.com/aseprite/aseprite/pull/5439

dacap:
I know the ColorPicker does an extra work to get the color, but it's not important (it's not a critical hot path). In a future we could introduce an extra parameter to do the cel picking only, but it doesn't worth it. I'll cherry pick the change and close both PRs.

Issue #3130, opened 2022-01-14
User data with multiple lines


dacap:
https://community.aseprite.org/t/userdata-enhancements/4831/1

Nrosa01:
How's this issue going? I'm reading the ui:Entry func to see if I can help someway to this issue as it has been open for a long while and would be really useful to have

ckaiser:
@Nrosa01 Hi! I'm working on creating a new multiline text entry widget at the moment, it's turning out to be a bit more complicated than I hoped it'd be but I'm making good progress and it should be available soon.

PR #5455, opened 2025-10-06
User data with multiple lines of text (new TextEdit widget) (#3130, #3131, #4743, #4968)

dacap:
This PR is a squashed and rebased version of #4968 to continue working in the multiple lines text editor.


dacap:
@ckaiser would you mind to give a test to this? I think all missing parts are implemented and it's working pretty well (even inserting emojis/using IMEs).

The only missing part is for RTL languages, but that is something true for `ui::Entry` too.

ckaiser:
Something I noticed, testing it on the user data input, fresh config, on macOS Tahoe:

It seems to not be excluded from the keyboard shortcut context, so I can't type capital letters since there's a bunch of Shift+[KEY] shortcuts, or even just the the C key (opens the canvas size selector)

This kinda makes it hard to test anything else since trying to type steals my focus and bounces me all over the place 😵

ckaiser:
Testing now on Windows:

 * Typing multiple lines and then selecting everything with Shift-A and pressing delete, leaves the caret in the second line, instead of at 0,0 as you'd expect, although I couldn't reproduce this consistently
 * The horizontal scrollbar isn't showing up when there's only one line (unless I create a new line with enter and then delete it)
 * Using shortcuts like ctrl-z undoes text but also undoes things on the canvas (same with pressing tab toggling the timeline)

dacap:
> It seems to not be excluded from the keyboard shortcut context, so I can't type capital letters since there's a bunch of Shift+[KEY] shortcuts, or even just the the C key (opens the canvas size selector)
> 
> This kinda makes it hard to test anything else since trying to type steals my focus and bounces me all over the place 😵

Nice catch. There is something to avoid this for ui::Entry, just added the missing part for ui::TextEdit: https://github.com/aseprite/aseprite/pull/5455/commits/050126b0d57c09fa459927b6802cbd2b0fdbdfe1

> * Typing multiple lines and then selecting everything with Shift-A and pressing delete, leaves the caret in the second line, instead of at 0,0 as you'd expect, although I couldn't reproduce this consistently

I wasn't able to reproduce it, but probably something to look in a future iteration.

> * The horizontal scrollbar isn't showing up when there's only one line (unless I create a new line with enter and then delete it)

Fixed 👍 (also fixed the whole scrolling/needed viewport size).

> * Using shortcuts like ctrl-z undoes text but also undoes things on the canvas (same with pressing tab toggling the timeline)

Known issue. It's a little hard to fix, as the undo is done for the whole sprite, the same happen in the (e.g.) Layer name field (ui::Entry fields). Something to fix in a future in some way 🤔 

I'm merging this into beta 🚀

PR #5459, opened 2025-10-09
Improve rotated rectangles and square brush drawing

martincapello:
This PR improves how the rectangles (made using any of the rectangle tool) and the square brush are drawn in rotated positions.


dacap:
It looks great 👍 I'll merge it, I'm not sure if it would make sense to move the `std::swap()` inside the `rotateRectangle` function in the future.

PR #5461, opened 2025-10-12
Fix rotate command creating copy of selection

### LINKED ISSUE #1330: Edit > Rotate > 90 CW/CCW work differently when using keyboard shortcuts
Hello!

Using v1.2-Beta3 (Steam) on Windows 10, I'm seeing something a bit weird when using custom keyboard shortcuts for Edit > Rotate > 90 CW/90CCW.

p_0

With the entire canvas selected, using the menu item rotates as expected:

p_1

However, using the keyboard shortcut seems to rotate a copy instead:

p_2

I've checked the keyboard shortcut configuration and something a little weird is that there are two actions which seem to share shortcuts (Edit > Rotate > 90 CW, and Rotate Selection 90°); if I edit the shortcut for one, the other changes to match.  I suspect triggering both actions simultaneously might be causing this.

p_3

JoshuaL03:
Fix #1330.

This PR fixes an issue where using a keyboard shortcut to rotate a selection would create a copy of the selection if certain keys were used. Copies are now only made if a handle is passed to the related code (i.e. rotation was performed through the GUI).

### DIFF
--- a/src/app/ui/editor/standby_state.cpp
+++ b/src/app/ui/editor/standby_state.cpp
@@ -749,9 +749,9 @@ void StandbyState::transformSelection(Editor* editor, MouseMessage* msg, HandleT
                                                         document->mask(),
                                                         "Transformation"));
 
-    // If the Ctrl key is pressed start dragging a copy of the selection
+    // If the Ctrl key is pressed and there is a handle, start dragging a copy of the selection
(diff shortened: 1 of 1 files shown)


dacap:
It works great @JoshuaL03, thanks for this patch. Yeah, it looks like the <kbd>Ctrl</kbd> key was creating the copy.

Issue #3558, opened 2022-10-13
User Customizable Double Clicking Behavior


Unknow0059:
Feature Request
In the Editor, when Double Clicking with a selection tool, a grid tile gets selected.
It would be convenient if the action taken upon double clicking was customizable in the keyboard shortcut menu.

### Aseprite and System version
1.2.34.1

* System: Windows 10

Xtendera:
Seems like a cool thing, I might mess around with it in a fork.

joaofranciscoguarda:
I've implemented this feature in [PR 5473](https://github.com/aseprite/aseprite/pull/5473), waiting for approval

PR #5473, opened 2025-10-16
Feature/double and triple click modifiers for shortcuts

joaofranciscoguarda:
# Keyboard Double/Triple-Click Shortcuts - Feature Contribution

## Overview

This contribution implements keyboard double-click and triple-click detection as modifiers for keyboard shortcuts in Aseprite. It also implements API for scripting and testing. 

## Testing

Run the test suite:

```bash
ASEPRITE=path/to/bin/aseprite ./tests/run-tests.sh shortcuts
```

Or manually test:

1. Open Aseprite
2. Go to Edit > Keyboard Shortcuts
3. Select any command
4. Press a key to assign
5. Check the "2x" or "3x" checkbox
6. Click OK
7. Rapidly press the key 2-3 times to trigger the shortcut
8. Reset values
9. Check out the shortcut for Set Ink Type: shading and simple, if they are still there if removed and reset to default.

## Demo:

Switch between Brush (Simple) -> Change color -> Brush (Shading) -> Change shading order -> Brush (Simples) using only ONE key

[aseprite-double-click-demo.webm](https://github.com/user-attachments/assets/1d21ae7e-6df0-49db-9ef3-a7ff28086dfc)

Mutually exclusive:

Screenshot_20251015_175340

## My idea:

I was struggling to make a toggle between simple brush and shading, the outcome for my idea using the key X was way beyond my original intention, because the X key is used for changing color and I didn't know that the X would change the shading order, with a single key this feature achieves a lot of possibilities and I believe that very experienced artists and Aseprite users will enjoy it! Allowing way more command while keeping one hand in the keyboard and another one in the pencil/mouse. Also, providing the API for scripting will allow users to share custom settings with the community. Thanks for your reading, I'll be available for any request, but I hope I did a good job and that this feature may be released soon.


dacap:
@joaofranciscoguarda it looks like #3558 asks to customize the mouse double-click and this PR adds the possibility to customize keys-repetitions (I'm not sure if I'd call it "click", as it's not common to call "click" to a "key press" event).

Gasparoken:
Congratulations @joaofranciscoguarda , very well documented PR!

joaofranciscoguarda:
A friend of mine was testing the feature, and I found an inconvenient **bug**  -> the double press also triggers the single press action. Because of that, I’m going to move this **PR to Draft for now.**

Bug demo:

[Screencast_20251017_165108.webm](https://github.com/user-attachments/assets/0c76c37e-d64b-4448-8af7-85fc370b08b3)

joaofranciscoguarda:
After a few tries, the best approach, in my opinion, is still the input delay to handle double and triple press down.

I believe that, as it will impact the whole aseprite, it would be better to put it as an experimental option, and when setting a key with a double/triple press down functionality to alert the user about this minimal delay, so only those who are willing to use it are going to be impacted by a delay. Another good thing would be to register the keys that are under the assignment of double/triple, so only those keys would be impacted.

I will be working on that soon.

joaofranciscoguarda:
# Keyboard Double/Triple-Click Detection - Experimental Feature

## Problem Solved
The original implementation triggered both single-click AND double-click events when a key was pressed twice quickly because:
1. First key press → Single-click event fired immediately
2. Second key press → Double-click modifier added
3. Result: Both events executed

Now it works as intended.

## Testing Scenarios

1. **Feature OFF**: All keyboard input works exactly as before 
2. **Feature ON - Single click**: 180ms delay, then fires once 
3. **Feature ON - Double click**: Only double-click fires, no single 
4. **Feature ON - Triple click**: Only triple-click fires, no double/single 
5. **Keyboard to mouse**: Pending keys flush on mouse click 
6. **Preferences change**: Updates Manager configuration in real-time 

## Migration Path
- **Existing users**: Feature OFF, zero impact
- **New users**: Discover feature through greyed-out 2x/3x checkboxes with helpful tooltips
- **Power users**: Enable feature, customize timings, create multi-click shortcuts

## Future Extensibility
- Per-key customization
---

Priorities: **user experience**, **backward compatibility**, and **discoverability**

---

@dacap As you mentioned before, what should be renamed for "press_down" instead of "click"? Only the internal code references or for the end user as well? When I get your answer I'll do the necessary changes for your approval, so we can continue with this feature and hopefully deploy it.

dacap:
Hi @joaofranciscoguarda, thanks for your contribution and work with this PR. Anyway I have several points about this:

1. This feature was not requested before by other users (#3558 is not about this). I've searched in other graphics software and I don't know if something similar is available and nobody asked for this feature before. So we don't have any survey about how many users would like to use this feature.
2. This gives me some doubts to add such complexity for a feature that wasn't requested before.
3. The description/comments in this PR looks generated by AI, not sure if that is true, but it looks like everything is over explained. Please don't do this for future PRs: Sometimes the source code is enough to explain what the PR does, and just a little paragraph about what is solved at a higher level is enough. Try to communicate the minimum possible to explain what your patches are about.

I'm sorry but I'm going to close this PR by now. If in the future other users ask for this feature, we can try to integrate it in some way.

joaofranciscoguarda:
Following Gasparoken instructions I will be creating the discussion in the Aseprite community.

Thank you so much for maintaining this awesome project for so long!

dacap:
Hi @joaofranciscoguarda, I'll comment here and in the PR: I'm not a big fan of this double-click / triple-click because it adds an extra delay for the single key case and I think it will give more problems of what it solves.

I'd happily merge a PR to integrate some kind of prefix-key if that helps you in your case. It's common in text editor, like <kbd>Ctrl+K</kbd> in VSCode, etc.

image

The idea is that any key could be defined as a prefix, but when it's a prefix, it cannot be used as a "single tap/click" case.

joaofranciscoguarda:
> Because it adds an extra delay for the single key case and I think it will give more problems of what it solves.

Totally agree, even though the feature I did is good for my case, it is not even close to perfect, because as my friend suggested and I wasn't able to implement it (because it would also take a lot of time), it needs to be bound to the keys that set this functionality, because while typing in a text field, it also has the delay, so yeah, terrible UX...

As a experimental feature, anyone using would be aware of potential weirdness, but this would never be a permanent solution, which makes it waste of time to implement. 

Anyways, the prefix key is a awesome alternative, because the possibility of making more combination with Alt or right Alt would already help, because is a key used by the thumbs, and this would provide multiple combinations which is awesome, and as you mentioned, would remove the delay. 

I’m not a genius in C++, BUT I’m a dev and I can try to implement your idea and have a “starting point” for the feature or maybe a almost final version for review, so it would save you some time, possibly a faster deploy and achieve my goal that is contributing significantly to the open-source projects that I use and like.

PR #5474, opened 2025-10-17
Fix pixel-perfect intertwine tile painting (fix #5394)

### LINKED ISSUE #5394: Cannot paint in flipped tiles correctly with pixel-perfect mode
Reported here: https://community.aseprite.org/t/bug-with-drawing-on-tiles/27126

In Manual mode when we paint in the flipped version of the tile it doesn't paint pixels correctly, example file: [flipped-tiles.zip](https://github.com/user-attachments/files/22044790/flipped-tiles.zip)

Image

I'm pretty sure that this was working originally, so I'm trying to find the regression (or probably it never worked?).

ckaiser:
Fixes #5394.

This fix was a bit trial and error because I'm not 100% sure how everything in the surrounded systems work, but the main thing was that the `forEachTilePos` was modifying the destination image directly and then that was being saved, which might've worked for non-flipped tiles and gone unnoticed, but I think goes against what the logic here is trying to do (using a temporary image and then restoring it).

### DIFF
--- a/src/app/tools/intertwiners.h
+++ b/src/app/tools/intertwiners.h
@@ -614,24 +614,25 @@ private:
     loop->getTiledModeHelper().wrapPosition(rgn);
     loop->getTiledModeHelper().collapseRegionByTiledMode(rgn);
 
-    for (auto a : rgn) {
+    for (auto& a : rgn) {
       a.offset(-loop->getCelOrigin());
+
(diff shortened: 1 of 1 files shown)


dacap:
It looks like the part of updating all tiles is fixed :+1: but the pixel perfect tool is still failing in some way when it's used from a flipped tile (probably peeking colors from a non-flipped region?):

pixelperfect

PR #5476, opened 2025-10-20
[lua] Add error message param to onreceive handler

martincapello:
Now script developer can define websocket's onreceive handlers with an additional parameter that holds error information when messages of type WebSocketMessageType.ERROR are received

NOTE: This PR was not requested explicitly, but I saw the opportunity to add this while trying to debug a script that used websockets.


dacap:
I'll merge it as it is.

It's a shame that the API design is not prepared for extensibility, i.e. just one "ev" argument with message type/data/error inside. I'll create a new issue to redesign the API in some way in the future.

martincapello:
Yeah, that would be nice. It would break current scripts though.

PR #5478, opened 2025-10-21
Add Widget Zoo, fix various widget consistency issues

ckaiser:
Adds a new command, WidgetZoo, to launch a window containing most of the usable widgets:

image

This is useful to check the consistency of the widgets through different states and variations (there's a toggle for normal/disabled/readonly), and to see how things look with different themes at a glance (there's a theme selector on the top-right).

The work creating this made a few issues come up and those have also been fixed in this PR:

 * `ExprEntry` widgets could be created with any "text" property through XML and that would remain until it received and lost focus and was thus validated, it now performs a validation when the first text is set.
 * `TextEdit`, `FilenameField` and `FontEntry` now have a `readonly` mode.
 * `FontEntry` now disables the size dropdown when there's no font set, instead of just showing an empty popup (shouldn't be an issue since in most places we show this the font is already pre-set to Aseprite by default anyway).
 * `TextEdit` now has the correct colors for disabled and readonly, and now supports placeholders, although they cannot be multi-line and do not affect the view/scrolling (if there's a place where we need to use this, then it can be tackled, but the implementation would add complexity that might not be necessary atm) 
 * The color consistency of `Entry` and `TextEdit` has been improved when switching through the different states:
   * Suffixes now aren't basically invisible when in readonly mode
   * Placeholders and suffixes don't show when the widget is disabled.
 * Fixed being able to paste text into disabled/readonly Entries.
 * Removed some deprecated/unused code from WidgetLoader.
 * Fixed a crash when setting the wrong grid column span (for example 4 on a table with 3 columns), now it's an exception.
 * Fixed `mini_button` not using the disabled text color

Setting this as draft to validate that these are the changes we want and the design of the window itself (it's very "developer-y" but since it's a hidden command you can only access through the runner or scripting I think that's fine IMO).

Only widget that isn't here is the `splitter`, ran into some weird issues when trying to use it relating to size, will look into it.


data/widgets/zoo.xml
```diff
@@ -0,0 +1,140 @@
+<!-- Aseprite -->
+<!-- Copyright (C) 2018-2025  Igara Studio S.A. -->
+<gui>
```
dacap: We can disable the i18n warnings as all strings are not translated and we can keep it in that way:
```suggestion
<gui i18nwarnings="false">
```

data/widgets/zoo.xml
```diff
@@ -0,0 +1,140 @@
...
+            <combobox id="static_combo_box" suffix="with stuffix" expansive="true">
+              <listitem text="Combobox" />
+              <listitem text="Combobox" />
+              <listitem text="Combobox" />
+            </combobox>
+            <combobox id="editable_combo_box" editable="true" expansive="true">
+              <listitem text="Editable Combobox" />
+              <listitem text="Editable Combobox" />
+              <listitem text="Editable Combobox" />
+            </combobox>
```
dacap: Not sure why but these comboboxes are too tall:

Image

src/app/commands/cmd_widget_zoo.cpp
```diff
@@ -0,0 +1,162 @@
...
+          changeReadOnly(false);
+        } break;
+        case 1: {
+          mainContainer()->setEnabled(false);
+          changeReadOnly(false);
+        } break;
+        case 2: {
+          mainContainer()->setEnabled(true);
+          changeReadOnly(true);
+        } break;
```
dacap:
```suggestion
        case 0:
          mainContainer()->setEnabled(true);
          changeReadOnly(false);
          break;
        case 1:
          mainContainer()->setEnabled(false);
          changeReadOnly(false);
          break;
        case 2:
          mainContainer()->setEnabled(true);
          changeReadOnly(true);
          break;
```

src/app/commands/cmd_widget_zoo.cpp
```diff
@@ -0,0 +1,162 @@
...
+      }
+      invalidate();
+    });
+
+    themeSelector()->Click.connect(&ZooWindow::showThemeMenu, this);
+
+    reset()->Click.connect([&] {
```
dacap: Catch only the `this` pointer
```suggestion
    reset()->Click.connect([this] {
```

src/app/commands/cmd_widget_zoo.cpp
```diff
@@ -0,0 +1,162 @@
...
+    for (const auto* extension : App::instance()->extensions()) {
+      if (!extension->isEnabled())
+        continue;
+
+      if (extension->themes().empty())
+        continue;
+
+      for (const auto& [id, info] : extension->themes()) {
+        auto* item = new MenuItem(id);
+        item->Click.connect([&, id] {
```
dacap: Prefer to capture as few variables as possible and use `&` or `=` only in extreme cases (i.e. there are a lot of vars to capture):
```suggestion
        item->Click.connect([theme, id] {
```

src/app/commands/cmd_widget_zoo.cpp
```diff
@@ -0,0 +1,162 @@
...
+  // Reset any themes/scaling to the default when leaving the window.
+  if (startingTheme != App::instance()->preferences().theme.selected() ||
+      guiscale() != startingScale) {
+    App::instance()->preferences().theme.selected(startingTheme);
+    ui::set_theme(theme, startingScale);
+  }
+
+  // Used when the user pressed the reset button, easier to just re-create everything.
+  if (window.shouldReset())
+    onExecute(context);
```
dacap: I think we can just
```c++
  bool retry;
  do {
    ZooWindow window;
    ...
    // Used when the user pressed the reset button, easier to just re-create everything.
    retry = window.shouldReset();
  } while (retry);
```
this code instead of recursive calls.

src/app/ui/expr_entry.cpp
```diff
@@ -25,15 +25,19 @@ ExprEntry::ExprEntry() : ui::Entry(1024, ""), m_decimals(0)
 {
 }
 
+void ExprEntry::validateText()
+{
+  std::string buf;
+  onFormatExprFocusLeave(buf);
+  if (text() != buf)
+    setTextQuiet(buf);
```
dacap: I see that `setText()` was changed by `setTextQuiet()` on `kFocusLeaveMessage`, this introduces several new bugs (as there is no `Entry::onSetText() / recalcCharBoxes()` call). E.g. File > New File, write "32+32" en Width: field and then unfocus and click the field:

Image

src/app/widget_loader.cpp
```diff
@@ -473,8 +440,12 @@ Widget* WidgetLoader::convertXmlElementToWidget(const XMLElement* elem,
   else if (elem_name == "buttonset") {
     const char* columns = elem->Attribute("columns");
 
-    if (!widget && columns)
-      widget = new ButtonSet(strtol(columns, NULL, 10));
+    if (!widget) {
+      if (columns)
+        widget = new ButtonSet(strtol(columns, NULL, 10));
```
dacap:
```suggestion
        widget = new ButtonSet(strtol(columns, nullptr, 10));
```

src/ui/grid.cpp
```diff
@@ -381,8 +382,11 @@ void Grid::expandStrip(std::vector<Strip>& colstrip,
           if (cell_span == current_span) {
             // Calculate the maximum (expand_count) in cell's columns.
             int max_expand_count = 0;
-            for (i = col; i < col + cell_span; ++i)
+            for (i = col; i < col + cell_span; ++i) {
+              if (colstrip.size() - 1 < i)
+                throw base::Exception("Grid cell failed to span correctly");
```
dacap: Could we add some extra information to the exception? (grid id? number of columns? why did this happen?)

src/ui/textedit.cpp
```diff
@@ -123,8 +139,9 @@ bool TextEdit::onProcessMessage(Message* msg)
     }
     case kFocusLeaveMessage: {
       stopTimer();
+      m_selection.clear();
```
dacap, line 143: We need to reset the `m_lockedSelectionStart` var too so when we click the TextEdit back we don't bring the selection back:
```suggestion
      m_lockedSelectionStart.clear();
      m_selection.clear();
```

src/ui/textedit.cpp
```diff
@@ -311,9 +339,17 @@ void TextEdit::onPaint(PaintEvent& ev)
     point.y += line.height;
   }
 
+  // TODO: Placeholder text should basically be normal text wrapped in proper lines
+  // and there should be flag.
+  if (text().empty() && !m_placeholder.empty() && isEnabled() && !isReadOnly())
+    g->drawText(m_placeholder,
+                m_colors.placeholderText.color(),
+                backgroundPaint.color(),
+                clientChildrenBounds().origin());
```
dacap:
```suggestion
  if (text().empty() && !m_placeholder.empty() && isEnabled() && !isReadOnly()) {
    g->drawText(m_placeholder,
                m_colors.placeholderText.color(),
                backgroundPaint.color(),
                clientChildrenBounds().origin());
  }
```

src/ui/theme.cpp
```diff
@@ -231,6 +231,8 @@ Theme::TextColors Theme::getTextColors(Widget* widget)
   c.background = Paint(kBgColor);
   c.selectedText = Paint(kBgColor);
   c.selectedBackground = Paint(kFgColor);
+  c.disabledText = Paint(kFgColor);
+  c.disabledBackground = Paint(kBgColor);
```
dacap, line 235: It looks like the `placeholderText` is not being initialized.

data/widgets/zoo.xml
ckaiser: It was `expansive="true"`, removing it fixes it

src/app/ui/expr_entry.cpp
ckaiser: I think I'm gonna leave out the changes to `ExprEntry` as out of scope for this PR, might be worth refactoring it in the future.

src/ui/grid.cpp
ckaiser: Made it slightly more clear, problem is I'm not 100% sure why it's crashing here since I haven't gone through the entire grid logic, but it does happen when there's an uneven amount of cells and there's an hspan, you can reproduce it in a release build (debug will assert much earlier) by adding two widgets before the font input.
```xml
          <label text="hello"></label>
          <label text="world"></label>
          
          <!-- Font -->
          <vbox expansive="true" cell_hspan="3">
```

ckaiser:
Rebased, updated and removed some stray things (textbox without view), ready for the full review.

dacap:
LGTM :+1:

PR #5480, opened 2025-10-22
Fix export file window not starting out centered

ckaiser:
We were missing the call to `centerWindow()` so when loading the window for the first time it defaulted to the top left.


dacap:
LGTM 👍

PR #5481, opened 2025-10-23
Fix dialog w/scrollbar wrong size in a particular case (fix #5396)

### LINKED ISSUE #5396: Modifying widget visibility on a dialog with autoscrolls shrink the dialog size
Reported here: https://community.aseprite.org/t/modifying-widget-visibility-with-autoscrollbars-shrinks-dialog-size/27128

Gasparoken:
fix #5396

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -32,6 +32,8 @@
 #include "base/fs.h"
 #include "base/paths.h"
 #include "base/remove_from_container.h"
+#include "os/screen.h"
+#include "os/system.h"
 #include "ui/app_state.h"
 #include "ui/box.h"
 #include "ui/button.h"
@@ -100,6 +102,8 @@ std::vector<Dialog*> all_dialogs;
 
 struct Dialog {
   DialogWindow window;
+  // Main view that holds the grid and the scrollbars
+  ui::View* view = nullptr;
   // Main grid that holds the dialog content.
   ui::Grid grid;
   // Pointer to current grid (might be the main grid or a tab's grid).
@@ -260,6 +264,20 @@ struct Dialog {
     }
   }
 
+  // Workarea of the screen which contains the dialog bounds
+  gfx::Rect getCurrentWorkarea()
+  {
+    os::Window* nativeWindow = parentDisplay()->nativeWindow();
+    const gfx::Point mainOrigin = (nativeWindow->contentRect() / nativeWindow->scale()).origin();
+    const gfx::Rect bounds = getWindowBounds().offset(mainOrigin);
+    const int scale = nativeWindow->scale();
+    os::ScreenList screens;
+    os::System::instance()->listScreens(screens);
+    for (const auto& screen : screens)
+      if ((screen->bounds() / scale).contains(bounds))
+        return screen->workarea() / scale;
+  }
+
   // TODO merge this code with add_scrollbars_if_needed() from
   //      ui/menu.cpp (creating a new function in the ui library)
   void addScrollbarsIfNeeded(const gfx::Rect& workarea, gfx::Rect& bounds)
@@ -294,13 +312,13 @@ struct Dialog {
     if (newRc == window.bounds())
       return;
 
-    View* view = new View();
-    view->InitTheme.connect([view] { view->noBorderNoChildSpacing(); });
+    if (!view)
+      view = new View();
+    view->InitTheme.connect([this] { this->view->noBorderNoChildSpacing(); });
     view->initTheme();
 
     if (vscrollbarsAdded) {
       int barWidth = view->verticalBar()->getBarWidth();
-      ;
       if (get_multiple_displays())
         barWidth *= window.display()->scale();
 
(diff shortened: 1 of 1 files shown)


src/ui/viewport.cpp
```diff
@@ -59,6 +59,11 @@ void Viewport::onResize(ResizeEvent& ev)
 void Viewport::onSizeHint(SizeHintEvent& ev)
 {
   ev.setSizeHint(gfx::Size(1 + border().width(), 1 + border().height()));
+  gfx::Size sz = ev.sizeHint();
+  for (auto child : children()) {
+    sz += child->sizeHint();
+    ev.setSizeHint(sz);
+  }
```
dacap: I'm not quite sure if we should change the viewport size here (as it affects all viewports in the app), and have make this change explicit so the user must change the minimum size of the parent View to specify a minimum size (check that several <view> elements in .xml widgets change the `minwidth/height` fields).

I'd prefer if we revert this change and in someway change the minimum size of the [view](https://github.com/aseprite/aseprite/blob/1faf96c2531ed1bd2c89b33da5a742bc88de7dd2/src/app/script/dialog_class.cpp#L297) explicitly (the view could be stored in a field of the `Dialog` class).

Just in case we need this change, we could do something like:
```
  gfx::Size sz(1, 1);
  for (auto* child : children())
    sz |= child->sizeHint();
  ev.setSizeHint(sz);
```
But I'm not sure to apply this change to `main` directly (as it might break something else)

src/app/script/dialog_class.cpp
```diff
@@ -1691,23 +1686,62 @@ int Dialog_modify(lua_State* L)
...
+          oldBounds = dlg->getWindowBounds().offset(mainWindowBounds.origin());
+          workarea = nativeWindow->screen()->workarea() / nativeWindow->scale();
+        }
+        else {
+          oldBounds = dlg->window.bounds();
+          workarea = App::instance()->mainWindow()->bounds();
+          mainWindowBounds = workarea;
+        }
+        const bool allowVResize = bool(dlg->autofit & ui::TOP) ^ bool(dlg->autofit & ui::BOTTOM);
+        const bool allowHResize = bool(dlg->autofit & ui::LEFT) ^ bool(dlg->autofit & ui::RIGHT);
```
dacap, line 1734: Is this right? This logic looks confusing, like it wants to say "allow vertical resize if top or bottom, **but not** both are enabled", but probably you want to specify it (a comment might help in these cases).

Should be this change to?
```suggestion
        const bool allowVResize = bool(dlg->autofit & ui::TOP) | bool(dlg->autofit & ui::BOTTOM);
        const bool allowHResize = bool(dlg->autofit & ui::LEFT) | bool(dlg->autofit & ui::RIGHT);
```

src/app/script/dialog_class.cpp
```diff
@@ -1691,23 +1686,62 @@ int Dialog_modify(lua_State* L)
...
+
+        gfx::Rect newBounds(oldBounds.x,
+                            oldBounds.y,
+                            allowHResize ? dlg->window.sizeHint().w : oldBounds.w,
+                            allowVResize ? dlg->window.sizeHint().h : oldBounds.h);
+
+        if (auto* view = dynamic_cast<View*>(View::getView(&dlg->grid))) {
```
dacap: Probably it'd be better to have the view pointer directly in the `Dialog` class and doing something like:
```suggestion
        if (View* view = dlg->view) {
```

src/app/script/dialog_class.cpp
```diff
@@ -1691,23 +1686,62 @@ int Dialog_modify(lua_State* L)
...
+          if (newBounds.y < workarea.y) {
+            newBounds.h = oldBounds.y2() - workarea.y;
+            newBounds.y = workarea.y;
+          }
 
+          // Restore newBounds refered to the mainWindows
+          if (get_multiple_displays())
+            newBounds.offset(-mainWindowBounds.origin());
+        }
         dlg->setWindowBounds(newBounds);
```
dacap: I'll test this tomorrow, I've detected that with multiples monitors it doesn't work as expected. Wouldn't be possible to use `fit_bounds()` function in this case?

src/app/script/dialog_class.cpp
Gasparoken: Maybe I'm confused in interpreting the[ API documentation](https://github.com/aseprite/api/blob/main/api/dialog.md#dialogautofit) that says: "...(eg. when autofitting with Align.LEFT | Align.TOP, the dialog bounds will shrink to the minimum size anchored at the top left of the dialog)". Therefore, I assume that if we anchor, for example, TOP and BOTTOM, the size would remain fixed given the anchoring concept. I'll need your confirmation this the behavior I'm describing is correct.
On the other hand, but following the logic of anchoring, how should a dialogue behave without Align?

dacap: It's as the API says, but never thought about the possibility of having TOP|BOTTOM or LEFT|RIGHT as anchors. The anchor should be one corner/edge (0, TOP|LEFT, TOP, TOP|RIGHT, LEFT, RIGHT, BOTTOM|LEFT, BOTTOM, BOTTOM|RIGHT), but probably we can support the case you say too.

Without autofit the default is TOP|LEFT (just changing the size without changing the origin).

dacap:
I have to test this a little more this, but I'm not sure if we should modify the dialog size when it has a view/autoscrollbars, as the user might have changed the dialog size to something preferred, and modifying that size is just making things worse (changing the user preference / viewport size).

dacap:
I've cherry-picked the commit in 71dd3835db576ecfcd60b84595adc787a24c393a (as laf has conflicts), and I'll continue from there.

dacap:
There are a couple of issues I'm trying to fix of this patch (now merged in main): 1) `getCurrentWorkarea()` doesn't return a rectangle on all paths, 2) a crash where dlg->view used but it's nullptr.

Issue #4758, opened 2024-11-03
Enabling Dialog:show property 'autoscrollbars' will reset previously set bounds


MapleGecko:
(Made for [v1.3.x (features for extensions)](https://github.com/aseprite/aseprite/milestone/59) milestone)

If a dialog's bounds were set before calling the [Dialog:show()](https://www.aseprite.org/api/dialog#dialogshow) function, they will be reset if the autoscrollbars property is set to true.

Setting the bounds using Dialog:show{bounds= Rectangle()} (or once the dialog is visible using dialog.bounds as a setter) works, but will cause this issue instead : #4757 

Example script:

``` Lua

function generateDialogMain()
	if DialogMain ~= nil then
		DialogMain:close()
	end
	if DialogTwo ~= nil then
		DialogTwo:close()
	end
	DialogMain = Dialog{
		id="DialogMain",
		title="DialogMain"
	}

	DialogMain:button{
		id="DialogMain_withbutton",
        text="With autoscrollbars",
        onclick=function()
			generateDialogTwo(true)
        end
	}
	DialogMain:newrow{}

	DialogMain:button{
		id="DialogMain_withoutbutton",
        text="Without autoscrollbars",
        onclick=function()
			generateDialogTwo(false)
        end
	}

	DialogMain:show{wait=false}

	return DialogMain
end

function generateDialogTwo(haveautoscrollbars)
	if DialogTwo ~= nil then
		DialogTwo:close()
	end
	DialogTwo = Dialog{
		id="DialogTwo",
		title="DialogTwo"
	}
	DialogTwo:button{
		id="DialogTwo_button",
        text="--- Button ---",
        onclick=function()
            print("Button was pressed")
        end
		
	}
	DialogTwo.bounds = Rectangle(200,10,400,200)

	if haveautoscrollbars then
		DialogTwo:show{wait=false, autoscrollbars=true}
	else
		DialogTwo:show{wait=false}
	end
end

generateDialogMain()
```

### Aseprite and System version

* Aseprite version: v1.3.9.2
* System: Windows

PR #5485, opened 2025-10-24
BrushPopup and Dialog autoscrollbars fixes

martincapello:
This PR contains 2 fixes in three commits:
- 126f42564cb5ff5a03ac3a3a169e6bd71d2acf6e and e2a77f8eed449eef2ae899d2b166a358a24af504 make the BrushPopup refresh when the "Saving Brush Here" button is pressed and an unlocked slot is present.
- 4c92828af11a33fa736df3ec3efbaf1e5a227512 fixes Dialog's behavior when autoscrollbars property were set to true in Lua scripts. This is because of the reversion of the fix_bounds change (126f42564cb5ff5a03ac3a3a169e6bd71d2acf6e), that made the dialogs to show up uncentered when `autoscrollbars=true`

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -470,7 +470,6 @@ int Dialog_show(lua_State* L)
   dlg->refShow(L);
 
   bool wait = true;
-  bool autoScrollbars = false;
   obs::scoped_connection conn;
   if (lua_istable(L, 2)) {
     int type = lua_getfield(L, 2, "wait");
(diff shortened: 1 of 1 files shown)


dacap:
How I can test both fixes? (step by step / script example)
Is the autoscrollbars fix related to #4758 or #5481?

martincapello:
> How I can test both fixes? (step by step / script example)

Try the following in the current Aseprite version:
BrushPopup not refreshing:
1. In any sprite you have open, click the "brush type" button.
2. In the BrushPopup set to "unlocked" to any custom brush you have.
3. Press the "Save Brush Here" button.
4. You will see that the "unlocked" custom brush is not updated.

Then, after applying **just the two first commits** try the above again. In step 4 you should see that the BrushPopup gets updated. Also try the following script to see the issue that will be fixed after applying the remaining commit:
```lua
dlg = Dialog { title="Test dialog" }
for i = 1, 39 do
  dlg:entry { label="Entry " .. i}
end

dlg:button    { text="Apply"}
   :button    { text="Cancel"}

dlg:show{wait = false, autoscrollbars=true}
```
Once run you will see that the dialog is positioned at (0,0) instead of being centered. Make sure you are using the multiple displays option. This is fixed after applying the last commit.

> Is the autoscrollbars fix related to https://github.com/aseprite/aseprite/issues/4758 or https://github.com/aseprite/aseprite/pull/5481?

No, it is not related to any of those issues, it is just a fix that makes the dialog with autoscrollbars behave as before reverting the fit_bounds change.

dacap:
> Try the following [...]

Thanks! I was able to reproduce the bugs and test the fixes. Looks great :+1:

PR #5490, opened 2025-10-26
Make slice selection more clear (fix #5484)

### LINKED ISSUE #5484: Slice combo box search field might be confusing
From https://community.aseprite.org/t/slice-tool-issues/27705 some users might think the combobox is useful to rename the slice name, while the field is just a search/filter field, not sure what would be the best here, but we could:

1. Add a 🔍 icon to indicate this is a search field
2. If the Slice property dialog is non-modal the user could change the name of slices more quickly

ckaiser:
Fixes #5484 for the most part, turning the `ComboBox` entry into a `SearchEntry` would be ideal but that'd require a big refactor of how they work, so this PR is a simpler fix:

 * Adds a placeholder to the entry
 * Enables the properties/delete icons only when we have a valid slice selected so we have a visual indicator

### DIFF
--- a/src/app/ui/context_bar.cpp
+++ b/src/app/ui/context_bar.cpp
@@ -1642,7 +1642,10 @@ class ContextBar::SliceFields : public HBox {
     SliceFields* m_sliceFields;
 
   public:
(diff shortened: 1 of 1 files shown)


dacap:
LGTM :+1:

Issue #5452, opened 2025-10-04
Add `Ignore Empty` toggle to `Export As..` window


tonycoculuzzi:
**Is your feature request related to a problem? Please describe.**
When using the `Export As..` window, there's no option to skip empty frames. Without this, `Export As..` exports all frames in the timeline.

**Describe the solution you'd like**
Adding an `Ignore Empty` checkbox to the `Export As..` window would be pretty useful. The export CLI supports an `--ignore-empty` argument, but to access it in the main GUI. To work it into your workflow, you need to write your own bash script or exporter. It would be nice if this feature was available to toggle in the GUI itself too.

**Describe alternatives you've considered**
I've written shell scripts and import pipeline scripts to export aseprite files through the cli, but it would be nice to not need to do that for this simple thing.

tonycoculuzzi:
You're all awesome, that went in so fast. Much appreciated!

dacap:
Thanks @tonycoculuzzi! This is thanks to @JoshuaL03!

PR #5491, opened 2025-10-26
Add ignore empty frames toggle to Export As window

JoshuaL03:
Feature #5452.

This PR adds a toggle to ignore empty frames when using the "Export As..." window. Enabling the setting will prevent all empty frames from being saved/exported.

### DIFF
--- a/src/app/commands/cmd_save_file.cpp
+++ b/src/app/commands/cmd_save_file.cpp
@@ -353,6 +353,7 @@ void SaveFileCopyAsCommand::onExecute(Context* context)
   doc::AniDir aniDirValue = params().aniDir();
   bool isPlaySubtags = params().playSubtags();
   bool isForTwitter = false;
+  bool isIgnoreEmpty = params().ignoreEmpty();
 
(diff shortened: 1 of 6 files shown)


src/app/commands/cmd_save_file.cpp
```diff
@@ -353,6 +353,7 @@ void SaveFileCopyAsCommand::onExecute(Context* context)
   doc::AniDir aniDirValue = params().aniDir();
   bool isPlaySubtags = params().playSubtags();
   bool isForTwitter = false;
+  bool isIgnoreEmpty = false;
```
dacap: We need to use the original `ignoreEmpty` parameter so running from CLI (and CLI tests) work:
```suggestion
  bool isIgnoreEmpty = params().ignoreEmpty();
```

JoshuaL03: Updated, thank you for the help!

dacap:
Thanks @JoshuaL03 👍 just merged

PR #5502, opened 2025-10-30
Add options to customize File Explorer thumbnails on Windows

### LINKED ISSUE #2606: Support for "Display file icon on thumbnail"

dacap:
Fix #2606

More info: https://github.com/aseprite/aseprite/issues/2606#issuecomment-3470162917

Add option to associate other file kind with Aseprite:

image


martincapello (approved):
Works like a charm! I didn't even have to restart Windows nor explorer.exe to make the "Display little Aseprite icon along the thumbnail" work 👏 
The file association worked as well 🥳

dacap:
> EDIT: Just one comment/thought...wouldn't it be nice to refactor the code in `src/app/win/file_associations.cpp`‎ to make it part of LAF? It seems something useful for developing applications. Anyway, I guess this can be done in the future if needed.

Probably in a future, as the `src/desktop` should go to another library too but it requires quite a refactor (GUIDs to be specified via cmake flags, generate Info.plists via cmake, etc.).

Issue #2986, opened 2021-10-05
Add option to disable tooltips


dacap:
It was asked a couple of times before.

ilulillirillion:
I've tried to switch to this tool many times but on my screen the tooltips take up a ton of room, pop up way too quickly (such that just mousing over a portion of the screen while traveling to another is sure to reveal all along the way) and 99% of them are giving hyper basic information. I get why these are useful, but if they're going to be this aggressive, a toggle would make a huge difference.

It was marked as high priority 4 years ago.

Issue #5353, opened 2025-08-14
Add a search field in Preferences dialog


dacap:
Just like in *Edit > Keyboard Shortcuts* or a possible command palette #740 in the future, it'd be nice to have a search field in *Edit > Preferences*.

SusilRamarao:
Hi Aseprite is a awesome tool, I have used it previously to make sprites for an android game. I am also a C++ engineer by profession in Bosch. I was interested to try test develop this feature. I have set the development env in local. 

Just curious, is it similar to how in visual studio code settings we have "search settings" bar where we can search for settings with the keywords?

PR #5503, opened 2025-10-31
Add AppTooltipManager with preferences-based delay

ckaiser:
One possible solution for #2986, adds custom delay support for `TooltipManager` and a new app-level `AppTooltipManager` that automatically sets this delay to the new preference `editor.tooltipDelay`, this allows the user to customize how fast they want all the app tooltips to appear.

https://github.com/user-attachments/assets/6639c847-e8f9-4e70-800a-5190743b77d8


ckaiser:
Rebased and updated:

* Moved to the experimental section, feels more natural there with all the other "advanced" UI things, maybe once a few of these graduate we can add an "Interface" section?
* Depending on if #5353 lands before or after this, it'll have to be changed to use the new app-tooltips in widget_loader

src/app/ui/app_tooltips.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "app/app.h"
+#include "app/pref/preferences.h"
+#include "ui/tooltips.h"
+
+namespace app {
+
+class AppTooltipManager : public ui::TooltipManager {
+public:
+  explicit AppTooltipManager()
```
dacap: The `explicit` keyword is only for copy ctors (AFAIR):
```suggestion
  AppTooltipManager()
```

src/app/ui/app_tooltips.h
```diff
@@ -0,0 +1,33 @@
...
+  {
+    int delay = App::instance()->preferences().experimental.tooltipDelay();
+    m_delayConn = App::instance()->preferences().experimental.tooltipDelay.AfterChange.connect(
+      [this] { setDelay(App::instance()->preferences().experimental.tooltipDelay()); });
+    setDelay(delay);
```
dacap: The lambda func and the ctor are doing the same, we can add a private member like `updateDelay()`:

```c++
  AppTooltipManager()
  {
     m_delayConn = App::instance()->preferences().experimental.tooltipDelay.AfterChange.connect(
      [this] { updateDelay(); });
     updateDelay();
  }

private:
  void updateDelay()
  {
    setDelay(App::instance()->preferences().experimental.tooltipDelay()); 
  }

```

src/app/ui/toolbar.cpp
```diff
@@ -89,6 +89,12 @@ ToolBar::ToolBar() : Widget(kGenericWidget), m_openedRecently(false), m_tipTimer
 {
   m_instance = this;
 
+  auto tooltipDelay = [&] {
+    m_tipTimer.setInterval(App::instance()->preferences().experimental.tooltipDelay());
+  };
```
dacap: We prefer capturing explicitly if there are few captures (even more if it's only `this`):
```suggestion
  auto tooltipDelay = [this] {
    m_tipTimer.setInterval(App::instance()->preferences().experimental.tooltipDelay());
  };
```

src/ui/tooltips.h
```diff
@@ -46,6 +47,7 @@ class TooltipManager : public Widget {
     TipInfo(const std::string& text, int arrowAlign) : text(text), arrowAlign(arrowAlign) {}
   };
 
+  int m_delay;
   typedef std::map<Widget*, TipInfo> Tips;
   Tips m_tips;                            // All tips.
```
dacap: Reorder to leave the typedef before all membres:
```c++
  using Tips = ...;

  int m_delay;
  Tips m_tips;
  ...
```

PR #5507, opened 2025-11-04
Add Preferences search (fix #5353)

### LINKED ISSUE #5353: Add a search field in Preferences dialog

ckaiser:
Fixes #5353, adds a search to the preferences menu like so:

(**Note**: Outdated video)

https://github.com/user-attachments/assets/8c292720-b55c-4edf-85d8-6e5a1064cc3c

### DIFF
--- a/data/widgets/keyboard_shortcuts.xml
+++ b/data/widgets/keyboard_shortcuts.xml
@@ -7,7 +7,7 @@
       <splitter horizontal="true" expansive="true" noborders="true" childspacing="2"
                 by="pixel" position="80" id="section_splitter">
         <vbox expansive="true">
-          <search id="search" magnet="true" />
+          <search id="search" magnet="true" clearonesc="true" debounce="250" />
           <view expansive="true">
             <listbox id="section" expansive="true">
               <listitem text="@.section_menus" />
--- a/src/ui/tooltips.cpp
+++ b/src/ui/tooltips.cpp
@@ -73,6 +73,15 @@ void TooltipManager::setDelay(int delay)
     m_timer->setInterval(m_delay);
 }
 
+std::string TooltipManager::getTooltipFor(Widget* widget)
+{
+  const auto it = m_tips.find(widget);
+  if (it != m_tips.end())
+    return m_tips[widget].text;
+
+  return {};
+}
+
 bool TooltipManager::onProcessMessage(Message* msg)
 {
   switch (msg->type()) {
--- a/data/extensions/aseprite-theme/dark/theme.xml
+++ b/data/extensions/aseprite-theme/dark/theme.xml
@@ -561,7 +561,7 @@
         <style id="list_header_label" padding="2">
             <text color="text" align="left" x="2"/>
         </style>
-        <style id="link" padding="1">
+        <style id="link" extends="label" padding="1">
             <text color="link_text" align="left"/>
             <text color="link_hover" align="left" state="mouse"/>
         </style>
--- a/data/extensions/aseprite-theme/theme.xml
+++ b/data/extensions/aseprite-theme/theme.xml
@@ -554,7 +554,7 @@
         <style id="list_header_label" padding="2">
             <text color="text" align="left" x="2"/>
         </style>
-        <style id="link" padding="1">
+        <style id="link" extends="label" padding="1">
             <text color="link_text" align="left"/>
             <text color="link_hover" align="left" state="mouse"/>
         </style>
--- a/data/widgets/options.xml
+++ b/data/widgets/options.xml
@@ -5,31 +5,35 @@
   <window id="options" text="@.title" help="preferences">
   <vbox>
     <hbox expansive="true">
-      <view maxsize="true">
-        <listbox id="section_listbox">
-          <listitem text="@.section_general" value="section_general" />
-          <listitem text="@.section_files" value="section_files" />
-          <listitem text="@.section_color" value="section_color" />
-          <listitem text="@.section_alerts" value="section_alerts" />
-          <listitem text="@.section_editor" value="section_editor" />
-          <listitem text="@.section_selection" value="section_selection" />
-          <listitem text="@.section_timeline" value="section_timeline" />
-          <listitem text="@.section_cursors" value="section_cursors" />
-          <listitem text="@.section_background" value="section_bg" />
-          <listitem text="@.section_grid" value="section_grid" />
-          <listitem text="@.section_guides_and_slices" value="section_guides_and_slices" />
-          <listitem text="@.section_undo" value="section_undo" />
-          <listitem text="@.section_theme" value="section_theme" />
-          <listitem text="@.section_extensions" value="section_extensions" />
-          <listitem text="@.section_aseprite_format" value="section_aseprite_format" />
-          <listitem text="@.section_experimental" value="section_experimental" />
-          <separator text="@.section_windows_specific" horizontal="true" style="separator_in_view" id="section_windows_specific_separator" />
-          <listitem text="@.section_tablet" value="section_tablet" />
-          <listitem text="@.section_file_explorer" value="section_file_explorer" />
-          <separator horizontal="true" style="separator_in_view" />
-          <listitem text="@general.reset" value="section_reset" />
-        </listbox>
-      </view>
+
+      <vbox expansive="true">
+        <search id="search" magnet="true" clearonesc="true" debounce="250" />
+        <view maxsize="true" expansive="true">
+          <listbox id="section_listbox">
+            <listitem text="@.section_general" value="section_general" />
+            <listitem text="@.section_files" value="section_files" />
+            <listitem text="@.section_color" value="section_color" />
+            <listitem text="@.section_alerts" value="section_alerts" />
+            <listitem text="@.section_editor" value="section_editor" />
+            <listitem text="@.section_selection" value="section_selection" />
+            <listitem text="@.section_timeline" value="section_timeline" />
+            <listitem text="@.section_cursors" value="section_cursors" />
+            <listitem text="@.section_background" value="section_bg" />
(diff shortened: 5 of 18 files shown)


Gasparoken:
Great, cool feature! Some comments:
- The Dirty flag hack doesn't work in some cases (try searching for "p" and then go to the Timeline section; the "First Default Frame" label will be incorrectly enabled). It might be better to collect all Label widgets that have buddies during the "for" loop, then reprocess the labels collection to determine whether the label and its corresponding buddy should be disabled or not (I don't think the DIRTY flag will be needed. Just check if the label or its associated buddy is enabled; if so, both will be enabled).
- Please rebase the PR to the latest commits.

ckaiser:
Removed the use of the `DIRTY` flag and fixed some corner cases.

Gasparoken:
Great!! I didn't mention before but `Slider` and `ColorButton` widgets should be disabled too. I found that disable all `Slider` and `ColorButton` widget before to process the labels will be enough (similar case that`uiWindows()->setEnabled(false);`
```cpp
        // Always disable the Slider and ColorButton widgets; they will be enabled through
        // their label.
        if (widget && widget->type() == kSliderWidget || widget->type() == ColorButton::colorbutton_type())
          widget->setEnabled(false);
```
`ColorButton::colorbutton_type()` isn't accesible, I moved `colorbutton_type()` to the header file inside `ColorButton` class. This would require approval from @dacap.

dacap:
In case we need access to `colorbutton_type()`, you can refactor it to `static WidgetType ColorButton::Type()` as a public member of `ColorButton`.

ckaiser:
Fixed some issues with the disabling/re-enabling and added a disabled state to the `ColorButton` basically making it greyscale, since it had no disabled feedback.

@dacap Made a change to Slider so it's also setting the `text()` to the value, I think it makes sense semantically but I could also just work around that in my code and not change a core class.

ckaiser:
Fixed accidentally breaking everything (oops) and added two new things:

 * Made the search a focus magnet like in keyboard shortcuts
 * `LinkLabel` now inherits the regular label style which means they have a "disabled" color
 * Added the `closeOnEsc` property to `SearchEntry`, which came from me trying to naturally "clear" the search using the escape key. Made it optional by default and I've added it for big windows with a search like this, keyboard shortcuts and the command runner.

Gasparoken:
Almost perfect, but...:
```cpp
          if (label->isEnabled()) {
            label->buddy()->setEnabled(true);

            // Special case to handle buttonsets
            if (label->buddy()->type() == kGridWidget) {
              for (auto* child : label->buddy()->children())
                child->setEnabled(true);
            }
          }
          else if (label->buddy()->isEnabled() && !label->buddy()->text().empty()) {
            label->setEnabled(true);
          }
```
I found that the ComboBox labels are disabled regardless, even though the string "pn" belongs to one of the available options ("png") in the default file association ComboBox under "Files" section.

The problem lies in the following "else if" statement:
```cpp
  else if (label->buddy()->isEnabled() && !label->buddy()->text().empty())
```
In the mentioned ComboBox `label->buddy()->isEnabled()` is "true", however, `!label->buddy()->text().empty()` is "false" since `!label->buddy()` is a ComboBox type with no text. Finally, in the UI, the ComboBox label will incorrectly change to a disabled state.

ckaiser:
@Gasparoken Fixed! It was old code that I forgot to take out when I changed the approach to combo boxes 🤦

Gasparoken:
I think it's fine now!
Just one observation for the future: When I search for "co", the "Colors:" label in the "Background" section has disabled buttons, but they should be enabled. This happens because the "Colors:" label doesn't have a specific "buddy" (it's in nullptr). I see that this could be resolved with one of two alternatives:
1. My preferred solution would be to allow "Labels" to have a "Grid" as their "buddy", so `label->buddy()` could return an array with all the displayable widgets.
2. Address this specific case in some way, since it only happens in this instance.

In any case, I think it could be mixed in as is, and a separate issue could be opened for this detail.

ckaiser:
Yeah, ideally I think we should be searching and highlighting "rows" of widgets but that required a way to tag them or completely reorganize the entire XML somehow to get that to work, so I went for the approach that had the least friction with the current code and layout. I'll see about a way to either set the buddy for this or handle the case specifically.

ckaiser:
### Latest polish pass

 * Adjusted the color buttons so the text would also be slightly transparent to more accurately portray the fact it's disabled
 * Added a `setEnabled` call to correspond to most of the other `setVisible` calls in the code, this way we're not accidentally searching inside things like the font warning, OS things that don't exist for the current platform, etc.
 * Added a `resetSearch` function and started calling on some disruptive events like things that spawn extra controls (custom cursor color, etc).
 * Added exceptions for the new document/active document combo boxes.
 * Added some hboxes with IDs for things that are closely connected, this way if you search for the label, you'll be able to see all the relevant things highlighted instead of just the first one, for example:
image


### Related to this
@dacap I found some weird behavior regarding the checkered background size option, presumably complicated by the logic of it having to switch between active or new documents. Right now you can't change this option and then switch to another section without losing what you did, which feels really inconsistent with other options, this is happening in the current stable release, figured I'd mention it, seems like the logic handling this could use a refactor.

dacap:
Found this bug giving a try (cannot change font after a search and canceling the search):

search-pref-bug

I've also noticed that we have to cancel the search to enable some item that we might click. Probably something to see in the future how we could fix it (e.g. just clicking an element that is "temporarily disabled" cancels the search / or re-enable all items in the section).

dacap:
> @dacap I found some weird behavior regarding the checkered background size option, presumably complicated by the logic of it having to switch between active or new documents. Right now you can't change this option and then switch to another section without losing what you did, which feels really inconsistent with other options, this is happening in the current stable release, figured I'd mention it, seems like the logic handling this could use a refactor.

We can leave a fix for the future, probably with a new "Set as default" option, but it's other issue.

ckaiser:
Fixed the weird issues, it was a byproduct of me relying on `isEnabled()` instead of using the raw flags, now things work more smoothly.

With regards to what happens when you click on things, my first solution was to just reset the search manually in all of the different `onSomething[..]()` functions, for things that might have a potential side effect, but that presents too many limitations and weird quirks, so my new solution is a filter where any focus change that isn't to the section list or the search box will just clear the search and reset things to normal. This works with mouse and keyboard selection as well.

Presumably by focusing on something specific the user has finished searching and this works pretty well because the focus message runs before anything else

Other changes:
 * Simplified code a fair bit
 * Now using the fuzzy search from the runner (although a bit more strict)
 * We can search widget tooltips (@dacap  This PR adds a way for generated widgets to access tooltip managers, figured this was the cleanest way to implement it, lmk what you think).
 * Labels are better at finding their buddies, and at asserting when they can't (this was the cause of some issues)

ckaiser:
Added a cache of the lower case searchable text for the widgets, this speeds up the processing on each keystroke by around 40%~, I've also added debouncing to `SearchEntry` which further minimizes the load at the expense of a bit of latency, but it should be helpful on lower-end systems and not particularly noticeable otherwise. Also renamed `closeOnEsc` to `clearOnEsc` since that's what it's doing.

dacap:
I've rebased this to latest beta to avoid conflicts with the merged PR #5503

You can find the rebased version here https://github.com/dacap/aseprite/tree/preferences-search in case you want to move your own branch. Anyway there are some comments to make.

data/widgets/keyboard_shortcuts.xml
```diff
@@ -7,7 +7,7 @@
       <splitter horizontal="true" expansive="true" noborders="true" childspacing="2"
                 by="pixel" position="80" id="section_splitter">
         <vbox expansive="true">
-          <search id="search" magnet="true" />
+          <search id="search" magnet="true" clearonesc="true" debounce="250" />(
```
dacap: An extra `(`, it looks like a typo
```suggestion
          <search id="search" magnet="true" clearonesc="true" debounce="250" />
```

data/widgets/run_command.xml
```diff
@@ -3,7 +3,7 @@
 <gui>
 <window id="run_command" minwidth="256" maxwidth="256" text="@.title" help="run-command">
   <vbox>
-    <search id="search" placeholder="@.command_placeholder" />
+    <search id="search" placeholder="@.command_placeholder" clearonesc="true" />
```
dacap: Probably this search field shouldn't `clearonesc`, so <kbd>Esc</kbd> just close the Run dialog, just like in VSCode if we <kbd>Ctrl+Shift+P</kbd>, write something, and press <kbd>Esc</kbd>.

dacap:
> @dacap Made a change to Slider so it's also setting the `text()` to the value, I think it makes sense semantically but I could also just work around that in my code and not change a core class.

I'm not sure why [this change](https://github.com/aseprite/aseprite/pull/5507/changes#diff-cf981206c9f7c13f97d033e3a4af7027eb29c7f96c82f31cc81c2081af73b4daR68) was required. Is it necessary for this feature?

src/ui/tooltips.cpp
```diff
@@ -67,6 +67,15 @@ void TooltipManager::removeTooltipFor(Widget* widget)
     m_tips.erase(it);
 }
 
+std::string TooltipManager::getTooltipFor(Widget* widget)
+{
+  const auto it = m_tips.find(widget);
+  if (it != m_tips.end())
+    return m_tips[widget].text;
+
+  return "";
```
dacap:
```suggestion
  return {};
```

dacap:
Looks great @ckaiser 👍 I'll merge it right now, and make a little change in the Grid section that it looks something has changed.

PR #5524, opened 2025-11-09
Vcpkg demonstration

winterheart:
This PR demonstrates vcpkg usage on external dependency handling. See #5313 for rationale.

I agree that my contributions are licensed under the Individual Contributor License Agreement V4.0 ("CLA") as stated in https://github.com/igarastudio/cla/blob/main/cla.md

I have signed the CLA following the steps given in https://github.com/igarastudio/cla#signing

Issue #5313, opened 2025-07-31
Proposals for handling external dependencies


winterheart:
Hello.

I'd like to propose a new way to handling external dependencies by managing by [CPM.cmake](https://github.com/cpm-cmake/CPM.cmake) module. This module simplifies fetching and building cmake-based libraries just declaring it as like:

```
CPMAddPackage("gh:fmtlib/fmt@7.1.3")
```

This approach allows to use unified and clean external dependencies handling. See PR #5312 for demonstration.

dacap:
Thanks @winterheart for sharing this. At the moment we're not interested to change the way third-party libraries are handled, but we might check this later in a future if it's required.

dacap:
Something to check later (and might be another option to replace dependency management): https://github.blog/changelog/2025-08-12-dependabot-version-updates-now-support-vcpkg/ (it's something released today).

winterheart:
Well, vcpkg is good option too.
I can prepare some demo on integration vcpkg for dependency management.

dacap:
@winterheart if you would like to give a try to test vcpkg with a couple of packages you can go forward; we'd need a way to compile it with vcpkg or without it: just doing as we do right now like "submodules", but probably replacing "submodules" with some "clone third-party" repo script (I'm not sure what would be the best way to bypass vcpkg and use our own clones).

winterheart:
See #5524 for demonstration.

winterheart:
@dacap any news about vcpkg integration?

dacap:
@winterheart I have to test it (and reopen the PR, we have an automatic action that closes any PR that want to change the actions, even if I'm the author 🫠)

winterheart:
@dacap Any feedback? I see that PR #5565 passed checks, but closed without additional comments. Can I go further?

dacap:
Hi @winterheart, sorry but right now we are working in the backend (website) a little so the desktop is in standby at the moment. I’ll check this later with more time.

dacap:
I'm trying to do a little of refactor to handle third party dependencies, but mainly to group all the directives in one .cmake file: https://github.com/dacap/aseprite/tree/refactor-third-party (still a work-in-progress/force-pushing a lot).

PR #5525, opened 2025-11-10
Trying vcpkg to manage dependencies

dacap:
Thanks to @winterheart (this PR was re-created from #5524) to test a possible vcpkg integration to manage dependencies. I'd like to give a try to this possibility to check if we can reduce compilation times (and share cached compiled modules between different builds/reduce disk space).

See #5313 for more information.


winterheart:
Currently built-in vcpkg binary caching is probably broken, but some workarounds exists (https://github.com/lukka/run-cmake/issues/152#issuecomment-2833729132).

PR #5526, opened 2025-11-10
Show an error in the console when dropping an image with an unsupported format

### LINKED ISSUE #5514: Crash when an image or path is dropped but in the callback there is no surface (?)
Sentry Issue: [ASEPRITE-3E4](https://igara-studio.sentry.io/issues/6751352995/?referrer=github_integration)

```
OS Version: Windows 10.0.26200 (6899)
Report Version: 104

Crashed Thread: 33808

Application Specific Information:
Fatal Error: EXCEPTION_ACCESS_VIOLATION_READ / 0x0

Thread 33808 Crashed:
0   Aseprite.exe                    0x7ff7d5e46d88      <lambda>::operator() (timeline.cpp:4637)
1   Aseprite.exe                    0x7ff7d60e9cbb      [inlined] std::_Func_class<T>::operator() (functional:926)
2   Aseprite.exe                    0x7ff7d60e9cbb      [inlined] os::Event::execCallback (event.h:140)
3   Aseprite.exe                    0x7ff7d60e9cbb      ui::Manager::generateMessagesFromOSEvents (manager.cpp:600)
4   Aseprite.exe                    0x7ff7d60e937a      ui::Manager::generateMessages (manager.cpp:364)
5   Aseprite.exe                    0x7ff7d611a170      ui::MessageLoop::pumpMessages (message_loop.cpp:24)
6   Aseprite.exe                    0x7ff7d60ec209      ui::Manager::run (manager.cpp:312)
7   Aseprite.exe                    0x7ff7d5ae45ab      app::App::run (app.cpp:566)
8   Aseprite.exe                    0x7ff7d5ada49c      app_main (main.cpp:138)
9   Aseprite.exe                    0x7ff7d636fd73      wWinMain (main.cpp:40)
10  Aseprite.exe                    0x7ff7d62c90b1      [inlined] invoke_main (exe_common.inl:118)
11  Aseprite.exe                    0x7ff7d62c90b1      __scrt_common_main_seh (exe_common.inl:288)
12  KERNEL32.DLL                    0x7ffc8fbae8d6      BaseThreadInitThunk
13  ntdll.dll                       0x7ffc9102c53b      RtlUserThreadStart
```

martincapello:
Fix #5514

### DIFF
--- a/src/app/console.cpp
+++ b/src/app/console.cpp
@@ -269,6 +269,8 @@ void Console::printf(const char* format, ...)
 
   // Update the textbox
   m_console->addMessage(msg);
+  m_console->invalidate();
+  m_console->flushRedraw();
 }
 
 // static


src/app/ui/timeline/timeline.cpp
```diff
@@ -4629,25 +4629,37 @@ void Timeline::onDrop(ui::DragEvent& e)
...
+        tx.commit();
+        m_document->notifyGeneralUpdate();
       }
-      else {
-        txmsg = "Drop File";
-        docProvider = std::make_unique<DocProviderFromPaths>(m_document->context(), paths);
+      catch (const std::exception& e) {
+        Console::showException(e);
+        Console::window()->invalidate();
+        Console::window()->flushRedraw();
```
dacap: I think we shouldn't give access to the internal console window pointer (at least not for this) and just fix the `showException()` if it needs these function calls inside.

src/app/console.h
```diff
@@ -23,6 +27,7 @@ class Console {
 
   static void showException(const std::exception& e);
   static void notifyNewDisplayConfiguration();
+  static ui::Window* window();
```
martincapello: I wasn't sure what would be better...if creating a redraw method here instead of exposing the ui::Window of the console...or just create this method and call what I needed. Let me know if you prefer the former.

src/app/ui/timeline/timeline.cpp
martincapello: The thing is that it is only needed for this case I believe.

martincapello: Well, actually it is needed for any case where the showException is called from within the onDrop method. For instance, try to drag and drop a pdf file into the timeline. You will see that the console shows up, but it is empty until you click on it. On the other hand, if you drag and drop a pdf into Aseprite, it will show the console window properly.

martincapello: So, if adding those lines is fine for all cases, then I'll do it.

dacap: Test if it makes sense to add those lines on all cases.

dacap:
I wasn't able to test this, so merging as it is.

PR #5527, opened 2025-11-10
Add key bindable "Hide other layers" command (fix #5291)

### LINKED ISSUE #5291: Feature - Native "Hide all layers but selected" toggle
This is a request to have a native option to add a (keybindable) option under View or Layer that hides all layers but the selected layer, similar to how Blockbench has this feature.

I made a script that has this functionality, but it's not ideal as layers have to be renamed in order to remember which ones to hide and when to hide and unhide.

https://github.com/phantomeye1931/Aseprite-Layer-Focus-Toggle

Example in Blockbench:

Image

Example in Aseprite with script:

Image

Gasparoken:


src/app/commands/toggle_other_layers_visibility.cpp
```diff
@@ -0,0 +1,133 @@
...
+};
+
+ToggleOtherLayersVisibilityCommand::ToggleOtherLayersVisibilityCommand()
+  : CommandWithNewParams<ToggleOtherLayersVisibilityParams>(
+      CommandId::ToggleOtherLayersVisibility(),
+      CmdUIOnlyFlag)
+{
+}
+
+void showChildrenLayers(const LayerGroup* layerGroup, Doc* doc)
```
dacap: It looks like a static standalone function used only in this file:
```suggestion
static void show_children_layers(const LayerGroup* layerGroup, Doc* doc)
```

dacap:
Going to rename the command to SoloLayer + restore the possibility to soloing several layers dragging the mouse with Alt key (this was lost with this PR).

PR #5529, opened 2025-11-11
Add power of two sheet size flag on CLI 

### LINKED ISSUE #2289: [CLI] [feature request] Power of two sheet size flag
Aseprite v1.2.15-x64 and earlier seemed to always export sprite atlases with dimensions that were powers of two (e.g., 128, 256, ...). This was very useful happenstance behavior for WebGL v1 applications as textures must otherwise be clamped. As of v1.2.16.3-x64, I've seen packed textures output with non-powers of two (e.g., a height of 184px). This ticket is a feature request to add a command line option to always round up to the nearest power of two width and height dimensions (e.g., a 100x155 image would round up to 128x256) when packing.

If the flag is broadly applicable, maybe it could be called `--round-to-power-of-two-sizes`. If the flag was only specific to sheets, maybe it could be called `--sheet-pack-power-of-two` and be used instead of `--sheet-pack`. Then again, maybe it makes more sense to have a `--sheet-width` / `--sheet-height`-like option called `--sheet-pow2`.

The command used is (sample input, [flag.zip](https://github.com/aseprite/aseprite/files/4714796/flag.zip)):

```sh
$ aseprite -b --sheet-pack --sheet atlas.png flag.aseprite | grep size
  "size": { "w": 6, "h": 8 },
```

Thank you for writing awesome software :+1:

* Aseprite version: v1.2.16.3-x64, v1.2.18-x64, 1.2.19.2-x64
* System: Ubuntu v19.10

Gasparoken:
Fix #2289

### DIFF
--- a/tests/cli/sheet.sh
+++ b/tests/cli/sheet.sh
@@ -451,3 +451,33 @@ end
 EOF
 $ASEPRITE -b -script-param file=$d/tilesets.png \
           -script "$d/compare.lua" || exit 1
+
+
+# Testing -power-of-two-size flag
+# https://github.com/aseprite/aseprite/issues/2289
+d=$t/issue-2289
+# Without this flag:
+$ASEPRITE -b "sprites/point4frames.aseprite" \
+          -trim \
+          -sheet-pack \
+          -data "$d/atlas.json" \
+          -sheet "$d/sheet.png" || exit 1
+cat >$d/size_verification.lua <<EOF
+local sheet = Sprite{ fromFile="$d/sheet.png" }
+assert(sheet.width == 10)
+assert(sheet.height == 5)
+EOF
+$ASEPRITE -b -script "$d/size_verification.lua" || exit 1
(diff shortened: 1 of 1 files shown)


src/app/cli/app_options.cpp
```diff
@@ -77,6 +77,8 @@ AppOptions::AppOptions(int argc, const char* argv[])
       m_po.add("sheet-width").requiresValue("<pixels>").description("Sprite sheet width"))
   , m_sheetHeight(
       m_po.add("sheet-height").requiresValue("<pixels>").description("Sprite sheet height"))
+  , m_powerOfTwoSize(m_po.add("power-of-two-size")
+                       .description("Round up the size of the sprite sheet to a power of 2"))
```
martincapello: I think I would change the description for something like:
"Round up the width and height of the sprite sheet to the nearest power of 2"

Gasparoken:
I'll add a simple automatic test

PR #5540, opened 2025-11-17
Add snap line to more angles (fix #1641)

### LINKED ISSUE #1641: Ability to snap line tool to more angles

Gasparoken:
This PR is a cherry-pick and updates the original PR [#4241](https://github.com/aseprite/aseprite/pull/4241) from [tsone](https://github.com/tsone)
fix #1641


Gasparoken:
I'll review the tests that are failing.
I can change `MAX_M` from 8 to 2 to allow immediate inclusion of this pull request, then I can send a new commit that will add an additional checkbox to enable/disable more perfect slopes in _Edit > Preferences > Editor > Line Tool with more snap slopes_ (disabled by default).

dacap:
I think we talked about adding an option to enable this. Probably I'm missing something from the PR.

martincapello (dismissed):
It works great! The only thing that I was just thinking about is that the last step is the only step that might not be the same size as the others...so I was wondering if we should force the last step size to be the same as the others, so the line would always have all their steps of the same size...just something to think about. Anyway I'm approving this!

martincapello (commented):
Gave a second look to the changes (because Gaspar made me notice about the static variable in algo.h). So I think we should remove the `fixed_step_tilt` static variable.

Gasparoken:
This last change removes the static variable `fixed_step_tilt` in `algo.h` and replaces it with explicit parameter passing through `std::function` callback.
@martincapello  suggested this approach, which I wasn’t familiar with. Thank you!

martincapello (approved):

dacap:
I have just rebased this to latest beta in https://github.com/aseprite/aseprite/tree/snap-to-more-angles

I've also added the possibility to just leave one angle (45 degrees). I didn't make any further review of the PR, you can reset your branch to snap-to-more-angles and I'll merge it and then see if the configuration UI can be improved a little (e.g. a slider instead of a text entry field).

PR #5554, opened 2025-11-22
 [lua] Improve Dialog labels and buddy enabled handling (fix #5117) 

### LINKED ISSUE #5117: Labels of disabled dialog widgets should appear in a different color to match their content

At the moment, dialog labels will always appear as "active", even if their widget is disabled. This makes complex and dynamic dialogs harder to use and process visually.

Proposed change (Current -> Edited):

Image

* Aseprite version: v1.3.13
* System: Windows 11

ckaiser:
Fixes #5117, and makes some changes to default behaviors.

Marking as draft because of `Widget` changes, decided to go for the "simple but disruptive" approach of just adding a new signal to Widget and modifying the default `onEnabled`, it's definitely the simplest approach but it does have ripple effects on the entire app.

Decided to leave the "buddy enabled sync" feature disabled by default behind a flag, but we could easily just simplify it and make this the default behavior and void having an extra bool - in fact I think that's probably the way to go before the merge, but I figured I'd first validate this is what we want without any side-effects.

Added the `placeholder` and `labelalign` propertes to `Dialog_add_widget`, the former is self-explanatory, the latter is because I noticed all of the labels were aligned to the top-right and it looks kinda bad, but maybe some things rely on the old behavior so I changed the default to `ui::LEFT | ui::CENTER` and added a way to change the alignment if desired to go back to the previous default or any other.

Here's a little test script:

```lua
local d = Dialog()
local b = true

d:entry{ id="entry_1", label="Entry 1", text="", placeholder="Placeholder" }
:entry{ id="entry_2", label="Entry 2", text="rrrrr" }
:entry{ id="entry_3", label="Entry 3", labelalign=Align.TOP | Align.LEFT, text="old align default" }
:entry{ id="entry_4", label="Entry 4", text="bbbbb" }
:button{ id="button_1", label="Button 1", labelalign=Align.BOTTOM, text="aligned bottom" }
:button{ id="button_2", label="Button 2", labelalign=Align.RIGHT, text="aligned right" }
:button{ id="button_3", text="3" }
:button{ id="button_4", text="4" }

:button{ id="button_5", label="Button 5", text="5" }
:button{ id="button_6", text="6" }
:button{ id="button_7", text="7" }

:separator()

:button{ id="disable", text="Toggle Disabled", onclick=function()
    b = not b
    d:modify{ id="entry_2", enabled=b }
    :modify{ id="entry_4", enabled=b }
    :modify{ id="button_1", enabled=b }
    :modify{ id="button_3", enabled=b }
    --
    :modify{ id="button_5", enabled=b }
    :modify{ id="button_6", enabled=b }
    :modify{ id="button_7", enabled=b }
    --
end}
:button{ id="ok", text="Close" }

d:show()
```

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -65,6 +65,7 @@ namespace app { namespace script {
 using namespace ui;
 
 static constexpr const int kDefaultAutofit = ui::LEFT | ui::TOP;
+static constexpr const int kDefaultLabelAlign = ui::LEFT | ui::CENTER;
 
 namespace {
 
@@ -129,8 +130,8 @@ struct Dialog {
   ui::HBox* hbox = nullptr;
   bool autoNewRow = false;
   WidgetsList mainWidgets;
-  std::map<std::string, ui::Widget*> dataWidgets;
-  std::map<std::string, ui::Widget*> labelWidgets;
+  std::map<std::string, ui::Widget*, std::less<>> dataWidgets;
+  std::map<std::string, ui::Widget*, std::less<>> labelWidgets;
   int currentRadioGroup = 0;
   int autofit = kDefaultAutofit;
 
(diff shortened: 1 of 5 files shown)


dacap:
LGTM 👍  As we talked I'll try to investigate a way to reduce the obs::signal size or create a new lightweight signal for these cases.

PR #5558, opened 2025-11-27
Fix ButtonSet mouse events propagating incorrectly (fix #5547)

### LINKED ISSUE #5547: Picking ink causes the color palette to drag
From: https://community.aseprite.org/t/picking-ink-causes-the-color-palette-to-drag/27779

ckaiser:
Also fixes a potential case of this happening in LinkLabel.

The event pausing with the popupMenu was making things extra weird here I think, hopefully nothing relies in this behavior, I tested all the buttonsets I remembered existing and they were fine.

Edit: I always forget I can't make multi-line commit messages 🥲, I miss 'em

### DIFF
--- a/src/app/ui/button_set.cpp
+++ b/src/app/ui/button_set.cpp
@@ -124,6 +124,7 @@ bool ButtonSet::Item::onProcessMessage(ui::Message* msg)
 
       if (static_cast<MouseMessage*>(msg)->left() && !buttonSet()->m_triggerOnMouseUp) {
         onClick();
+        return true;
       }
       break;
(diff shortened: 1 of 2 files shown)


dacap:
Just tested on Linux, it works great! :+1: 

> Edit: I always forget I can't make multi-line commit messages 🥲, I miss 'em

You can add any extra description in the commit message but you need an extra empty line between summary and description: 

```
Summary of the change in the first line

After an empty line, you can write more lines / comments about
why the change is needed (generally what changed is in the same
patch).
```

PR #5559, opened 2025-11-27
Use the actual string when measuring shortcut "add" button (fix #5523)

### LINKED ISSUE #5523: "Add" Button Displayed Next to Keyboard Shortcuts Doesn't Stretch to Text Length When Localised
I'm unsure if I should call this a bug as it might be something that is hard-coded, but it is something that could be quite inconvenient for people who use Aseprite in different languages.

The "Add" Button in the "Keyboard Shortcuts" window is a constant width, which, for many languages, doesn't allow the UI to actually show the full text of the translations for "Add". You can see some examples of this further down.

**To Reproduce**
(Using the English display language)
1. Go to "Edit"
2. Click on "Keyboard Shortcuts".
3. Hover your mouse over a shortcut.
4. Look at the button that appears to the left of the shortcut, "Add" should be fully visible.
5. Switch to a language such as Dutch, German, Indonesian or any language where the word "Add" is longer than the English default.
6. You should see that the button size is the same as before but the text is being cut off at both sides.

**Expected behavior**
Ideally, the button should scale to fit the current language's text length. I see that the button is placed to the left of the shortcut, so I assume that there is an anchor on the right side of the button. The scaling of the button would ideally scale towards the left-hand side of the application until the text is properly visible.

**Screenshots**
English Default ("Add" fully visible next to "Ctrl+N" at the top)
Image

Dutch (should say "Toevoegen" but can only see "evoeg")
Image

German (should say "Hinzufügen" but can only see "zufüg")
Image

Spanish (should say "Agregar" but can only see "grega":
Image

**Aseprite & System:**
- Aseprite v1.3.15.5-x64, portable AppImage
- System: Linux, Fedora 43
- Extensions: Default extensions, Lyutria's Studio Theme (disabled for the screenshots), and a WIP translation pack (also disabled for the screenshots, but this was why I noticed that the button doesn't scale to size for the text)

ckaiser:


dacap:
LGTM :+1:

PR #5560, opened 2025-11-27
Fix scaling sprite causes coloured transparent pixels (fix #5552)

### LINKED ISSUE #5552: Scaling a sprite causes coloured transparent pixels
Image

Image

inspected the same image before and after using sprite scaling, with "identify -verbose [img]". The pixels remain 'broken' even when reverting the changes, but can be correctly floodfilled/bucketfilled and overwritten with another color or uniform transparency.

The pixels cause issues when naively loading them in a program later, and assuming them to be (0, 0, 0, 0), and there doesn't seem to be any indicator inside aseprite that there is differently colored transparent pixels. (Spent a few hours trying to find why i am getting aliasing artifacts ^^' )

Gasparoken:
fix #5552


martincapello (commented):
The fix works. I just have some doubts about if it is actually needed to call `remove_fake_transparent_colors` for ROTSPRITE method or we can actually just avoid calling `fixup_image_transparent_colors` for that method as we currently do for the NEAREST_NEIGHBOR method. In other words by calling `fixup_image_transparent_colors` and `remove_fake_transparent_colors` for BILLINEAR method only, it seemed to fix this issue as well.
 
However, after talking with Gaspar, he found out that some sprites scaled using ROTSPRITE might give different results depending if the `fixup_image_transparent_colors` was called first or not. So I started wondering what should be the correct result for the ROTSPRITE scaling method.
For instance, for the following image (by Gaspar):
Image

I get this when scaling the image 250%  and`fixup_image_transparent_colors` is called:
Image

But then I get this when scaling the image 250%  and`fixup_image_transparent_colors` is NOT called:
Image

Gasparoken:
@dacap Could you take a look at this fix? While the fix works, I don't really like the solution of clearing both images (src and dst) at the end of the function (especially the 'src' one).

I see that the original 'resize_image' function modifies the source image (src), which then affects the 'undo' command (it incorrectly restores the image from the modified 'src' with the invisible pixels). I think the correct approach of the 'resize_image' function would have been to copy the 'src' image and work with that copy, however, I want to confirm if this was done from the beginning for performance reasons or if there was another reason.

dacap:
I'm testing this. Why is it called `remove_fake_transparent_colors(src)`? and not only for `dst` image?

dacap:
Sorry I noticed the `fixup_image_transparent_colors(src)` at the beginning and your comment. So undoing will restore the middle `src` used for the algorithm with the broken alphas.

I'll merge this PR as it is, and do the copy of `src` as you suggested, so we don't touch the source image. But probably this will need a refactor to avoid the copy in certain cases (e.g. when the copy was already done).

PR #5562, opened 2025-11-28
Fix console re-centering too much (fix ##5349)

ckaiser:
Makes the code in `notifyNewDisplayConfiguration` be consistent with the one on `addMessages` which checks for text before re-centering the window, so this preserves the original behavior but makes relocating the already visible console stick,


Gasparoken:
Great! LGTM.

PR #5564, opened 2025-11-29
Hardware acceleration for sprite resizing, more sprite resizing methods

Archonic944:
# Summary

This pull request uses Pixman, a library already present in Aseprite, to:

1. speed up existing sprite resizing methods (bilinear) using hardware acceleration
2. add new filtering methods using Pixman's built in convolution kernels

# Implementation

To do this, I first enabled aligned pixels by defining `DOC_USE_ALIGNED_PIXELS`. This aligns every row buffer to a multiple of 16 in the address space in order to allow for the SIMD/NEON optimizations used by Pixman. 

Next, I changed the implementation of RGBA bilinear filtering in `resize_image.cpp` to use the Pixman library. This took minimal hassle since the existing pixel buffer is compatible with the memory alignment expected by Pixman. Finally, I added some more filtering methods (Lanczos3, Bicubic, Gaussian) to the sprite resizing menu. These methods use Pixman's convolution filter with a different kernel (all built into Pixman) for each.

# Verification

I realize that changing the memory layout of our pixel buffer to use aligned pixel rows may cause issues. That's why I tested every tool, as well as copy/paste, marquee + drag, canvas resizing, etc. I noticed no abnormalities. I tested with various different resolutions of images. I am on MacOS 15; while looking in Activity Monitor, I noticed no memory leaks or other significant issues like high CPU. Memory usage hovered at around 150mb with a single on a 50x100 file.

I compared the Pixman bilinear output to the former implementation. I noticed a very slight difference between the two, particularly that the Pixman implementation was more "spread out" (pixels occupied a larger space and blended into each other more). It was a very small difference, though, and I would just chalk this up to slight differences in the implementations.

# Limitations

While an option for nearest neighbor scaling exists in the Pixman library, I decided not to use it since Aseprite's nearest neighbor algorithm is a. not that compute intensive and b. could have intricacies related to pixel art.

The implementation only works for RGBA; for grayscale, bilinear falls back on the old implementation. For indexed pallettes it is obviously not possible.


dacap:
Thanks @Archonic944 for this PR. You're uncommenting the `DOC_USE_ALIGNED_PIXELS` flag, but there might be parts of code that stop working if we enable the flag as it is.

I have to continue reviewing some parts of code that might expect a continuos buffer (rowstride == image width), so this change must wait.

Archonic944:
All good. I did do a bit of testing but I understand if you need to do more.

Archonic944:
Hey @dacap, no stress, but have you looked at the code any more?

PR #5573, opened 2025-12-05
[lua] Add Sprite.isValid (fix #5017)

### LINKED ISSUE #5017: Add `isValid(Object)` to detect deleted objects

ckaiser:
Fixe [#5017](https://github.com/aseprite/aseprite/issues/5017)

### DIFF
--- a/src/app/script/sprite_class.cpp
+++ b/src/app/script/sprite_class.cpp
@@ -1066,6 +1066,12 @@ int Sprite_get_undoHistory(lua_State* L)
   return 1;
 }
 
+int Sprite_get_isValid(lua_State* L)
+{
+  lua_pushboolean(L, may_get_docobj<Sprite>(L, 1) ? true : false);
+  return 1;
+}
+
 const luaL_Reg Sprite_methods[] = {
   { "__eq",              Sprite_eq                },
   { "resize",            Sprite_resize            },
@@ -1132,6 +1138,7 @@ const Property Sprite_properties[] = {
   { "tileManagementPlugin", Sprite_get_tileManagementPlugin, Sprite_set_tileManagementPlugin },
   { "useLayerUuids",        Sprite_get_useLayerUuids,        Sprite_set_useLayerUuids        },
   { "undoHistory",          Sprite_get_undoHistory,          nullptr                         },
+  { "isValid",              Sprite_get_isValid,              nullptr                         },
   { nullptr,                nullptr,                         nullptr                         }
 };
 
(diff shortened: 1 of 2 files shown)


src/app/script/sprite_class.cpp
```diff
@@ -341,6 +341,18 @@ int Sprite_flatten(lua_State* L)
   return 0;
 }
 
+int Sprite_isValid(lua_State* L)
+{
+  const auto* id = static_cast<ObjectId*>(luaL_testudata(L, 1, "doc::Sprite"));
```
dacap: I think you can use `may_get_docobj<Sprite>(L, 1);`, see `Sprite_eq` for an example.

src/app/script/sprite_class.cpp
```diff
@@ -1078,6 +1090,7 @@ const luaL_Reg Sprite_methods[] = {
   { "assignColorSpace",  Sprite_assignColorSpace  },
   { "convertColorSpace", Sprite_convertColorSpace },
   { "flatten",           Sprite_flatten           },
+  { "isValid",           Sprite_isValid           },
```
dacap: It must be a property just like `Sprite.id`, so we can use 
```lua
if sprite.isValid then ... end
```
Instead of
```lua
if sprite:isValid() then ... end
```

dacap:
Oh something I forgot to mention: this change can go to `main` branch.

dacap:
Looks good 👍

PR #5593, opened 2025-12-13
Check if extension files are present to avoid crashes & issues

ckaiser:
Having an installed extension that refers to a non-existing keys file can cause aseprite to crash on load, as `AppMenu` attempts to load the XML file without checking if it exists. Other non-existing files can also cause annoying issues (like language entries appearing but doing nothing, etc).

This adds some checks to avoid this. The only only _necessary_ one is the `KeyboardShortcuts` one, but I think the others won't hurt and make sense.

Test exception to reproduce it: [test-extension.zip](https://github.com/user-attachments/files/24143216/test-extension.zip)

### DIFF
--- a/src/app/extensions.cpp
+++ b/src/app/extensions.cpp
@@ -274,6 +274,9 @@ void Extension::executeExitActions()
 
 void Extension::addKeys(const std::string& id, const std::string& path)
 {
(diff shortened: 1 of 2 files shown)


dacap:
The fix works great! :+1: thanks for the example test extension file.

PR #5596, opened 2025-12-14
[win]Fix Windows path handling in build.sh

alexzobi:
This change fixes multiple issues in build.sh where paths containing whitespace (e.g. "Program Files") were being split by the shell.

The script previously relied on unquoted variable expansion and word-splitting constructs such as `for x in $(...)`, which break on Windows and in any environment where directories include spaces.

Key fixes include:
- Quoting all path variables when used in commands and tests
- Replacing unsafe `for ... in $(...)` loops with `while read` loops
- Fixing mkdir, rm, dirname, and executable invocation to be whitespace-safe
- Correcting Skia directory creation and persistence logic, which was the primary source of failures on Windows

These changes preserve existing behavior while making the build script robust across Windows (Git Bash/MSYS), macOS, and Linux environments.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,505 +1,166 @@
 #! /usr/bin/env bash
-#
-# This is a script to help users and developers to build Aseprite.
-# Usage:
-#
-#   ./build.sh
-#   ./build.sh --reset
-#   ./build.sh --auto [--norun]
-#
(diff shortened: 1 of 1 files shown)


dacap:
Why comments were removed?

PR #5601, opened 2025-12-16
Fix Alt-tabbing while using hand tool locks the hand tool (fix #325)

### LINKED ISSUE #325: Alt-tabbing while using hand tool locks mouse in other windows.
_From [bangs.j...@gmail.com](https://code.google.com/u/114714790844789003241/) on January 27, 2014 01:51:43_

What steps will reproduce the problem? 1.Hold down hand tool key (Middle mouse click)
2.Alt-Tab to another application
3.mouse will only move in area restricted by aseprite (See video) What is the expected output? What do you see instead? Not to lock my mouse (see video) What version of the product are you using? On what operating system? Aseprite 0.9.6-beta1 for WinXP/Vista/7 

**Attachment:** [mouse lock.zip](http://code.google.com/p/aseprite/issues/detail?id=325)

_Original issue: http://code.google.com/p/aseprite/issues/detail?id=325_

Gasparoken:
fix #325

### DIFF
--- a/src/app/ui/editor/editor.cpp
+++ b/src/app/ui/editor/editor.cpp
@@ -2101,6 +2101,27 @@ bool Editor::onProcessMessage(Message* msg)
       break;
     }
 
+    case kAppLostFocusMessage: {
+      ui::Manager* manager = ui::Manager::getDefault();
+      if (manager && m_state) {
+        ui::MouseMessage* msg = new ui::MouseMessage(ui::MessageType::kMouseUpMessage,
+                                                     os::PointerType::Mouse,
+                                                     ui::MouseButton::kButtonLeft,
+                                                     os::KeyModifiers::kKeyNoneModifier,
+                                                     manager->getMouse()->mousePosInDisplay());
+        m_state->onMouseUp(this, msg);
+        ui::MouseMessage* msg2 = new ui::MouseMessage(ui::MessageType::kMouseUpMessage,
+                                                      os::PointerType::Mouse,
+                                                      ui::MouseButton::kButtonMiddle,
+                                                      os::KeyModifiers::kKeyNoneModifier,
+                                                      manager->getMouse()->mousePosInDisplay());
+        m_state->onMouseUp(this, msg2);
+        os::SystemRef sys = os::System::instance();
+        sys->resetKeyPressed();
+      }
+      break;
+    }
+
     case kMouseEnterMessage:
       // Do not update tool loop modifiers when the mouse exits and/re-enters
       // the editor area while we are inside the same tool loop (hasCapture()).
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -487,6 +487,15 @@ void Manager::generateMessagesFromOSEvents()
         break;
       }
 
+      case os::Event::AppLostFocus: {
+        if (mouse_widget) {
+          auto msg = new AppLostFocusMessage();
+          msg->setRecipient(mouse_widget);
+          enqueueMessage(msg);
+        }
+        break;
+      }
+
       case os::Event::KeyDown:
       case os::Event::KeyUp:   {
         Message* msg = new KeyMessage(
--- a/src/ui/message.h
+++ b/src/ui/message.h
@@ -276,6 +276,11 @@ private:
   base::paths m_files;
 };
 
+class AppLostFocusMessage : public Message {
+public:
+  AppLostFocusMessage() : Message(kAppLostFocusMessage) {}
+};
+
 class DndMessage : public Message {
 public:
   os::DragEvent& event() { return m_event; }
--- a/src/ui/message_type.h
+++ b/src/ui/message_type.h
@@ -18,10 +18,11 @@ enum MessageType {
   kCloseMessage,        // Windows is closed.
   kCloseDisplayMessage, // The user wants to close the entire application.
   kResizeDisplayMessage,
-  kPaintMessage,     // Widget needs be repainted.
-  kTimerMessage,     // A timer timeout.
-  kDropFilesMessage, // Drop files in the manager.
-  kWinMoveMessage,   // Window movement.
+  kPaintMessage,        // Widget needs be repainted.
+  kTimerMessage,        // A timer timeout.
+  kDropFilesMessage,    // Drop files in the manager.
+  kWinMoveMessage,      // Window movement.
+  kAppLostFocusMessage, // App lost focus.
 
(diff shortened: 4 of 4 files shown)


dacap:
@Gasparoken please give a proper title to the PR

Gasparoken:
> @Gasparoken please give a proper title to the PR

Ok, no problem, by the way: @dacap If you're available, could you check if the approach is correct? I'm still testing on Windows and trying to address the hand tool issue with the mouse middle button.

src/ui/manager.cpp
```diff
@@ -486,6 +499,10 @@ void Manager::generateMessagesFromOSEvents()
         enqueueMessage(msg);
         break;
       }
+      case os::Event::CancelActions: {
```
martincapello: I'm not sure about introducing this new os.Event in laf as is...because it doesn't seem  to map well to an actual OS event. Should this event be fired when the program's window is not the main/focused/active window any more?

src/ui/manager.cpp
```diff
@@ -115,6 +115,19 @@ static bool first_time = true; // true when we don't enter in poll yet
 // when the os::Display is resized by the user.
 static bool auto_window_adjustment = true;
 
+static CancelActionsCallback cancelActionsCallback = nullptr;
```
martincapello: I think we should avoid using a static callback for this.

src/ui/manager.cpp
```diff
@@ -486,6 +499,10 @@ void Manager::generateMessagesFromOSEvents()
         enqueueMessage(msg);
         break;
       }
+      case os::Event::CancelActions: {
+        cancelActions();
```
martincapello: I think we should follow the same principles as for the other events: translating the os Event into a Message. Then you will be able to handle that Message in the Editor's onProcessMessage method and call your Editor::cancelOperation() method.

dacap:
I'm testing this one with https://github.com/gasparoken/laf/tree/fix-alt-tab (which is different from https://github.com/aseprite/laf/tree/fix-alt-tab).

Probably I'll cherry-pick https://github.com/gasparoken/laf/tree/fix-alt-tab and rename the events to `AppEnter`/`AppLeave`. I always wanted a couple of `WindowEnter`/`WindowLeave` events, but probably those would be different e.g. in a multiple-windows scenario a window could lose focus, but another window of the same app be focused (so there is a `WindowLeave` for the old window, `WindowEnter` for the new one, but no `AppEnter/Leave` events).

I'll try to make these events work on laf, and then go back to this PR.

dacap:
Merged the laf patch in https://github.com/aseprite/laf/commit/c0a84275254ddf0b634c5387eb886e063225f66e but then implemented AppEnter/AppLeave/WindowEnter/WindowLeave in all platforms: https://github.com/aseprite/laf/commit/592136b38c964c56053675eee9c554bf60fb17ca

I'll continue with this PR tomorrow.

src/app/ui/editor/editor.cpp
```diff
@@ -2101,6 +2101,27 @@ bool Editor::onProcessMessage(Message* msg)
...
+                                                     ui::MouseButton::kButtonLeft,
+                                                     os::KeyModifiers::kKeyNoneModifier,
+                                                     manager->getMouse()->mousePosInDisplay());
+        m_state->onMouseUp(this, msg);
+        ui::MouseMessage* msg2 = new ui::MouseMessage(ui::MessageType::kMouseUpMessage,
+                                                      os::PointerType::Mouse,
+                                                      ui::MouseButton::kButtonMiddle,
+                                                      os::KeyModifiers::kKeyNoneModifier,
+                                                      manager->getMouse()->mousePosInDisplay());
+        m_state->onMouseUp(this, msg2);
```
dacap, line 2118: Both `new ui::MouseMessage(...)` generate memory leaks as they are not queued in the manager, I'm going to replace both of them with:
```suggestion
        ui::MouseMessage msg(ui::MessageType::kMouseUpMessage,
                             os::PointerType::Mouse,
                             ui::MouseButton::kButtonLeft,
                             os::KeyModifiers::kKeyNoneModifier,
                             manager->getMouse()->mousePosInDisplay());
        m_state->onMouseUp(this, &msg);

        ui::MouseMessage msg2(ui::MessageType::kMouseUpMessage,
                              os::PointerType::Mouse,
                              ui::MouseButton::kButtonMiddle,
                              os::KeyModifiers::kKeyNoneModifier,
                              manager->getMouse()->mousePosInDisplay());
        m_state->onMouseUp(this, &msg2);
```
Another important thing, after merging this I'll try to refactor the ui-lib in some way automatically or to handle this in a more generic way (e.g. the Timeline has the same problem)

src/ui/manager.cpp
```diff
@@ -487,6 +487,15 @@ void Manager::generateMessagesFromOSEvents()
         break;
       }
 
+      case os::Event::AppLostFocus: {
```
dacap, line 490: I'll try to use the `WindowLeave` event instead of the app one (both messages should be received, but I have to test on all platforms yet):
```suggestion
      case os::Event::WindowLeave: {
```

src/ui/manager.cpp
```diff
@@ -487,6 +487,15 @@ void Manager::generateMessagesFromOSEvents()
...
+        if (mouse_widget) {
+          auto msg = new AppLostFocusMessage();
```
dacap, line 492: There is no need for a new Message sub-class if the message doesn't contain any extra information, I'm going to replace this with:
```suggestion
          auto* msg = new Message(kWinLeaveMessage);
```
The `kWinLeaveMessage` name is temptative, but probably we could add something like `kMouseLostMessage` or see if we can just send a `kMouseLeaveMessage`.

dacap (commented):
Leaving some comments about the code and how I'm going to rebase this PR with these patches. There is no need for action from your side.

src/ui/manager.cpp
```diff
@@ -487,6 +487,15 @@ void Manager::generateMessagesFromOSEvents()
...
+          msg->setRecipient(mouse_widget);
```
dacap, line 493: After furthering testing, this `AppLostFocus` (or another one like `WindowLeave` message) cannot be received by the `mouse_widget` only (we should `setPropagateToParent(true)`), but the message should be sent to the display window.

For this specific fix, I'll try another approach, probably a `MouseLeave` message when `os::Event::WindowLeave` is received (instead of creating a new `AppLostFocusMessage`), something to solve every case of all widgets that have the mouse captured.

PR #5625, opened 2026-01-05
Fix accent + character results in character without accent (fix #5618)

### LINKED ISSUE #5618: Combining accent + character results in character without accent
When I enter an accent + character combination, it results in the character only, NOT the expected accented character.

Example: I enter ¨ + u which should result in ü. Instead I get a plain u.

Accented characters that have a dedicated key, like é à ù è ç on French keyboards, work as expected. The issue happens only when two keys are needed (diacritic + letter) to output the expected character.

In Aseprite:
Image

In other apps:
Image

Tested with 1.3.16.1 (downloaded) and 1.3.17-beta2 (from Steam) on macOS Sequoia 15.7.1

Diacritic marks used to work in previous versions, but I'm not exactly sure when it stopped. I _think_ it worked up to 1.3.13.

Enter any diacritic mark + letter combination in any font that offers such characters (any system font should cover common languages).
¨ + u
^ + e
or any other.

Gasparoken:
fix #5618

### DIFF
--- a/src/ui/entry.cpp
+++ b/src/ui/entry.cpp
@@ -60,7 +60,8 @@ Entry::Entry(const int maxsize, const char* format, ...)
   , m_recent_focused(false)
   , m_lock_selection(false)
   , m_persist_selection(false)
-  , m_translate_dead_keys(true)
+  , m_translate_dead_keys(false)
+  , m_lastDeadKey(0)
   , m_scale(1.0f, 1.0f)
 {
   enableFlags(CTRL_RIGHT_CLICK);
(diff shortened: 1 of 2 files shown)


dacap:
I'm not sure about adding those functions to detect dead keys and do the combination by ourselves. Generally it's something you will want the OS to solve (and probably the IME in case of CJK languages and other languages).

On Linux it works (in Windows it should work too), so I'm not sure if this is an macOS-specific problem (something to solve in laf-os).

Gasparoken:
Thanks @dacap, you're right. I think the solution we're looking for is aseprite/laf#173

PR #5627, opened 2026-01-09
Extend the "keep closed sprite on memory" time (fix #5603)

### LINKED ISSUE #5603: Longer Keep Closed Sprite On Memory Options
Having the option to keep closed files on memory for a while is great, but it caps at eight hours and that's a bit brief for me. It would be great if we could prolong that to 24 or 48 hours.

Naturally, it uses a fair bit of memory to keep files around like this, but it would be hugely helpful for me to have the option at least.

Thanks for considering it! :D 

..using ASEprite 1.3.15.4 on Win10.

ckaiser:
Implements #5603. *Technically* the forever option isn't *forever* forever, but I don't think it counts if the earth isn't gonna be inhabitable by then anyhow and I didn't want to change the underlying code or logic too much.


dacap:
Looks great :+1: I'll add the 48hs option to the combobox which is missing.

PR #5639, opened 2026-01-16
Tooltip for corner radius

Syrux64:
Hi. I updated the en.ini for the following result.

Image

BUT the same tool tip is applied for the slider as well.

Image

Do I have to make changes to the pref.xml file? Please let me know. I would like to have some directions here.


Syrux64:
Selected wrong 'merge into' branch. My bad.

Issue #5635, opened 2026-01-14
Crash when exporting an animation as two files and target files are read only


RockyMullet:
**Describe the bug**
I have a aseprite file that I used to export as a single file (.tga), but decided to make a 2 frame animation with it. I use source control, so to not rename everything, the initial file still has the name of the file, but the second frame is "FileName2.tga" as auto generated by Aseprite.

Because of source control, my files were read only, trying to export the 2 frame animation, when I press export, instead of being met with the usual "Do you want to export the animation in 2 files?" message box, Aseprite instantly close (I guess it crashes).

Aseprite do not crash if my files (the export files) are not read only.

**To Reproduce**
1. Have an aseprite file with a 2 frame animation
2. "Export As" the animation in 2 files
3. Set the 2 exported files as read only
4. "Export As" the animation in 2 files again
5. Will crash before the message box asking if you want to export the animation in 2 files

**Expected behavior**
Aseprite should scold me that my files are read only and not crash.

**Screenshots**
Attached a video

**Aseprite & System (please complete the following information):**
- Aseprite: v1.3.16.1-x64 from Steam
- System: Windows 10
- Extensions: None

**Additional context**
I got 11 layers, some hidden, 2 frames

Video reproducing the crash
https://github.com/user-attachments/assets/5c2fb286-02d0-48a8-aa96-57f67d1c41e9

PR #5657, opened 2026-02-03
Fixes issue #5635 fixed a crash when doing 'Export As' for animation and the target file is read-only

nelson-steve:
I am fairly new to the codebase and this is my first pull request so I am not certain if the fix is this simple or there is more to it. But it seems to have fixed the crash for me and I am getting the appropriate message saying the file is read-only.
Fixes [aseprite/aseprite#5635](https://github.com/aseprite/aseprite/issues/5635)


dacap:
Thanks @nelson-steve! The fix works great :+1:

PR #5676, opened 2026-02-25
Fix multi-layer selection movement preview isn't working (fix #2865)

### LINKED ISSUE #2865: Moving a selection from multiple layers doesn't show anything moving (or moves the thing from one layer)
Problem, if you have something drawn over multiple layers, you use the selection tool and select the everything, select all the layers and start moving the selection it doesn't show anything moving, or it shows some part of it moving (probably the last selected layer). When you press space it does as intented, eg it moves everything on all layers selected.    

Steps to reproduce:
1) Create 3 layers
2) Draw something next to eachother on all of them
3) Use selection tool to select everything you've drawn
4) Select all the layers
5) Start moving the selection
6) Press enter after you finished moving

* Aseprite version: v1.3-dev self compiled
* System: ubuntu 20.04

Gasparoken:
This fix solves the multi-layer/multi-cel transformation preview.

- It can manage a mix of image and tilemap layers in TilemapMode::Pixels or TilemapMode::Tiles.
- Considers opacities of each cel.
- Considers the blend modes of each layer.
- Considers the Z-order of each cel.
- Consider the overlaps between selected and unselected layers.
- Correct behavior of Skew/Rotate/Resize/Move/Flip/Shift transformations.
- In Tilemap Mode::Tiles, the tilemap layers that are transformed (Resize/Move) are restricted to those whose tile grid matches that of the active layer.
- When generating new tiles in a tilemap layer by breaking tiles because transformation in TilemapMode::Pixels, it was decided to do so in TilesetMode::Stack mode to avoid accidental destruction of tiles in the tileset.
- Fixed the Shift Pixels in a multi-layer transformation when it is the first transformation command.
- Consider non-rectangular selections.
- Consider linked cels.

fix #2865

### DIFF
--- a/src/app/ui/editor/pixels_movement.h
+++ b/src/app/ui/editor/pixels_movement.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-2026  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -16,12 +16,14 @@
 #include "app/tx.h"
 #include "app/ui/editor/handle_type.h"
 #include "doc/algorithm/flip_type.h"
+#include "doc/cel_list.h"
 #include "doc/frame.h"
 #include "doc/image_ref.h"
 #include "gfx/size.h"
 #include "obs/connection.h"
 
 #include <memory>
+#include <vector>
 
 namespace doc {
 class Image;
@@ -138,14 +140,17 @@ private:
   bool editMultipleCels() const;
   void stampImage(bool finalStamp);
   void stampExtraCelImage();
+  void stampExtraCelImages();
   void onPivotChange();
   void onRotationAlgorithmChange();
   void redrawExtraImage(Transformation* transformation = nullptr);
+  void redrawExtraImages(Transformation* transformation = nullptr);
   void redrawCurrentMask();
   void drawImage(const Transformation& transformation,
                  doc::Image* dst,
                  const gfx::PointF& pt,
                  const bool renderOriginalLayer);
+  void drawImages(const Transformation& transformation);
   void drawMask(doc::Mask* dst, bool shrink);
   void drawParallelogram(const Transformation& transformation,
                          doc::Image* dst,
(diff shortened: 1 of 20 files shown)


Gasparoken:
Pending:
- Rotation/Skew preview and stamp is incorrect.

Gasparoken:
This fix solves the multi-layer/multi-cel transformation preview.

- It can manage a mix of image and tilemap layers in `TilemapMode::Pixels` or `TilemapMode::Tiles`.
- Considers opacities of each cel.
- Considers the blend modes of each layer.
- Considers the Z-order of each cel.
- Consider the overlaps between selected and unselected layers.
- Correct behavior of Skew/Rotate/Resize/Move/Flip/Shift transformations.
- In `Tilemap Mode::Tiles`, the tilemap layers that are transformed (Resize/Move) are restricted to those whose tile grid matches that of the active layer.
- When generating new tiles in a tilemap layer by breaking tiles because transformation in `TilemapMode::Pixels`, it was decided to do so in `TilesetMode::Stack` mode to avoid accidental destruction of tiles in the tileset.
- Fixed the Shift Pixels in a multi-layer transformation when it is the first transformation command.

src/app/ui/editor/pixels_movement.h
```diff
@@ -138,14 +140,17 @@ class PixelsMovement {
   bool editMultipleCels() const;
   void stampImage(bool finalStamp);
   void stampExtraCelImage();
+  void stampExtraCelImages();
   void onPivotChange();
   void onRotationAlgorithmChange();
   void redrawExtraImage(Transformation* transformation = nullptr);
+  void redrawExtraImages(Transformation* transformation = nullptr);
```
ckaiser, line 147: If we're not using the old `redrawExtraImage` function we can get rid of it

src/doc/util.cpp
```diff
@@ -95,4 +95,68 @@ Mask make_aligned_mask(const Grid* grid, const Mask* mask)
   return maskOutput;
 }
 
+template<typename ImageTraits>
```
ckaiser: `mask_img_templ` and `mask_image ` are duplicated from `src/app/cel_ops.cpp`, can probably use these in the latter

src/doc/util.h
```diff
@@ -16,6 +17,8 @@ class Mask;
 class Image;
 class Tileset;
 
+// typedef uint32_t color_t;
+
```
ckaiser:
```suggestion
```

ckaiser (commented):
LGTM! Left some minor comments but overall it works great. The only thing that still feels strange is having to build two separate maps for the extra cel data, and the separate function to assign it specifically. I understand the separation because they're in different libraries but it feels like there should be a more elegant way to solve it (I just can't think of it 😅, we'll see what @dacap thinks)

dacap:
I'm trying to start testing this PR, it looks like the wrap mode is broken:

[Kooha-2026-03-18-14-56-40.webm](https://github.com/user-attachments/assets/909524b8-158a-416c-bb7a-f1fe56acb237)

Gasparoken:
Thank you @ckaiser and @dacap. @dacap if you don't mind, I can review the wrap mode (for TiledMode) and push again (I can also squash for the latest commits, and applying @ckaiser's recommendations)

dacap:
@Gasparoken you can go ahead :+1:

Gasparoken:
- Tiled mode has been restored for multi-cel transformations.
- The paste command has been fixed (discovered during testing).
- Fixed a double drawn of the current cel's image during multi-cel transformation preview.

dacap:
Continues here: https://github.com/aseprite/aseprite/pull/5967

PR #5684, opened 2026-03-02
Keep empty cels

### LINKED ISSUE #1303: Associate user data to an empty cel

dacap:
Should fix all issues from #1303.


dacap:
@ckaiser would you mind to take a look to this? It's a quite large change, I'd like to add some tests and probably change some APIs as the code looks like a mess for all the given cases.

ckaiser (commented):
Some things I found while testing:
 * FlattenLayers crashed when merging down in some cases
 * Linked Tilemaps behave differently in the UI than normal layers, if you link two tilemap cels that have no image data, then draw something in one of them and then undo that, the timeline won't go back to showing the "hollow" cel indicator.
 * When merging a layer with a cel with user data into a layer with an empty cel, the user data is not preserved regardless of it had an image before, I think this happens without this patch, so it's default behavior but it feels strange to not merge user data.

dacap:
Thanks for testing @ckaiser. As we talked I'll see how we can automate all these kind of tests before merging this PR.

PR #5687, opened 2026-03-04
Fix eyedropper behavior without a reference layer (fix #5685)

### LINKED ISSUE #5685: Avoid confusion with Eyedropper and "First Reference Layer" option
"First Reference Layer" option could behave like "All Layers" when there is no "Reference Layer" present in the sprite, so in this way the user doesn't think that the eyedropper is not working anymore.

Image

From this issue: https://www.reddit.com/r/aseprite/comments/1rij0qu/how_to_fix_color_picker_always_going_to/

ckaiser:
Fixes #5685 with the simplest possible method. I experimented with adding a warning to the context bar or changing the status of the combobox in some way (disabling the item widget when it's not available) but it all resulted in a bunch of code that had to run for a lot of state changes since everything is global and I ended up dropping it. I think it would still be good to at least a tooltip explaining exactly what "sample" does and the three options though.

### DIFF
--- a/src/app/color_picker.cpp
+++ b/src/app/color_picker.cpp
@@ -82,7 +82,7 @@ ColorPicker::ColorPicker() : m_tile(doc::notile), m_alpha(0), m_layer(nullptr)
 void ColorPicker::pickColor(const Site& site,
                             const gfx::PointF& _pos,
                             const render::Projection& proj,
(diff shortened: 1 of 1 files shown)


dacap:
LGTM :+1: probably in a future we can try to add some feedback in the context bar.

PR #5691, opened 2026-03-06
Release capture after switching to other app w/Alt+tab (fix #325, #5601)

### LINKED ISSUE #325: Alt-tabbing while using hand tool locks mouse in other windows.

dacap:
This should fix #325, after receiving an `os::Event::WindowLeave` message (e.g. because another native window is activated by the user), we check if there is a widget with the mouse capture and generate an `kMouseUpMessage` without buttons (kButtonNone) for that specific widget, which can be processed just like a regular `kMouseUpMessage`.

It's similar to #5601 but doesn't require a new `ui::MessageType` and this fixes issues with the sprite Editor, Timeline, and color bar (PaletteView, e.g. when we are dragging colors).

### DIFF
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -461,6 +461,25 @@ void Manager::generateMessagesFromOSEvents()
         break;
       }
 
+      case os::Event::WindowLeave:
+        if (capture_widget) {
+          const gfx::Point mousePos = display->nativeWindow()->pointFromScreen(
+            get_mouse_position());
+          auto* msg = newMouseMessage(kMouseUpMessage,
+                                      display,
+                                      nullptr,
+                                      mousePos,
+                                      PointerType::Unknown,
+                                      // No button indicates the capture was lost
+                                      kButtonNone,
+                                      kKeyUninitializedModifier);
+          msg->setRecipient(capture_widget);
+          enqueueMessage(msg);
+
+          setMouse(nullptr);
+        }
+        break;
+
       case os::Event::CloseWindow: {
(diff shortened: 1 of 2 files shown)


dacap:
I'm taking this PR back because https://github.com/aseprite/aseprite/actions/runs/22776507662/job/66071133540 is failing, it looks like on Windows a WA_INACTIVE message is received when the HWND is being destroyed and a reference to a deleted os::Window used in some place.

dacap:
> I'm taking this PR back because https://github.com/aseprite/aseprite/actions/runs/22776507662/job/66071133540 is failing, it looks like on Windows a WA_INACTIVE message is received when the HWND is being destroyed and a reference to a deleted os::Window used in some place.

Fixed with: https://github.com/aseprite/laf/commit/57b22eb66b1a15dbf96d3c02f60f2b3378153670

PR #5692, opened 2026-03-06
[Lua] Add eyedropper action to useTool

Gasparoken:


src/app/script/app_object.cpp
```diff
@@ -448,6 +449,32 @@ int App_useTool(lua_State* L)
     }
   }
 
+  // Eyedropper (pick the color from the first point only)
+  if (params.ink->isEyedropper()) {
+    type = lua_getfield(L, 1, "points");
+    if (type == LUA_TTABLE) {
```
dacap: It looks like these lines are duplicated, probably checking for the "points" tablet is the first thing we can do, and after that decide the eyedropper path or the ToolLoop one.

PR #5697, opened 2026-03-07
Allow installing extensions by dropping them into the preferences dialog

ckaiser:
Right now if you double click an extension from the OS, install it, and then attempt to install another one, it does nothing because "dropping" files only works when the foreground window is the main window.

This change allows the event to propagate so we can have modal dialogs that accept file drops, in this case the Preferences dialog accepts `.aseprite-extension` and `.zip` (since if you're dropping something onto here, it'll most likely be an extension).


dacap:
LGTM :+1:

PR #5699, opened 2026-03-08
Fix Palette unlock icon being drawn over Palette lock icon

thkwznk:
This PR fixes the Palette unlock icon being drawn over the Palette lock icon. This is a very minor issue as it is invisible on the default dark theme and barely visible in the default theme but does become visible when editing themes.

Preview of the before and after:
image

---


dacap:
Thanks @thkwznk for the PR :+1: LGTM

PR #5702, opened 2026-03-10
Fix color picker selecting below alpha bar

### LINKED ISSUE #2006: [colorbar] RGB zero values extend several pixels below alpha bar
When adjusting alpha on a pixel from the color picker it's quite easy to "whiff" the alpha box and replace your current color with black because the SV box's area for some reason extends several pixels below the alpha bar's.  Demonstration in the below image:

image

* 1.2.9 Steam
* System: Windows

luke-henry-04:
Fixes #2006 

I added a check in the color selector, making sure the user can't select from below the alpha bar. The outside border of the ui elements are included in the widget, but the alpha bar and bottom bar bounds are contained strictly inside of the color selector's border. One check makes sure the user isn't selecting the boundary underneath of the alpha bar bounds, and the other prevents the eyedropper tool from previewing colors beneath the alpha bar.

This fixes the issue in the parent `ColorSelector` class rather than in each subclass individually, which also covers the RGB and RYB wheels not addressed in #2522 

### DIFF
--- a/src/app/ui/color_selector.cpp
+++ b/src/app/ui/color_selector.cpp
@@ -273,9 +273,6 @@ app::Color ColorSelector::getColorByPosition(const gfx::Point& pos)
   if (rc.isEmpty())
     return app::Color::fromMask();
 
-  if (!hasCapture() && pos.y >= rc.y2())
-    return app::Color::fromMask();
-
   const int u = pos.x - rc.x;
   const int umax = std::max(1, rc.w - 1);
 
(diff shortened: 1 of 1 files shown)


luke-henry-04:
Found that when you click and drag the eyedropper from the primary or secondary color up towards the color selector from below, it sets the color to mask on the border, which then disables the alpha bar functionality until the mouse moves off of the alpha bar. This doesn't feel very clean, so I changed it to match the functionality of the other edges, which just extend the color picker values to the border.

Instead of returning the mask, it now selects from the lowest visible color selector bar (usually the alpha). This has the effect of the eyedropper appearing a few pixels out of bounds, although that mirrors the functionality of the other three edges.

dacap:
Thanks for the patch @luke-henry-04 :pray: LGTM

PR #5707, opened 2026-03-14
Fix saving a sprite with an empty palette corrupts file (fix #5673)

### LINKED ISSUE #5673: Saving a sprite with an empty palette produces an invalid file.
**Describe the bug**
When saving a sprite while a palette with zero (not one) colors is loaded, the resulting `.aseprite` file is invalid.

**To Reproduce**
1. Create or open an existing sprite
2. Load a palette with zero colors. For example, save the following as `empty.gpl` and load it:

```text
GIMP Palette
```

3. Save the sprite
4. Close and attempt to re-open the sprite

**Expected behavior**
The sprite should be openable.

**Screenshots**

Image

The palette after clicking "Close" on the above error:

Image

- Aseprite: ~1.3.16 (compiled from source)
- System: Linux Mint 22.2
- Extensions: None

**Additional context**
This causes an old palette chunk to be written with `0` colors, but old palette chunks use `0` to mean `256` colors. Initially I thought this could be easily fixed like so: https://github.com/grind086/aseprite/commit/74065f8d82eb4cd5d77b78742906712de7097dd8

But that causes `palTo` on [this line](https://github.com/aseprite/aseprite/blob/efc30e24a55a01d0666358a0475af4e43fd7e2d2/src/app/file/ase_format.cpp#L426) to be `-1`, which then underflows when it's [written as a DWORD](https://github.com/aseprite/aseprite/blob/main/docs/ase-file-specs.md#palette-chunk-0x2019). And clamping to `0` results in both start and end indices being `0`, which is incorrect because that's a non-empty range. I think you could set `palFrom` to `1` and `palTo` to `0` to get an empty range, but I'm not sure if that would break something else.

This is probably a low priority issue, as getting Aseprite to have a completely empty palette in the first place requires intention, and (afaik) is pointless. I just ran into this while trying to create a file to test a theoretical edge case on a file parser, and figured I'd look into it a bit.

XavierCunha:
Before, when loading an empty palette and saving the file, the file would become corrupted upon opening it again.

This fixes it by handling the case where the palette size is 0 when saving the file, so it no longer causes underflow.
Also fixed the decoder, as before it would only resize palettes with newSize > 0, which would not work for loading empty palettes.
Finally, added a new test to verify if the file doesn't break.

### DIFF
--- a/src/app/file/ase_format.cpp
+++ b/src/app/file/ase_format.cpp
@@ -417,7 +417,7 @@ bool AseFormat::onSave(FileOp* fop)
     // is the first frame or did the palette change?
     Palette* pal = sprite->palette(frame);
     int palFrom, palTo;
-    
+
     // If palette has no colors, palFrom is set to 1 and palTo to 0, avoiding
     // palTo being -1 when written as a DWORD, which would underflow
     if (pal->size() == 0) {


XavierCunha:
Hi @dacap, just checking in on this! Is there anything I should adress?

dacap:
I wasn't able to test and review this yet.

dacap:
I'm not sure about writing "from > to" in the palette chunk. I think that in the best case, for the first palette (frame = 0) if it doesn't contain colors, we shouldn't write a palette chunk at all, and in case that a new palette change appears in a frame > 0 where the palette size = 0, we could write a chunk with ncolors = 0 (and from=to=0, just ignoring these).

Anyway, a patch for this would be preferred in the `beta` branch, as it contains a refactor where the encoder was moved to src/dio.

XavierCunha:
@dacap Alright, that makes sense. Do you think I could still try to implement the patch in this branch using your suggestion? Maybe if it works well I could then try to do it also in the beta branch.

XavierCunha:
Meanwhile I decided to go ahead and implement the suggested fix in the beta branch, and opened a new PR (#5779) to reflect these changes.

PR #5714, opened 2026-03-21
Fix Shift no longer adding to selection when bound as shortcut (fix #5686)

### LINKED ISSUE #5686: Holding shift no longer adds selections when selecting
**Describe the bug**
Holding shift no longer adds selections when selecting

**To Reproduce**
1. Select Selection tool
2. Make a selection
3. Hold shift and make another selection

**Expected behavior**
Add to selection, like it used to

**Screenshots**

- Aseprite v1.3.17
- Windows 11

leonormoru:
Previously, if the Shift key was assigned to a command shortcut, pressing Shift while making a selection (e.g. Shift+drag) would not add to the selection as expected. This happened because Shift-only shortcuts could consume mouse-down events, preventing additive selection.

After some testing, I came to the conclusion that the issue itself wasn't only related to Shift shortcuts, but also affected any keyboard-only shortcut (such as shortcuts bound to modifier keys like Ctrl, Alt, or combinations of them) when they were assigned to commands. In these cases, pressing the modifier key could swallow mouse-down events, breaking expected behaviors such as additive selection or other mouse interactions that rely on modifiers.

This patch adds a guard to ignore keyboard-only shortcuts (with no mouse button) on mouse-down events. This ensures Shift+mouse actions continue to work for additive selection, even if Shift is bound to a command, for example. Mouse-button shortcuts are unaffected.

The commit also includes a regression test to verify correct handling of this issue. The test was implemented in C++ because the bug involves low-level GUI event processing and shortcut handling logic that is only accessible and verifiable within the C++ application layer.

### DIFF
--- a/src/app/ui/key.cpp
+++ b/src/app/ui/key.cpp
@@ -529,6 +529,13 @@ const AppShortcut* Key::isPressed(const Message* msg, const KeyContext keyContex
       const auto pressedMouseButton = (pressed ? mouseMsg->button() : kButtonNone);
 
       for (const AppShortcut& shortcut : shortcuts()) {
+        // Command shortcuts that are keyboard-only (e.g. Shift) must
+        // not match mouse events; otherwise they can consume
+        // interactions that rely on modifier actions in specific
+        // contexts (e.g. Shift+drag in selection tools).
+        if (m_type == KeyType::Command && shortcut.mouseButton() == kButtonNone)
+          continue;
+
         if ((shortcut.modifiers() == mouseMsg->modifiers()) &&
             ((shortcut.scancode() == kKeyNil && shortcut.unicodeChar() == 0) ||
              (shortcut.modifiers() == kKeySpaceModifier && shortcut.scancode() == kKeySpace &&
--- a/src/app/ui/keyboard_shortcuts_tests.cpp
+++ b/src/app/ui/keyboard_shortcuts_tests.cpp
@@ -168,6 +168,38 @@ TEST(KeyboardShortcuts, FramesSelection)
   EXPECT_COMMAND_FOR_KEY(SetLoopSection, kKeyF2, KeyContext::FramesSelection);
 }
 
+TEST(KeyboardShortcuts, KeyboardOnlyCommandDoesNotWinOnMouseDown)
+{
+  ks->clear();
+
+  // Define a keyboard-only shortcut (Shift with no key/button):
+  // This should NOT trigger on mouse events.
+  KeyPtr keyboardOnlyCommand = ks->command(CommandId::Zoom(), {}, KeyContext::Any);
+  keyboardOnlyCommand->add(ui::Shortcut(kKeyShiftModifier, kKeyNil, 0),
+                           KeySource::UserDefined,
+                           *ks);
+
+  // Define a mouse shortcut (Shift + Left Mouse Button):
+  // This should trigger on mouse events with Shift pressed.
+  KeyPtr mouseCommand = ks->command(CommandId::Undo(), {}, KeyContext::Any);
+  mouseCommand->add(ui::Shortcut(kKeyShiftModifier, kButtonLeft), KeySource::UserDefined, *ks);
+
+  // Simulate a mouse down event with Shift pressed.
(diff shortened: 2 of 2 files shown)


dacap:
I'm trying to repro the bug, but given the keyboard configuration shared here: https://github.com/aseprite/aseprite/issues/5686#issuecomment-4014765017 there is no command associated with the <kbd>Shift</kbd> key, so I'm not sure if this is the actual bug.

The test is also is quite basic: it's only testing the `if` condition inside `should_ignore_keyboard_only_shortcut_on_mousedown`, so I'm not sure if it adds too much value to the PR.

Could you please share some `aseprite.user-keys` file to reproduce the bug and how this PR fixes it?

leonormoru:
Hi @dacap, thanks for the feedback!

Regarding reproduction: while the original reporter's configuration didn't explicitly show a Shift-bound shortcut, the behavior described matches what @Gasparoken identified in the issue thread. By following those steps (assigning Shift to a command like _View > Show > Tile Numbers_), I was able to consistently reproduce the swallowed mouse-down event that prevents additive selection.

This PR addresses the underlying logic where keyboard-only shortcuts can interfere with mouse interactions. Here is a sample [aseprite.user-keys](https://github.com/user-attachments/files/26581235/user.zip) file that reproduces the issue on the current main branch and behaves correctly with this patch applied. If it would help, I can share a short video demonstrating the issue and the fix.

Regarding the testing, could you clarify on what kind of coverage you'd find most helpful here?

dacap:
> Regarding reproduction: while the original reporter's configuration didn't explicitly show a Shift-bound shortcut, the behavior described matches what @Gasparoken identified in the issue thread. By following those steps (assigning Shift to a command like _View > Show > Tile Numbers_), I was able to consistently reproduce the swallowed mouse-down event that prevents additive selection.

I'll give a try to that configuration, anyway I'm worried about that this is not going to solve the original user's issue.

> Regarding the testing, could you clarify on what kind of coverage you'd find most helpful here?

An unit test for such a simple function doesn't add something of value. Probably you'd like to give a look at `keyboard_shortcuts_tests.cpp` and see if there is a case we can add there with a specific keyboard configuration to test these cases.

dacap:
> This PR addresses the underlying logic where keyboard-only shortcuts can interfere with mouse interactions. Here is a sample [aseprite.user-keys](https://github.com/user-attachments/files/26581235/user.zip) file that reproduces the issue on the current main branch and behaves correctly with this patch applied.

Actually this looks like a problem about `KeyContext` preferences. In your configuration the Shift key is configured to execute `ShowTileNumbers` command for `KeyContext::Any` context. And the AddSelection is pre-configured to be enabled with the Shift key in the `KeyContext::SelectionTool` context.

Probably a bug related to `Key::fitsContext()`, but I'm not sure.

leonormoru:
Hi @dacap, that makes sense. The context priority between `KeyContext::Any` and `KeyContext::SelectionTool` definitely seems like a good direction to investigate.

I'll dig into that and check out `keyboard_shortcuts_tests.cpp` to add a better test case for these context conflicts. I'll update the PR as soon as I have something!

leonormoru:
Hi @dacap, I’ve just updated the PR.

I spent some time looking into the `KeyContext` and `Key::fitsContext()` logic as you suggested, but I couldn't find a clear bug there that could cause the behavior described in the original issue.

Because of that, I decided to focus on the event-swallowing issue itself. Even if we haven't pinpointed why the original user's config triggers this, this guard prevents command shortcuts from hijacking modifiers during mouse actions, which we know is happening in some cases and causes the same behavior described in the issue.

I moved the logic from the GUI layer to `key.cpp` since I realized that was where the issue was actually originating, and I also updated the tests in `keyboard_shortcuts_tests.cpp` to properly cover this conflict, following your suggestions.

Meanwhile, I messaged the original reporter in the issue thread to see if they can give us more info, which will hopefully give us more clues on what could be causing their issue. Let me know what you think of this approach!

leonormoru:
Hi @dacap, I was just wondering if you’ve had some time to look over the new changes. Let me know if there's anything else needed on my end!

PR #5720, opened 2026-03-26
Fix regression crash when removing cel in tool loop (fix #5717)

### LINKED ISSUE #5717: Reproducable Crash Bug (Non-Idx0 Transparency)
**Describe the bug**
When setting the transparent color to an Idx other than 0 in Index Color Mode, new Tilemap layers will set the default "empty" tile as being full of Idx-0. Attempting to fix this causes an immediate crash with no error message.

**To Reproduce**
1. Create a new file in Index Color mode.
2. Through either Middle Click or Sprite/Properties/Transparent Color, set the transparent color to any Idx other than 0.
3. The canvas will now be filled with Idx-0. This is expected behavior. Select the full canvas and use Ctrl-X to remove the canvas full of Idx-0.
4. Create a new tilemap layer. The default tile should appear to be filled with Idx-0.
5. With the tilemap layer active, set tilemap editing to "Manual: Modify Existent Tiles."
6. With the paint bucket selected, attempt to fill the canvas with whatever color Idx you've set to be the transparent color. This will cause an immediate crash.

**Expected behavior**
Either the tilemap layers should be using the set transparent color on creation, or the program should not crash when attempting to fix this issue.

- Aseprite: v1.3.18-beta1
- System: Windows 10 Pro 10.0.19045
- Extensions: Aseprite 98 Theme

**Additional context**
I believe this to be related to an issue I posted prior regarding several aspects of the program that incorrectly assume Idx-0 to be transparency regardless of the current project settings.

ckaiser:
Regression from 05f2a1e75b1d42ae5b6af5e1f83150145e4dd2f2, because of the reworked logic we were calling `getDestTileset()` even when we created a new cel, so the new cel was never attached to a layer and we crashed when [we removed it later](https://github.com/aseprite/aseprite/blob/61ccd144bc6d2d5bf11d87815b8f509bc7177d8a/src/app/util/expand_cel_canvas.cpp#L222).

This logic could probably be simplified a bunch with #5684. Related: #5678

### DIFF
--- a/src/app/util/expand_cel_canvas.cpp
+++ b/src/app/util/expand_cel_canvas.cpp
@@ -157,7 +157,7 @@ ExpandCelCanvas::ExpandCelCanvas(Site site,
             m_grid.origin(),
             m_grid.tileSize());
 
-  if (isTilesetPreview()) {
+  if (!m_celCreated && isTilesetPreview()) {
     getDestTileset();
   }
   else if (m_celCreated || (m_layer && m_layer->isTilemap())) {


dacap:
I'll merge this one as it fixes the crash, and then continue with #5732 as I think the Manual mode shouldn't create new tiles.

PR #5722, opened 2026-03-27
Fix TextEdit selection clearing on right click menu

ckaiser:
Fixes a regression introduced in 02abd73234c94dbeaadf0bdbf4f1b24d8bbffb86 that broke cut/copy/paste from the menu, it also introduces he "readonly" XML property for parity with Entry.


dacap:
LGTM :+1:

PR #5728, opened 2026-04-02
Fix CLI --oneframe command

### LINKED ISSUE #5709: CLI argument --oneframe still exports a file per frame
**Describe the bug**
The CLI argument `--oneframe` exports multiple files rather than the 1 specified. It suffixes frame numbers to them, and all files but the first one are just the background layer (usually transparent or black).

**To Reproduce**
1. Get a `input.aseprite` with multiple frames. 
2. Run `aseprite -b --oneframe input.aseprite --save-as thumbnail.png`. (as suggested by the wiki) 
3. Check output files.

**Expected behavior**
Should result in a single output file `thumbnail.png`. The same behavior as `--frame-range 0,0` (which works correctly)

**Screenshots**
`input.aseprite`: (with a different image on each frame)
Image

Command:
Image

Output files:
Image

- Aseprite: v1.3.18-beta1-x64
- System: Arch linux

mrgreenapple24:
Fix #5709 

Get a input.aseprite with multiple frames.
Run aseprite -b --oneframe input.aseprite --save-as thumbnail.png. (as suggested by the wiki)
Check output files.

Results in a single output file thumbnail.png.

Old Behaviour:
image

Fixed Behaviour:
image

### DIFF
--- a/src/app/cli/cli_processor.cpp
+++ b/src/app/cli/cli_processor.cpp
@@ -584,6 +584,8 @@ int CliProcessor::process(Context* ctx)
         // --oneframe
         else if (opt == &m_options.oneFrame()) {
           cof.oneFrame = true;
+          cof.fromFrame = 0;
+          cof.toFrame = 0;
         }
         // --export-tileset
         else if (opt == &m_options.exportTileset()) {
--- a/src/app/cli/cli_processor.cpp
+++ b/src/app/cli/cli_processor.cpp
@@ -299,6 +299,7 @@ int CliProcessor::process(Context* ctx)
 
           cof.fromFrame = base::convert_to<frame_t>(splitRange[0]);
           cof.toFrame = base::convert_to<frame_t>(splitRange[1]);
+          cof.oneFrame = false;
         }
         // --ignore-empty
         else if (opt == &m_options.ignoreEmpty()) {
@@ -584,6 +585,8 @@ int CliProcessor::process(Context* ctx)
         // --oneframe
         else if (opt == &m_options.oneFrame()) {
           cof.oneFrame = true;
+          cof.fromFrame = 0;
+          cof.toFrame = 0;
         }
         // --export-tileset
         else if (opt == &m_options.exportTileset()) {


Gasparoken:
Thank you @mrgreenapple24. Works great👍! Just two comments:
Recommended commit message:
```
Fix CLI --oneframe command (fix #5728)

Before this fix, the `--oneframe` argument in the following CLI line:
"aseprite -b --oneframe input.aseprite --save-as thumbnail.png"
saved the entire sequence of frames instead of just the first frame.
```
I also tested the following command:
`aseprite -b --oneframe --frame-range 2,3 input.aseprite --save-as thumbnail.png`
While this combination does not make much sense, the expected behavior is that the last option takes precedence, in this case, `--frame-range`, but I found the output frames were empty.
To fix this we'll need to explicitly set `oneframe` to false when `--frame-range` is used.

src/app/cli/cli_processor.cpp
Gasparoken:
```suggestion
  cof.toFrame = base::convert_to<frame_t>(splitRange[1]);
  cof.oneFrame = false;
}
```

Gasparoken:
@mrgreenapple24, would you mind implementing this suggestion?

Gasparoken:
Sorry, I think my suggestion had two mistakes: the indentation was wrong, and it duplicated the line immediately above my suggestion. It's my first time using GitHub's suggestion feature, so I'll pay closer attention next time.
My intention was to insert the line as described in the picture
Screenshot 2026-06-02 at 4 57 09 PM

Please create a single commit containing both your changes and the suggestion. **From the tip of your fix/cli-oneframe branch**, you can combine the two commits with:

(before running the next command, make sure you're on commit [5310832](https://github.com/aseprite/aseprite/pull/5728/commits/5310832bb0ce590b093ceb35cb6caf406f1b63c5))
```
git reset --soft HEAD~2
```

then:
```
git commit -m "Fix CLI --oneframe command (fix #5728)

Before this fix, the \`--oneframe\` argument in the following CLI line:
\"aseprite -b --oneframe input.aseprite --save-as thumbnail.png\"
saved the entire sequence of frames instead of just the first frame."
```

Finally, force-push to update this PR:
```
git push <your remote> fix/cli-oneframe --force
```

Just in case: There's no need to include me as a co-author.

Gasparoken:
Perfect! LGTM👍

PR #5732, opened 2026-04-04
Fix crash drawing on empty tile layer in manual mode (fix #5620)

### LINKED ISSUE #5620: Crash after attempting to draw in an empty tile layer in Manual mode
**Describe the bug**
Aseprite is crashing after attempting to draw in an empty tile layer in Manual mode. If there is even one tile, then all is good. Likewise, if you first draw a tile, and then delete it, then there all is good too.

**To Reproduce**
1. Create new Tilemap layer
2. Change the drawing mode to Manual
3. Click on the canvas
4. Crash

**Expected behavior**
Sprite will remain unchanged

**Screenshots**
Image

- Aseprite: 1.13.7-beta2-x64, Steam/beta
- System: Windows

miguelfabiao125-bit:
Upon investigating the cell-addition logic for tilemap layers, I found that Manual Mode was creating new cels without properly attaching them. While PR #5720 resolved the immediate crash, the first interaction in Manual Mode was still incorrectly committing a tile.

To resolve this, I implemented an early exit within the commit function to bypass the data-writing process when in Manual Mode. However, this initially prevented the preview layer from being cleared.

To ensure a consistent UI state, I updated drawing_state to explicitly clean the preview buffer when the drawing loop is destroyed. This ensures that "ghost" pixels are removed even when a commit is opted out.

### DIFF
--- a/src/app/ui/editor/drawing_state.cpp
+++ b/src/app/ui/editor/drawing_state.cpp
@@ -14,6 +14,7 @@
 #include "app/commands/command.h"
 #include "app/commands/commands.h"
 #include "app/commands/params.h"
+#include "app/modules/gui.h"
 #include "app/tools/controller.h"
 #include "app/tools/ink.h"
 #include "app/tools/tool.h"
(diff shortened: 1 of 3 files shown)


dacap:
Thanks @miguelfabiao125-bit for this PR. Although this fixes the manual mode to avoid committing/adding new tiles, I'm not sure if we should show the preview at all. For example, if we already have a tilemap defined, the manual mode doesn't paint outside the tilemap area, so in the initial case we shouldn't show/preview nothing, probably showing a forbidden icon just like if the layer is locked.

PR #5738, opened 2026-04-07
New standalone Lua Engine

### LINKED ISSUE #5530: dofile() doesn't work for relative files when used in Dialog:show{wait=false}
A dialog running in background (`Dialog:show{wait=false}`) loses the "current script dir", so a `dofile()` call doesn't know where the relative dir is located.

E.g.

`main.lua`
```lua
Dialog("test dofile")
  :button{text="click", onclick=function() dofile("./sub.lua") end }
  :show{wait=false}
```

`sub.lua`:
```lua
print("hi")
```

ckaiser:
Replaces the singleton Lua Engine with a new standalone one, this comes with a few refactors and new classes.

The biggest change is that now scripts and extensions do not share globals, and extension timers and events are properly unregistered and disposed of when they're disabled.

Running a script that doesn't have anything "lingering" (aka a timer, dialog or any registered events) will leave the Engine running and will function as before, otherwise a new Engine is created and then disposed of after script execution. In theory we could also monitor for changes to the timers/dialogs/events and then dispose of the engines, but already the fact we're immediately deleting them is a big change and in the end it's not that much memory to keep them around.

The Engine itself has been updated with a few extra features as well:

 * Memory usage tracker, with a memory limit set to 256 mb atm.
 * Optimized startup times, replaced some raw Lua code with the API equivalent
 * Fixes #5530 due to better tracking of the "current" script
 * Modifying package.path we're able to get rid of the custom require function, it should now function the same without having to register and use that.
 * Added Engine tests to validate things.

This PR also removes the debugger and dev console. I'll be re-introducing the debugger in a separate PR.


dacap:
I'll be reviewing this in these days and leaving some comments. At a first glance, all the impl code of `events.h` should go to a new `events.cpp` file and reduce `#include`s/code to a minimum in that header file.

src/app/commands/cmd_run_command.cpp
```diff
@@ -264,7 +264,10 @@ class RunnerWindow final : public gen::RunCommand {
 
     const base::Chrono timer;
     const std::string& text = search()->text();
-    const bool result = App::instance()->scriptEngine()->evalCode(text.substr(1, text.length()));
+    if (!m_evalEngine)
+      m_evalEngine = std::make_unique<script::Engine>();
+
+    const bool result = m_evalEngine->evalCode(text.substr(1, text.length()));
```
dacap: A couple of things here, first we should remove "text.length()" and use the `npos` as second argument for `substr` (because the length should be `text.length()-1` to be precise):
```suggestion
    const bool result = m_evalEngine->evalCode(text.substr(1));
```
And as a second point, just testing something like RunCommand > `@print "hi"` is not working.

src/app/extensions.cpp
dacap, line 951: This executes the `init()` function several time (one time for each script specified in the `package.json` file). This is because we are not resetting the `init()/exit()` globals now.

Probably we can create a test case for extensions with multiple scripts for this specific case.

src/app/console.h
```diff
@@ -19,7 +19,8 @@ class Console {
   Console(Context* ctx = nullptr);
   ~Console();
 
-  void printf(const char* format, ...);
+  static void printf(const char* format, ...);
+  static void print(std::string message);
```
dacap: Could we change it to `const std::string&` or `std::string_view` and print the extra `'\n'` without modifying the given string?
```suggestion
  static void print(const std::string& message);
```

src/app/console.cpp
```diff
@@ -250,8 +250,15 @@ void Console::printf(const char* format, ...)
   std::string msg = base::string_vprintf(format, ap);
   va_end(ap);
 
-  if (!m_withUI) {
-    fputs(msg.c_str(), stdout);
+  print(msg);
```
dacap, line 226: Here `printf()` function behavior changed, a new `'\n'` is added. Before it were printing just the `msg` string as it was. Search for `console.printf` and you will notice a lot of strings already include the `'\n'` char.

E.g. Testing `aseprite -b -save-as hi.png` will output:
```
A document is needed before --save-as argument

```
instead of
```
A document is needed before --save-as argument
```

src/app/console.cpp
```diff
@@ -250,8 +250,15 @@ void Console::printf(const char* format, ...)
...
+}
+
+void Console::print(std::string message)
+{
+  message.push_back('\n');
+
+  if (!isUIAvailable()) {
```
dacap, line 231: Probably
```suggestion
  if (!m_withUI) {
```

src/app/app.cpp
```diff
@@ -88,8 +88,8 @@
 #include <memory>
 
 #ifdef ENABLE_SCRIPTING
-  #include "app/script/engine.h"
   #include "app/shell.h"
+  #include "script/engine_manager.h"
```
dacap:
```suggestion
  #include "app/script/engine_manager.h"
```

src/app/app.h
dacap: We can remove this forward declaration:
```diff
-#ifdef ENABLE_SCRIPTING
-namespace script {
-class Engine;
-}
-#endif
-
```

dacap (changes requested):
Here is a first batch of review comments, still several files to be checked, but already some changes to be made:

src/app/commands/cmd_run_command.cpp
ckaiser: I forgot to connect `ConsolePrint` and `ConsoleError`. Now that we can actually redirect this, I can improve things so we the output of commands in here can go to an inline console in the run dialog and make it a nicer REPL interface similar to what you'd get running in shell mode. I think I'll leave that for a separate PR though.

src/app/extensions.cpp
ckaiser: Whoops, I did not realize that you could have multiple init/exit entry points in a single extension, fixed.

src/app/console.cpp
ckaiser: Hm, yeah the issue here is that before we had a bunch of different places sending different things to the Console, from the engine itself, through the old delegate and direct usage like this, and my idea was to simplify things by just making the print functions static, but I also retained the old usage of instanciating an object.

Personally I prefer just having it be static, I went through all the usages and added a println version (and also found a bunch of places where we using printf without a newline).

src/app/console.cpp
ckaiser: Now that it's static calling the function is the only way to check. We could cache the value, but it's kinda nice that we check if we have a display and use the console otherwise? Not sure if there's a case where that could change.

src/app/script/engine_manager.h
```diff
@@ -0,0 +1,58 @@
+#ifndef ASEPRITE_ENGINEMANAGER_H
+#define ASEPRITE_ENGINEMANAGER_H
```
dacap: We should follow this names (probably some file is not following this, but this is the way):
```suggestion
#ifndef APP_SCRIPT_ENGIME_MANAGER_H_INCLUDED
#define APP_SCRIPT_ENGIME_MANAGER_H_INCLUDED
```

src/app/script/engine_manager.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include "app/console.h"
+#include "engine.h"
```
dacap:
```suggestion
#include "app/script/engine.h"
```

src/app/script/engine_manager.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace app { namespace script {
+
+using Engines = std::map<std::string, std::shared_ptr<Engine>>;
+class EngineManager {
```
dacap, line 15: Move the impl to its own engine_manager.cpp file.

src/app/script/engine_manager.h
```diff
@@ -0,0 +1,58 @@
...
+    auto engine = std::make_unique<Engine>();
+    engine->ConsolePrint.connect(&Console::println);
+    engine->ConsoleError.connect(&Console::println);
+    return engine;
+  }
+
+  // Creates a new Engine and evals the given filename, if the filename entry point was already
+  // used, the engine is reused.
+  // This engine's memory is handled by the manager and is released on exit.
+  static void createForScript(const std::string& filename, const Params& params)
```
dacap: I've found confusing this line:
```
script::EngineManager::createForScript(m_filename, m_params);
```
Because this "create" function doesn't return nothing.

Probably changing this name to `evalScriptInItsOwnEngine()` or `createEngineEvalScript()` or something like that.

src/app/script/engine_tests.cpp
```diff
@@ -0,0 +1,351 @@
...
+#ifdef ENABLE_SCRIPTING
+  #include "app/app.h"
+  #include "app/cli/app_options.h"
+  #include "app/commands/commands.h"
+  #include "app/context.h"
+  #include "app/doc.h"
+  #include "app/script/luacpp.h"
+  #include "base/fs.h"
+  #include "base/time.h"
+  #include "engine.h"
```
dacap:
```suggestion
  #include "app/script/engine.h"
```

src/app/script/engine_benchmark.cpp
```diff
@@ -0,0 +1,34 @@
+// Aseprite
+// Copyright (C) 2026-present  Igara Studio S.A.
+//
+// This program is distributed under the terms of
+// the End-User License Agreement for Aseprite.
+
+#include <benchmark/benchmark.h>
+
+#ifdef ENABLE_SCRIPTING
+  #include "engine.h"
```
dacap:
```suggestion
  #include "app/script/engine.h"
```

src/app/shell.cpp
dacap: This is an old `#include` from several refactors between scripting engines (21921fdfd53d54e0e76f3fe224a16bff0b9d0ca6). This can be fixed from `script/engine.h` to `app/script/engine.h`, we can include a fix for this `#include` here too.

src/app/commands/cmd_save_file.cpp
```diff
@@ -234,8 +234,7 @@ void SaveFileBaseCommand::saveDocumentInBackground(const Context* context,
   }
 
   if (fop->hasError()) {
-    Console console;
-    console.printf(fop->error().c_str());
+    Console::printf(fop->error().c_str());
```
dacap: As fop->error() is already a `const std::string&`, it'd be great if we can change all these calls to the `print(const std::string&)` directly, something like:
```suggestion
    Console::print(fop->error());
```
to avoid calling -> `c_str` -> `printf` -> `base::string_vprintf` -> `std::string` -> `print`

src/app/script/engine.cpp
```diff
@@ -32,6 +30,7 @@
 #include "doc/algorithm/flip_type.h"
 #include "doc/anidir.h"
 #include "doc/color_mode.h"
+#include "events.h"
```
dacap:
```suggestion
#include "app/script/events.h"
```

src/app/console.cpp
dacap: I think there is no way to change, but we can use `isUIAvailable()` now that it's a static function.

src/app/console.cpp
```diff
@@ -208,38 +208,11 @@ class Console::ConsoleWindow final : public Window {
...
-  if (!m_withUI)
-    return;
-
-  TRACE_CON("CON: -> withUI=", m_withUI);
-
-  if (!m_console)
-    m_console = new ConsoleWindow;
-}
-
 Console::~Console()
```
dacap: As we talked, now this destructor is not called as there is no instance of `Console` created. We guess we can just remove it.

src/app/console.cpp
```diff
@@ -208,38 +208,11 @@ class Console::ConsoleWindow final : public Window {
...
 {
-  TRACE_CON("CON: } ~Console this=", this, "withUI=", m_withUI);
-
-  if (!m_withUI)
-    return;
-
   if (m_console && m_console->hasConsoleText() && !m_console->isVisible()) {
     m_console->manager()->attractFocus(m_console);
```
dacap: Probably this `attrachFocus()` is not needed anymore, but please check if this e905fa0266f8b049c2a71e28ce94df79a6ca95c7 still works in most cases.

src/app/script/engine.cpp
```diff
@@ -214,72 +132,206 @@ void register_websocket_class(lua_State* L);
...
+    }
+    return nullptr;
+  }
+
+  // Memory usage limiter
+  if (tracker->limit > 0 && tracker->usage + nsize >= tracker->limit) {
+    LOG(ERROR,
+        "Script memory limit hit: Using %d, attempted to allocate %d.\n",
+        tracker->usage,
+        nsize);
```
dacap: Should we use `base::get_pretty_memory_size`?

src/app/script/engine.cpp
```diff
@@ -214,72 +132,206 @@ void register_websocket_class(lua_State* L);
...
 
-  // Standard Lua libraries
-  luaL_openlibs(L);
+  // Register a pointer to the current engine object
+  *static_cast<Engine**>(lua_getextraspace(L)) = this;
+
+  // Load Lua libraries
+  for (const auto& [name, func] : lua_libraries) {
+    luaL_requiref(L, name, func, 1);
+    lua_pop(L, 1); /* remove lib */
```
dacap:
```suggestion
    lua_pop(L, 1); // remove lib
```

src/app/extensions.cpp
```diff
@@ -539,7 +539,7 @@ class ExtensionCustomFormat : public FileFormat {
     lua_pushstring(L, m_definition.binary ? "wb" : "w");
 
     if (lua_pcall(L, 2, 1, 0) != LUA_OK) {
-      Console().printf("Failed to open file for writing");
+      Console::println("Failed to open file for writing\n");
```
dacap: Fix `println` with `\n`
```suggestion
      Console::println("Failed to open file for writing");
```

src/app/app.cpp
```diff
@@ -263,15 +242,6 @@ int App::initialize(const AppOptions& options)
...
 
-  // Notify the scripting engine that we're going to enter to GUI
-  // mode, this is useful so we can mark the stdin file handle as
-  // closed so no script can hang the program if it tries to read from
-  // stdin when the GUI is running.
-#ifdef ENABLE_SCRIPTING
-  if (m_isGui)
-    m_engine->notifyRunningGui();
-#endif
-
```
dacap, line 274: It's great that we can delete a lot of lines for these special cases.

src/app/script/luacpp.cpp
```diff
@@ -11,41 +11,94 @@
 
 #include "app/script/luacpp.h"
 
-#include <cstdio>
-
 namespace app { namespace script {
 
-static const char mt_index_code[] =
```
dacap, line 18: It's nice to have this code converted to C.

dacap (changes requested):
My 2nd batch of comments. I have some files to review yet, but there are some work to be done with this. I'll try to continue tomorrow (you can take/solve these new comments if you wish).

src/app/script/engine_manager.h
ckaiser: I gotta disable automatic includes or something, this always gets me 😭

src/app/script/engine_manager.h
ckaiser: Originally it did do a return, and then it didn't, and then briefly in the DAP branch it did, but now it doesn't... so I'm changing it to `runUserScript`, feels like the most relevant.

src/app/script/events.h
```diff
@@ -0,0 +1,171 @@
...
+                     public DocObserver {
+public:
+  enum : EventType {
+    Unknown = -1,
+    Change,
+    FilenameChange,
+    AfterAddTile,
+#if ENABLE_REMAP_TILESET_EVENT
+    RemapTileset,
+#endif
```
dacap: As all reference to `ENABLE_REMAP_TILESET_EVENT` are gone, we can remove these lines.

src/app/script/engine.cpp
```diff
@@ -214,72 +133,206 @@ void register_websocket_class(lua_State* L);
...
+  if (!g_orig_loadfile) {
     lua_getglobal(L, "loadfile");
-    orig_loadfile = lua_tocfunction(L, -1);
+    g_orig_loadfile = lua_tocfunction(L, -1);
     lua_pop(L, 1);
   }
-  lua_register(L, "loadfile", loadfile);
+  lua_register(L, "loadfile", &wrap<&Engine::lua_loadfile>);
 
+  // TODO: Do we want a global clock or are we keeping one per engine?
```
dacap: I'm not sure if there is a benefit having a clock by engine, or if having a global clock might be helpful for some kind of synchronization.

We can leave this with one clock per engine, which seems nice to have the "running time of a engine", and see in the future if we need a global clock.

src/app/script/engine.cpp
```diff
@@ -214,72 +133,206 @@ void register_websocket_class(lua_State* L);
...
-  register_app_os_object(L);
-  register_app_command_object(L);
-  register_app_preferences_object(L);
-  register_json_object(L);
-
   // Register constants
-  lua_newtable(L);
-  lua_pushvalue(L, -1);
-  lua_setglobal(L, "ColorMode");
+  lua_createtable(L, 0, 5);
```
dacap, line 319: We can leave all these global as they are now, but in a future we should change this with some automation, it looks quite fragile to specify the number of items in lua_createtable beforehand.

src/app/script/engine.h
```diff
@@ -86,56 +84,68 @@ class DebuggerDelegate {
 
 class Engine {
 public:
+  struct MemoryTracker {
+    size_t initialUsage = 0;
+    size_t usage = 0;
+    size_t limit = 256'000'000;
```
dacap, line 90: Probably already planned but it'd be nice to be able to 1) configure this limit, and 2) be asked when an extension reach the limit to increase the available memory for that specific extension.

src/app/script/engine.h
```diff
@@ -86,56 +84,68 @@ class DebuggerDelegate {
...
+  std::unique_ptr<WindowEvents> m_windowEvents;
+  std::map<doc::ObjectId, std::unique_ptr<SpriteEvents>> m_spriteEvents;
 
-private:
-  Engine* m_engine;
-  EngineDelegate* m_oldDelegate;
+  // Holds the base "entry point" of this engine, the last filename with which evalUserFile was
+  // called, remains after execution has finished, for any events/dialogs that might need to run
+  // relative-path scripts.
+  std::string m_baseScript;
```
dacap, line 139: Should we offer an API to access this dir? e.g. `app.fs.baseScript` / `app.scriptName` / something else?

src/app/script/engine_manager.cpp
```diff
@@ -0,0 +1,49 @@
+// Aseprite
+// Copyright (C) 2026-present  Igara Studio S.A.
+//
+// This program is distributed under the terms of
+// the End-User License Agreement for Aseprite.
+
+#include "app/script/engine_manager.h"
+#include "app/console.h"
```
dacap, line 9: Separate these lines, as if we sort the includes it will be the inverse:
```suggestion
#include "app/script/engine_manager.h"

#include "app/console.h"
```
Strange that pre-commit isn't sorting includes, I thought it was doing so 🤔

src/app/script/events.cpp
```diff
@@ -19,537 +19,430 @@
 #include "app/pref/preferences.h"
 #include "app/script/docobj.h"
 #include "app/script/engine.h"
+#include "app/script/events.h"
```
dacap, line 22: We can rename `events.h` to `events_class.h`, or this `events_class.cpp` to `events.cpp`, the idea is that `.cpp` <-> `.h` must match.

src/app/script/events_class.cpp
```diff
@@ -19,537 +19,430 @@
...
 
-  Doc* doc()
-  {
-    Sprite* sprite = doc::get<Sprite>(m_spriteId);
-    if (sprite)
-      return static_cast<Doc*>(sprite->document());
-    else
-      return nullptr;
-  }
+bool Events::empty() const
```
dacap: Put this `Events::empty()` impl after `AppEvents::eventType()` (to follow the same order as in the .h file).

src/app/script/engine_tests.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+  EXPECT_TRUE(engine->evalCode(R"(
+BeforeCommandID = app.events:on('beforecommand',
+  function()
+    print('event.beforecommand')
+  end)
+AfterCommandID = app.events:on('aftercommand',
+  function()
+    print('event.aftercommand')
+  end)
```
dacap: A minor thing, but probably checking the event name too? I'm not sure, anyway it's great to have tests for the engine 👍 
```suggestion
  function(ev)
    print('event.beforecommand ' .. ev.name)
  end)
AfterCommandID = app.events:on('aftercommand',
  function(ev)
    print('event.aftercommand ' .. ev.name)
  end)
```

dacap (changes requested):
Finally some last comments 🫠 it's quite a huge refactor but it looks great 🚀

src/app/script/engine.cpp
ckaiser: It's not too bad since the table will be created properly regardless, it's like calling reserve() on a container

src/app/script/engine.h
ckaiser: Yeah, I'm planning a followup PR with configurable extension limits (for execution time & memory)

src/app/script/engine.h
ckaiser: Not sure in what context that'd be useful, for includes relative paths work fine, idk 🤔

dacap: Probably a `app.scriptDir` would be the most useful, because we could do something like:
```lua
local resource = Image{ fromFile=app.fs.joinPath(app.scriptDir, "resource.png") }
```
In this way we have a possibility to a dofile-like behavior but for other functions like `Image{ fromFile }`

ckaiser: Sounds good, though we can probably leave it for a follow-up PR once we merge this behemoth

dacap: Yes, it can be other PR. Just a question, would you prefer if I squash this PR or keep all the commits?

dacap:
Finally merged! :clinking_glasses:

PR #5742, opened 2026-04-10
Fix #5659: Square constraint doesn't consider the aspect ratio

### LINKED ISSUE #5659: Square constraint of shapes (SHIFT key modifier) ​​does not consider the aspect ratio of the sprite
**Describe the bug**
With a 2:1 aspect ratio sprite, if you hold down the SHIFT key, for example, to draw a circle, an ellipse with a 2:1 aspect ratio will be drawn. The same problem occurs with the Rectangle tool or any element that would be expected to be square (1:1 aspect ratio); it becomes rectangular.

Sprite aspect ratio 2:1, circle tool with SHIFT key modifier pressed:
Image

Reported here: https://community.aseprite.org/t/double-width-pixels-holding-shift-to-make-1-1-aspect-constraint-makes-2-1/27982

**To Reproduce**
1. New sprite with 2:1 aspect ratio
2. Select Ellipse tool
3. Press and hold SHIFT, then left click and drag the mouse to draw a circle
4. Result: the shape is an 2:1 ellipse instead a perfect circle.

**Expected behavior**
The square aspect ratio restriction must take into account the sprite's current aspect ratio to compensate for the distortion it generates.

**Additional context**
Aseprite version 1.3.6.1

dartlord345:
When choosing a different aspect ratio than 1:1, when holding the shift key while trying to draw a shape like a circle,for example, it draws an ellipse instead. This behaviour happens since it doesn't take in account the aspect ratio of the sprite.

What I did to fix was using the ratio on the calculations, fixing the error. I also implemented this on the 45º snap of the line tool, since it showed the same problem. I also added tests to prevent this bug to happen again.

### DIFF
--- a/src/app/script/app_object.cpp
+++ b/src/app/script/app_object.cpp
@@ -448,6 +448,14 @@ int App_useTool(lua_State* L)
     }
   }
 
+  // Square aspect (SHIFT key constraint)
+  type = lua_getfield(L, 1, "squareAspect");
+  if (type != LUA_TNIL && lua_toboolean(L, -1))
+    params.modifiers = static_cast<tools::ToolLoopModifiers>(
+      static_cast<int>(params.modifiers) |
+      static_cast<int>(tools::ToolLoopModifiers::kSquareAspect));
+  lua_pop(L, 1);
+
   // Do the tool loop
   type = lua_getfield(L, 1, "points");
   if (type == LUA_TTABLE) {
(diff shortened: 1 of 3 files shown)


dartlord345:
Hi @dacap , just a quick ping on this. Let me know if you need anything!

dacap:
@dartlord345 I'll try to take a look to this next week :+1:

dacap:
Just testing it works pretty well, the only thing I noticed is that it looks like one angles is lost. E.g., when pixel ratio = 1:1:

image

But with 2:1 we have these angles:

image

I think we lost -64 degrees in this example.

Also I'm not sure if we should offer an option to apply/ignore the pixel ratio to this transformation.

dacap:
Probably part of this should go (in `beta` branch) after https://github.com/aseprite/aseprite/pull/5540 is merged, because extra angles will be available.

PR #5746, opened 2026-04-13
Fix problem of quick tool shortcut not working when in select tool

### LINKED ISSUE #5743: Ctrl+Space as a quick tool for zooming doesn't work on selection tools
**Describe the bug**
Configuring <kbd>Ctrl+Space</kbd> as a Quick Tool for Zoom doesn't work when we are in a selection tool.

**To Reproduce**
1. Configure your `user.aseprite-keys` with the following:
   ```xml
   <?xml version="1.0" encoding="utf-8"?>
   <keyboard version="1">
     <quicktools>
       <key tool="zoom" shortcut="Ctrl+Space"/>
     </quicktools>
   </keyboard>
   ```
2. New File
3. Check that <kbd>Ctrl+Space</kbd> can be used for zooming
4. Select the Rectangular Marquee tool (<kbd>M</kbd> key)
5. Check that <kbd>Ctrl+Space</kbd> doesn't for zooming

From: https://community.aseprite.org/t/quick-zoom-shortcut-stops-working-while-on-selection-tools/28098/1

Rudra-Pakhare:
fixes #5743 

Before: 

https://github.com/user-attachments/assets/28793184-4ee1-4066-9f51-b0cd07c4b99c

After:

https://github.com/user-attachments/assets/42cc632f-13e1-47a4-b40c-fb739dc7fa1d

Also this if the PR #5590 that introduced this bug, with this fix it doesnt regress on that 

https://github.com/user-attachments/assets/c4aca290-36ee-4cd8-abcf-0fa769f28f33

### DIFF
--- a/src/app/ui/editor/editor.cpp
+++ b/src/app/ui/editor/editor.cpp
@@ -1779,12 +1779,9 @@ void Editor::updateQuicktool(const ui::Message* msg)
     const tools::Tool* selectedTool = atm->selectedTool();
 
     // Don't change quicktools if we are in a selection tool and using
-    // the selection modifiers (or Ctrl key to start a copy of the
-    // selection).
+    // the selection modifiers (or msg is not a key message, so we are using the mouse buttons).
     if (selectedTool->getInk(0)->isSelection()) {
-      if ((int(m_customizationDelegate->getPressedKeyAction(KeyContext::SelectionTool)) != 0) ||
-          (int(m_customizationDelegate->getPressedKeyAction(KeyContext::TranslatingSelection)) &
-           int(KeyAction::CopySelection))) {
+      if ((int(m_customizationDelegate->getPressedKeyAction(KeyContext::SelectionTool)) != 0) || msg->type() > 11) {
         if (atm->quickTool())
           atm->newQuickToolSelectedFromEditor(nullptr);
         return;


src/app/ui/editor/editor.cpp
```diff
@@ -1779,12 +1779,9 @@ void Editor::updateQuicktool(const ui::Message* msg)
...
 
     // Don't change quicktools if we are in a selection tool and using
-    // the selection modifiers (or Ctrl key to start a copy of the
-    // selection).
+    // the selection modifiers (or msg is not a key message, so we are using the mouse buttons).
     if (selectedTool->getInk(0)->isSelection()) {
-      if ((int(m_customizationDelegate->getPressedKeyAction(KeyContext::SelectionTool)) != 0) ||
-          (int(m_customizationDelegate->getPressedKeyAction(KeyContext::TranslatingSelection)) &
-           int(KeyAction::CopySelection))) {
+      if ((int(m_customizationDelegate->getPressedKeyAction(KeyContext::SelectionTool)) != 0) || msg->type() > 11) {
```
dacap, line 1784: Using a hard-coded constant value like `11` instead a `MessageType` to compare the message type value is not a good practice.

What happened to the TransformationSelection+CopySelection case?

dacap (changes requested):
Thanks for your PR @Rudra-Pakhare, I've made a comment.

Just in case I have #5743 because I was working on it. Trying to create new test cases for these kind of situations.

src/app/ui/editor/editor.cpp
Rudra-Pakhare: What happened to the TransformationSelection+CopySelection case?
The third video is the one that shows it is working.

The idea behind the solution is as follows:
If the user is still typing the keyboard that means they have not completely typed out their shortcut so we are still in matching phase
If there is a mouse msg that means user has their shortcut typed in and now want to interact with whatever tool they arrive at using the mouse

assumptions are mouse is not any part of shortcut combination, there are no two combination of keys leading to same shortcut.


eg. shortcut for quick zoom is ctrl + space and CopySelection is ctrl
so when user just types ctrl we dont enter the selection tool, depending on next action that is if space we switch quick tool or mouse interaction we do copy selection

Yes instead of hard coding i should use the enum here, If you are ok with this approach will make the changes or we can find another solution

PR #5748, opened 2026-04-14
Allow the CliProcessor to attempt to install extensions (fix #5744)

### LINKED ISSUE #5744: Aseprite can't load .aseprite-extension files
**Describe the bug**

When opening an Aseprite Extension file (.aseprite-extension) on Windows 11, either by double-clicking the file in the File Explorer or using the command line, an error is thrown and the extension file doesn't load. This only happens when starting a new instance of Aseprite, if Aseprite is already open double-clicking on an extension file will open it correctly in the existing process.

Image

- Aseprite: v1.3.17
- System: Windows 11
- Extensions: had the same issue on an install with 20+ extensions and on a portable executable

ckaiser:
Fixes #5744, also useful for portable installs so the user can just drag and drop an extension file onto the executable.

### DIFF
--- a/src/app/cli/cli_processor.cpp
+++ b/src/app/cli/cli_processor.cpp
@@ -595,8 +595,15 @@ int CliProcessor::process(Context* ctx)
         cof.document = nullptr;
         cof.filename = base::normalize_path(value.value());
 
-        if ( // Check that the filename wasn't used loading a sequence
-             // of images as one sprite
+        if (ctx->isUIAvailable() &&
+            base::string_to_lower(base::get_file_extension(cof.filename)) == "aseprite-extension") {
+          Params params = {
+            { "installExtension", cof.filename }
+          };
+          ctx->executeCommand(Commands::instance()->byId(CommandId::Options()), params);
+        }
+        else if ( // Check that the filename wasn't used loading a sequence
+                  // of images as one sprite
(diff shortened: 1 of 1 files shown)


src/app/cli/cli_processor.cpp
```diff
@@ -595,8 +595,15 @@ int CliProcessor::process(Context* ctx)
         cof.document = nullptr;
         cof.filename = base::normalize_path(value.value());
 
-        if ( // Check that the filename wasn't used loading a sequence
-             // of images as one sprite
+        if (ctx->isUIAvailable() &&
+            base::get_file_extension(cof.filename) == "aseprite-extension") {
```
dacap: Probably we could compare with the lower case version of the extension (`base::string_to_lower(base::get_file_extension(cof.filename))`).

dacap (changes requested):
Looks good so far, just a minor change.

As a side note, we could probably include this fix directly to main branch.

ckaiser:
Added the lowercase call, I didn't include it on the main branch because Params doesn't have the initializer_list there so when I did the rebase to main it exploded on me😅, I can change it to not use it but I'd rather use the newer constructor

dacap:
> I can change it to not use it but I'd rather use the newer constructor

Don't worry, we can integrate this in beta as it is, if everything goes ok, `main` will be `beta` in a couple of weeks anyway.

PR #5760, opened 2026-04-20
Fix selection tool performance in tilemap layers in 'Tiles' mode

Gasparoken:
This fix has no associated issue; the improvement was discovered following the finding that @ckaiser mentioned during a review:
Large tiles (e.g. 128x128) caused a noticeable delay when releasing the mouse button during a selection action in tilemap mode 'Tiles'. The issue becomes more noticeable with larger tile sizes.
How to reproduce the original issue:
- Active tilemap layer with large tile size (e.g., 128x128)
- TilemapMode::Tiles
- Rectangular Marquee Tool
- Make a selection.

### DIFF
--- a/src/app/tools/point_shapes.h
+++ b/src/app/tools/point_shapes.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-present  Igara Studio S.A.
 // Copyright (C) 2001-2017  David Capello
 //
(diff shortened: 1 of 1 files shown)


dacap:
I guess this fixes a performance issue as I noticed a heavy CPU usage without this patch, but I'm not sure if this is related to that (as the PR and the commit don't mention the performance issue).

dacap:
I've assigned back this PR @Gasparoken so you can improve a little the commit message just to know what this patch does at the user level (not at the internal level)

PR #5762, opened 2026-04-21
Fix gradient tool doesn't work when used inside a selection (fix #5750)

### LINKED ISSUE #5750: Gradient tool doesn't work when used inside a selection
Reported here: https://community.aseprite.org/t/gradient-tools-coordinates-are-incorrect-based-on-how-far-right-the-selection-is/28264/1

https://github.com/user-attachments/assets/7b7772b1-b5ca-43a1-b346-e8b81759b324

Gasparoken:
The wrong gradient results happened because m_tmpAddress was initialized before the clip process of x1 inside the processScan base function which initialices m_srcAddress/m_dstAddress in such clipped coordinates. Finally, m_tmpAddress goes shifted (misaligned) refered to m_srcAddress/m_dstAddress.

fix #5750

### DIFF
--- a/src/app/tools/ink_processing.h
+++ b/src/app/tools/ink_processing.h
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-present  Igara Studio S.A.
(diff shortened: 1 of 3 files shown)


Gasparoken:
I'll include some test

PR #5771, opened 2026-04-24
Fix cannot select one tile in a tilemap layer in tiles mode (fix #5769)

### LINKED ISSUE #5769: Cannot select one tile in a tilemap layer in tiles mode
Reported: https://community.aseprite.org/t/selecting-1x1-tile-immediately-de-selects-it/28282/1

We cannot select one tile clicking and dragging the mouse a little in the same tile as before.

Regression from https://github.com/aseprite/aseprite/pull/5760

Gasparoken:
fix #5769


dacap:
LGTM :+1:

PR #5776, opened 2026-04-28
Fix zoom level 20% distorts layers (fix #5774)

### LINKED ISSUE #5774: Zoom level 20% distorts layers
**Describe the bug**
Large images become distorted at zoom level 20%. Partially-transparent cels in upper layers with cel boundaries that do not match the canvas size or are off-canvas become stretched at this specific zoom level. This is not a contrived stress test: it has directly impacted my work for a few weeks now and makes it difficult to work on and preview images of this size.

**To Reproduce**
I cannot determine an easy reproduction. This frequently crops up as I'm working with very large images and resizing them such that the aspect ratio changes or the cels get cropped off-canvas. The best I can do is this .ase file I edited until it reproduced the error:
[Zoom Glitch.zip](https://github.com/user-attachments/files/27138744/Zoom.Glitch.zip)

**Expected behavior**
The 20% zoom level looks as expected without cropping / stretching cels.

**Screenshots**

This is the exact same image at two different zoom levels:
Image

- Aseprite: v1.3.17.1 (Steam)
- System: Windows 10
- Extensions: None

**Additional context**
This bug tends to stretch layers horizontally if I resized the canvas to be horizontally smaller, and same for vertical resizes.

Gasparoken:
The problem was that `sx` was 0.20000000000000001 and then `1.0/sx` gave 5.0 which naturally was truncated to 5. In the other hand, `sy` was 0.20000000000000004 and then `1.0/sy` gave 4.9999999999999991 which was truncated to 4.
https://github.com/aseprite/aseprite/blob/e0536556f4f254c317b44fa47becf7ace5100a37/src/render/render.cpp#L277-L278

It would seem better to remove the multiplication and division of `celBounds.w / double(cel_image->width())` and `celBounds.h / double(cel_image->height())` that resulted in 0.20000000000000004 in the first place:
https://github.com/aseprite/aseprite/blob/e0536556f4f254c317b44fa47becf7ace5100a37/src/render/render.cpp#L1368-L1369
However, if I'm not mistaken, this is intended for reference layers. Therefore, I'll leave it as is.

Fix #5774

### DIFF
--- a/src/render/render.cpp
+++ b/src/render/render.cpp
@@ -1,5 +1,5 @@
 // Aseprite Render Library
-// Copyright (C) 2019-2025  Igara Studio S.A.
+// Copyright (C) 2019-present  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This file is released under the terms of the MIT license.
@@ -274,8 +274,8 @@ void composite_image_scale_down(Image* dst,
     return;
 
   BlenderHelper<DstTraits, SrcTraits> blender(dst, src, pal, blendMode, newBlend);
-  int step_w = int(1.0 / sx);
-  int step_h = int(1.0 / sy);
+  int step_w = int(std::round(1.0 / sx));
+  int step_h = int(std::round(1.0 / sy));
   if (step_w < 1 || step_h < 1)
     return;
 


dacap:
It works great :+1: I'll merge it as it is (probably a release in `main` will include this fix tomorrow)

There are some crashes with the render logic which might be due to some similar issues, i.e. src/dst bounds being calculated outside the valid image bounds, e.g. in `composite_image_general` [ASEPRITE-4AZ](https://igara-studio.sentry.io/share/issue/162397ba0a9241309a9bb94bc40d4287/)

PR #5777, opened 2026-04-29
Use format auto-detection only on built-in formats (fix #5772)

### LINKED ISSUE #5772: New custom file format loader is skipped on open if the file looks like a built-in format
**Describe the bug**

I’m testing the new custom file format API added in `v1.3.18-beta2` via #5174 fixing #4950.

My extension registers `.dmi` as a custom file format. However, `.dmi` files are `PNG`-backed, and when I open one, Aseprite loads it as a normal PNG instead of calling the extension’s custom onload handler. This prevents the use of this feature with any formats that extend upon auto-identified ones.

**To Reproduce**

1. Install an extension that registers .dmi with `plugin:newFileFormat()`
2. In the extension’s onload, show an alert so it is obvious when it runs
3. Rename any valid PNG file to test.dmi
4. Open test.dmi in Aseprite

Example:
``` lua
function init(plugin)
  plugin:newFileFormat{
    name = "DMI Repro",
    extension = "dmi",
    onload = function(args)
      app.alert("custom loader called")
      return true
    end,
    onsave = function(args)
      app.alert("custom saver called")
      return true
    end,
  }
end
```

**Expected behavior**
Opening a .dmi file should call the extension’s custom loader for .dmi.

**Screenshots**

**Aseprite & System:**
- Aseprite: `1.3.18-beta2`
- System: Windows
- Extensions: Just my DMI extension.

**Notes:**
The current load path appears to choose the file format from `dio::detect_format(filename)` first, and only if content detection returns `UNKNOWN` does it fall back to extension-based detection.

ckaiser:
Fixes #5772 by first checking the file by the extension and then using content detection only when the detected extension is not registered as a custom format.

### DIFF
--- a/src/dio/detect_format.cpp
+++ b/src/dio/detect_format.cpp
@@ -49,9 +49,16 @@ void unregister_custom_format_extension(const std::string& ext, const FileFormat
 
 FileFormat detect_format(const std::string& filename)
 {
-  FileFormat ff = detect_format_by_file_content(filename);
-  if (ff == FileFormat::UNKNOWN)
-    ff = detect_format_by_file_extension(filename);
+  FileFormat ff = detect_format_by_file_extension(filename);
+
+  if (ff < FileFormat::FIRST_CUSTOM) {
+    // For non-custom formats, we use content detection and override the detection if we find
+    // something valid, to account for incorrect extensions.
+    const FileFormat contentFormat = detect_format_by_file_content(filename);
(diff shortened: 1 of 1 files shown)


dacap:
LGTM 👍 I'd have prefer `contentFormat != UNKNOWN` instead of `contentFormat > UNKNOWN` but it works. I'll merge it as it is so we can release this today.

PR #5778, opened 2026-04-29
Extension installation and command fixes

ckaiser:
Fixes a few assorted extension issues:

 * When an extension with subdirectories was uninstalled, we were leaving the directories behind, since they were never added to installedFiles.
 * [An assert](https://github.com/aseprite/aseprite/blob/793fb6526540b4c6f5cf8381ae189241cb003fdf/src/app/app_menus.cpp#L602) could be triggered when exiting or disabling an extension because a menu item that was created by a command that was deleted lingered.
 * Extension could delete any command either by using deleteCommand or creating a new command with the same name, this includes built-in commands like "About" or... anything else really. Now it checks.
 * Before you could copy paste the extension directory and have a bunch of the same extension with the same name, this is a more "controversial" change since we don't really have an extension `id` per-se, just a "name". Ideally through normal installation this would never happen because the installer checks, but it's IMO good practice to avoid weird cases with extensions registering the same commands, etc.

### DIFF
--- a/src/app/extensions.cpp
+++ b/src/app/extensions.cpp
@@ -1100,7 +1100,7 @@ Extensions::Extensions()
         loadExtension(dir, fullFn, isBuiltinExtension);
       }
       catch (const std::exception& ex) {
-        LOG("EXT: Error loading JSON file: %s\n", ex.what());
+        LOG(ERROR, "EXT: Error loading extension from '%s': %s\n", dir.c_str(), ex.what());
       }
     }
   }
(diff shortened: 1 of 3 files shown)


src/app/extensions.cpp
```diff
@@ -1313,6 +1320,12 @@ Extension* Extensions::installCompressedExtension(const std::string& zipFn,
           continue;
       }
 
+      std::string subpath = base::get_file_path(fn);
+      if (subpath != currentSubpath) {
+        currentSubpath = subpath;
+        if (!subpath.empty())
+          installedFiles.push_back(subpath);
```
dacap: Is it possible in a `.zip` file to have the entries not in order? E.g.
```
subdir1/a.lua
subdir2/b.lua
subdir1/c.lua
```
So that would add two times the same subpath in installFiles.

ckaiser:
Fixed the potential duplicates by just deduplicating every install file, and also added a guard to prevent extensions from deleting another extensions menu groups

dacap:
LGTM :+1:

PR #5782, opened 2026-04-29
Add safe save (atomic file write) (fix #5770)

### LINKED ISSUE #5770: Cancelling `Save` action overwrites original `.aseprite` file with partially empty file, causing data loss
**Describe the bug**
When you make changes to a (large) aseprite file and attempt to save it again to disk, you can cause your aseprite file to be partially emptied out. **This results in a large amount of frames' cels getting completely removed from the file, and your original (saved-on-disk) file to be overwritten with this (near) emptied file, causing you to lose a lot of data in your original Aseprite file, without a way to recover it without backups**.

**To Reproduce**
1. Open a large aseprite file, with multiple frames, cels, etc. The size is important, because you need to be given time to cancel the save action.
2. Make changes to the file.
3. Save the file by pressing CTRL+S
4. Immediately cancel the Save action. This should keep the file on-disk as it was before you made changes in Step 1.
5. Close the Aseprite file. Confirm you want to close without saving.
6. Re-open the Aseprite file.
7. You will see that the file will: 1) Contain the changes you made in Step 1, despite not being saved when the file was closed. 2) A large amount of frames/cels have been removed. I suspect the amount of data lost depends on how quickly you cancel your save action.

**Expected behavior**
When cancelling a Save action, I would expect the on-disk file to stay exactly the same, and to not be overwritten. Only once the save action fully completes can the file be overwritten, until that time, the original should not be touched.

**Screenshots**
Image
_A resized gif, showing the all the reproduction steps leading to a large file losing 10+ frames of data_

- Aseprite: Aseprite v1.3.18-beta2 (original bug happened on v.1.3.17)
- System: Windows 10
- Extensions: A lot of custom-made ones.

**Additional context**
This happened to me yesterday, after almost saving over a file that wasn't properly synced from my PC to the laptop yet, due to me working outside of wifi range for a bit. When I cancelled the save, and closed the file (so I could get the latest file before making the changes), I noticed the file had almost completely wiped. I was a little worried and confused, so I walked up to my PC, which still had the full file. Luckily for me, the two versions were conflicting and caused a sync issue, so no data was lost. I also still had the file opened on my PC, and had recent backups and Aseprite had recently closed recoverable files from a few days before)

I did some testing today to see if I could reproduce the issue, and I could.

Make some backups people!

Gasparoken:
Original issue name: "Cancelling Save action overwrites original .aseprite file with partially empty file, causing data loss"

Example of how it works:
output

fix #5770

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -1680,6 +1680,7 @@ set_cursor_fix_tooltip = Sets the mouse position to the pen location when\nyou h
 wintab_more_info = (More Information)
 flash_selected_layer = Flash layer when it is selected
 use_selection_tool_loop = New selection tools implementation
+safe_save = Safe save (saves to a temp file first)
 non_active_layer_opacity = Opacity for non-active layers:
 cel_content_format = Cel content format:
 cel_format_compressed = Compressed
--- a/data/widgets/options.xml
+++ b/data/widgets/options.xml
@@ -651,6 +651,8 @@
                  pref="experimental.flash_layer" />
           <check id="use_selection_tool_loop" text="@.use_selection_tool_loop"
                  pref="experimental.use_selection_tool_loop" />
+          <check id="safe_save" text="@.safe_save"
+                 pref="experimental.safe_save" />
           <hbox>
             <label for="tooltip_delay" text="@.tooltip_delay" />
             <expr id="tooltip_delay" suffix="@.tooltip_delay_unit" />
--- a/src/app/file/file.cpp
+++ b/src/app/file/file.cpp
@@ -51,9 +51,91 @@
 #include <cstdarg>
 #include <cstring>
 
+#if LAF_LINUX || LAF_MACOS
+  #include <sys/stat.h>
+  #include <unistd.h>
+#elif LAF_WINDOWS
+  #include <windows.h>
+#endif
+
 namespace app {
 
 using namespace base;
+
+class SafeSave {
+public:
+  SafeSave(std::string& filename, FileOp* fop, bool enabled)
+    : m_filename(filename)
+    , m_fop(fop)
+    , m_enabled(enabled)
+  {
+    if (!m_enabled)
+      return;
+
+#if LAF_LINUX || LAF_MACOS
+    if (base::is_file(m_filename)) {
+      struct stat sts;
+      if (stat(m_filename.c_str(), &sts) == 0) {
+        m_restoreAttrs = true;
+        m_mode = sts.st_mode;
+      }
+    }
+#elif LAF_WINDOWS
+    if (base::is_file(m_filename))
+      m_fileAttributes = ::GetFileAttributes(base::from_utf8(m_filename).c_str());
+#endif
+
+    auto now = std::chrono::steady_clock::now().time_since_epoch().count();
+    m_tempFilename = base::join_path(
+      base::get_file_path(m_filename),
+      fmt::format("{}_{}.tmp", base::get_file_name(m_filename), now));
+    std::swap(m_filename, m_tempFilename);
+  }
+
+  ~SafeSave()
+  {
+    if (!m_enabled)
+      return;
+    std::swap(m_filename, m_tempFilename);
+    if (!m_fop->hasError() && !m_fop->isStop()) {
+      try {
+        base::move_file(m_tempFilename, m_filename, true);
+#if LAF_LINUX || LAF_MACOS
+        if (m_restoreAttrs)
+          chmod(m_filename.c_str(), m_mode);
+#elif LAF_WINDOWS
+        if (m_fileAttributes != INVALID_FILE_ATTRIBUTES)
+          ::SetFileAttributes(base::from_utf8(m_filename).c_str(), m_fileAttributes);
+#endif
+      }
+      catch (const std::exception& e) {
+        m_fop->setError("Error overwriting file: %s\n", e.what());
+      }
+    }
+    else {
+      if (base::is_file(m_tempFilename)) {
+        try {
+          base::delete_file(m_tempFilename);
+        }
+        catch (const std::exception& e) {
+          m_fop->setError("Error deleting temporary file: %s\n", e.what());
+        }
+      }
+    }
+  }
+
+private:
+  std::string& m_filename;
+  std::string m_tempFilename;
+  FileOp* m_fop;
+  bool m_enabled;
+#if LAF_LINUX || LAF_MACOS
+  bool m_restoreAttrs = false;
+  mode_t m_mode = 0;
+#elif LAF_WINDOWS
+  DWORD m_fileAttributes = INVALID_FILE_ATTRIBUTES;
+#endif
+};
 
 class FileOp::FileAbstractImageImpl : public FileAbstractImage {
 public:
@@ -996,6 +1078,8 @@ void FileOp::operate(IFileOpProgress* progress)
           // Make directories
           makeDirectories();
 
+          SafeSave safeSave(m_filename, this, m_config.safeSave);
+
           // Call the "save" procedure... did it fail?
           if (!m_format->save(this)) {
             setError("Error saving frame %d in the file \"%s\"\n",
@@ -1021,6 +1105,8 @@ void FileOp::operate(IFileOpProgress* progress)
       if (m_abstractImage) {
         m_abstractImage->setSpecSize(m_roi.fileCanvasSize(), m_roi.fileCanvasSize());
       }
+
+      SafeSave safeSave(m_filename, this, m_config.safeSave);
 
       // Call the "save" procedure.
       if (!m_format->save(this)) {
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -260,6 +260,7 @@
       <option id="nonactive_layers_opacity" type="int" default="255" />
       <option id="nonactive_layers_opacity_preview" type="int" default="255" />
       <option id="tooltip_delay" type="int" default="300" />
+      <option id="safe_save" type="bool" default="false" />
     </section>
     <section id="developer">
       <option id="debug_paint_messages" type="bool" default="false" />
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -97,6 +97,7 @@ add_library(app-lib ${generated_files})
 # These specific-platform files should be in an external library
 # (e.g. "base" or "os").
 if(WIN32)
+  target_compile_options(app-lib PRIVATE -DNOMINMAX)
   target_sources(app-lib PRIVATE fonts/font_path_win.cpp)
 elseif(APPLE)
   target_sources(app-lib PRIVATE fonts/font_path_osx.mm)
--- a/src/app/file/file_op_config.cpp
+++ b/src/app/file/file_op_config.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-present  Igara Studio S.A.
 //
 // This program is distributed under the terms of
 // the End-User License Agreement for Aseprite.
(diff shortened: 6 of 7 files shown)


src/app/file/file.cpp
```diff
@@ -1018,9 +1018,39 @@ void FileOp::operate(IFileOpProgress* progress)
         m_abstractImage->setSpecSize(m_roi.fileCanvasSize(), m_roi.fileCanvasSize());
       }
 
+      std::string originalFilename;
+      std::string tempFilename;
+      if (m_safeSave) {
+        originalFilename = m_filename;
+        tempFilename = m_filename + ".tmp";
```
ckaiser: We should be using a temporary path with `base::get_temp_path()`, that way we can also count on the OS deleting it eventually in case we fail to do that. Generating a random file name or enumerating it might also be worth it? Maybe not.

data/strings/en.ini
```diff
@@ -1633,6 +1633,7 @@ set_cursor_fix_tooltip = Sets the mouse position to the pen location when\nyou h
 wintab_more_info = (More Information)
 flash_selected_layer = Flash layer when it is selected
 use_selection_tool_loop = New selection tools implementation
+safe_save = Safe save (atomic file write)
```
ckaiser: It's an advanced feature, but still 99% of users have no idea what an "atomic file write" is (also some parts aren't _technically_ as atomic as they could be at the moment 🤓)

```suggestion
safe_save = Safe save (saves to a temp file first)
```
maybe? it's a hard one to word

src/app/file/file.cpp
```diff
@@ -1018,9 +1018,39 @@ void FileOp::operate(IFileOpProgress* progress)
...
-        setError("Error saving the sprite in the file \"%s\"\n", m_filename.c_str());
+        setError("Error saving the sprite in the file \"%s\"\n",
+                 (m_safeSave ? originalFilename : m_filename).c_str());
+      }
+
+      if (m_safeSave) {
+        m_filename = originalFilename;
+        if (!hasError() && !isStop()) {
+          try {
+            std::string backupFilename = m_filename + ".bak";
```
ckaiser: Doing it this way works, but on Windows we do have access to [SetFileInformationByHandle](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfileinformationbyhandle) with a `ReplaceIfExists` in the `FILE_RENAME_INFO` which is an atomic operation.

Delegating as much of this as we can to the OS is the ideal, would be nice to add an "overwrite" parameter to `base::move_file` or a new function for this. Unfortunately on other platforms `std::rename` is implementation dependent, worth investigating.

ckaiser (commented):
Left some comments, but I'm looking forward to what IMO should just be the default behavior after a brief testing period

src/app/file/file.cpp
Gasparoken: Perfect. I like it.

data/strings/en.ini
Gasparoken: Initially, the comment was similar to "Safe save (saved via temporary file)." But after searching online to better summarize, I saw that the term "atomic file write" fit the saving scheme quite well. And finally, I thought that what the user was interested in was already present: "Safe save," but if they want more detail, I included the more technical term in case they want to look it up.

src/app/file/file.cpp
Gasparoken: Perfect, I'll implement it this way.

src/app/file/file.cpp
```diff
@@ -1018,9 +1018,34 @@ void FileOp::operate(IFileOpProgress* progress)
         m_abstractImage->setSpecSize(m_roi.fileCanvasSize(), m_roi.fileCanvasSize());
       }
 
+      std::string originalFilename;
+      std::string tempFilename;
+      if (m_safeSave) {
+        originalFilename = m_filename;
+        tempFilename = base::get_temp_path() + base::get_file_name(m_filename) + ".tmp";
```
dacap: Should we write an existing file in the temp path that matches this file? Why not checking if the file exists/create an unique name for the temp file?

There is a bug here, we have to join paths:
```suggestion
        tempFilename = base::join_path(base::get_temp_path(), base::get_file_name(m_filename) + ".tmp");
```

Image

src/app/file/file.cpp
```diff
@@ -1018,9 +1018,34 @@ void FileOp::operate(IFileOpProgress* progress)
...
+        m_filename = tempFilename;
+      }
+
       // Call the "save" procedure.
       if (!m_format->save(this)) {
-        setError("Error saving the sprite in the file \"%s\"\n", m_filename.c_str());
+        setError("Error saving the sprite in the file \"%s\"\n",
+                 (m_safeSave ? originalFilename : m_filename).c_str());
```
dacap: I'm not sure if it makes sense to show the temporal file name without the original name to the user. I'd prefer to tell the user "error saving your file X, we weren't able to save temp file Y".

src/app/file/file.cpp
```diff
@@ -1018,9 +1018,34 @@ void FileOp::operate(IFileOpProgress* progress)
...
+      }
+
+      if (m_safeSave) {
+        m_filename = originalFilename;
+        if (!hasError() && !isStop()) {
+          try {
+            base::move_file(tempFilename, m_filename, true);
+          }
+          catch (const std::exception& e) {
+            setError("Error replacing the original file: %s\n", e.what());
```
dacap: I'd just show an error like:
```suggestion
            setError("Error overwriting file: %s\n", e.what());
```

src/app/file/file.cpp
```diff
@@ -1018,9 +1018,34 @@ void FileOp::operate(IFileOpProgress* progress)
...
+          }
+        }
+        else {
+          if (base::is_file(tempFilename))
+            base::delete_file(tempFilename);
```
dacap: `delete_file()` can throw another exception.

data/widgets/options.xml
```diff
@@ -625,6 +625,7 @@
                  pref="experimental.hue_with_sat_value_for_color_selector" />
           <check id="flash_layer" text="@.flash_selected_layer" />
           <check id="use_selection_tool_loop" text="@.use_selection_tool_loop" />
+          <check id="safe_save" text="@.safe_save" />
```
dacap: Please use the `pref=` attribute for boolean values (this is the "new" way to handle `<check>` options).

data/strings/en.ini
```diff
@@ -1633,6 +1633,7 @@ set_cursor_fix_tooltip = Sets the mouse position to the pen location when\nyou h
 wintab_more_info = (More Information)
 flash_selected_layer = Flash layer when it is selected
 use_selection_tool_loop = New selection tools implementation
+safe_save = Safe save (save process via temporary file)
```
dacap: I think @ckaiser fits quite well:
```suggestion
safe_save = Safe save (saves to a temp file first)
```

src/app/commands/cmd_options.cpp
```diff
@@ -742,6 +742,7 @@ class OptionsWindow : public app::gen::Options {
 #endif
 
     useSelectionToolLoop()->setSelected(m_pref.experimental.useSelectionToolLoop());
+    safeSave()->setSelected(m_pref.experimental.safeSave());
```
dacap: These lines are not required if you use the `pref=` attribute in `<check>`, it should be handled automatically.

laf
dacap, line 1: Would you mind to create a PR for the laf part first?

dacap (changes requested):
Looks good so far, some changes to make.

dacap:
Another issue I've found is that this works only for .aseprite file (non-sequencing files), but it should work in the same way for .png files, etc.

In the case of sequences I'm not sure if we should backup the whole sequence or file by file (I'd do it just file by file just to keep it simple).

src/app/commands/cmd_save_file.cpp
```diff
@@ -221,6 +221,8 @@ void SaveFileBaseCommand::saveDocumentInBackground(const Context* context,
   if (!fop)
     return;
 
+  fop->setSafeSave(Preferences::instance().experimental.safeSave());
```
dacap: Other issue here is that we cannot access Preferences in a background thread, there should be an `ASSERT` in `Preferences::instance()` checking this.

dacap:
@Gasparoken don't worry to update this PR until we can agree in https://github.com/aseprite/laf/pull/174 impl

src/app/commands/cmd_options.cpp
```diff
@@ -974,6 +974,7 @@ class OptionsWindow : public app::gen::Options {
 
     // Experimental features
     m_pref.experimental.useSelectionToolLoop(useSelectionToolLoop()->isSelected());
+    m_pref.experimental.safeSave(safeSave()->isSelected());
```
dacap: This is not required with the `pref=` argument (the load/save of the property is automatically handled).

src/app/commands/cmd_save_file.cpp
dacap: @Gasparoken I think I didn't express well here, you cannot access `Preference::instance()` in this place, you have to pass it through `FileOpConfig`.

src/app/file/file.cpp
```diff
@@ -990,13 +990,22 @@ void FileOp::operate(IFileOpProgress* progress)
           // Make directories
           makeDirectories();
 
+          std::string tempFilename;
+          if (m_safeSave)
+            initSafeSave(tempFilename);
```
dacap: It would be nice if we can create an auxiliary RAII-style class to do this init/finish to avoid errors/missing paths (e.g. in case of exceptions the destructor will be called).

You can find an example of something similar in https://github.com/aseprite/aseprite/blob/main/src/app/ui/editor/brush_preview.h#L146 (but we use this a lot).

dacap (changes requested):

Gasparoken:
Personal opinion: It would be good to release the beta version with Safe Save enabled, and the stable version with it disabled.

dacap:
> Personal opinion: It would be good to release the beta version with Safe Save enabled, and the stable version with it disabled.

I think it will be good to release this in v1.3.19-beta disabled first (as these changes will be in beta only first), then v1.3.19 disabled, and v1.3.20-beta enabled by default for first time.

dacap:
I'm going to test this on Linux, there is a conflict in this PR with the laf module as it was updated. Could you please rebase the PR to beta.

dacap:
Owner and permissions are lost after saving.

Steps on Linux (probably the same on macOS, and should be tested on Windows too):

Change the owner and flags of the file:
```
sudo chown david:www-data Sprite-0001.aseprite
chmod 0655 Sprite-0001.aseprite
```
`ls -la Sprite-0001.aseprite` before saving:
```
-rw-r-xr-x 1 david www-data 552054 Jul  3 10:35 Sprite-0001.aseprite
```
And after saving (owner and flags are overwritten):
```
-rw-rw-r-- 1 david david 552728 Jul  3 10:36 Sprite-0001.aseprite
```
This doesn't happen in the current version.

Gasparoken:
Through testing, I've found that the group inevitably changes with this new method. I tried another program (Affinity Photo), and the group also changes when saving.

Yes, it's true that the group doesn't change using the current method, so it seems relevant to leave the choice of method to the artist (if it's even worthwhile).

On the other hand, I've seen that on macOS and Linux, it's recommended to create temporary files in the same folder as the original file when using `rename()`. This is because if the temporary folder, and therefore the temporary file, is in a different filesystem than the original file, `rename()` throws an EXDEV error (there's a slight difference between Linux and macOS, but the idea is the same: if the temporary file is in a sensitive location, `rename()` can throw an error).
There is an alternative, but it makes `rename()` no longer atomic on Linux and macOS. Personally, I would prioritize atomicity. I would need your perspective to go ahead.

Note: ignore the last couple of commits.

PR #5787, opened 2026-04-30
Replace pre-commit with prek

dacap:


dacap:
It's funny because I think I cannot keep this PR open.

PR #5789, opened 2026-05-04
Replace pre-commit with prek

dacap:
Use [prek](https://prek.j178.dev/) instead of pre-commit by default.


ckaiser (approved):
Been using it locally for a while with no issues so LGTM!

PR #5792, opened 2026-05-05
Move from dropbox/json11 to nlohmann/json

ckaiser:
Replaces the unmaintained [dropbox/json11](https://github.com/dropbox/json11) with [nlhomann/json](https://github.com/nlohmann/json), the only functionality change is that now we require extensions to have a name in package.json, because before you could create a "valid" extension with just an empty file.


dacap:
"nlohmanns-sky" :laughing:

dacap:
It looks great! :rocket:

Issue #5513, opened 2025-11-04
Improve data recovery to avoid deleting old versions / keep alive important sprite backups


dacap:
Related to #4130, sometimes we receive empty [sessions folders](https://www.aseprite.org/docs/data-recovery/), or incomplete sessions, or sessions that are not enough to recover users' work. There are some things that we could do to improve or mitigate this:

- [ ] [Session::isOldSession()](https://github.com/aseprite/aseprite/blob/7cbc4f8d168d1f1c0d56ee1f7c86fec3f40139a7/src/app/crash/session.cpp#L135) does a terrible job to indicate if it's an old session (a session to be deleted). Instead of using the folder's filename, e.g. `20251104-165047-23372` (or the `ver` file modification time as fallback), we should save a file when the session is closed/last used.
  * I can imagine some cases where the user open Aseprite in a notebook (the session is created), and they never shut Aseprite down, just go to sleep and wake up the notebook. The session date is never updated, and just a crash might mark the session as "old".
- [ ] Some sessions might contain important files to "keep alive". These files could be flagged by specific event
  * If the "File > Save" operation is canceled, it'd be a good idea to mark the sprite as "keep alive"
  * If the user was editing the sprite for "a considerable amount of time" (related to #5512) we can mark the sprite as "keep alive"
- [ ] Investigate if the the sessions folder can be saved in other place by mistake (e.g. not in `%APPDATA%\Aseprite` when launched from Steam)

PR #5805, opened 2026-05-11
Improve "Recover Files" UI (related issues #4130, #5513)

dacap:
This PR adds the "Active Session" in the "Recover Files" tab:

image

#### When we close a sprite:

image

#### When a version of the sprite is fully backed up in disk and we close the sprite we can see both versions (the one closed still in memory with undo info, and the backed up sprite in disk):

image

#### After the file in memory is freed only the backed up data is displayed:

image

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -1743,9 +1743,12 @@ refresh = Refresh
 raw_images_as_frames = Raw Images as Frames
 raw_images_as_layers = Raw Images as Layers
 loading = Loading...
+active_session = Active Session
 crash_sessions = Crashed Sessions
 old_sessions = Previous Sessions
 incompatible = [MIGHT BE INCOMPATIBLE v{1}] {0}
+no_closed_docs = When you close a file it will appear here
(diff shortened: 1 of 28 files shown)


Gasparoken:
Perfect! It works great, I like it.

The only note is: "in memory" implies that the sprite is still in RAM, which is correct, but perhaps it would be better to report "undo history" or "w/undo history" or "can undo" to express the advantages of "in memory".

dacap:
> The only note is: "in memory" implies that the sprite is still in RAM, which is correct, but perhaps it would be better to report "undo history" or "w/undo history" or "can undo" to express the advantages of "in memory".

I thought something similar, but I'm as #3718 is planned, the idea is that after that all backups (even the one in disk) will contain the undo information.

dacap:
One thing I don't like is when both versions appear: the one in memory (with undo) and the one in disk (without undo, at the moment). So probably just hiding the one in disk while we have the one in memory would be enough to avoid confusion to the user.

PR #5808, opened 2026-05-11
Fix Drag-and-drop image into layers loses one color (treated as transparency) in RGB sprites (fix #5786)

### LINKED ISSUE #5786: Drag-and-drop image into layers loses one color (treated as transparency) in RGB sprites.
## **Description of the bug**
When dragging and dropping a PNG image into the layers timeline of a project, one color from the imported image is always missing, with is usually the lightest color in the sprite’s palette, and it gets replaced by transparency.

This happens even though the sprite project where i'm trying to drop it is in RGB Color mode (not indexed), so there should be no palette limitations.
‎ 
## **To reproduce**
Dragging a PNG file directly into the layers timeline of a RGBA sprite causes the issue: one of the colors (usually the lightest) is lost and replaced with transparency.
‎ 
## **Expected Behavior**
The image should be imported with all its original colors intact, without any color being replaced or lost!
‎ 
## **Media**
[Here’s a video I made showing exactly what I mean.](https://www.youtube.com/watch?v=48tZR6mPqqQ)
‎ 
## **My Aseprite & System**
- Aseprite: 1.3.17.2-x64
- System: Windows 11 Pro 25H2
- Extensions: None

## **Additional context**
I wasn’t initially sure whether this behavior was intended or a bug. However, since the sprite where i'm trying to drop the image is already in RGB color mode, there shouldn’t be any palette-related limitations, so this appears to be a bug.

If the PNG is dragged into the workspace (creating a new file), it imports correctly with all colors, but it is automatically set to Indexed color mode.

What I’ve been doing as a workaround is opening the image separately, converting it to RGB color mode, converting the background into a layer, and then copying and pasting it into the project sprite.

This workaround works, but it’s inefficient when importing multiple PNG files, which is why dragging directly into the layers timeline is preferred! n.n

Gasparoken:
This fix includes two types of color loss related with background layer:
- Dragging files without transparent color to the timeline.
- Copy-paste images which source image corresponds to a background layer from a sprite.

I'm not entirely sure about the fix since I had to remove a flag that didn't seem to make much sense to me at the moment (`FILE_LOAD_AVOID_BACKGROUND_LAYER`).

A separate issue (discovered during testing #5809): I noticed that a background layer can be inserted in the middle of the timeline hierarchy if:
- You have a sprite with transparent layers
- You drag a sprite file with a background layer to a layer in the middle of the layer hierarchy.

Need advise of the first intention about `FILE_LOAD_AVOID_BACKGROUND_LAYER`.

fix #5786

### DIFF
--- a/src/app/cmd/set_pixel_format.cpp
+++ b/src/app/cmd/set_pixel_format.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-present  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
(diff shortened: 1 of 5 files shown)


dacap:
@Gasparoken all builds failed.

PR #5813, opened 2026-05-12
Fix combo box lists do not expand in the Preference dialog on MacOS

Gasparoken:
Regression of commit d9785e5c363a555fcda30ad0a256abc59f5f6550: 'Fix inactive window passes events to window below (#4561, #4835)'

On MacOS, this fixes combo box lists in the Preferences dialog when the 'UI with multiple windows' option was enabled.

### DIFF
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -1483,6 +1483,15 @@ void Manager::_openWindow(Window* window, bool center)
       spec.transparent(window->isTransparent());
       spec.modal(window->isForeground());
 
+      // Convert the current window to modal if the parent window is also modal
+      // so the child window is not placed behind its modal parent.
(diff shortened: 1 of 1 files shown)


dacap:
I think with this change we cannot create a child background window for the current modal, this breaks that possibility, e.g. a modal dialog in scripts with multiple mini-windows:
```lua
local subs = {}
local modal = Dialog("Modal")
modal:button{ text="Create Sub-window", onclick=function()
  table.insert(subs, Dialog("subwindow"):show{wait=false})
end }
modal:show()
for _,sub in pairs(subs) do
  sub:close()
end
```
Test this with a single window, because with multiple windows this doesn't work as the modal possibility is not implemented on all platforms.

Gasparoken:
Thanks for your feedback! I'll try a different approach.

PR #5816, opened 2026-05-14
Fix mouse events reaching wrong UI widget on macOS (fix #5814)

### LINKED ISSUE #5814: Sometimes an overlapped dialog at the backgroud avoids mouse events in the foreground dialog
**Describe the bug**
Sometimes an overlapped dialog at the backgroud avoids mouse events in the foreground dialog. This happens with 'UI with multiple dialogs'.

**To Reproduce**
1. Execute the following script:
```lua
local subs = {}
local modal = Dialog("Modal")
modal:button{ text="Create Sub-window", onclick=function()
  table.insert(subs, Dialog("subwindow"):show{wait=false})
end }
modal:show()
for _,sub in pairs(subs) do
  sub:close()
end
```
2. Be sure _Edit > Preferences > Experimental > UI with multiple dialogs_ ON
3. Run the script, and press "Create Sub-window"
4. Arrange windows as:
Image

5. Try to move the foreground dialog clicking and moving in the circled area.
6. Result: sometimes, the foreground dialog doesn't move

- Aseprite: 1.3.17.2
- System: MacOS

Gasparoken:
The MacOS native z-order of multiple displays did not match the widget list maintained by the Aseprite UI manager.

To fix this, widgets associated with the display receiving the mouse event are now checked first when searching for the widget that should consume the event. If no widget is found, the whole widget list is iterated as a fallback.

### DIFF
--- a/src/ui/manager.cpp
+++ b/src/ui/manager.cpp
@@ -837,12 +837,13 @@ void Manager::updateMouseWidgets(const gfx::Point& mousePos, Display* display)
   for (auto mouseWidget : mouse_widgets_list) {
     if (get_multiple_displays()) {
       if (display) {
-        if (display != mouseWidget->display()) {
-          widget = mouseWidget->display()->containedWidget()->pickFromScreenPos(screenPos);
-        }
-        else {
-          widget = mouseWidget->pick(mousePos);
-        }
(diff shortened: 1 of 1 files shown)


Gasparoken:
It may not be necessary after applying https://github.com/aseprite/laf/pull/181

PR #5824, opened 2026-05-17
Add "About Extensions" dialog in Preferences

ckaiser:
Adds a new "About Extensions" dialog and button:

https://github.com/user-attachments/assets/1299f879-963f-40ed-a9e0-64533b1f8a5c

Moves the "Open Folder" button in here to avoid clutter, attempts to show as much information as possible that was previously hidden away in the package.json file.

### DIFF
--- a/src/ui/theme.cpp
+++ b/src/ui/theme.cpp
@@ -896,6 +896,7 @@ void Theme::drawTextBox(Graphics* g,
   gfx::Point scroll;
   const text::FontRef& font = widget->font();
   const int lineheight = font->lineHeight();
+  const int lineseparation = (widget->type() == kTextBoxWidget) ? 2 * guiscale() : 0;
   char *beg_end, *old_end;
   int width;
   gfx::Rect vp;
@@ -1016,7 +1017,7 @@ void Theme::drawTextBox(Graphics* g,
     if (w)
       *w = std::max(*w, len);
 
-    y += lineheight;
+    y += lineheight + lineseparation;
 
     if (end) {
       *end = chr;
@@ -1025,7 +1026,7 @@ void Theme::drawTextBox(Graphics* g,
   }
 
   if (h)
-    *h = (y - y1 + scroll.y);
+    *h = (y - y1 - lineseparation + scroll.y);
 
   if (w)
     *w += widget->border().width();
--- a/data/extensions/aseprite-theme/dark/theme.xml
+++ b/data/extensions/aseprite-theme/dark/theme.xml
@@ -543,7 +543,7 @@
             <background color="textbox_face"/>
             <text color="textbox_text" align="left"/>
         </style>
-        <style id="textbox_label" extends="textbox_text">
+        <style id="textbox_label" border="0" extends="textbox_text">
             <background color="window_face"/>
             <text color="text" align="left"/>
         </style>
--- a/data/extensions/aseprite-theme/theme.xml
+++ b/data/extensions/aseprite-theme/theme.xml
@@ -539,7 +539,7 @@
             <background color="textbox_face"/>
             <text color="textbox_text" align="left"/>
         </style>
-        <style id="textbox_label" extends="textbox_text">
+        <style id="textbox_label" border="0" extends="textbox_text">
             <background color="window_face"/>
             <text color="text" align="left"/>
         </style>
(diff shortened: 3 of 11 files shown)


src/ui/theme.cpp
```diff
@@ -896,6 +896,7 @@ void Theme::drawTextBox(Graphics* g,
   gfx::Point scroll;
   const text::FontRef& font = widget->font();
   const int lineheight = font->lineHeight();
+  const int lineseparation = (widget->type() == kTextBoxWidget) ? 2 * guiscale() : 0;
```
ckaiser, line 899: @dacap I've added this because the text looked kinda terrible without any separation, with the kTextBoxWidget condition to make sure I didn't break any other word-wrapped text widget but I'm not sure I should keep the conditional. I also didn't want to touch this code too much because it's.... _vintage_.

martincapello: Hey there, I couldn't resist and I made a [PR](https://github.com/ckaiser/aseprite/pull/2) to avoid this.

martincapello (commented):
Looks great! Just one observation that I'm not sure if it was intentional or not. When resizing the dialog the description is not being expanded (I think), then it looks like aligned to the left:

Image

ckaiser:
@martincapello I've made the changes to keep the label stretching properly and removed some border jank. I think we should keep the hardcoded line separation value until we merge the code for the TextBox multi-line handling with the one in Graphics ([which also has a hardcoded line separation](https://github.com/aseprite/aseprite/blob/a2e803d794c1e5284fb7c880a66a10d7f04d4840/src/ui/graphics.cpp#L529) and just a similar, cleaner implementation).

The Widget parameter is also not the approach I'd go for because this kind of thing should be handled by the theming variables, ideally this should be in the `<dimensions>` XML definition but until we can unify `TextBox` with the rest it feels out of scope for a small cosmetic change to make this widget fit in better. Especially when it's only used in one other place.
The main issue is it's internally using `Theme` to draw and calculate the size hint so we couldn't pull a variable from `SkinTheme`, it needs a bigger refactor. I'm gonna be looking in the future into making the theming variable stuff more consistent since right now we also shouldn't be using Widget::setBorder, etc, when not using a style (there's a warning in debug mode for it, even).

martincapello:
> @martincapello I've made the changes to keep the label stretching properly and removed some border jank. 

Great!

> I think we should keep the hardcoded line separation value until we merge the code for the TextBox multi-line handling with the one in Graphics ([which also has a hardcoded line separation](https://github.com/aseprite/aseprite/blob/a2e803d794c1e5284fb7c880a66a10d7f04d4840/src/ui/graphics.cpp#L529) and just a similar, cleaner implementation).
> The Widget parameter is also not the approach I'd go for because this kind of thing should be handled by the theming variables, ideally this should be in the `<dimensions>` XML definition but until we can unify `TextBox` with the rest it feels out of scope for a small cosmetic change to make this widget fit in better. Especially when it's only used in one other place. The main issue is it's internally using `Theme` to draw and calculate the size hint so we couldn't pull a variable from `SkinTheme`, it needs a bigger refactor. 

I think that the line spacing should be something that we can configure by using styles (`<styles>` section in the theme's XML) and also by overriding it by attributes on the element itself. I just implemented that last part, which IMO it will be needed at some point and that is why I'd decided to go for it. We might not need that property for Aseprite right now (or actually we would just need it for this case), but, as far as I know, the stuff in src/ui, in the future, will be separated into a dedicated repository to make a library so we can reuse it in other projects or just make it accessible to people interested in the UI used by Aseprite, so having the line spacing would be nice.

Anyway, I'm not going to push harder for the change, just wanted to share my opinion.

EDIT: I agree that the themes needs a bigger refactor, I'm just saying that by having the line spacing now could help us not forgetting about it when that refactor comes.

dacap:
Just in case we have to avoid adding new properties to `Widget` as much as we can, even more if it's just for one widget (`TextBox` in this case). We can add a new `ui::Theme` property in this case and a new dimension in the `SkinTheme`.

Even if this might be useful for some future hypothetical application, please let's keep things simple until such thing happens. We are still far away to be able to separate the ui-lib with an useful default `ui::Theme` (non-SkinTheme).

martincapello (approved):

dacap:
@ckaiser I can merge this as it is and leave the line separation as a dimension for the future.

ckaiser:
@dacap I'm good to merge this as is

Issue #5624, opened 2026-01-05
Packed Spritesheet in wrong order (Linux)


DeltaTimed:
On Linux, the Packed mode for Export Spritesheet does not export files in the correct order.
However when doing the same thing on Windows, it exports in the correct order.

1. Go to File >> Export >> Export Spritesheet
2. Select Packed mode
3. Export
4. Sprites are in the wrong order on Linux!

Behavior:
I've used the Windows version as up until recently I was using Windows 11. 
This problem did not occur on Windows so I used a test animation and Wine to export the same file on both Linux and Windows.
Windows exports the packed spritesheet in the correct order while Linux jumbles them around incoherently.

Screenshots:
https://drive.google.com/drive/u/3/folders/1-Go-PIWRAnEMy3SEkhooiQvOVVLKBujn

- Aseprite: [1.3.16, Steam + Portable + Dev]
- System: [Windows + Linux, Windows 11 + Ubuntu 24 LTS]
- Extensions: [None]

It works perfectly fine on Windows so I don't know why it wouldn't do the same on Linux

DeltaTimed:
Push

DeltaTimed:
Push

Javier-AN:
I am having the exact same issue. Very frustrating.

Issue #3608, opened 2022-11-29
ExportSpriteSheet CLI Bug - mergeDuplicates option does not work


snarlynarwhal:
I believe there is a bug with the ExportSpriteSheet CLI command:

https://www.aseprite.org/api/command/ExportSpriteSheet#exportspritesheet

I set mergeDuplicates to false, but duplicate frames are still merged.

### Aseprite and System version

* Aseprite version: v1.2.40-x64 - Steam
* System: Windows 11

snarlynarwhal:
Here is a zip with the aseprite file, the exported spritesheet and animation data, and the lua script I use to export:

File -> Scripts -> ExportSpritesheet

[HeroSprites.zip](https://github.com/aseprite/aseprite/files/10125341/HeroSprites.zip)

snarlynarwhal:
Hi, upon further testing I discovered that it fails when `type` is set to `SpriteSheetType.PACKED`

I was able to reproduce with these options:

```
app.command.ExportSpriteSheet {
  ui=false,
  type=SpriteSheetType.PACKED,
  textureFilename="/Users/gasparoken/Aseprite/tests/sheet.png",
  mergeDuplicates=true,
}
```

Gasparoken:
Oh, I see 😄. The PACKED option is designed for minimize memory (optimizing sheet space). So "Merge Duplicates" is always true (also "Trim Cels" should be true to take advantage of PACKED). Furthermore, this type of sprite sheet ALWAYS needs the JSON file to know which frame is at which location on the sheet.

So, this issue, isn't a bug. If you want to preserve the  duplicates, you need to select HORIZONTAL or VERTICAL, and define a fixed number of columns or rows.

The "PACKED + No merge duplicates" feature was requested in the past here: https://github.com/aseprite/aseprite/issues/2375 and here: [community topic](https://community.aseprite.org/t/export-packed-sprite-sheet-with-out-merging/6201). But I think PACKED worked for game devs because the sprite sheet resulted in a fixed grid of sprites with constant frame size (without saving memory), then each frame has a predictable location and a Json file isn't needed.

I guess this issue should be closed and https://github.com/aseprite/aseprite/issues/2375 as well.

Gasparoken:
[...]

Issue #2375, opened 2020-05-08
Allow unchecking Merge Duplicates with Packed


connynordlund:
I understand the reasoning why we want to check "Merge Duplicates" when exporting spritesheets with Packed. But I really like that Packed makes the columns and everything for me. Having it be an option to uncheck Merge Duplicates is something that would be preferrable for the ones who wants it. Me being one of them.

Aseprite version: 1.2.18-x64
System: Windows 10

dacap:
Hi @connynordlund could you please provide an example what is the expected result of a packed sprite sheet but without merging duplicates. As far I as know if you are using the packed format you will need to read the JSON data, so basically the source of each sprite doesn't matter if it's duplicated or not. But I might be missing something here.

Unknow0059:
I would too like this. It's breaking my animations.
Also, I don't use the json data at all when I export my frames from Aseprite, so that part is irrelevant for me.

asthesus:
I don't understand the point of forcing this. Of course it's optimal, and it may not be obvious why one might want to uncheck it, but that doesn't mean it should be forced.

I created my own animation system for my game. I just want to test the animation I currently have. I may make the system more efficient in the future, but adding a few redundant frames would be fine for now. By removing this option, you are essentially saying to me _"no, this is the most important thing in your entire game now and must be solved first because we decided this way is always more optimal"_. I don't want to change the animation system right now, I just want to test an animation in the system. Please change this.

erniel:
Hi @dacap i also need this feature as im creating animation in Godot Engine using frames and not via JSON data. anytime you export spirtesheet, some frames are missing because of merging duplicates.

Gasparoken:
As a workaround, try this:

1. Sheet Type: `By Rows`
2. Constraints: `Fixed # of Columns`
3. On # columns you can use this expression: `ceil(sqrt(11))` in the entry widget. Where 11 is the total frame count (as an example) of your sprite, in this case it will result in '4' columns x '3' rows sprite sheet.

<img width="250" alt="Screenshot 2023-05-31 at 09 12 39" src="https://github.com/aseprite/aseprite/assets/32165475/7e76f5dc-b7af-4ced-a478-04ac4eb7e463">

PR #5829, opened 2026-05-20
Fix wrong frame sequence on Packed sprite sheet (fix #4474)

### LINKED ISSUE #4474: Wrong tile sequence in exported tilesets when using Packed sheet type
Exported tileset in horizontal strip (CORRECT ORDER):

right

Same tileset in packed (WRONG ORDER): 

wrong

Packing size doesn't change the sequence.

The biggest problem is that the exported JSON is useless, because the tileset sequence doesn't match with the tile index in the tilemap.

Aseprite v1.3.6-dev on Arch Linux.

Gasparoken:
Users frequently misunderstand the purpose of the 'Packed' sprite sheet type. Some Aseprite importers have even assumed that this mode preserves frame or tile ordering.

'Packed' does not guarantee any specific order, and the resulting layout can appear visibly disorganized.
This mode is primarily intended to maximize compression by discarding unused transparent space around frames. Because of this, it should always be used together with a data file describing the position of each frame or tile inside the sprite sheet (note also that the 'Merge duplicates' flag is always checked to promote data compression).

To address this recurring confusion, 'Packed Ordered' has been added. This mode preserves the order of appearance while keeping an approximately square row-to-column ratio.

Internally, it behaves like the 'By Rows' sprite sheet type using the constraint 'Fixed # of columns' = ceil(sqrt(Nframes)) or 'Fixed # of columns' = ceil(sqrt(Ntiles)) when exporting tilesets.

Original issue description:
'Wrong tile sequence in exported tilesets when using Packed sheet type #4474'

This solution could be the solution to other problems:
- Packed Spritesheet in wrong order (Linux) #5624
- Allow unchecking Merge Duplicates with Packed #2375
- ExportSpriteSheet CLI Bug - mergeDuplicates option does not work #3608

fix #4474


Gasparoken:
Some decisions have to be evaluated when 'Source' = 'Tilesets':
- Removed the "duration" field from the JSON output
- Default tile name in JSON is now: "{title} {layer} {tile}"
- Renamed the tile group in JSON from "frames" to "tiles"
- Renamed 'spriteSourceSize' to 'tilesetSourceSize'
- Renamed 'spriteSize' to 'tileSize'
- Automatic switching between the default templates '{title} {layer} {tile}' and '{title} {frame}.{extension}' depending on the selected 'Source' (without string memorization between changes)

martincapello (commented):
It works! I'm not sure if the "Packed ordered" will be needed after https://github.com/aseprite/laf/pull/188 is merged. The only reason I see to keep it is to give users a way to uncheck the "Merge duplicates" box, because the regular "Packed" doesn't allow to do it.

Gasparoken:
> It works! I'm not sure if the "Packed ordered" will be needed after [aseprite/laf#188](https://github.com/aseprite/laf/pull/188) is merged. The only reason I see to keep it is to give users a way to uncheck the "Merge duplicates" box, because the regular "Packed" doesn't allow to do it.

I tested the new stable_sort and work great! I agree that "Packed ordered" isn't needed anymore. I'll open a new PR with the part of the code that solves the other part of problem of #4474. (i.e.: "...The biggest problem is that the exported JSON is useless, because the tileset sequence doesn't match with the tile index in the tilemap...")

PR #5831, opened 2026-05-21
Fix ⌘ key is held down using ⌘O to open native file dialog (fix #4817)

### LINKED ISSUE #4817: Command key is held down when we use ⌘O to open the native file dialog
This happens on macOS, if we press ⌘O to open a file with the native dialog (the default option) is highly probable that then the key will be pressed in the internal state of Aseprite and the Move tool enabled.

Gasparoken:
fix #4817


dacap:
It looks good :+1: I don't like the fact that this is a macOS-specific bug and should be in laf-dlgs, but laf-dlgs doesn't depend on laf-os so at the moment we can leave this as it is but I'll create a new laf issue.

Issue #5801, opened 2026-05-10
Show alert when assigning an already assigned keyboard shortcut to a different command


crashtestjava:
**Did other user suggested a similar idea?**
- [x] No
- [ ] Yes/Links to similar ideas
  > You can try to find a similar feature requests before in:
  > - GitHub issues: https://github.com/aseprite/aseprite/issues?q=label%3Afeature
  > - Community site: https://community.aseprite.org/c/features/7
  > - Steam community: https://steamcommunity.com/app/431730/discussions/1/
  > In case you find a similar feature request, making a comment there will be useful to give some traction and show interest in the feature.

**Is your feature request related to a problem? Please describe.**
If the user assigns an already assigned keyboard shortcut to another command, there is not a warning/alert in place telling them that this will override the already assigned shortcut. Aseprite only tells the user that the shortcut is already assigned, not that the action will override the old shortcut.

**Describe the solution you'd like**
An alert should appear when they reassign the shortcut to a different command, asking if they want to overwrite the old command.

**Describe alternatives you've considered**
I am unsure if it is possible with scripting. 

**Additional context**
x

PR #5833, opened 2026-05-22
Fix Issue #5801

WereWolf1x1:
Implemented the logic to check and give an alert to user if they try to assign a shortcut key that is already assigned. I used AI for understand the codebase and finding those files where I had to make changes. I asked it for logic flow and further fixing my implementation if there were some errors. All of the code was written by me, and AI (Github Copilot) was used as a guidance tool. The required feature if I understood it correctly is working correctly on my locally build aseprite.exe

### DIFF
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -107,6 +107,7 @@ run_script = Run Script\n<<Do you want to run the following script?\n<<\n<<  {0}
 save_sprite_changes = Warning\n<<Saving changes to the sprite\n<<"{0}" before {1}?\n||&Save||Do&n't Save||&Cancel
 save_sprite_changes_quitting = quitting
 save_sprite_changes_closing = closing
+shortcut_overwrite_alert = Warning\n<<Do you really want to overwrite "{0}" keyboard shortcut? The previous assigned shortcut will be deleted\n||&Yes||&No
 overwrite_existent_file = Warning<<File exists, overwrite it?<<{0}||&Yes||&No||&Cancel
 overwrite_files_on_export_sprite_sheet = Export Sprite Sheet Warning\n<<Do you want to overwrite the following file(s)?\n{0}\n||&Yes||&No
 overwrite_files_on_export = Export Warning\n<<Do you want to overwrite the following file?\n<<{0}\n||&Yes||&No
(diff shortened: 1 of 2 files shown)


data/strings/en.ini
```diff
@@ -107,6 +107,7 @@ run_script = Run Script\n<<Do you want to run the following script?\n<<\n<<  {0}
 save_sprite_changes = Warning\n<<Saving changes to the sprite\n<<"{0}" before {1}?\n||&Save||Do&n't Save||&Cancel
 save_sprite_changes_quitting = quitting
 save_sprite_changes_closing = closing
+shortcut_overwrite_alert = Warning\n<<Do you really want to overwrite "{0}" keyboard shortcut? The previous assigned shortcut will be deleted\n||&Yes||&No
```
dacap, line 110: It'd be nice if we can show the current assignment of the shortcut in some way:
```suggestion
shortcut_overwrite_alert = Warning\n<<Do you really want to overwrite "{0}" keyboard shortcut?\nIt's assigned to "{1}", and it will change to "{2}".\n||&Yes||&No
```

src/app/commands/cmd_keyboard_shortcuts.cpp
```diff
@@ -205,13 +205,28 @@ class KeyItem : public KeyItemBase {
...
+        (key->type() != KeyType::DragAction || m_key.get() == nullptr ||
+        m_key->type() != KeyType::DragAction)) {
+        conflict = true;
+        break;
+      }
+    }
+    if (conflict) {
+      if (ui::Alert::show(Strings::alerts_shortcut_overwrite_alert(window.shortcut().toString())) != 1)
+          return;
+    }
```
dacap, line 225: All this code is duplicated in both places, probably most of the logic could be inside `SelectShortcut`.

I didn't check/test the logic, but some simplifications could be made (as `m_key` cannot be `nullptr` at this point).
```suggestion
    for (KeyPtr& key : m_keys) {
      if (key != m_key && key->keycontext() == m_key->keycontext() && key->hasShortcut(window.shortcut()) &&
        // Tools can contain the same keyboard shortcut
        (key->type() != KeyType::Tool || m_key->type() != KeyType::Tool) &&
        // DragActions can share the same keyboard shortcut (e.g. to
        // change different values using different DragVectors)
        (key->type() != KeyType::DragAction || m_key->type() != KeyType::DragAction)) {
        conflict = true;
        break;
      }
    }
    if (conflict) {
      if (ui::Alert::show(Strings::alerts_shortcut_overwrite_alert(window.shortcut().toString())) != 1)
          return;
    }
```

dacap (changes requested):
Thanks for the PR @WereWolf1x1, some minor changes are required to merge this (I'll test this after the changes are made).

PR #5835, opened 2026-05-26
Update Entry and ResouceFinder to avoid using raw char buffers

ckaiser:
Updated Entry: removed the format parameter entirely since nothing was using it, and did a small refactor of ResourceFinder to switch it over to std::string instead of raw strings.


dacap:
Looks great :+1: some of this code comes from the [very beginning](https://github.com/aseprite/aseprite/blob/57e978b102f3ae0924d47b5f6b5f7d1305933363/src/core/dirs.c#L124) :older_man:

Issue #5395, opened 2025-08-29
Rethink "output file name" widget design


dacap:
Talking with @ckaiser we could redesign the file name widget in some way (as a combo box? as a text field + "..." + "⚙️" buttons?), I'm not sure. The current file name widget is used to select an output file name (generally relative to the folder where the current sprite is located). Some issues solved by the current design (#4483):

1. By default, it only shows the output file name relative to the active sprite's folder (no path is displayed)
1. We can use the file dialog to select the output folder/file (although it needs an extra click or a mouse drag/opening the popup)
1. We can write the output file name relatively to the sprite location (including sub folders)
1. We can show/write the absolute path
1. We can explicitly see the output folder (if we are in relative mode, pressing the "..." we can see the absolute folder)

Some problems still to be solved:

1. One extra click to show the file dialog
1. No familiar design with other desktop apps
1. We have to yet solve the "streamer mode": how to avoid showing absolute paths (when *Edit > Preferences > Files > Show full file name path* is unchecked)
1. https://community.aseprite.org/t/make-file-widget-sub-menu-configurable-v-1-3-14-beta1/24806 (improve navigation with the keyboard)
1. https://community.aseprite.org/t/focus-property-no-longer-works-for-file-widget/27130 (fix focus issue)

If we redesign this widget, the new design should solve all these issues.

ckaiser:
Something to keep in mind for a new design would be a way to have a property/flag to switch between file and folder mode, to accommodate stuff like #5399, perhaps a base widget that just has the browse button and opens the file selector, and then a more intricate selector derived from that.

PR #5836, opened 2026-05-26
[lua] Add Dialog:folder, onchange event values, fix onresize event

### LINKED ISSUE #5399: Add a `Dialog:folder()` function
I am implementing a simple "project manager" tool, and the first step of this is to allow the user to select which folder they would like to save their project in.

Aseprite has a `Dialog:file()`, and I would like pretty much the same thing, but for folders.

**Alternatives I've considered**
I've used a regular `entry` in the meantime, but it has multiple drawbacks :
- the user must type it fully or copy-paste it
- the verification that the folder exists, or its creation, must be done through Lua, which is quite bad at doing things like that (see [this stack overflow answer](https://stackoverflow.com/a/40195356) ...)

ckaiser:
Fixes #5399, adds a new Dialog widget, the folder widget with the following API:

```lua
:folder{
    id="string",
    title="string",
    placeholder="string!",
    onchange=function,
    onselected=function
}
```
image

It looks basically just like the file selector but with no option to hide the entry. I'm looking at this as the first step towards #5395, starting from a more traditional base and then iterating towards a version of this where we can keep the old functionality.

In order to implement it I had to do some modifications to the Aseprite file picker to add a folder-only mode, I tried to keep the code modifications to a minimum. It'd be good to have t[he companion LAF PR](https://github.com/aseprite/laf/pull/183) merged before we merge this but it doesn't depend on code from it, it'll just not work on linux and work a bit weird on OS X.

Along with it and to encourage API consistency, I've added a `value` parameter to all of the `onchange` and other signals for Dialog widgets.

While doing some testing for this I also noticed that the Dialog onresize event was not firing for resizes that occur when we're in single-window mode, because the code path is different, so I added a little "fake" event to trigger the signal regardless.

Here's a script for testing:
```lua
local dlg = Dialog{
    onresize=function(ev)
        print("dialog resize:", ev.size)
    end
}
dlg:entry{
    id="entry",
    placeholder="Entry!",
    onchange=function(ev)
        print("entry change:", ev.value)
    end
}
:newrow()
:number{
    id="number",
    placeholder="Number!",
    onchange=function(ev)
        print("number change:", ev.value)
    end
}
:newrow()
:slider{
    id="slider",
    onchange=function(ev)
        print("slider change:", ev.value)
    end,
    onrelease=function(ev)
        print("slider release:", ev.value)
    end
}
:newrow()
:combobox{
    id="combobox",
    options={ "one", "two", "three" },
    onchange=function(ev)
        print("combobox change:", ev.index, " - ", ev.value)
    end
}
:newrow()
:file{ 
    id="file",
    title="File",
    placeholder="File!",
    open=true,
    entry=true,
    onchange=function(ev)
        print("file change:", ev.value)
    end
}
:newrow()
:folder{ 
    id="the_folder",
    title="Folder",
    placeholder="Folder!",
    onchange=function(ev)
        print("folder change:", ev.value)
    end,
    onselected=function(ev)
        print("folder selected:", ev.value)
    end
}
dlg:show()
print("Folder at exit:", dlg.data.the_folder)
dlg:modify{ id="the_folder", path="/another/folder" }
print("Folder after dlg.modify:", dlg.data.the_folder)
dlg.data = { the_folder = "/again/another/folder" }
dlg:show()
```

### DIFF
--- a/src/app/script/dialog_class.cpp
+++ b/src/app/script/dialog_class.cpp
@@ -11,28 +11,27 @@
 
 #include "app/app.h"
 #include "app/color.h"
-#include "app/color_utils.h"
 #include "app/context.h"
 #include "app/doc.h"
 #include "app/file_selector.h"
+#include "app/i18n/strings.h"
 #include "app/script/canvas_widget.h"
 #include "app/script/engine.h"
 #include "app/script/graphics_context.h"
 #include "app/script/keys.h"
 #include "app/script/luacpp.h"
 #include "app/script/tabs_widget.h"
-#include "app/ui/button_set.h"
 #include "app/ui/color_button.h"
 #include "app/ui/color_shades.h"
 #include "app/ui/editor/editor.h"
 #include "app/ui/expr_entry.h"
 #include "app/ui/filename_field.h"
 #include "app/ui/main_window.h"
+#include "app/ui/path_field.h"
+#include "app/ui/skin/skin_theme.h"
 #include "app/ui/window_with_hand.h"
 #include "base/fs.h"
 #include "base/paths.h"
-#include "base/remove_from_container.h"
-#include "os/screen.h"
 #include "os/system.h"
 #include "ui/app_state.h"
 #include "ui/box.h"
@@ -46,6 +45,7 @@
 #include "ui/manager.h"
 #include "ui/menu.h"
 #include "ui/message.h"
+#include "ui/resize_event.h"
 #include "ui/scale.h"
 #include "ui/scroll_window.h"
 #include "ui/separator.h"
@@ -405,8 +405,9 @@ int Dialog_new(lua_State* L)
 
     type = lua_getfield(L, 1, "onresize");
     if (type == LUA_TFUNCTION) {
-      Dialog_connect_signal(L, -2, dlg->window.Resize, [](lua_State*, ResizeEvent&) {
-        // Do nothing
+      Dialog_connect_signal(L, -2, dlg->window.Resize, [](lua_State* L, ResizeEvent& event) {
+        push_obj(L, event.bounds());
+        lua_setfield(L, -2, "size");
       });
     }
     lua_pop(L, 1);
@@ -873,8 +874,9 @@ int Dialog_entry(lua_State* L)
   if (lua_istable(L, 2)) {
     int type = lua_getfield(L, 2, "onchange");
     if (type == LUA_TFUNCTION) {
-      Dialog_connect_signal(L, 1, widget->Change, [](lua_State* L) {
-        // Do nothing
+      Dialog_connect_signal(L, 1, widget->Change, [widget](lua_State* L) {
+        lua_pushstring(L, widget->text().c_str());
+        lua_setfield(L, -2, "value");
       });
     }
     lua_pop(L, 1);
@@ -908,10 +910,21 @@ int Dialog_number(lua_State* L)
     }
     lua_pop(L, 1);
 
+    type = lua_getfield(L, 2, "placeholder");
+    if (type == LUA_TSTRING) {
+      if (auto p = lua_tostring(L, -1))
+        widget->setPlaceholder(p);
+    }
+    lua_pop(L, 1);
+
     type = lua_getfield(L, 2, "onchange");
     if (type == LUA_TFUNCTION) {
-      Dialog_connect_signal(L, 1, widget->Change, [](lua_State* L) {
-        // Do nothing
+      Dialog_connect_signal(L, 1, widget->Change, [widget](lua_State* L) {
+        if (widget->decimals() == 0)
+          lua_pushinteger(L, widget->textInt());
+        else
+          lua_pushnumber(L, widget->textDouble());
+        lua_setfield(L, -2, "value");
       });
     }
     lua_pop(L, 1);
@@ -951,16 +964,18 @@ int Dialog_slider(lua_State* L)
   if (lua_istable(L, 2)) {
     int type = lua_getfield(L, 2, "onchange");
     if (type == LUA_TFUNCTION) {
-      Dialog_connect_signal(L, 1, widget->Change, [](lua_State* L) {
-        // Do nothing
+      Dialog_connect_signal(L, 1, widget->Change, [widget](lua_State* L) {
+        lua_pushnumber(L, widget->getValue());
+        lua_setfield(L, -2, "value");
       });
     }
     lua_pop(L, 1);
 
     type = lua_getfield(L, 2, "onrelease");
     if (type == LUA_TFUNCTION) {
-      Dialog_connect_signal(L, 1, widget->SliderReleased, [](lua_State* L) {
-        // Do nothing
+      Dialog_connect_signal(L, 1, widget->SliderReleased, [widget](lua_State* L) {
+        lua_pushnumber(L, widget->getValue());
+        lua_setfield(L, -2, "value");
       });
     }
     lua_pop(L, 1);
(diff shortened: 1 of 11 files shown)


src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
   return Dialog_add_widget(L, widget);
 }
 
+namespace {
+class PathSelector : public HBox {
```
martincapello: I think we can move this class into its own file in the same folder where the filename_field.h is. Also I would rename it to PathnameField (or we could rename the FilenameField to FilenameSelector).

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
...
+
+      if (base::is_directory(path))
+        m_lastPath = path;
+
+      Selected(path);
+      Change(path);
+    }
+  }
+
+private:
```
martincapello: There is an issue in how the path selector looks, it is a bit taller than the file selector, and also the "..." button looks different: 

Image

This can be fixed by creating the following class and use it instead of a Button:

```suggestion
private:
  class FoldernameButton : public ButtonSet {
  public:
    FoldernameButton(const std::string& text) : ButtonSet(1) { addItem(text); }
  };
```

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
...
+  std::string m_title;
+  std::string m_lastPath;
+
+  Entry* m_entry;
+  Button* m_button;
```
martincapello:
```suggestion
  FoldernameButton* m_button;
```

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
...
+  PathSelector(const std::string& title, const std::string& path) : m_title(title)
+  {
+    setFocusStop(true);
+    setChildSpacing(0);
+
+    m_entry = new Entry(256, "");
+    m_entry->setExpansive(true);
+    m_entry->setText(path);
+    m_lastPath = path;
+    m_button = new Button(Strings::select_file_browse());
```
martincapello:
```suggestion
    m_button = new FoldernameButton(Strings::select_file_browse());
```

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
...
+    m_button->Click.connect(&PathSelector::onClickBrowse, this);
+    m_entry->Change.connect([this] { Change(m_entry->text()); });
+
+    addChild(m_entry);
+    addChild(m_button);
+  };
```
martincapello:
```suggestion
    initTheme();
  };
```

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
...
+  }
+  std::string path() const { return m_entry->text(); }
+  Entry* getEntryWidget() { return m_entry; }
+
+  obs::signal<void(const std::string&)> Selected;
+  obs::signal<void(const std::string&)> Change;
+
+protected:
+  virtual void onClickBrowse()
+  {
```
martincapello:
```suggestion
  {
    m_button->setSelectedItem(nullptr);
```

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
...
+      m_entry->setText(path);
+
+      if (base::is_directory(path))
+        m_lastPath = path;
+
+      Selected(path);
+      Change(path);
+    }
+  }
+
```
martincapello:
```suggestion

  void onInitTheme(ui::InitThemeEvent& ev)
  {
    HBox::onInitTheme(ev);
    setChildSpacing(0);
  }

```

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1209,123 @@ int Dialog_file(lua_State* L)
...
+    Change(path);
+  }
+  std::string path() const { return m_entry->text(); }
+  Entry* getEntryWidget() { return m_entry; }
+
+  obs::signal<void(const std::string&)> Selected;
+  obs::signal<void(const std::string&)> Change;
+
+protected:
+  virtual void onClickBrowse()
```
martincapello:
```suggestion
  virtual void onClickBrowse(ButtonSet::Item*)
```

martincapello (commented):
Works like a charm! Just left some changes and comments to try to keep code and widget look consistency.
However, if the idea is to make the filename selector look like the path selector, then just omit my code suggestions (except the one about moving the class into its own file).

martincapello (approved):

src/app/ui/path_field.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "app/pref/preferences.h"
+#include "base/fs.h"
+
+namespace app {
+
+using namespace ui;
+
+PathField::PathField(const std::string& title, const std::string& path) : m_title(title)
+{
+  m_entry = new Entry(256, "");
```
dacap: Just removing this to sync with new b6f0381e0ac04395706582a06a1df30bb2352099 Entry ctor:
```suggestion
  m_entry = new Entry(256);
```

src/app/script/dialog_class.cpp
```diff
@@ -1180,6 +1207,56 @@ int Dialog_file(lua_State* L)
...
+    lua_pop(L, 1);
+
+    type = lua_getfield(L, 2, "onselected");
+    if (type == LUA_TFUNCTION) {
+      Dialog_connect_signal(L, 1, widget->Selected, [](lua_State* L, const std::string& selected) {
+        lua_pushstring(L, selected.c_str());
+        lua_setfield(L, -2, "value");
+      });
+    }
+    lua_pop(L, 1);
```
dacap: Why this signal is needed? generally we are using `onchange` to know if the value of the widget has changed (like in `Dialog:file()`). In case is needed, it should be called `onselect`, but I think we'd remove it.

src/ui/window.cpp
```diff
@@ -554,6 +554,10 @@ bool Window::onProcessMessage(Message* msg)
 
             moveWindow(rect, false);
             invalidate();
+
+            // Send a "fake" resize event since this resize didn't come from the window manager
+            ResizeEvent ev(this, rect);
+            Resize(ev);
```
dacap, line 560: I have to test a little more this to give some comment (e.g. check that this is not generated when we're in multiple windows / every OS).

dacap (changes requested):
Looks good so far 👍  I've been testing and it looks like we cannot select the current folder with the Aseprite file dialog (e.g. the first item/folder is selected and filled in the "folder name", so it looks hard to select the current folder as the folder).

src/app/script/dialog_class.cpp
ckaiser: The main difference is that "onselect" only fires when the user actually selects a folder from the picker, it gives the script a way to react to that and perform some action instead of checking the folder on every keystroke, it's mostly to give options, but I can remove it if we don't want it.

ckaiser:
Changed the behavior to be able to select the current folder better, and fixed the fact that you could still see files (in a pretty hacky way but I have no regrets I don't really think refactoring this dialog is super worth it)

dacap:
~~I'm reviewing this one, I can close this PR and fix the commit, or you can try to rebase it and fix it~~

Oh sorry, "gh pr checkout" failed so I was using an old version of the PR.

dacap:
I will merge this but fix our file selector for the folder case, it doesn't work correctly to select folders.

PR #5842, opened 2026-05-29
Avoid updating pref value from <check pref="..." /> if user didn't interact w/the widget (#5219)

### LINKED ISSUE #5219: Ability to export settings compactly
I have a backup of a settings file that I use frequently. I would like to be able to export settings to a file, but to export only those settings that I have changed, that is, that the export does not just copy Aseprite.ini, but copies it without the default settings (they are restored when I open the program anyway).

For example, this is how Aseprite.ini looks like after minimal customization and how it should look when exported:

dacap:
We should try to avoid saving modified values of preferences when the user didn't change those parameters.

This doesn't fully fix #5219 in the sense that we're not offering a way to export the Aseprite.ini file, but probably something to add in the Preferences dialog in a near future.

### DIFF
--- a/src/app/commands/cmd_options.cpp
+++ b/src/app/commands/cmd_options.cpp
@@ -761,8 +761,6 @@ public:
     firstFrame()->setTextf("%d", m_globPref.timeline.firstFrame());
 
     // Others
-    expandMenubarOnMouseover()->setSelected(m_pref.general.expandMenubarOnMouseover());
-
     enableDataRecovery()->setSelected(m_pref.general.dataRecovery());
 
     if (m_pref.general.dataRecovery() && m_pref.general.keepEditedSpriteData())
@@ -773,7 +771,6 @@ public:
     }
 
     keepClosedSpriteOnMemory()->setSelected(m_pref.general.keepClosedSpriteOnMemory());
-    showFullPath()->setSelected(m_pref.general.showFullPath());
 
     dataRecoveryPeriod()->setSelectedItemIndex(dataRecoveryPeriod()->findItemIndexByValue(
       base::convert_to<std::string>(m_pref.general.dataRecoveryPeriod())));
@@ -784,15 +781,6 @@ public:
     keepClosedSpriteOnMemoryFor()->setSelectedItemIndex(
       keepClosedSpriteOnMemoryFor()->findItemIndexByValue(
         base::convert_to<std::string>(m_pref.general.keepClosedSpriteOnMemoryFor())));
-
-    zoomFromCenterWithWheel()->setSelected(m_pref.editor.zoomFromCenterWithWheel());
-    zoomFromCenterWithKeys()->setSelected(m_pref.editor.zoomFromCenterWithKeys());
-    autoOpaque()->setSelected(m_pref.selection.autoOpaque());
-    keepSelectionAfterClear()->setSelected(m_pref.selection.keepSelectionAfterClear());
-    autoShowSelectionEdges()->setSelected(m_pref.selection.autoShowSelectionEdges());
-    moveEdges()->setSelected(m_pref.selection.moveEdges());
-    modifiersDisableHandles()->setSelected(m_pref.selection.modifiersDisableHandles());
-    moveOnAddMode()->setSelected(m_pref.selection.moveOnAddMode());
 
     // If the platform supports native cursors...
     if ((int(m_system->capabilities()) & int(os::Capabilities::CustomMouseCursor)) != 0) {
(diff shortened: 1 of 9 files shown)


dacap:
Just rebased resolving conflicts.

src/app/widget_loader.cpp
```diff
@@ -328,7 +328,10 @@ Widget* WidgetLoader::convertXmlElementToWidget(const XMLElement* elem,
     Splitter::Type type = (by && strcmp(by, "pixel") == 0 ? Splitter::ByPixel :
                                                             Splitter::ByPercentage);
 
-    Splitter* splitter = new Splitter(type, horizontal ? HORIZONTAL : vertical ? VERTICAL : 0);
+    Splitter* splitter = new Splitter(type,
+                                      horizontal ? HORIZONTAL :
+                                      vertical   ? VERTICAL :
```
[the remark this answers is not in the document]
dacap: By default it's nested in false, so we can keep these cases as they are, and sincerely it looks a lot better in this way that with parenthesis.

src/app/commands/cmd_options.cpp
```diff
@@ -1022,47 +1044,50 @@ class OptionsWindow : public app::gen::Options {
...
+    }
 
     // Aseprite format preferences
-    m_pref.asepriteFormat.celFormat(gen::CelContentFormat(celFormat()->getSelectedItemIndex()));
+    set_edited_value(m_pref.asepriteFormat.celFormat, celFormat());
 
     // Experimental features
-    m_pref.experimental.useSelectionToolLoop(useSelectionToolLoop()->isSelected());
-    m_pref.experimental.flashLayer(flashLayer()->isSelected());
-    m_pref.experimental.nonactiveLayersOpacity(nonactiveLayersOpacity()->getValue());
```
ckaiser, line 1069: Non-active layer opacity is missing the set_edited_value

src/ui/style.h
```diff
@@ -43,27 +45,23 @@ class Style {
     };
 
     // Flags (in which state the widget must be to show this layer)
-    enum { kMouse = 1, kFocus = 2, kSelected = 4, kDisabled = 8, kCapture = 16 };
+    enum StyleFlags : uint8_t {
+      kNoFlag = 0,
```
ckaiser: Feels like if it's getting a name it should also be an enum class to match `Type` and without the `k` prefix (which goes for Type as well, not a big fan of enums with a k prefix but we have a lot of them 😅)

src/ui/widget.h
```diff
@@ -459,7 +465,7 @@ class Widget : public Component {
 
   WidgetType m_type; // Widget's type
   std::string m_id;  // Widget's id
-  int m_flags;       // Special boolean properties (see flags in ui/base.h)
+  fa_t m_fa;         // Properties/status (WidgetFlags) + alignment/layout flags (WidgetAlign)
```
ckaiser: It's not a property we see a lot in the code itself but I'm not a big fan of `m_fa` as a variable name, it's not very descriptive. IDK that I have a good suggestion for it, but I'd even prefer m_flags over "fa"[.](https://www.youtube.com/watch?v=-vH2b55TSHI)

src/ui/combobox.cpp
```diff
@@ -595,8 +595,10 @@ void ComboBoxListBox::onChange()
   ListBox::onChange();
 
   int index = getSelectedIndex();
-  if (isValidItem(index))
+  if (isValidItem(index)) {
+    m_comboBox->enableFlags(EDITED);
```
ckaiser: This will make it so that the widget is always market as edited even when it just gets opened and then closed again, should probably check the index changed or move the flag to the item selection code inside combobox?

ckaiser (commented):
Pointed out some minor things but overall I'm not 100% sold on the `EDITED` flag as a part of Widget itself, feels like we're gonna be setting this flag all over the place and repeatedly on a bunch of widgets (but only some of them, since in in some it's inside methods that up until now did nothing so they don't get called on overriding, like `onChange()`) and we're adding a new thing we have to do for widget subclasses that is not obvious unless you remember it exists.

It's probably a less clean implementation but we could get the same functionality with a custom flag + signals and then that can get set up by the widget loader or inside preferences? And if that's too fiddly to apply to every widget in preferences then... that's fine IMO. It feels like an over-correction for the relatively minor issue of "we're saving all the preferences even if they don't change" which is not, in my opinion, worth the cost of extra methods Widget, new flag operations on every change for every widget everywhere, extra burden for subclasses to implement `EDITED`, etc.

Pref XML stuff is great though, code looks a lot cleaner now 🥳

src/app/commands/cmd_options.cpp
dacap: Nice catch :+1: fixed.

src/ui/style.h
dacap: At the moment I don't want to do that refactor (too many renames to include in this PR). Probably in the future, anyway in some kind of enums I like the shorter `ui::Style::kDisabled` instead of `ui::Style::StyleFlags::kDisabled`.

src/ui/widget.h
dacap: I think I'll keep `m_fa` because it's the shortest form to specify an union between flags+align. `m_flags` made things more confusing as it excludes the align concept. And as you saw, this `m_fa` is only used in the set of flags/align member functions in a very narrow region of code.

src/ui/combobox.cpp
dacap: I'll be testing this a little more, the idea would be to set the value only when the user clicks or press enter key in a specific item. So the EDITED flag should indicate "user edited value", like a way to indicate that the user explicitly selected the option.

dacap:
> Pointed out some minor things but overall I'm not 100% sold on the `EDITED` flag

I know it's not the more "clean" implementation, but probably a similar behavior would need an extra signal or a new argument on every `onChange()` event/signal to know if the edit was done by the user or programmatically.

I think an extra signal (e.g. `UserChange`) is the worse because it's too much extra memory/overhead (a new signal call) for every single widget (even if there is no observers, I'd prefer just setting the flag which is just one CPU operation).

The alternative to add a new parameter to all change events (e.g. `class ChangeEvent`, and `enum ChangeSource::ByUser`,  `ChangeSource::ByCode`) would be the "clean solution." At the moment I'm more inclined towards this `EDITED` flag.

The flag should be implemented "on demand". If some use case needs the "know if the user edited a widget" behavior, the flag is already there to be used. In other case it can be just ignored.

As a side note: I'll try to enable a UI debugging flag in this same PR to see the edited widgets in real-time and check when it's "edited". Probably something to enable in DEVMODE/DEBUG initially, but I'd like to make these "debugging flags" in release mode too (EDITED / [paint baseline](https://github.com/aseprite/aseprite/blob/ea2c6e23d1c8e324c96e54b1e6047db7e7168cd9/src/ui/widget.cpp#L1290-L1303), etc.).

dacap:
@ckaiser I'll merge the less controversial part of this PR (without the EDITED flag), and continue other refactors in other PR draft for future work, it will be a WIP.

PR #5852, opened 2026-06-04
[lua] Fix Tile_get_color() returning wrong tile indexes (fix #5838)

### LINKED ISSUE #5838: [lua] tilemapLayer.tileset:tile(N).color.index equal to 0 for all 'N'
**Describe the bug**
`Color{ tile = N }.index` different to `tilemapLayer.tileset:tile(N).color.index`

**To Reproduce**
1. New RGBA sprite
2. New tilemap layer 16x16
3. Draw until 3 tiles were generated.
4. Run the following script:
```lua
local tilemapLayer = app.sprite.layers[2]
if Color{tile = 2}.index ~= tilemapLayer.tileset:tile(2).color.index then
  print("ERROR")
end
```

**Expected behavior**
`Color{tile = 2}.index` == `tilemapLayer.tileset:tile(2).color.index`
The line `tilemapLayer.tileset:tile(2).color.index` should return the correct color index for the sampled tile.

- Aseprite: 1.3.17.2
- System: macOS

mrgreenapple24:
Fix #5838

The `Tile_get_color()` function was converting tile data to RGB color using `fromRgb()`, which lost the tile index information. This caused `tilemapLayer.tileset:tile(N).color.index` to return 0 or incorrect values instead of the actual tile index N.

Changed the implementation to use `fromTile()` to preserve the tile index, matching the behavior of `Color{ tile = N }.index`.

### DIFF
--- a/src/app/script/tile_class.cpp
+++ b/src/app/script/tile_class.cpp
@@ -83,10 +83,7 @@ int Tile_get_color(lua_State* L)
 
   auto& ud = ts->getTileData(tile->ti);
   doc::color_t docColor = ud.color();
-  app::Color appColor = app::Color::fromRgb(doc::rgba_getr(docColor),
-                                            doc::rgba_getg(docColor),
-                                            doc::rgba_getb(docColor),
-                                            doc::rgba_geta(docColor));
+  app::Color appColor = app::Color::fromTile(doc::tile(tile->ti, 0));
   if (appColor.getAlpha() == 0)
     appColor = app::Color::fromMask();
   push_obj<app::Color>(L, appColor);


Gasparoken:
Thanks for the PR. I've been reviewing my issue again and I see I made a mistake interpreting `Tile.color`. On second reading, I see that the `Tile` object contains user data, which classically consists of "color" and "data" (this user data was originally intended for cels, layers and tags but has been extended to other objects).
Currently, it's not possible to modify this user data in the Tiles from the Aseprite editor (but it's possible via scripting).
Therefore, `tilemapLayer.tileset:tile(N).color.index` always returns 0 because `tilemapLayer.tileset:tile(N).color` corresponds to the "color" of the user data which is black as default, not the tile index. To get the tile index, you should use `tilemapLayer.tileset:tile(N).index`.
So, I'm sorry, I'll have to close this PR and also the issue I created due to my misunderstanding.

PR #5853, opened 2026-06-05
Refactor MiniEditor enabled state to use Preferences API

zoldyck13:
=### Description
This pull request migrates the "MiniEditor" configuration from the legacy ini_file system to the modern app::Preferences API. 

- Updated pref.xml to include the miniEditor section.
- Refactored src/app/ui/preview_editor.cpp to use the Preferences API for the enabled state.
- Cleaned up redundant legacy config calls, improving configuration consistency and aligning the feature with current codebase standards.

---

### Contribution Agreement

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -181,6 +181,9 @@
       <option id="allow_nonlinear_history" type="bool" default="false" />
       <option id="show_tooltip" type="bool" default="true" />
     </section>
+    <section id="miniEditor" text="Mini Editor">
+      <option id="enabled" type="bool" default="true" />
+    </section>
     <section id="editor" text="Editor">
       <option id="zoom_with_wheel" type="bool" default="true" />
       <option id="zoom_with_slide" type="bool" default="false" />
--- a/src/app/ui/preview_editor.cpp
+++ b/src/app/ui/preview_editor.cpp
@@ -188,7 +188,7 @@ PreviewEditorWindow::PreviewEditorWindow()
   setAutoRemap(false);
   setWantFocus(false);
 
-  m_isEnabled = get_config_bool("MiniEditor", "Enabled", true);
+  m_isEnabled = app::Preferences::instance().miniEditor.enabled();
 
   m_centerButton->Click.connect([this] { onCenterClicked(); });
   m_playButton->Click.connect([this] { onPlayClicked(); });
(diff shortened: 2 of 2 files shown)


data/pref.xml
```diff
@@ -181,6 +181,9 @@
       <option id="allow_nonlinear_history" type="bool" default="false" />
       <option id="show_tooltip" type="bool" default="true" />
     </section>
+    <section id="miniEditor" text="Mini Editor">
```
dacap, line 184: We're using another kind of standard for these preferences, it should be:
```
    <section id="mini_editor" text="Mini Editor">
```
Anyway we should move this preference to the already existing `"preview"` section. E.g.
```
    <section id="preview" text="Preview">
      <option id="enabled" type="bool" default="true" />
      <option id="play_once" type="bool" default="false" />
      <option id="play_all" type="bool" default="false" />
      <option id="play_subtags" type="bool" default="true" />
    </section>
```

src/app/ui/preview_editor.cpp
```diff
@@ -188,7 +188,7 @@ PreviewEditorWindow::PreviewEditorWindow()
   setAutoRemap(false);
   setWantFocus(false);
 
-  m_isEnabled = get_config_bool("MiniEditor", "Enabled", true);
+  m_isEnabled = app::Preferences::instance().miniEditor.enabled();
```
dacap, line 191: Actually if we refactor this, we should impl this "Preview" visibility switching listening the `app::Preferences::instance().preview.enabled()` changes, and if one script changes the value the Preview should change its visibility too.

src/app/ui/preview_editor.cpp
```diff
@@ -202,13 +202,16 @@ PreviewEditorWindow::PreviewEditorWindow()
 
 PreviewEditorWindow::~PreviewEditorWindow()
 {
-  set_config_bool("MiniEditor", "Enabled", m_isEnabled);
+  app::Preferences::instance().miniEditor.enabled(m_isEnabled);
+  save_window_pos(this, "MiniEditor");
```
dacap, line 206: `save_window_pos()` is called from `kCloseMessage`, is this call needed? What case does it fix?

dacap (changes requested):
Thanks for your PR @zoldyck13, actually if we're going to refactor the `[MiniEditor]` to Preferences, it'd be nice to use the "preview" section and start listening to the changes of the `preview.enabled` option so we sync the "Preview" window visibility with the option value.

Issue #5575, opened 2025-12-05
Add mouse button + mouse wheel customization


dacap:
To continue #598, we should be able to press a mouse button and move the mouse wheel to customize a parameter: sprite editor scroll, timeline scroll, scroll through color bar like alt+mouse wheel, scroll through a customized set of tools (?), scroll through undo history, customize scroll through frames like Ctrl+mouse wheel, related to #1789

leonormoru:
Hi! Me and a colleague would like to work on this feature. If there isn't anyone currently working on it, we would like to try to implement it.

Issue #598, opened 2015-02-17
Make special mouse buttons customizable


dacap:
There are mouses with more buttons that should be nice to be able to customize through `Edit > Keyboard Shortcuts...`

fahseltc:
+1 for this feature request.
My issue is that I use Mouse5 as my Push-To-Talk button for discord. When using Aseprite, it recognizes my mouse5 press as a left mouse press, and draws all over my image.
So I would like to add an additional criteria to this feature: **Ability to disable mouse buttons**

ballerburg9005:
This is really and mostly only important to draw one-handed. Therefore care should be taken to not simply make mouse buttons mapable like keyboard shortcuts, but that you can also use mouse buttons as modifiers, as well as in the sense of toggle or push-release switches. Focus should lie on the top most used functions while drawing, which are in my book: switching tools, switching colors, undo/redo, instant eyedropper and pen or eraser tool.

Example: typical gaming mouse with two thumb buttons and one wheel

- pressing and holding the first thumb button can be made a modifier to scroll in conjunction with the wheel through a customizable smaller selection of tools (e.g. brush, marquee, eraser, fill) which are then made the primary tool for the first button 
- this smaller selection could furthermore also include other actions, such as flippling the image or toggling symmetry drawing
- pressing and holding the secondary thumb button can be made the eraser, but it erases only as long as the button is pressed it doesn't switch the primary tool (push-release)
- pressing and holding the secondary thumb button can additionally be made a modifier to scroll in conjunction with the wheel through undo/redo history if it couldn't fullfill the first function to erase, e.g. because it didn't hit the drawable canvas
- pressing the first then the second thumb button and then releasing both can momentarily invoke and apply the eyedropper tool (push-release)
- pressing and holding both thumb buttons at the same time while scrolling with the mouse wheel can be made to switch through the color palette, or another customizable selection of actions/tools
- in certain tools like the marquee tool, holding ctrl, shift or alt modifies the tool behavior in a pretty vital manner that is not even always accessible via the tool bar. In those tools, the two thumb buttons could function the same way as ctrl and shift. In case the custom mappings described above would conflict with this, the user should be able to change priorities.

This is the power of what can be done by just fully utilizing two additional thumb buttons. Most people own such a mouse today. Some people even own mice with multiple scroll wheels and many more buttons, but those are by far not as integral as what is available in the standard setup.

[...]

PR #5855, opened 2026-06-06
Add side buttons support to wheel actions and add new actions

XavierCunha:
**Co-authored-by: @leonormoru <maria.leonor.morujao@tecnico.ulisboa.pt>**

Previously, if a mouse side button (X1 or X2) was assigned to a mouse wheel shortcut, using the side button in combination with the scroll wheel would not trigger the mapped action as expected. This happened because the internal mouse wheel event handling logic was only comparing bindings against standard keyboard modifiers (such as Ctrl, Shift, or Alt) and completely missed checking if the side buttons matched the configured wheel shortcuts.

While looking into the main input issue, we also explored expanding mouse wheel actions to reduce keyboard dependency. This directly addresses issues #598 and #5575 by allowing mouse buttons to be used alongside the wheel. By supporting this, mapping actions like undo history, playback speed, or tool cycling to mouse combos makes the whole workflow much faster.
To fix the input bug and implement these improvements, this PR adds logic to intercept mouse wheel events and correctly include side mouse buttons in the shortcut-matching behavior. Furthermore, we implemented several new wheel-based actions and structural features:

* **Undo/Redo History Navigation:** Reused `UndoCommand` and `RedoCommand` to allow scrolling backward and forward through project changes.
* **Playback Speed Customization:** Integrated a new action using `setAnimationSpeedMultiplier` to dynamically adjust the animation speed by scrolling through Aseprite's existing playback speed presets.
* **Visual Feedback Popup:** Since changing playback speed via shortcuts originally paused animations and provided no visual feedback, we built a live popup UI that displays the exact speed value in real time as the user scrolls.
* **Custom Tool Set:** We added a new "Custom Tool Set" section under the Preferences menu, enabling users to define and cycle strictly through their most-used tools. Please note that this new action is currently a basic implementation meant to demonstrate how the feature works, leaving room for future improvements and expansion.

This PR also includes automated integration tests to ensure these new features work reliably. Since the repository had no prior tests for mouse wheel behavior, we wrote those foundational tests from scratch. On top of that, our new test suite verifies that side button and wheel combinations trigger actions properly, validates correct push-release behavior, and ensures all the new wheel actions function as expected.

We declare that our contributions are not co-authored using a generative AI technology.

We agree that our contributions are licensed under the Individual Contributor License Agreement V4.0 ("CLA") as stated in https://github.com/igarastudio/cla/blob/main/cla.md

We have signed the CLA following the steps given in https://github.com/igarastudio/cla#signing


dacap:
Thanks for this PR @XavierCunha, just ignore the error in GitHub actions about `scripts/events.lua` failing, it's unrelated to the PR.

It will take some time to review this PR, although some features were planned (customization of toolbar), I have to check where to save that information (e.g. it was planned to store that information in the `user.aseprite-layout` file).

leonormoru:
> Thanks for this PR @XavierCunha, just ignore the error in GitHub actions about `scripts/events.lua` failing, it's unrelated to the PR.
> 
> It will take some time to review this PR, although some features were planned (customization of toolbar), I have to check where to save that information (e.g. it was planned to store that information in the `user.aseprite-layout` file).

Okay, thank you! Let us know if anything else is needed.

data/widgets/options.xml
```diff
@@ -486,6 +487,65 @@
           </vbox>
         </vbox>
 
+        <!-- Custom Toolset -->
+        <vbox id="section_custom_toolset">
+          <separator text="@.section_custom_toolset" horizontal="true" />
+          <grid columns="5" id="toolset_list">
```
dacap, line 493: Although it does the job, the presentation is not great to be sincere. Also it makes the dialog wider (which is not desirable). In case I merge this PR, I'll modify this section with something else, e.g. we'd like to have the chance to customize the set of visible tools + customize groups, but probably something to think about when #5933 is merged.

data/pref.xml
```diff
@@ -476,6 +476,34 @@
     <section id="aseprite_format">
       <option id="cel_format" type="CelContentFormat" default="CelContentFormat::COMPRESSED" />
     </section>
+    <section id="custom_toolset">
```
dacap, line 479: I'd change how to save these customized toolsets (probably in the layout file or creating a new `.aseprite-toolsets` file). But again something to think in a near future to be added with a possible customization of the default toolset.

src/app/ui/editor/state_with_wheel_behavior.cpp
```diff
@@ -64,6 +77,184 @@ static inline void adjust_unit(PreciseWheel preciseWheel, double dz, double& v)
   v = std::clamp<double>(v + (preciseWheel == PreciseWheel::On ? dz / 100.0 : dz / 25.0), 0.0, 1.0);
 }
 
+static bool isInCustomToolset(tools::Tool* tool)
```
dacap, line 80: There should be a better way to handle this tool name/key <-> custom toolset association. We should avoid using preferences for this. And avoid hard-coding tool names (they are from `gui.xml`).

src/app/ui/editor/state_with_wheel_behavior.cpp
```diff
@@ -64,6 +77,184 @@ static inline void adjust_unit(PreciseWheel preciseWheel, double dz, double& v)
...
+    return custom.jumble();
+  if (id == "text")
+    return custom.text();
+
+  return false;
+}
+// Retrieve playback speed presets from the Animation menu (gui.xml).
+static std::vector<double> get_playback_speed_presets()
+{
+  std::vector<double> presets;
```
dacap, line 143: Probably `static` to cache the results?
```suggestion
  static std::vector<double> presets;
  
  if (!presets.empty())
    return presets;
```

src/app/ui/editor/state_with_wheel_behavior.cpp
```diff
@@ -64,6 +77,184 @@ static inline void adjust_unit(PreciseWheel preciseWheel, double dz, double& v)
...
+
+  if (!presets.empty()) {
+    std::sort(presets.begin(), presets.end());
+    presets.erase(std::unique(presets.begin(), presets.end()), presets.end());
+  }
+
+  return presets;
+}
+
+class PlaybackSpeedPopupWindow : public ui::PopupWindow {
```
dacap, line 176: Probably we could just use `StatusBar::showTip()` and remove this popup window.

src/app/ui/editor/state_with_wheel_behavior.cpp
```diff
@@ -64,6 +77,184 @@ static inline void adjust_unit(PreciseWheel preciseWheel, double dz, double& v)
...
+    setTransparent(true);
+    addChild(&m_label);
+    makeFixed();
+    initTheme();
+  }
+
+  void setInterval(int msecs)
+  {
+    if (!m_timer)
+      m_timer = new ui::Timer(msecs, this);
```
dacap, line 191: It looks like nobody is deleting this timer. Using tooltips and removing this class should fix this.

src/app/ui/editor/state_with_wheel_behavior.cpp
```diff
@@ -507,6 +779,20 @@ void StateWithWheelBehavior::processWheelAction(Editor* editor,
...
+        if (command)
+          UIContext::instance()->executeCommand(command, Params());
+      }
+      else if (dz > 0.0) {
+        Command* command = Commands::instance()->byId(CommandId::Redo());
+        if (command)
+          UIContext::instance()->executeCommand(command, Params());
+      }
+      break;
+    }
```
dacap, line 795: Although it works for mouse wheel, it's not reliable with Drag Value.

src/ui/manager.h
```diff
@@ -232,6 +234,9 @@ class Manager : public Widget,
   // Last pressed mouse button.
   MouseButton m_mouseButton;
 
+  // Pressed side buttons (X1/X2).
+  bool m_sideButtons[2] = { false, false };
```
dacap, line 238: Probably I'd prefer something like `m_pressedButtons[kButtons]` and add 
```c++
enum MouseButton {
  ...
  kButtons,
};
```
and keep track of all mouse buttons.

src/ui/message.h
```diff
@@ -224,6 +230,7 @@ class MouseMessage : public Message {
   PointerType m_pointerType;
   MouseButton m_button;    // Pressed button
   gfx::Point m_pos;        // Mouse position
+  bool m_sideButtons[2];   // X1/X2 pressed state
```
dacap, line 233: I'd try to replace this with some kind of `MouseModifiers`
```suggestion
  MouseButton m_button;       // Pressed button
  MouseModifiers m_mouseModifiers;
  gfx::Point m_pos;           // Mouse position
```
and create a new enum in `ui/mouse_button.h` like:
```c++
enum MouseModifiers { 
  kButtonNoneModifier = 0,
  kButtonLeftModifier = 1,
  kButtonRightModifier = 2,
  kButtonMiddleModifier = 4,
  kButtonX1Modifier = 8,
  kButtonX2Modifier = 16,
};

LAF_ENUM_FLAGS(MouseModifiers)
```

dacap (changes requested):
First of all thanks both of you for the work in this PR. It looks like an interesting implementation. I think the most controversial part of it is the custom toolset, I'd prefer if you remove it from the PR to make it easier/possible to merge the PR as it is (with some minor changes).

Please check the following comments:

leonormoru:
Hey @dacap, thank you for reviewing our PR and for your feedback! @XavierCunha and I are a bit busy with other projects at the moment, but we will try to update our PR soon with your suggested changes so that they are ready to be merged!

dacap:
Thank you @leonormoru for your contributions!

Just in case, some work from https://github.com/aseprite/aseprite/pull/6027 will help to specify/customize the active set of groups/tools which will finally helpful for this PR.

PR #5858, opened 2026-06-08
Feat aseprite#5747: Implement Yin Yang symmetry

dartlord345:
The current current horizontal/vertical/diagonal symmetry mirrors the drawn pixels.
The yin yang symmetry option works like a yin yang drawing. When drawing in the top left side of the canvas, the same pixels appear in the bottom right side, and vice versa. When drawing in the bottom left, the pixels appear in the top right. 
We added tests to complement the feature.


src/doc/brush.h
```diff
@@ -31,7 +31,8 @@ enum class SymmetryIndex {
   ROT_FLIP_90 = 5,
   ROTATED_90 = 6,
   ROT_FLIP_270 = 7,
-  ELEMENTS = 8
+  ELEMENTS = 8,
+  ROTATED_180 = 9
```
dacap: The idea of this `ELEMENTS` item is to be the last one in the collection, so `ROTATED_180` can be `= 8` and `ELEMENTS = 9`.

src/app/tools/point_shapes.h
```diff
@@ -105,7 +105,8 @@ class BrushPointShape : public PointShape {
   bool m_firstPoint;
   Brush* m_lastBrush;
   BrushType m_origBrushType;
-  std::array<std::shared_ptr<CompressedImage>, int(SymmetryIndex::ELEMENTS)> m_compressedImages;
+  std::array<std::shared_ptr<CompressedImage>, int(SymmetryIndex::ROTATED_180) + 1>
+    m_compressedImages;
```
dacap: So with `ELEMENTS` being always the latest item in `SymmetryIndex` enum, we don't need this change each time a new kind of symmetry is introduced.

dacap (changes requested):
Thanks for your PR @dartlord345! It looks great so far, I wasn't able to test it yet, but I'm leaving some minor comments to fix.

dacap:
Just testing and the symmetry doesn't work correctly with a custom brush (part 0:12) and all other symmetries don't work when used combined (part 0:27):

https://github.com/user-attachments/assets/5dc04b93-3147-4bab-bd3e-610c4d786879

PR #5860, opened 2026-06-10
Implement undoable toggle all layers in the timeline

### LINKED ISSUE #5658: Think about a solution to undo/restore showing/hiding all layers at once
An accidental click could show/hide all layers, and there is no easy way to restore the previous configuration. When we hide/show all layers at once clicking the eye icon in the timeline header we could:

1. make the action undoable, or
2. add a confirmation dialog

Probably I would prefer the first solution if we don't find something better.

Requested in ticket: https://igarastudio.zendesk.com/agent/tickets/8840

Darth712:
Clicking the eye icon in Aseprite's timeline header to show or hide all layers at once can accidentally destroy a carefully configured layer visibility setup. This implementation allows users to restore their
exact configuration via Ctrl+Z.

Example:
Before click:
- Layer A: visible
- Layer B: hidden
- Layer C: visible

After accidental click:
- Layer A: hidden
- Layer B: hidden
- Layer C: hidden No way to go back to previous configuration.

Closes #5658


Darth712:
Hi @dacap, it's been a week since you self-assigned for this pull request. Is there something you need from my side? Thank you!

dacap:
@Darth712 I'm just reviewing other PRs, but I'll take these one now :+1:

dacap:
@Darth712 We'll need the CLA of the co-author Francisco.

src/app/commands/cmd_toggle_all_layers.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+ToggleAllLayersCommand::ToggleAllLayersCommand() : Command(CommandId::ToggleAllLayers())
+{
+}
+
+bool ToggleAllLayersCommand::onEnabled(Context* context)
+{
+  return context->checkFlags(ContextFlags::ActiveDocumentIsWritable);
+}
+void ToggleAllLayersCommand::onExecute(Context* context)
```
dacap, line 36:
```suggestion
}

void ToggleAllLayersCommand::onExecute(Context* context)
```

src/app/commands/cmd_toggle_all_layers.cpp
```diff
@@ -0,0 +1,57 @@
...
+{
+  ContextWriter writer(context);
+  app::Doc* doc = writer.document();
+  doc::Sprite* sprite = writer.sprite();
+
+  if (!doc || !sprite)
+    return;
+
+  bool isCurrentlyVisible = true;
+  bool newState = !isCurrentlyVisible;
```
dacap, line 46: Sorry to ask again, but did you use AI to generate this PR? These two statements makes no sense for a "toggle" command, it's like writing:
```suggestion
  bool newState = false;
```
Why `isCurrentlyVisible` doesn't contain a similar logic as in the timeline? (even more, the timeline could just be calling this command if implemented correctly).

src/app/commands/cmd_toggle_all_layers.cpp
```diff
@@ -0,0 +1,57 @@
...
+  Transaction transaction(writer.context(), doc, "Toggle All Layers");
+  transaction.execute(new app::cmd::ToggleAllLayers(sprite, newState));
+  transaction.commit();
```
dacap, line 49: We use `Tx` instead of `Transaction` (and replace the hard coded text with the command text):
```
  Tx tx(writer, onGetFriendlyName());
  tx(new app::cmd::ToggleAllLayers(sprite, newState));
  tx.commit();
```

dacap (changes requested):
To be sincere, doing this review I felt like I was reviewing the code of an AI. I have just make some comments, but creating a command that has a "toggle" logic but it just hide all layers doesn't make any sense for me. Sorry if this is not the case, but is what I felt doing the review.

PR #5861, opened 2026-06-10
Implement "power of two sheet size" UI (#2289, #5529)

dacap:
I was just trying to fix this compiler warning:
```
src/app/doc_exporter.cpp:713:5: warning: field 'm_sampleBuf' will be initialized after field 'm_powerOfTwoSize' [-Wreorder-ctor]
  712 |   : m_docBuf(std::make_shared<doc::ImageBuffer>())
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     m_powerOfTwoSize(false)
  713 |   , m_sampleBuf(std::make_shared<doc::ImageBuffer>())
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     m_docBuf(std::make_shared<doc::ImageBuffer>())
  714 |   , m_powerOfTwoSize(false)
      |     ~~~~~~~~~~~~~~~~~~~~~~~
      |     m_sampleBuf(std::make_shared<doc::ImageBuffer>())
```
But after realizing that it'd make sense to move `m_powerOfTwoSize = false` to the `DocExporter::reset()` function, I ended up implementing the complete UI for this option:

image

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!-- Aseprite -->
-<!-- Copyright (C) 2018-2025  Igara Studio S.A. -->
+<!-- Copyright (C) 2018-present  Igara Studio S.A. -->
 <!-- Copyright (C) 2014-2018  David Capello -->
 <preferences>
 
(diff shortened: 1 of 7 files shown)


Gasparoken:
Great! LGTM👌🏼

I almost recommend that this option should be available in the Constraints combo box. However, I see that it's compatible if "Fixed # of Columns" or "Fixed # of Rows" is enabled. On the other hand, it might seem a bit odd to configure "Fixed Width" or "Fixed Height" and also enable "Size=x2" (I would expect "Fixed Width" or "Fixed Height" to actually be the final width).
In short, I think this option, as it is, offers more flexibility.

PR #5873, opened 2026-06-13
Ensure the console is always parented to the main window (fix #5872)

### LINKED ISSUE #5872: Printing from within a `Dialogs` `onclick` function causes the console window to be closed permanently when the Dialog is closed
**Describe the bug**
If you call `print()` from within a Dialog function (`onclick`) the console that shows the prints will close when the Dialog it spawned from is closed. Any future prints, from within a Dialog or otherwise, will never show, because the console will never open again, until Aseprite is restarted completely.

**To Reproduce**
1. Run the following script:
```
local dialog = Dialog()
dialog:button{id="ok", text = "OK", onclick = function() print("The console opens!") end}
dialog:button{id="close", text = "Close", onclick = function() dialog:close() print("The console has closed, and won't appear again.") end}
dialog:show()
```
2. Press the OK button, which simply prints a string. This causes the Console to be parented to the current Dialog
3. Press the Close button, closing the Dialog
4. The console closes as well
5. The subsequent prints have not shown in the console
6. Run any script that prints anything to the log
7. These prints will not show up.

**Expected behavior**
The console is always its own window, and does not close when other windows get closed. At the very least it shouldn't be permanently closed when its parent gets closed.

**Screenshots**
Image

- Aseprite: Aseprite v1.3.18-beta3
- System: Windows 10
- Extensions: A lot of custom ones

**Additional context**
This happens with specific `onclick = function() dialog:close end`commands, but also when `onclick` is left empty (the default way to close). This presumably also happens with other Dialog functions. 

This does NOT happen, if the console was first spawned with a normal print, and the Dialog print was added to the same console afterwards.:

```
print("This avoids the console disappearing!")
local dialog = Dialog()
dialog:button{id="ok", text = "OK", onclick = function() print("The console opens!") end}
```
Presumably it would disappear if you were to close this console before pressing 'OK' to spawn the print, which would cause the print to spawn its own console window.

ckaiser:
Fixes #5872, the window was getting automatically parented to the current foreground window when shown, which could be an ephemeral script dialog.

### DIFF
--- a/src/app/console.cpp
+++ b/src/app/console.cpp
@@ -53,6 +53,8 @@ public:
     // the initialization.
     if (auto* mainWin = App::instance()->mainWindow()) {
       m_mainWindowClosedConn = mainWin->Close.connect([this] { closeWindow(nullptr); });
+      // Ensure we are parented to the main window when we are opened
+      setParentDisplay(mainWin->display());
     }
 
     // When the window is closed, we clear the text


dacap:
LGTM :+1:

PR #5874, opened 2026-06-13
Fix crash when undoing transactions involving MovingPixels

ckaiser:
Undoing or other transaction actions involving MovingPixels could end up with a crash or other strange behavior, this adds a new doc observer that monitors transactions and drops the pixels before any commits.

Repro script:
```lua
local sprite = Sprite(128, 128)
app.transaction("Crash me on undo", function()
    sprite.selection:selectAll()
    app.command.MoveMask{quantity = 10, target = "content", direction = "down"}
    sprite.selection = Selection()
end)
-- Time to crash
app.command.Undo()
```


dacap:
Looks good :clap:

PR #5875, opened 2026-06-13
Make Command tooltips consistent when inside transactions (fix #5862)

### LINKED ISSUE #5862: Custom name of `transaction` gets overwritten by `app.command.RemoveLayer` name in popup
**Describe the bug**
When calling a transaction that includes a `app.command.RemoveLayer` call, the first time the blue command name popup appears in the bottom right of the screen, it overwrites the transaction name with a generic `RemoveLayer` name, such as 'Layer "Layer 1" removed`

**To Reproduce**
1. Create a script with the following lines:
```
app.transaction("this has a custom name!", function() app.command.RemoveLayer()
end)
```
2. Open a sprite with two or more layers
3. Run the script
4. The blue popup will show 'Layer "Layer Name" removed" rather than the custom set transaction name.

**Expected behavior**
The transaction name gets priority and is shown, since it encapsules the called command.

**Screenshots**
Image

- Aseprite:Aseprite v1.3.18-beta3
- System: Windows 10
- Extensions: A lot of custom ones!

**Additional context**
Undoing and Redoing the transaction do show the correct names, and the Undo History is also correct. 

I mostly noticed this bug due to needing to spawn a temporary layer and then remove it again within a transaction, for a different script/plugin I'm working on. The little popup instead only mentioning a layer got deleted gives me a little heart attack each time, thinking something went wrong and data got deleted.

ckaiser:
Fixes #5862 and other instances of this happening, I've also taken the liberty of fixing an instance of potential `StatusBar` use when no UI is available, localized strings for transactions and improved the Remove Layer tooltip when coming from a timeline right click > remove.

### DIFF
--- a/src/app/commands/cmd_duplicate_slice.cpp
+++ b/src/app/commands/cmd_duplicate_slice.cpp
@@ -82,10 +82,15 @@ void DuplicateSliceCommand::onExecute(Context* context)
   }
 
   ContextWriter writer(context);
-  Tx tx(writer, "Duplicate Slice");
   Sprite* sprite = writer.site().sprite();
-
   Doc* doc = static_cast<Doc*>(sprite->document());
+
+  // Show the tooltip feedback only if we are not inside a transaction
+  // (e.g. we can be already in a transaction if we are running in a
+  // Lua script app.transaction()).
+  const bool showTooltip = (doc->transaction() == nullptr);
+
+  Tx tx(writer, Strings::commands_DuplicateSlice());
   doc->notifyBeforeSlicesDuplication();
   for (auto* s : selectedSlices) {
     Slice* slice = new Slice(*s);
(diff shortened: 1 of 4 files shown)


dacap:
With this solution the tooltip is not displayed at all, which is something I prefer for complex commands/script transactions.

@Arecher from your report in #5862 you'd expect the transaction name in the tooltip, but probably it'd be better to avoid the tooltip at all (just like other complex commands in Aseprite that don't display a tooltip, e.g. Flatten Layers).

ckaiser:
Yeah, no tooltip is best for these during transactions and if the script needs one for execution it can show one with `app.tip()`

Arecher:
I never noticed certain commands do not show the tooltip. I don't believe users rely on them, so as long as they are not displaying the wrong information that should be fine. I think the command tooltip for 'Undo [command X]' is more vital, but that one already worked!

PR #5876, opened 2026-06-13
Allow MoveMask quantity to be 0 (fix #5843)

### LINKED ISSUE #5843: Calling `app.command.MoveMask{quantity = 0}` still moves the `Selection`
**Describe the bug**
When you call `app.command.MoveMask{quantity = 0}` the selection is instead moved by 1 pixel. This seems to be the minimum.

**To Reproduce**
1. Create a script with the line 
```app.command.MoveMask{quantity = 0, target = "content", direction = "right"}```
2. Select part of a sprite
3. Run the script
4. The selection is moved

**Expected behavior**
I would expect it to not move at all. Makes sense that it has a minimum, since you do call `MoveMask` without wanting it to move, but in my case I was doing automatic movement value calculations, and in very rare instances the amount it moved was incorrect. Turns out it was this little bug.

**Screenshots**
Image

- Aseprite: Aseprite v1.3.18-beta3
- System: Windows 10
- Extensions: A lot of custom ones

**Additional context**
After I tried to Undo and Redo the change during the gif to showcase the change, I noticed only the `Selection` `boundaries` were removed, rather than the content. This might be a related bug? Upon other testing by changing the script to move 100px, undoing and redoing, nothing happened at all. The behaviour seems inconsistent.

This bug can obviously be worked around with a simple check, but if that is the preferred solution, the documentation should probably have a little note on this behaviour!

ckaiser:
Fixes #5843 allowing MoveMask `quantity` to be zero, where it does nothing. Recommended to test this with #5874 otherwise attempting to undo things results in weird locks.

### DIFF
--- a/src/app/commands/cmd_move_mask.cpp
+++ b/src/app/commands/cmd_move_mask.cpp
@@ -70,14 +70,20 @@ bool MoveMaskCommand::onEnabled(Context* context)
 
 void MoveMaskCommand::onExecute(Context* context)
 {
+  if (m_moveThing.quantity <= 0)
+    return;
+
   gfx::Point delta = m_moveThing.getDelta(context);
+  const auto& moveString = m_moveThing.getFriendlyString();
 
   switch (m_target) {
     case Boundaries: {
       ContextWriter writer(context);
       Doc* document(writer.document());
       {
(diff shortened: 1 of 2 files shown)


dacap:
LGTM :+1:

PR #5887, opened 2026-06-24
Ensure BoolPrefWidget selection counts as modification (fix #5885)

### LINKED ISSUE #5885: Regression: Changing the file dialog type from preferences is not working [beta branch]

ckaiser:
Fixes #5885, I noticed this on my own while working on #5836 and included the fix in there but then I noticed the issue 😅, so I extracted it here. It's mainly the fault of having two different checkboxes, one in experimental and the other one in the general options, but it's probably better for consistency of setting the selection triggers the change on bool prefs.

### DIFF
--- a/src/app/ui/pref_widget.h
+++ b/src/app/ui/pref_widget.h
@@ -47,6 +47,7 @@ public:
 
       // Load option value
       this->setSelected((*m_option)());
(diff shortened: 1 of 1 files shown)


dacap:
Thanks for the fix :+1: I was going to take this.

PR #5901, opened 2026-07-02
Fix RunCommand matching regression

ckaiser:
Fix a regression introduced with commit a2459af, I had assumed the searchable text was already pre-lowercased so I changed the call to the faster function and... well, it wasn't lowercase 💀

### DIFF
--- a/src/app/commands/cmd_run_command.cpp
+++ b/src/app/commands/cmd_run_command.cpp
@@ -522,7 +522,7 @@ void RunCommandCommand::loadDatabase()
         base::split_string(base::string_to_lower(key->triggerString()), labelWords, " ");
 
         const RunnerDB::Item item{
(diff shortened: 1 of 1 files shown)


dacap:
Yeah, changing `match()` call to `match.word()` removed the `base::string_to_lower()` part. From `MatchWords` definition:
```
class MatchWords {
public:
...
  bool operator()(const std::string& item) const
  {
    const std::string& lowerItem = base::string_to_lower(item);
    return word(lowerItem);
  }

  bool word(const std::string& lowerItem) const
  {
    std::size_t matches = 0;

    for (const auto& part : m_parts) {
      if (lowerItem.find(part) != std::string::npos)
        ++matches;
    }

    return (matches == m_parts.size());
  }
```

Issue #3112, opened 2021-12-28
Test that the crashpad handler is working (uploading crash reports) on all platforms


dacap:
Test that the Sentry crash handler is working on all platform before the release of the next version.

dacap:
At least we've already received reports from Windows (#3132), the macOS version should be working too (tested in dev versions, but we didn't receive a crash yet from an user). And the Linux version is disabled as we've to find a cross-distro solution to support SSL with libcurl (one possibility would be to [load OpenSSL dynamically](https://github.com/curl/curl/discussions/8224), but it might have [some ABI problems](https://abi-laboratory.pro/?view=timeline&l=openssl)).

dacap:
At least we've already received reports from Windows (#3132), the macOS version should be working too (tested in dev versions, but we didn't receive a crash yet from an user). And the Linux version is disabled as we've to find a cross-distro solution to support SSL with libcurl (one possibility would be to [load OpenSSL dynamically](https://github.com/curl/curl/discussions/8224), but it might have [some ABI problems](https://abi-laboratory.pro/?view=timeline&l=openssl)).

Gasparoken:
Can this issue be closed?

dacap:
I'll change the title of the issue as the Linux version doesn't use Sentry yet.

PR #5911, opened 2026-07-08
Use https in URLs on all platforms

### LINKED ISSUE #5340: The exported sprite sheet JSON content is not the exact same across OSes
When passing the [--data](https://www.aseprite.org/docs/cli/#data) argument to the Aseprite CLI, the generated .json file has a small difference between Windows and Linux.

On Windows, the app URL uses the https protocol:
```
 "meta": {
  "app": "https://www.aseprite.org/",
```

On Linux, it uses the http protocol:
```
 "meta": {
  "app": "http://www.aseprite.org/",
```

This is not a big deal, but creates unnecessary differences when working on a project across different OSes.

SVendittelli:
When passing the --data argument to the CLI, the generated .json file has a small difference between Windows/macOS and Linux. The .meta.app URL uses https on macOS and Windows, but http on Linux. This means developers working on the same projects across multiple OSes end up with a permanent diff on those files when they are generated.

Fixes #5340

### DIFF
--- a/src/ver/info.c
+++ b/src/ver/info.c
@@ -10,20 +10,14 @@
 #define PACKAGE   "Aseprite"
 #define COPYRIGHT "Copyright (C) 2001-2025 Igara Studio S.A."
 
-#if defined(_WIN32) || defined(__APPLE__)
-  #define HTTP "https"
-#else
-  #define HTTP "http"
-#endif
-
 #ifdef CUSTOM_WEBSITE_URL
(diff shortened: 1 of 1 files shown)


dacap:
Hi @SVendittelli, thanks for your PR but actually the fix is a little more complex, related to #3112, we need to enable TLS for libcurl and that means adding a new dependency on Linux to handle TLS.

PR #5866, opened 2026-06-12
[theme] Improve more dark mode legibility issues, make checkbox easier to see on high-DPI screens

3vorp:
Pretty much just an extension of #5764, featuring very similar contrast improvements. I tried to make the layer panel selection styles look more similar to comparable programs (which tend to use a light highlight rather than a dark one) to improve legibility, and thickened up the checkbox a bit to look better on high DPI screens (on my 4k monitor it was nearly invisible in dark mode even with a border and shadow).

I'm not 100% sold on the design for the checkbox; feel free to suggest an improved version. I'd definitely caution away from keeping it at the original 1px width though for the reasons mentioned above.

## Full changelog

- Made dark mode disabled text styles more legible
- Fixed disabled arrow styles in dark mode being a completely different color from disabled text.
- Fixed arrows being very poorly aligned in both light and dark mode, and sometimes even shifting around between standby/hover/disabled states.
- Made the alert sprite more visible in dark mode, and made the colors consistent between dark and light mode
- Made dark mode command palette hotkeys legible and made the selected styles consistent with the rest of the app
- Improved dark mode timeline styles to make the selected layer more visible and improved the contrast between the icon glyphs and the background (this has also been a feature suggestion on the forums in https://community.aseprite.org/t/better-dark-theme/22229/2)
- Increased the contrast of status bar text in dark mode
- Fixed the coloration of the home, edit, close, and flag tab icons to match the text colors correctly in dark mode
- Fixed the incorrect lock icon being used in the palette editor in dark mode
- Fixed the main modal dialog and menu body colors being slightly different to the version used in the spritesheet.
- Made the checkbox more visible in both dark and light mode
- Rethemed the radio buttons to match the checkbox more closely in both themes, and made it visible in dark mode (it was previously completely unreadable)
- Fixed menu dividers being a completely illegible color in dark mode
- Made the entry suffix color distinct from the main text color in dark mode
- Fixed the spacing of the canvas size arrow sprites in both themes (they were previously very off-centered)
- Fixed symmetry icons being a completely different color to all other icon button sprites in dark mode
- Made closed and open folder icons a bit more visually distinct
- Made the tab highlight effect more legible

## Comparisons (before/after):

Disabled styles, modal dialog color fix, and alert sprite:
Screenshot 2026-07-20 at 7 29 27 PM
Screenshot 2026-07-20 at 7 29 42 PM

Command palette select/hotkey color fix:
Screenshot 2026-07-20 at 7 30 14 PM
[...]

PR #5933, opened 2026-07-21
Tree & DirectoryTree widgets

ckaiser:
Adds a new `Tree` widget with a `DirectoryTree` subclass.

https://github.com/user-attachments/assets/ebd528d3-0add-44fd-8082-f403c6fe8d71


data/extensions/aseprite-theme/theme.xml
```diff
@@ -29,6 +29,7 @@
         <dim id="context_bar_height" value="18"/>
         <dim id="brush_type_width" value="15"/>
         <dim id="color_selector_bar_size" value="8"/>
+        <dim id="tree_item_spacing" value="2"/>
```
dacap, line 32: This should be in the dark version too.

data/extensions/aseprite-theme/theme.xml
```diff
@@ -391,6 +392,8 @@
         <part id="icon_delta" x="266" y="264" w="6" h="8"/>
         <part id="icon_corner_radius" x="272" y="264" w="8" h="8"/>
         <part id="icon_add" x="184" y="200" w="5" h="5"/>
+        <part id="icon_tree_folder" x="192" y="192" w="5" h="5"/>
+        <part id="icon_tree_file" x="200" y="192" w="5" h="5"/>
```
dacap: try to keep the same order of XML elements between the light and dark versions of the theme.

data/extensions/aseprite-theme/theme.xml
```diff
@@ -1168,5 +1173,9 @@
         <style id="runner_command_checked" width="8" height="8">
             <icon part="check_selected" align="center middle"/>
         </style>
+        <style id="tree" border="1" border-left="4">
```
dacap, line 1183: Should be in the dark version too.

src/app/ui/tree.h
```diff
@@ -0,0 +1,117 @@
...
+  void onInitTheme(InitThemeEvent& ev) override;
+  void onSizeHint(SizeHintEvent& ev) override;
+  bool onMouse(const MouseMessage* mouseMsg);
+  virtual void toggleCollapse(TreeNode* node, bool recursive = false);
+
+private:
+  void selectLast();
+  int nodeHeight(const TreeNode* node) const;
+
+  TreeNode* m_root;
```
dacap: It looks like this could be an unique ptr:
```suggestion
  std::unique_ptr<TreeNode> m_root;
```

src/app/ui/tree.cpp
```diff
@@ -0,0 +1,505 @@
...
+            [[fallthrough]];
+          case kKeyDown:
+            if (m_selected)
+              setSelected(m_selected->getNextInTree(), true);
+            if (!m_selected)
+              setSelected(m_root, true);
+            return true;
+          case kKeyLeft:
+            if (m_selected && m_selected->hasChildren() && !m_selected->isCollapsed())
+              toggleCollapse(m_selected, msg->shiftPressed());
```
dacap: If we collapse the node, we shouldn't select the "prev in tree" (right now pressing "left" key will collapse and move the selection to the previous).

Even more, I think the left arrow should select the parent node if this one is collapsed.

src/app/ui/tree.h
```diff
@@ -0,0 +1,117 @@
...
+  void setNext(TreeNode* next) { m_next = next; }
+  void setPrev(TreeNode* prev) { m_prev = prev; }
+  void setParent(TreeNode* parent) { m_parent = parent; }
+
+  // Directly attaches this node, does not touch the parent
+  void attachTo(TreeNode* parent, TreeNode* prev, TreeNode* next);
+  void addChild(TreeNode* child);
+
+  TreeNode* getNextInTree() const;
+  TreeNode* getPrevInTree() const;
```
dacap: I think these could be renamed to:
```suggestion
  TreeNode* nextInTree() const;
  TreeNode* prevInTree() const;
```

src/app/ui/tree.h
```diff
@@ -0,0 +1,117 @@
...
+  virtual void toggleCollapse() { m_collapsed = !m_collapsed; }
+  void setCollapsed(const bool isCollapsed) { m_collapsed = isCollapsed; }
+  virtual bool hasChildren() const { return m_firstChild != nullptr; }
+
+  // Calculates how deep in its tree this node is
+  int depth() const;
+
+  void setNext(TreeNode* next) { m_next = next; }
+  void setPrev(TreeNode* prev) { m_prev = prev; }
+  void setParent(TreeNode* parent) { m_parent = parent; }
```
dacap: Can anybody call `setNext/setPrev()` directly without fixing prev->next and next->prev pointers? These look like private member functions (or just set the property value directly inside `TreeNode` functions like in `TreeNode::addChild()`).

dacap (commented):
This looks great so far! Some minor comments in the PR.

ckaiser:
Updated with the corrections and a way to store data inside tree nodes so we can avoid subclasses (I'm using this in the permissions branch for the UI... which now that I'm realizing I think I left some icons in here from that 😬)

data/extensions/aseprite-theme/dark/theme.xml
```diff
@@ -1057,10 +1066,10 @@
...
-            <background color="textbox_face" />
-            <background color="selected" align="left" state="selected" />
-            <text color="textbox_text" align="left" />
-            <text color="selected_text" align="left" state="selected" />
+            <background color="textbox_face"/>
+            <background color="selected" align="left" state="selected"/>
+            <text color="textbox_text" align="left"/>
+            <text color="selected_text" align="left" state="selected"/>
         </style>
         <style id="mini_slider" extends="slider" font="mini">
```
dacap, line 1077: Missing "textedit_static" element in dark version (as we talked, in a future we can see how to remove all styles from the dark version and keep only the light version, but after merging #5866)
```suggestion
        </style>
        <style id="textedit_static" extends="textedit">
            <background color="textbox_face" state="disabled"/>
        </style>
        <style id="mini_slider" extends="slider" font="mini">
```

src/app/ui/tree.h
```diff
@@ -0,0 +1,157 @@
...
+  const std::string& tooltip() const { return m_tooltip; }
+  void setTooltip(const std::string& tooltip) { m_tooltip = tooltip; }
+
+  SkinPartPtr icon() const { return m_icon; }
+  void setIcon(const SkinPartPtr& icon) { m_icon = icon; }
+
+  bool isCollapsed() const { return m_collapsed; }
+  virtual void toggleCollapse() { m_collapsed = !m_collapsed; }
+  void setCollapsed(const bool isCollapsed) { m_collapsed = isCollapsed; }
+  virtual bool hasChildren() const { return m_firstChild != nullptr; }
```
dacap: I'd prefer to avoid public virtual functions, generally it'd be better to have non-virtual functions for the public API, and protected virtual functions to extend the behavior in sub-classes.

For example, in this case `toggleCollapse()` is customizable but `setCollapsed()` no, so using `setCollapsed()` doesn't call the customization but with `toggleCollapse()` does. 

I'd prefer something like:
```suggestion
  bool isCollapsed() const { return m_collapsed; }
  void toggleCollapse() 
  { 
    m_collapsed = !m_collapsed;
    onCollapseChange();
  }
  void setCollapsed(const bool isCollapsed)
  {
    m_collapsed = isCollapsed;
    onCollapseChange();
  }
  bool hasChildren() const { return onHasChildren(); }
```
Or something similar.

src/app/ui/directory_tree.cpp
```diff
@@ -0,0 +1,172 @@
...
+using namespace ui;
+
+DirectoryTreeNode::DirectoryTreeNode(const std::string& text,
+                                     const std::function<void(TreeNode*)>& loadCallback,
+                                     const SkinPartPtr& icon)
+  : TreeNode(text, icon)
+  , m_loadCallback(loadCallback)
+  , m_loaded(false)
+{
+  setCollapsed(true);
```
dacap, line 27: Probably if we save the "expanded" flag instead of the "collapsed" one we could simplify some of the `DirectoryTreeNode` code (but I'm not sure). Ignore me if this is not the case.

src/app/ui/directory_tree.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+void DirectoryTree::setExtensionFilter(const std::vector<std::string>& extensions)
+{
+  if (m_extensions != extensions) {
+    m_extensions = extensions;
+    switchDirectory(m_path);
+  }
+}
+
+void DirectoryTree::toggleCollapse(TreeNode* node, bool)
```
dacap, line 76: Probably
```suggestion
void DirectoryTree::onToggleCollapse(TreeNode* node, bool)
```
To avoid calling `Tree:toggleCollapse()`, same case as in nodes.

src/app/ui/directory_tree.cpp
```diff
@@ -0,0 +1,172 @@
...
+  switchDirectory(path);
+
+  m_doubleClickConn = DoubleClickItem.connect(&DirectoryTree::showItem, this);
+  m_rightClickConn = RightClickItem.connect(&DirectoryTree::showPopup, this);
+}
+
+void DirectoryTree::switchDirectory(const std::string& path)
+{
+  auto* root = new TreeNode(base::get_file_path(path));
+  root->setProperty("path", path);
```
dacap: Not sure if I'm a big fan of this "properties" for nodes. I think that a FileTreeNode could have a path() member and get/build the full path from the parent (DirectoryTreeNode) + its text. Saving a lot of memory.

src/app/ui/directory_tree.h
```diff
@@ -0,0 +1,54 @@
...
+  std::function<void(TreeNode*)> m_loadCallback;
+  bool m_loaded;
+};
+
+class DirectoryTree : public Tree {
+public:
+  explicit DirectoryTree(const std::string& path, const std::vector<std::string>& extensions = {});
+
+  void switchDirectory(const std::string& path);
+  void setExtensionFilter(const std::vector<std::string>& extensions);
```
dacap, line 49: Although it's not yet used in this PR, I'm not sure if this extensions filter should be a std::map or similar in the future (for faster matching with each file item to filter). But I think you have something else in mind (I remember other path matching technique from a previous meet we had).

src/app/ui/directory_tree.cpp
```diff
@@ -0,0 +1,172 @@
...
+      if (base::list_files(full).empty()) {
+        newNode = new TreeNode(fn, fileIcon);
+      }
+      else {
+        auto load = [this, full](TreeNode* sub) { addNodesForPath(sub, full); };
+        newNode = new DirectoryTreeNode(fn, load, folderIcon);
+      }
+    }
+    if (newNode) {
+      newNode->setProperty("path", full);
```
dacap: Same here as before, with some kind of `FileTreeNode` we might avoid having a full path copy in the node itself.

src/app/ui/tree.cpp
```diff
@@ -0,0 +1,570 @@
...
+      if (m_selected) {
+        if (m_selected->hasChildren() && m_selected->isCollapsed())
+          toggleCollapse(m_selected, keyMsg->shiftPressed());
+        DoubleClickItem();
+      }
+      return true;
+    }
+    case kKeyRight:
+      if (m_selected && m_selected->hasChildren() && m_selected->isCollapsed())
+        toggleCollapse(m_selected, keyMsg->shiftPressed());
```
dacap: I think the first <kbd>Right arrow</kbd> key shouldn't go to the first child and only expand the node:
```suggestion
      if (m_selected && m_selected->hasChildren() && m_selected->isCollapsed()) {
        toggleCollapse(m_selected, keyMsg->shiftPressed());
        return true;
      }
```

src/app/ui/tree.cpp
```diff
@@ -0,0 +1,570 @@
...
+        else {
+          selectLast();
+        }
+        return true;
+      }
+      break;
+    }
+    case kKeyHome: setSelected(root(), true); return true;
+    case kKeyEnd:  selectLast(); return true;
+    default:       break;
```
dacap: It'd be nice to have `kKeyPageUp` and `kKeyPageDown` behavior here.

src/app/ui/tree.cpp
```diff
@@ -0,0 +1,570 @@
...
+  const Paint& backgroundPaint = isEnabled() ? textColors.background :
+                                               textColors.disabledBackground;
+  const Paint& selectedPaint = textColors.selectedText;
+  const Paint& selectedBackgroundPaint = textColors.selectedBackground;
+
+  const int itemSpacing = m_themeCache.itemSpacing;
+  const int depthSpacing = m_themeCache.depthSpacing;
+
+  const gfx::Rect rect = view->viewportBounds().offset(-bounds().origin());
+  g->drawRect(rect, backgroundPaint);
```
dacap: Simplify
```suggestion
  g->drawRect(g->getClipBounds(), backgroundPaint);
```

src/app/ui/tree.cpp
```diff
@@ -0,0 +1,570 @@
...
+  return Widget::onProcessMessage(msg);
+}
+
+void Tree::onPaint(PaintEvent& ev)
+{
+  Graphics* g = ev.graphics();
+  const auto* view = View::getView(this);
+  ASSERT(view);
+  if (!view)
+    return;
```
dacap: Remove
```suggestion
```

src/app/ui/tree.cpp
```diff
@@ -0,0 +1,570 @@
...
+                                  nodePoint.y);
+        textPoint.x += node->icon()->size().w + (itemSpacing * 2);
+      }
+
+      // Selection background rect
+      if (node == m_selected && isEnabled())
+        g->drawRect(
+          gfx::RectF(textPoint - gfx::Point(itemSpacing / 2, itemSpacing / 2),
+                     node->textBlob()->bounds().size() + gfx::SizeF(itemSpacing, itemSpacing)),
+          selectedBackgroundPaint);
```
dacap: Just a couple of `{ ... }`:
```suggestion
      if (node == m_selected && isEnabled()) {
        g->drawRect(
          gfx::RectF(textPoint - gfx::Point(itemSpacing / 2, itemSpacing / 2),
                     node->textBlob()->bounds().size() + gfx::SizeF(itemSpacing, itemSpacing)),
          selectedBackgroundPaint);
      }
```

dacap (changes requested):
The new `Tree` widget looks great!

Some comments about the properties. In a far future I'd love to have a kind of `std::vector<TreeNode>` for each level nodes (in that case I'd prefer properties), but having a linked-list of dynamic nodes, it seems unnecessary to have a list of properties on each node. Some extra comments about this below.

dacap (commented):
It looks great 🚀  I'll merge the PR as it is and make a minor change later.

Issue #4746, opened 2024-10-31
Aseprite always stamps GIF87a version in the file header when saving in GIF format


Gasparoken:
Internal report: https://igarastudio.zendesk.com/agent/tickets/6862
An user noticed that uploading a gif created from Aseprite to a Facebook account behaves differently compared to other gif files.

To find out the difference, we tried varying some saving parameters: we tried constant frame duration, we tried using only a global palette, but @martincapello found that the GIF version described in the GIF header made a difference. Aseprite saves gif files as GIF87a version instead of GIF89a. But GIF89a would be more convenient for a wide compatibility.

On code it was observed that `const char * EGifGetGifVersion(GifFileType *GifFile)` function always return GIF87_STAMP since:
- `GifFile->ImageCount` == 0 (in all cases)
- `GifFile->ExtensionBlockCount` == 0 (in all cases)
More research is needed to understand why `ImageCount` and `ExtensionBlockCount` are both 0.

One idea to solve this very easily would be to stamp GIF89a in all cases, assuming that all current (and older) viewers support this format. However, there might be some very old software (i'm talking MS-DOS) that only supports GIF87a (unlikely at this time).

@dacap what do you think is more convenient in this case?

PR #5942, opened 2026-07-23
Set the GIF version to 89a when exporting > 1 frame

s0ph0s-dog:
Hi! I ran into #4746. In one of my own projects, I implemented an animated GIF detection routine to suppress generating preview-sized images for animations. Based on the GIF specifications ([87a](https://www.w3.org/Graphics/GIF/spec-gif87.txt) and [89a](https://www.w3.org/Graphics/GIF/spec-gif89a.txt)), only GIF 89a supports the extensions which define animation support. In my original code, I therefore assumed that any GIF 87a file would not include animation and exited the routine early. This assumption turns out to be invalid, as proved by Aseprite :)

I have implemented a very small patch for Aseprite which tells giflib to write the GIF 89a header when the file has more than one animation frame. In my own testing, it works as expected: animated files export as GIF 89a and static files export as 87a.

### DIFF
--- a/src/app/file/gif_format.cpp
+++ b/src/app/file/gif_format.cpp
@@ -1073,7 +1073,9 @@ public:
 
   bool encode()
   {
-    writeHeader();
+    gifframe_t nframes = totalFrames();
+    // #4746: Use GIF 89a when there is more than one frame in the GIF.
+    writeHeader(nframes > 1);
     if (m_loop >= 0)
       writeLoopExtension();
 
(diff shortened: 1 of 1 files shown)


dacap:
Thanks for this patch @s0ph0s-dog, from what I can see from the 87a spec:
``` 
Appendix D - Multiple Image Processing

        Since a  GIF  data  stream  can  contain  multiple  images,  it  is
   necessary  to  describe  processing and display of such a file.  Because
   the image descriptor allows  for  placement  of  the  image  within  the
   logical  screen,  it is possible to define a sequence of images that may
   each be a partial screen, but in total  fill  the  entire  screen.   The
   guidelines for handling the multiple image situation are:
   1.  There is no pause between images.  Each is processed immediately  as
       seen by the decoder.
   2.  Each image explicitly overwrites any image  already  on  the  screen
       inside  of  its window.  The only screen clears are at the beginning
       and end of the  GIF  image  process.   See  discussion  on  the  GIF
       terminator.
```
it looks like multiple images from 87a spec were used only to create a bigger (non-animated) picture (that's why there is no delay between images).

We'll try to include the fix in the next stable release.

PR #5953, opened 2026-07-27
Fix break position in some switch/case statements

dacap:
We should always prefer:
```
case: {
  ...
  break;
}
case: ...
```
Instead of the compressed way:
```
case: {
  ...
} break;
case: ...
```
As it's easier to read/gives an extra separation between cases.


ckaiser (approved):
I think I'm partial to the break outside of the braces (all of the RunCommand ones are mine 😅), but we should pick one and stick to it, I tried to see if we could enforce it with `clang-format` or `clang-tidy` but I couldn't find anything.

PR #5957, opened 2026-07-28
Fix files larger than 2.15 GB can't be saved (fix #5935)

### LINKED ISSUE #5935: Files larger than 2.15 GB can't be saved
**Describe the bug**
Aseprite files can be as large as ~2.15 GB.
Internal example: https://igarastudio.zendesk.com/agent/tickets/9691

**To Reproduce**
1. Produce a large file with the following script:
```lua
local W = 1024
local H = 1024
local nframes = 510
local spr = Sprite(W, H, ColorMode.RGB)
spr:deleteCel(spr.cels[1])
for i = 2, nframes do
    spr:newEmptyFrame()
end
local layer = spr.layers[1]
for f = 1, nframes do
    local img = Image(W, H, ColorMode.RGB)
    for y = 0, H - 1 do
        for x = 0, W - 1 do
            img:drawPixel(x, y, math.ceil(math.random()*0xFFFFFFFF))
        end
    end
    spr:newCel(layer, spr.frames[f], img, Point(0, 0))
end
```
2. Save as `big.aseprite` (the file will have a size of 2.14GB)
3. Open `big.aseprite`
4. Copy and paste the layer to get 2 layers
5. Save as `big2.aseprite`
6.  `big2.aseprite` will have a size of ~2.15GB instead ~4.28GB

**Aseprite & System:**
- Aseprite: 1.3.17.2
- System: Windows and MacOS

Gasparoken:
fix #5935

### DIFF
--- a/src/app/file/file.cpp
+++ b/src/app/file/file.cpp
@@ -1109,18 +1109,34 @@ void FileOp::postLoad()
     // Creates a suitable palette for RGB images
     if (m_createPaletteFromRgba && sprite->pixelFormat() == IMAGE_RGB &&
         sprite->getPalettes().size() <= 1 && sprite->palette(frame_t(0))->isBlack()) {
-      std::shared_ptr<Palette> palette(
-        render::create_palette_from_sprite(sprite,
-                                           frame_t(0),
-                                           sprite->lastFrame(),
-                                           true,
-                                           nullptr,
-                                           nullptr,
-                                           m_config.newBlend,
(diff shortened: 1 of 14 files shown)


dacap:
I'll merge this PR as it is, doing some changes (in laf too) to avoid duplicating code and adding some abstraction/helper for cases where we have to calculate the file size.

For the .aseprite format, I'll add a note (and change the implementation) that if the size if 0xffffffff it means that it can be bigger than that.

PR #5958, opened 2026-07-28
Add new tile-based rendering (#3516)

### LINKED ISSUE #3516: Shader-based render engine
A planned effort we have (as an item of #960) is to rewrite the rendering engine to use shaders (SkSL language, skia shaders). This should fix several issues at the same time and can be a progressive effort:

- [ ] use SkShaders to use CPU or GPU accelerated rendering (similar as we've done with color selectors in #2900)
- [ ] cache rendered tiles (portion of the canvas in an editor/frame)
- [ ] rendering layer groups separately #3225
- [ ] mask layers https://github.com/aseprite/aseprite/issues/459#issuecomment-260010136

I have a minimal work-in-progress commit to start the refactor, but nothing serious yet.

dacap:
This is a work-in-progress for #3516. The new render engine can be enabled only when running Aseprite with the -dev parameter.

This is an experimental idea to render the canvas using tiles which:
1. can be cached for smooth scrolling
2. re-used for zooming (scaling tiles)
3. rendered in parallel (w/multiple threads for each tile)
4. rendering different qualities at different times (nearest neighbor first, bilinear later/in background, etc.)

https://github.com/user-attachments/assets/f128b400-d363-4f4a-9894-3074f033feac

Some of this work will be merged as it is. The tile-based renderer is still a work in progress, only available in `-dev` mode.

This new renderer is heavily related to #378, in a way that we need to fix every place where a modification/notification should be triggered (before only to remove `update_screen_for_document()`/repaint editors, but now to invalidate cached tiles).


dacap:
The new tile-based rendering is not ready, but with the latest refactoring we could integrate it and continue working on it in `-dev` mode.

dacap:
Some of this work will be merged as it is. The tile-based renderer is still a work in progress, only available in `-dev` mode.

This new renderer is heavily related to #378, in a way that we need to fix every place where a modification/notification should be triggered (before only to remove `update_screen_for_document()`/repaint editors, but now to invalidate cached tiles).

PR #5961, opened 2026-07-30
Add adjustable mouse wheel speed (fix #5103)

### LINKED ISSUE #5103: Scroll wheel zoom on MacOS is faster than light itself
https://cdn.discordapp.com/attachments/1279836614327144582/1359230331072548985/20250408_141451.mp4?ex=67f6b964&is=67f567e4&hm=849ae151dc45e25c573572bf0cb53a0d66723956fa3d2fa58f793c26ee91319d&

i have no idea what's happening please help

Gasparoken:
These changes make it possible to fine-tune the sensitivity/response speed of both the mouse wheel and trackpad wheel emulation.

During the implementation, I realized that each context using trackpad wheel emulation required its own baseline adjustment. As a result, I had to experimentally tune some values, which is why you may notice a few hardcoded constants.

There is a complementary PR to improve the behavior on Windows: https://github.com/aseprite/laf/pull/192

fix #5103

### DIFF
--- a/data/pref.xml
+++ b/data/pref.xml
@@ -190,6 +190,7 @@
       <option id="show_scrollbars" type="bool" default="true" />
       <option id="auto_scroll" type="bool" default="true" />
       <option id="auto_scroll_speed" type="int" default="100" />
+      <option id="wheel_scroll_speed" type="int" default="100" />
       <option id="right_click_mode" type="RightClickMode" default="RightClickMode::PAINT_BGCOLOR" />
       <option id="auto_select_layer" type="bool" default="false" />
       <option id="auto_select_layer_quick" type="bool" default="true" />
--- a/data/strings/en.ini
+++ b/data/strings/en.ini
@@ -1505,6 +1505,7 @@ show_scrollbars = Show scroll-bars in sprite editor
 show_scrollbars_tooltip = Show scroll-bars in all sprite editors
 auto_scroll = Auto-scroll on editor edges
 auto_scroll_speed = Speed [Slower / Faster]:
+wheel_scroll_speed = Wheel Speed [Slower / Faster]:
 auto_fit = Auto-fit on screen when a sprite is opened
 straight_line_preview = Preview straight line immediately with the Pencil tool
 straight_line_preview_tooltip = The Pencil tool can draw straight lines\nusing Shift + Click, with this option checked\nyou will see the preview immediately when\nthe Shift key is pressed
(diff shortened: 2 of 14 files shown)


dacap:
I have this one assigned but it looks like you are still working in https://github.com/aseprite/laf/pull/192

Gasparoken:
It's ready, you can check it. I'm not working on this anymore.

dacap:
In the past I've tried a lot of things to improve the trackpad situation on Windows, but the Win32 API always disappoints.

In this case I've tested a trackpad, a mouse wheel, and a Wacom touchpad (it's an Intuos CTH480, old drivers, I should check in something new like OpenTabletDriver as the latest Wacom drivers deprecated this tablet).

So far the results are pretty good for the panning (I've tested panviewport example too), but it start getting messy when we want to handle the regular mouse wheel for zooming. In this video you can see how both actions (scrolling/panning vs zooming) are competing, because when `(ABS(z) >= WHEEL_DELTA)` it's a not precise movement (like a regular mouse wheel notch = zoom by default), and when we move it slowly, it's handled as a panning action (because it's detected as a precise touchpad):

trackpad

As a side note: It looks like the horizontal wheel (`WM_MOUSEHWHEEL`) doesn't generate the zooming action and just panning.

With the Wacom touchpad, it has a similar problem, but horizontally it generates a `WM_HSCROLL` event, which makes things even worse (super fast horizontal scroll), and it's mixed with the slow vertical panning + zooming. Again, I'm not worry too much about this as probably we'd need to update to another kind of API to handle the Wacom/stylus cases (but generally it depends on the specific driver).

At the moment Windows offers several messages to create mouse wheel/panning information (we create a `Event::MouseWheel` from `WM_HSCROLL WM_VSCROLL WM_MOUSEWHEEL WM_MOUSEHWHEEL WM_POINTERWHEEL WM_POINTERHWHEEL INTERACTION_ID_MANIPULATION`); but none of them is enough to identify if we are receiving the information from a discrete mouse wheel notch, or from a precise touchpad.

In the case of https://github.com/aseprite/laf/pull/192, probably if we could identify that all consecutive mouse wheel events belong to the same pointer ID, it might help us to identify that all `MouseWheel` are precise (even when we reach the `z >= WHEEL_DELTA` threshold).

dacap:
I'm testing with `OS_USE_POINTER_API_FOR_MOUSE 1` (from https://github.com/aseprite/laf/blob/68259e1904fab548529e78583af71869fca1cb53/os/win/window.h#L27), and it works a lot better (the wacom touchpad and the trackpad).

A lot of time has passed [from this](https://github.com/aseprite/laf/blob/68259e1904fab548529e78583af71869fca1cb53/os/win/window.cpp#L193-L207), so probably we could give a try (even more now that we are not using wintab as the default impl to handling stylus input).

But it's something to test a lot in several devices (even in a Microsoft Surface, I have one [to test this](https://github.com/aseprite/laf/blob/68259e1904fab548529e78583af71869fca1cb53/os/win/window.h#L171-L201) again).

PR #5968, opened 2026-08-06
Fix multi-layer selection movement preview isn't working (fix #2865)

### LINKED ISSUE #2865: Moving a selection from multiple layers doesn't show anything moving (or moves the thing from one layer)

Gasparoken:
This fix solves the multi-layer/multi-cel transformation preview.

- It can manage a mix of image and tilemap layers in TilemapMode::Pixels or TilemapMode::Tiles.
- Considers opacities of each cel.
- Considers the blend modes of each layer.
- Considers the Z-order of each cel.
- Consider the overlaps between selected and unselected layers.
- Correct behavior of Skew/Rotate/Resize/Move/Flip/Shift transformations.
- In Tilemap Mode::Tiles, the tilemap layers that are transformed (Resize/Move) are restricted to those whose tile grid matches that of the active layer.
- When generating new tiles in a tilemap layer by breaking tiles because transformation in TilemapMode::Pixels, it was decided to do so in TilesetMode::Stack mode to avoid accidental destruction of tiles in the tileset.
- Fixed the Shift Pixels in a multi-layer transformation when it is the first transformation command.
- Consider non-rectangular selections.
- Consider linked cels.

fix #2865

### DIFF
--- a/src/app/commands/cmd_move_mask.cpp
+++ b/src/app/commands/cmd_move_mask.cpp
@@ -1,5 +1,5 @@
 // Aseprite
-// Copyright (C) 2019-2024  Igara Studio S.A.
+// Copyright (C) 2019-present  Igara Studio S.A.
 // Copyright (C) 2001-2018  David Capello
 //
 // This program is distributed under the terms of
@@ -10,7 +10,6 @@
 #endif
 
 #include "app/app.h"
-#include "app/cmd/shift_masked_cel.h"
 #include "app/commands/cmd_move_mask.h"
 #include "app/commands/command.h"
 #include "app/commands/params.h"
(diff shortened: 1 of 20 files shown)


dacap:
It works great! :rocket:

There are some structures that I'd like to improve in the future (probably merging ExtraCel+ExtraCelData+ExtraCelInfo in a common object), but it's something I'll be thinking along #3516 

Just as an optimization for the future: I'd consider to avoid transforming non-visible cels while we are in the middle of a `PixelsMovement` (i.e. just for preview purposes we could set `PixelsMovement::m_selectedCels` as the set of visible cels in all the current visible editors, and then in the final `PixelsMovement::dropImage()` use the full set of selected cels), but it's something to fix later.

PR #5981, opened 2026-08-14
Use new laf-base/file size API to avoid platform-specific code handling big files

dacap:
Refactors https://github.com/aseprite/aseprite/pull/5957

It moves all platform-specific code to handle file sizes/tell/seek to laf-base functions. There is a new `base::file_size(FILE*)` function to get the file size of an open `FILE` handle too.

This also fixes the `AsepriteDecoder::readCelChunk()` using the new `base::fileoff_t` type to limit the decompression of images to the expected chunk size.

It might be problems if one chunk is bigger than 4GB, but probably it's too much to ask to the file format. And finally the header file size will be `0xFFFFFFFF` when the whole file is bigger than 4GB (it is commented in the `.aseprite` spec).

### DIFF
--- a/docs/ase-file-specs.md
+++ b/docs/ase-file-specs.md
@@ -67,6 +67,7 @@ To read the sprite:
 A 128-byte header (same as FLC/FLI header, but with other magic number):
 
     DWORD       File size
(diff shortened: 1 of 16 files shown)


Gasparoken:
Works great!

PR #5989, opened 2026-08-19
Throw an exception when DocDestroyer can't unlock the document (fix #5988)

### LINKED ISSUE #5988: Aseprite crashes when calling `sprite:close()` and `app.command.OpenFile` in an `app.transaction`
**Describe the bug**
Aseprite crashes when calling `app.sprite:close()` followed by `app.command.OpenFile{ui=false, filename = filepath}` within an `app.transaction`

**To Reproduce**
1. Create a script:
```lua
-- Should close and reopen the current file
app.transaction("Test", function()
	local filepath = app.sprite.filename
	app.sprite:close()
	app.command.OpenFile{ui = false, filename = filepath}
end)
```
2. Open a file (.aseprite or .png works)
3. Run the script
4. Aseprite crashes to desktop

**Expected behavior**
Aseprite doesn't crash

**Screenshots**

- Aseprite: Aseprite v1.3.18.2
- System: Windows 10
- Extensions: A lot

**Additional context**
Crash doesn't happen when only calling `app:sprite:close()` nor when only calling `app.command.OpenFile{}`

I didn't expect the app.transaction to do much, as I don't think these actions are generally undoable, but a crash was a bit unexpected

ckaiser:
Fixes #5988 avoiding a crash by just throwing if we can't unlock the document, should also help in other places that use DocDestroyer since some of them are explicitly catching the lock exceptions but these would've never been throw with the current code., which I'm not sure if it's an oversight or maybe something that got refactored away at some point.

### DIFF
--- a/src/app/doc_access.h
+++ b/src/app/doc_access.h
@@ -194,6 +194,10 @@ public:
   {
     ASSERT(m_doc != nullptr);
 
+    m_doc->unlock(m_lockResult);
+    if (m_lockResult != LockResult::OK)
+      throw CannotWriteDocException();
+
     // Don't create a backup for destroyed documents (e.g. documents
     // are destroyed when they are used internally by Aseprite or by
     // a script and then closed with Sprite:close())
@@ -202,8 +206,6 @@ public:
 
     m_doc->close();
     Doc* doc = m_doc;
-    unlock();
-
     delete doc;
     m_doc = nullptr;
   }
@@ -212,11 +214,13 @@ public:
   {
     ASSERT(m_doc != nullptr);
 
+    m_doc->unlock(m_lockResult);
+    if (m_lockResult != LockResult::OK)
+      throw CannotWriteDocException();
+
     Context* ctx = (Context*)m_doc->context();
     m_doc->close();
     Doc* doc = m_doc;
-    unlock();
-
     ctx->closeDocument(doc);
     m_doc = nullptr;
   }


src/app/doc_access.h
```diff
@@ -194,6 +194,10 @@ class DocDestroyer : public DocWriter {
   {
     ASSERT(m_doc != nullptr);
 
+    m_doc->unlock(m_lockResult);
+    if (m_lockResult != LockResult::OK)
+      throw CannotWriteDocException();
+
```
dacap, line 200: Actually `unlock()` is not the part that can fail, only the lock can throw (or should throw) an exception. It means that `DocDestroyer` (or `DocWriter`) should throw an exception if the `m_lockResult != OK`. Probably patching `DocDestroyer` ctor:
```
  DocDestroyer(Context* context, Doc* doc, int timeout) : DocWriter(doc, timeout)
  {
    if (m_lockResult != LockResult::OK)
      throw CannotWriteDocException();
  }
```
As it cannot support a "re-entrant" lock (like `DocWriter` does).

The original bug is due DocDestroyer/DocWriter does a re-entrant lock (a lock from the same thread as the transaction, which is the other lock).

dacap:
> DocDestroyer since some of them are explicitly catching the lock exceptions but these would've never been throw with the current code., which I'm not sure if it's an oversight or maybe something that got refactored away at some point.

The locking mechanism is throwing exceptions only if the lock fails, it supports re-entrant locks. I guess you saw that `DocDestroyer` is "caching" (not throwing) an exception for re-entrants lock, but for other cases `DocWriter` is throwing just fine.

I think the bug is that `DocDestroyer` doesn't support "multiple writes in the same thread" (re-entrant writer locks). So the only way to close a sprite is if the lock is unique.

PR #5990, opened 2026-08-19
Use a ComboBox to select variants for themes with a lot of them

ckaiser:
variant-dropdown


dacap:
It works great :+1: and we keep the button set when there are a few variants.

PR #6000, opened 2026-08-25
Fix Snap to Grid alignment for brushes larger than 1 pixel (fix #5998)

### LINKED ISSUE #5998: First point misalignment when Snap to Grid is ON
When "Snap to Grid" is ON, the first brush position is calculated by finding the shortest distance between the brush center and the nearest grid point on the canvas, then placing the brush's top-left corner at that point. This causes the first brush stamp to be offset from the expected grid position.

**To Reproduce**
Steps to reproduce the behavior with an example:
1. New Sprite 100x100
2. Select Pencil tool with 10x10 square brush
3. Snap to Grid = ON (16x16 as default)
4. Go to coordinate x=25, y=25 (note the brush bound is x=20, y=20, w=10, h=10), and left click to stamp the brush
5. Result: the brush is stampped within the bounds x=32, y=32, w=10, h=10 (unexpected offset of 12 pixels)

**Expected behavior**
(For the example)
The brush would be stampped within the bounds x=16, y=16, w=10, h=10 (offset of 4 pixels)

**Screenshots**
Image

- Aseprite: 1.3.18.3

**Additional context**
Internal report: https://igarastudio.zendesk.com/agent/tickets/10071 
Related issue: https://github.com/aseprite/aseprite/issues/4373

Gasparoken:
fix #5998

### DIFF
--- a/src/app/tools/tool_loop_manager.cpp
+++ b/src/app/tools/tool_loop_manager.cpp
@@ -305,6 +305,7 @@ void ToolLoopManager::snapToGrid(Stroke::Pt& pt)
     return;
 
   gfx::Point point(pt.x, pt.y);
+  point -= m_toolLoop->getBrush()->center();
   point = snap_to_grid(m_toolLoop->getGridBounds(), point, PreferSnapTo::ClosestGridVertex);
   point += m_toolLoop->getBrush()->center();
   pt.x = point.x;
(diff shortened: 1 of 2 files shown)


dacap:
LGTM, shorter jumps when the brush is bigger :+1:

PR #6008, opened 2026-09-01
Avoid registering extensions for disabled formats (fix #6007)

### LINKED ISSUE #6007: Custom file formats cannot register .psd when built-in PSD support is disabled
**Describe the bug**

A plugin cannot register `.psd` or `.psb` through `Plugin:newFileFormat()` in an official Aseprite build that does not provide built-in PSD loading.

Registration fails because both extensions are considered already registered. This leaves neither the built-in application nor a plugin able to handle PSD files through the normal **File > Open** workflow.

**To reproduce**

Create an extension with the following script:

```lua
function init(plugin)
  plugin:newFileFormat{
    name="PSD Test",
    extensions={"psd"},
    onload=function(ev)
      error("Not reached")
    end,
  }
end
```

Install and enable the extension.

The extension fails during initialization because `.psd` is treated as an existing file format. The same occurs with `.psb`.

I also tested an otherwise identical registration using an unused extension as a control:

```text
psdtest = registration succeeds
psd     = registration fails
psb     = registration fails
```

The `onload` callback is never reached.

**Expected behavior**

When the built-in PSD handler is not included in the current build, plugins should be able to register `.psd` and `.psb`.

If Aseprite is built with `ENABLE_PSD`, rejecting the custom registration would be expected because the built-in handler is then available.

**Additional context**

The built-in `PsdFormat` is registered conditionally:

https://github.com/aseprite/aseprite/blob/main/src/app/file/file_formats_manager.cpp

However, `.psd` and `.psb` are recognized unconditionally by `dio::detect_format_by_file_extension()`:

https://github.com/aseprite/aseprite/blob/main/src/dio/detect_format.cpp

`dio::register_custom_format_extension()` uses this detection result to decide whether an extension is already occupied. Consequently, the extensions remain unavailable to plugins even when no corresponding built-in handler is registered.

It would be helpful if extension reservation reflected the file-format handlers enabled in the current build.

This currently prevents PSD importer extensions from integrating with **File > Open**. They must instead provide a separate import command as a workaround.

**Aseprite & System**

- Aseprite: 1.3.18.3-x64, Steam
- System: Windows x64
- Extensions: minimal test extension using `Plugin:newFileFormat()`

ckaiser:
Fixes #6007, added WebP as well since it had the same potential issue.

Had to add the ENABLE_WEBP definition to dio-lib, unsure as to why that one isn't getting defined but the PSD one is 🤔

### DIFF
--- a/src/app/CMakeLists.txt
+++ b/src/app/CMakeLists.txt
@@ -151,6 +151,7 @@ target_sources(app-lib PRIVATE
   file/svg_format.cpp
   file/tga_format.cpp)
 if(ENABLE_WEBP AND WEBP_FOUND)
+  target_compile_definitions(dio-lib PUBLIC -DENABLE_WEBP)
   target_compile_definitions(app-lib PUBLIC -DENABLE_WEBP)
   target_sources(app-lib PRIVATE
     file/webp_format.cpp)
--- a/src/dio/detect_format.cpp
+++ b/src/dio/detect_format.cpp
@@ -178,11 +178,15 @@ FileFormat detect_format_by_file_extension(const std::string& filename)
   if (ext == "css")
     return FileFormat::CSS_STYLE;
 
+#ifdef ENABLE_WEBP
   if (ext == "webp")
     return FileFormat::WEBP_ANIMATION;
+#endif
 
+#ifdef ENABLE_PSD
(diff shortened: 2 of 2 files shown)


dacap:
LGTM :+1:

PR #6014, opened 2026-09-02
Harden TLS, Lua sandbox, extensions, and file parsers

giobuilds:
Personal-fork hardening offered upstream. **Please close if you do not accept AI-assisted patches** — Igara's contributing guide says those will not be merged. This PR is not a CLA-signed contribution and does not claim otherwise.

Companion launcher fix: https://github.com/aseprite/laf/pull/200

This tree currently points the `laf` submodule URL at https://github.com/giobuilds/laf so the fork clones with the `xdg-open` fix; that `.gitmodules` URL change should not be merged as-is.

## What this changes

**Linux network**
- Build bundled curl with OpenSSL (the previous CMake path left Linux with no TLS backend).
- First-party website / news / update URLs use `https://` on all platforms.
- `HttpRequest` allows HTTPS (and loopback HTTP only), verifies TLS, sets timeouts, and caps the body at 8 MiB.

**Lua**
- `io.open` modes `a` / `r+` / `+` are treated as write.
- Native module searchers and `package.cpath` are disabled (`package.loadlib` was wrapped; `require` of a `.so` was not).
- `dofile` / `loadfile` / the Lua `require` searcher go through `ask_access(Read)`.
- Script grants are keyed off the real `evalFile`/`dofile` path, not spoofable `load(code, chunkname)` debug source.
- `os.rename` always asks write on the destination.
- `app.fs` list/stat require read; mkdir/rmdir prompt with an absolute path.
- `plugin:newCommand` cannot delete or replace a built-in command.

**Extensions**
- Reject `..` / absolute zip members, libarchive `SECURE_NODOTDOT|SYMLINKS|NOABSOLUTEPATHS`, and destination paths that escape the extensions dir.
- Escape non-identifier keys when serializing `__pref.lua`.
- Plugin script paths must stay under the extension directory.
- Adds a zip-slip gtest.

**Parsers**
- ASE chunk size < 6 or wrap is an error; compressed reads are capped to remaining chunk bytes (closes a `size_t` wrap into `fread`).
- `Palette::setEntry` is bounds-checked in all builds.
- Tileset `ntiles` is treated as unsigned and size-capped; null image after `Image::create` is handled.
- GIF rejects a 0×0 canvas; BMP size uses 64-bit math so the 128 MiB cap cannot wrap.

CLI `--script` is still unsandboxed when there is no UI (existing automation behavior).

## Test plan

- Linux RelWithDebInfo build with Skia; binary links `libssl`/`libcrypto`; strings show `https://www.aseprite.org/` and `https://blog.aseprite.org/rss`.
- `tests/run-tests.sh` Lua scripts including `os`, `app_fs`, and `require`.
- New `Extensions.ZipSlipRejected` test is in the tree; this local build had `ENABLE_TESTS=OFF` so that gtest was not run here.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -29,7 +29,7 @@
 	url = https://github.com/aseprite/undo.git
 [submodule "laf"]
 	path = laf
-	url = https://github.com/aseprite/laf.git
+	url = https://github.com/giobuilds/laf.git
 [submodule "third_party/cmark"]
 	path = third_party/cmark
 	url = https://github.com/aseprite/cmark.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,16 +111,17 @@ endif()
 
 # SSL/TLS support
 if(REQUIRE_CURL AND NOT USE_SHARED_CURL)
-  # Disable OpenSSL (use native libraries only)
-  set(CMAKE_USE_OPENSSL OFF CACHE BOOL "Use OpenSSL code. Experimental")
   set(CMAKE_USE_LIBSSH2 OFF CACHE BOOL "Use libSSH2")
 
   if(WIN32)
+    set(CMAKE_USE_OPENSSL OFF CACHE BOOL "Use OpenSSL code. Experimental")
     set(CMAKE_USE_SCHANNEL ON CACHE BOOL "enable Windows native SSL/TLS")
   elseif(APPLE)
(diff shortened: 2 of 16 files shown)


giobuilds:
Opened against the wrong repository — this was meant for the giobuilds fork only, not upstream.

Issue #6028, opened 2026-09-08
[Windows] Automated Pencil and marquee drags remain at the starting pixel


lokicik:
**Describe the bug**

On Windows, a drag issued through the `@oai/sky` Computer Use API draws only the initial pixel with the 1 px Pencil. A Rectangular Marquee drag similarly produces a 1 x 1 selection. Manually holding the mouse button and dragging worked on the same machine.

**To Reproduce**

1. Open the desktop Aseprite application and create a transparent RGB document. The matched local tests used a 1305 x 810 canvas.
2. Select a black 1 px Pencil.
3. Use the Windows `@oai/sky` Computer Use `drag` operation between two well-separated points inside the canvas, then release. In our window layout, the screen coordinates were (650, 350) to (1050, 500); these coordinates must be adjusted for a different window/zoom.
4. Observe that only the starting pixel is painted.
5. Repeat with Rectangular Marquee. Our screen-coordinate drag was (650, 650) to (1050, 850), and the resulting selection was 1 x 1.

This reproduction currently depends on that automation runtime. We have not captured its Windows input-event stream or established whether it sends a single movement, coalesced movements, or a different event sequence. We are not claiming all synthetic input tools or ordinary physical mouse drags are affected.

**Expected behavior**

The Pencil connects the drag positions, and Rectangular Marquee selects the dragged area.

**Aseprite & System**

- Source: official v1.3.18.3 source archive; the locally built executable reports 1.3.18.3-dev.
- Windows, OS build 10.0.26200.0.
- Build: Visual Studio 2022 x64 MSVC 14.44.35207, RelWithDebInfo, Skia backend.
- The comparison builds used separate portable profiles seeded from the same settings. The installed Steam application was not modified.
- No extension API was used to draw; the reproduction operates native Pencil and selection tools through the GUI. Installed extension inventory was not recorded.

**Local comparison**

We compiled both the unmodified release source and a candidate fix, then repeated the same two GUI actions:

| Check | Unmodified | Candidate fix |
| --- | --- | --- |
| Pencil PNG | 1 opaque pixel | 301 opaque pixels forming one continuous, 8-connected line |
| Rectangular Marquee | 1 x 1 | 301 x 151 |
| Compile/link | Passed | Passed |

The PNG pixel counts and connectivity were checked independently after saving through the GUI. There was one matched trial per tool/build. These are focused reproduction results, not a broad regression test.

**Code investigation**

[...]

PR #6029, opened 2026-09-08
Update drawing positions before delayed movement commits

lokicik:
Related to #6028. This PR is closed; the [debugging findings and test results](https://github.com/aseprite/aseprite/issues/6028#issuecomment-5593229236) are recorded in the issue.

On Windows, automated Pencil drags could leave just the starting pixel, and Rectangular Marquee could produce a 1 x 1 selection. With this patch, the same gestures produced a connected 301-pixel line and a 301 x 151 selection.

The patch updates DrawingState's pointer after auto-scroll returns the adjusted mouse position, before a delayed movement can commit. It uses the existing integer `screenToEditor` conversion and preserves the pointer metadata. Mouse-up also flushes a pending movement when the release position hasn't changed. The diff is four source files, +15/-6.

The first version updated the pointer at commit time, which could overwrite a newer scroll position. The revision removes that overwrite. A controlled test reproduced the regression in the first patch and verified the revised behavior. Negative-coordinate checks also confirmed consistent commit/release conversion. Details are in the issue.

Tested on locally built v1.3.18.3 sources with MSVC x64 and Skia:

- Build/link, repository formatting checks and `git diff --check` passed.
- Native GUI checks covered Pencil, Marquee, pixel/cel/text movement, undo/redo, cancellation after release, and Line/Rectangle edge auto-scroll. Saved PNGs were checked for pixel counts and preservation.
- A local probe passed 16 checks, including 84 coordinate/zoom/pixel-ratio combinations. It uses production methods with substitute editor/timer scheduling, so it isn't the upstream test suite.

I also ran additional manual checks, without a per-case log. A full current-main build and exhaustive hardware/GUI coverage haven't been verified.

The investigation and patch were developed with Codex assistance.

### DIFF
--- a/src/app/ui/editor/delayed_mouse_move.cpp
+++ b/src/app/ui/editor/delayed_mouse_move.cpp
@@ -123,6 +123,7 @@ bool DelayedMouseMove::updateSpritePos(const ui::MouseMessage* msg)
   // touch the viewport borders.
   const gfx::Point mousePos = m_editor->autoScroll(msg, AutoScroll::MouseDir);
   const gfx::PointF spritePos = m_editor->screenToEditorF(mousePos);
+  m_delegate->onUpdateMousePosition(m_editor, mousePos);
 
   // Avoid redrawing everything if the position in the canvas didn't
   // change.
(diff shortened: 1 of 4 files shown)


lokicik:
Withdrawing this AI-assisted candidate in accordance with the contribution policy. Updated root-cause findings, corrections to the initial approach, and test evidence are recorded in https://github.com/aseprite/aseprite/issues/6028#issuecomment-5593229236. Please keep the issue open for independent investigation; this PR remains available as experimental reference.

PR #6042, opened 2026-09-14
Curl WebSockets & app.net

ckaiser:
Reimplements WebSockets using curl, wich removes our dependency on ixwebsockets and adds a new `net` object with a new `fetch` function similar to what's available in browsers.

Functionality wise WebSocket should now work over HTTPs, API remains largely same which is a slight detriment since the old onreceive function params are not consistent with the rest of our APIs but I'm not sure if that's something that we should change right now.

The one thing that we cannot re-implement is the "deflate" option because curl does not support it (yet?), and the "ping" function is redundant since curl autopings to keep the connecting alive (and this function was providing a string parameter which seems to defeat the purpose of the ping? unclear to me how that used to work)

On the net object we add a few `fetch` function with the following signature:
```lua
app.net.fetch{
    url = string,
    method = "get" | "post" | "put" | "patch" | "options",
    body = string,
    onreceive = function(response)
	--[[
		{
			status = int,
			headers = table,
			json = json object or nil,
			text = string,
			blob = string
		}
	]]--
end
}
```
There's also a few functions to urlencode/decode and use base64 directly, it'd be useful to at some point add base64 utilities to some of the other classes like Image/Sprite, but that's for future consideration.

Network/curl errors are handled by returning status == 0 and putting the error string in the text/blob fields, but this could be improved with a dedicated field and/or error codes.

Marking as draft because it needs testing and API validation.


tests/scripts/net.lua
```diff
@@ -0,0 +1,32 @@
+-- Copyright (C) 2026-present  Igara Studio S.A.
+--
+-- This file is released under the terms of the MIT license.
+-- Read LICENSE.txt for more information.
+
+do
+  if (net ~= nil) then
```
dacap, line 7: I think `net` was used previously and now it's `app.net`:
```suggestion
  local net = app.net
  if net then
```

src/net/http_response.h
```diff
@@ -9,9 +9,7 @@
 #pragma once
 
 #include "base/disable_copying.h"
-
-#include <cstddef>
-#include <iosfwd>
+#include "http_headers.h"
```
dacap, line 12:
```suggestion
#include "net/http_headers.h"
```

src/net/http_request.h
```diff
@@ -20,11 +21,14 @@ class HttpResponse;
 
 class HttpRequest {
 public:
-  HttpRequest(const std::string& url);
+  enum class Method : uint8_t { GET, POST, PUT, PATCH, OPTIONS };
```
dacap, line 24:
```suggestion
  enum class Method : uint8_t { GET, POST, PUT, PATCH, OPTIONS, DELETE };
```

src/app/commands/cmd_open_file.cpp
```diff
@@ -189,7 +194,9 @@ void OpenFileCommand::onExecute(Context* context)
       Doc* doc = fop->document();
       if (doc) {
         if (context->isUIAvailable()) {
-          App::instance()->recentFiles()->addRecentFile(fop->filename().c_str());
+          if (m_addToRecents)
+            App::instance()->recentFiles()->addRecentFile(fop->filename().c_str());
```
dacap, line 198: Sometimes the bot is right 🙏 `c_str()` can be removed here:
```suggestion
            App::instance()->recentFiles()->addRecentFile(fop->filename());
```

src/app/script/debugger.cpp
```diff
@@ -241,8 +241,12 @@ dap::Variable Debugger::makeVariable(lua_State* L, const std::string& name)
       break;
     case LUA_TSTRING:
       var.type = "string";
-      if (auto* str = lua_tostring(L, -1))
+      if (const auto* str = lua_tostring(L, -1)) {
+        // TODO: Having a non-utf8 character in this string (such as raw bytes from an image or a
+        // network download) can crash nlohmann::json so we're gonna have to detect that and trim
+        // them or something like that
```
dacap, line 247: Could we use [base::utf8_decode()](https://github.com/aseprite/laf/blob/500768cccc6447e3b7816bedbfcb378a5c5222e5/base/utf8_decode.h#L33) to check if the string is a valid utf-8?

src/app/script/engine.cpp
```diff
@@ -1093,6 +1093,11 @@ bool Engine::hasLingeringObjects()
   return (m_appEvents && !m_appEvents->empty()) || (m_windowEvents && !m_windowEvents->empty());
 }
 
+bool Engine::isTemporaryFile(const std::string& filename)
```
dacap, line 1096:
```suggestion
bool Engine::isTemporaryFile(const std::string& filename) const
```

src/app/script/engine.h
```diff
@@ -74,6 +74,8 @@ class Engine final {
   // Checks if the Engine has any tracked objects or attached events
   bool hasLingeringObjects();
 
+  bool isTemporaryFile(const std::string& filename);
```
dacap, line 77:
```suggestion
  bool isTemporaryFile(const std::string& filename) const;
```

src/net/http_request.h
```diff
@@ -33,6 +38,10 @@ class HttpRequest {
   DISABLE_COPYING(HttpRequest);
 };
 
+bool is_valid_url(std::string_view url);
+std::string url_encode(std::string_view text);
+std::string url_decode(std::string_view text);
+
```
dacap, line 44: It looks like these functions could go to other files, like `net/url.h` / `cpp`.

tests/scripts/net.lua
```diff
@@ -0,0 +1,33 @@
...
+  if net then
+    assert(net.toBase64("hello_test!") == "aGVsbG9fdGVzdCE=")
+    assert(net.fromBase64("aGVsbG9fdGVzdCE=") == "hello_test!")
+    assert(net.urlEncode("h&llo=wr?d") == "h%26llo%3Dwr%3Fd")
+    assert(net.urlDecode("h%26llo%3Dwr%3Fd") == "h&llo=wr?d")
+    assert(net.urlEncode("テスト") == "%E3%83%86%E3%82%B9%E3%83%88")
+    assert(net.toBase64("") == "")
+    assert(net.fromBase64("") == "")
+    assert(net.urlEncode("") == "")
+    assert(net.urlDecode("") == "")
```
dacap, line 17: About this API I was thinking, I'd prefer something like:
```suggestion
    assert(base64.encode("hello_test!") == "aGVsbG9fdGVzdCE=")
    assert(base64.decode("aGVsbG9fdGVzdCE=") == "hello_test!")
    assert(url.encode("h&llo=wr?d") == "h%26llo%3Dwr%3Fd")
    assert(url.decode("h%26llo%3Dwr%3Fd") == "h&llo=wr?d")
    assert(url.encode("テスト") == "%E3%83%86%E3%82%B9%E3%83%88")
    assert(base64.encode("") == "")
    assert(base64.decode("") == "")
    assert(url.encode("") == "")
    assert(url.decode("") == "")
```
What do you think? Mainly to match the `json.encode/decode()` API. This way the new API would be:
* `app.net.fetch()`
* `base64.encode/decode`
* `url.encode/decode`
I know the url implementation requires curl API, but that would be a "minor" internal detail.

src/net/http_headers.cpp
```diff
@@ -8,13 +8,29 @@
...
   m_map[name] = value;
 }
 
+std::string HttpHeaders::getHeader(const std::string& name) const
+{
+  auto lowerName = base::string_to_lower(name);
+  const auto it =
+    std::find_if(m_map.begin(), m_map.end(), [&lowerName](const auto& header) -> bool {
+      return lowerName == base::string_to_lower(header.first);
+    });
```
dacap, line 29: Just a minor detail as it's not critical, but I'm not sure about the performance of this `std::find_if()` for maps. Probably we could use a `map.find()` if we store headers in lower case, but something just to think about in a possible future.

src/app/script/net_object.cpp
```diff
@@ -0,0 +1,330 @@
...
+  #include "base/base64.h"
+  #include "base/string.h"
+  #include "net/http_headers.h"
+  #include "net/http_request.h"
+  #include "net/http_response.h"
+  #include "ui/system.h"
+  #include "ver/info.h"
+
+  #include "fmt/format.h"
+#endif
```
dacap, line 23:
```suggestion
#ifdef HAVE_CONFIG_H
  #include "config.h"
#endif

#include "app/console.h"
#include "app/i18n/strings.h"
#include "app/script/engine.h"
#include "app/script/luacpp.h"
#include "app/task.h"
#include "base/base64.h"
#include "base/string.h"
#include "fmt/format.h"
#include "net/http_headers.h"
#include "net/http_request.h"
#include "net/http_response.h"
#include "ui/system.h"
#include "ver/info.h"
```

src/app/script/net_object.cpp
```diff
@@ -0,0 +1,330 @@
...
+std::set<FetchRequest*> g_requests;
+
+int AppNet_fetch(lua_State* L)
+{
+  FetchRequest::Params params;
+  std::string methodString = "get";
+
+  params.headers.setHeader(
+    "User-Agent",
+    fmt::format("Mozilla/5.0 (compatible; {}/{}; +https://www.aseprite.org/api/app#fetch)",
```
dacap, line 177: We could get the aseprite.org URL from `get_app_url()` and the final `/api/app#fetch` probably might be `/api/app_net#appnetfetch`

src/app/script/net_object.cpp
```diff
@@ -0,0 +1,330 @@
...
+        if (const auto* typeName = lua_tostring(L, -1); std::strcmp(typeName, "JsonObj") != 0) {
+          return luaL_error(L, "'body' arguments must be a string, table or json object");
+        }
+        lua_pop(L, 1);
+      }
+      else {
+        return luaL_error(L, "'body' arguments must be a string, table or json object");
+      }
+      size_t len;
+      const auto* s = luaL_tolstring(L, -1, &len);
```
dacap, line 241: We can add a comment here:
```suggestion
      // Here we are converting the Json object to a string, i.e. calling JsonObj_tostring()
      const auto* s = luaL_tolstring(L, -1, &len);
```

src/app/script/websocket_class.cpp
```diff
@@ -9,209 +9,536 @@
...
-  close_ws(ws);
-  delete ws;
+  auto* ws = get_obj<WebSocket>(L, 1);
+  ws->~WebSocket();
+  get_engine(L)->untrackObject();
   return 0;
 }
 
-int WebSocket_sendText(lua_State* L)
+std::string get_send_data(lua_State* L, WebSocket* ws)
```
dacap, line 482: Probably this could return a base::buffer directly:
```suggestion
base::buffer get_send_data(lua_State* L, WebSocket* ws)
```
And then move the buffer up to `WebSocketConnection::sendData()`

src/app/script/websocket_class.cpp
```diff
@@ -9,209 +9,536 @@
...
+
+  void sendText(const std::string& text)
+  {
+    if (!m_conn)
+      return;
+    const std::scoped_lock lock(m_mutex);
+    m_conn->sendText(text);
+  }
+
+  void sendBinary(std::vector<char>&& data)
```
dacap, line 396:
```suggestion
  void sendBinary(base::buffer&& data)
```

src/app/script/websocket_class.cpp
```diff
@@ -9,209 +9,536 @@
...
+  void sendText(const std::string& text)
+  {
+    ASSERT(m_outgoingData.empty() && m_outgoingOffset == 0);
+    m_outgoingData = std::vector<char>(text.begin(), text.end());
+    m_outgoingOffset = 0;
+    m_outgoingIsBinary = false;
+    curl_easy_pause(m_curl, CURLPAUSE_CONT);
+  }
+
+  void sendData(std::vector<char>&& data)
```
dacap, line 194:
```suggestion
  void sendData(base::buffer&& data)
```

src/app/script/websocket_class.cpp
```diff
@@ -9,209 +9,536 @@
...
+  void close()
+  {
+    m_shouldClose = true;
+    curl_easy_pause(m_curl, CURLPAUSE_CONT);
+  }
+
+private:
+  CURL* m_curl;
+  std::string m_url;
+  std::vector<char> m_outgoingData;
```
dacap, line 212:
```suggestion
  base::buffer m_outgoingData;
```

src/app/script/websocket_class.cpp
```diff
@@ -9,209 +9,536 @@
...
+  void abort()
+  {
+    curl_easy_setopt(m_curl, CURLOPT_TIMEOUT_MS, 1);
+    curl_easy_setopt(m_curl, CURLOPT_CONNECTTIMEOUT_MS, 1);
+
+    m_shouldClose = true;
+    curl_easy_pause(m_curl, CURLPAUSE_CONT);
+  }
+
+  obs::safe_signal<void(WebSocketMessageType, const std::vector<char>&)> Message;
```
dacap, line 91:
```suggestion
  obs::safe_signal<void(WebSocketMessageType, const base::buffer&)> Message;
```
And so on for each `std::vector<char>` -> `base::buffer`...

src/app/script/websocket_class.cpp
```diff
@@ -9,209 +9,536 @@
...
+
+    lua_rawgeti(L, LUA_REGISTRYINDEX, m_onreceiveRef);
+    lua_Debug ar;
+    if (!lua_getinfo(L, ">u", &ar))
+      return;
+
+    lua_rawgeti(L, LUA_REGISTRYINDEX, m_onreceiveRef);
+
+    int nargs = 1;
+    if (ar.nparams > 1 || ar.isvararg) {
```
dacap, line 308: Probably we should cache this information/decision in the `WebSocket()` ctor instead of asking for debug info on each `onMessage()`.

dacap (changes requested):
Looks great 👍 a lot of work but it'll be great to remove IXWebSocket dependency.

I've made just some final comments + API changes.

tests/scripts/net.lua
dacap: One problem that I'm thinking about `url.encode/decode()` API is that the "url" identifier might be a quite common one for a local variable/function parameter.

PR #6046, opened 2026-09-17
Reduce superfluous TextBlob creation

ckaiser:
We've been creating a lot more text blobs than we needed to, this PR mitigates that in two ways:

Before, we were resetting m_font and m_blob in onInitTheme, but this is premature, as Theme::initTheme will reset these for us by calling the widget's setFont() (which will only reset it if the font actually changes). This avoids many widgets re-creating their blobs when calling initTheme manually, which happens with a lot of widget subclasses and their inner parts.

The second one is through `paintLayer` and `PaintWidgetPartInfo`, since these store the textBlobRef directly by asking the widget for it, which feels unnecessarily if we're not actually performing the paint. The code preserves the ability to override the blob/baseline which is used particularly in the current RecentFiles that I'm replacing soon (which also does it in a particularly cursed way where it uses setTextQuietly multiple times with different parts on each onPaint, which completely thrashes the cache but this widget is going away with the new home so I haven't touched it).


dacap:
Looks good :+1: Yeah, the `setTextQuiet()` in the recent files list is a super hack to avoid creating a new `Theme::calcSizeHint()` with a `PaintWidgetPartInfo` ("Paint"WidgetPartInfo). I hope we can have something better in the future.

PR #6047, opened 2026-09-18
Fix build.sh path quoting when the repo path contains spaces.

Cerallin:
I have a space in my HOME path variable (Windows), so the build script cannot find the downloaded skia. I think quote all paths in the `build.sh` is not a bad idea. :)


dacap:
Thanks for the patch @Cerallin :+1:

PR #6049, opened 2026-09-18
Fix SkinTheme setting the font so that widgets can refresh their blobs

ckaiser:
Fixes a regression introduced in 55509ea where I'd assumed that `SkinTheme::initWidget()` behaved like `Theme::initWidget()` and always set the widget fonts to the default one, now it does this properly and blobs get refreshed when we change a theme and the font changes with them.


src/app/ui/skin/skin_theme.cpp
```diff
@@ -1145,6 +1145,8 @@ void SkinTheme::initWidget(Widget* widget)
 
   const int scale = guiscale();
 
+  widget->setFont(getWidgetFont(widget));
```
dacap: Is this needed for all kind of widgets? (e.g. it looks like it will be called for boxes, grids, etc.).

dacap: Could we restore a fixed version of the previous code like:
```c++
void Widget::onInitTheme(InitThemeEvent& ev)
{
  // Reset cached font and TextBlob
  if (m_font != ev.theme()->font()) {  // I'm not sure if this works
    m_font.reset();
    m_blob.reset();
  }

  ...
}
```

dacap: Or maybe:
```c++
if (m_font != ev.theme()->getWidgetFont(this)) {
```

ckaiser: Made the change and also tried to ensure we also call setFont when changing the font in other places (so onSetFont() is called and subclasses have the chance to invalidate their blobs as well, TextEdit for example)

dacap:
Looks good now! :+1:

PR #6052, opened 2026-09-21
Fix FG/BG color use when dynamics=ON between tools

Gasparoken:


dacap:
It looks like the issue is still present, as `ToolLoopManager::adjustPointWithDynamics` is not being called for the contour tool:

https://github.com/aseprite/aseprite/blob/b9d291bff7afb8e78a9fb3d38fcdf57a80951dc4/src/app/tools/tool_loop_manager.cpp#L383-L386

https://github.com/aseprite/aseprite/blob/b9d291bff7afb8e78a9fb3d38fcdf57a80951dc4/src/app/tools/tool_loop_manager.cpp#L394-L399

And `m_toolLoop->getFilled()` is true for the contour tool. Probably this patch should be merged, but I'm not sure which specific case is fixing.

